PR #1, opened 2019-03-20
f-vue-icons@0.2.0 — Adding icons and config.

DamianMullins:
### Added
- Changelog file.
- Licence files.
- `package.json` details.
- New icon components and config.

### Changed
- Updated `.gitignore` file.
- Set `useBuiltIns` to false in `f-vue-icons` babel config.
- Moved dependencies back into `f-vue-icons`.
- Replaced `for in` loop with `Object.values` in entry module.

### DIFF
--- a/packages/f-vue-icons/src/components/HotIcon/HotIcon.vue
+++ b/packages/f-vue-icons/src/components/HotIcon/HotIcon.vue
@@ -1,0 +1,13 @@
+<template>
+    <spicy-icon class="c-icon c-icon--vegetarian" />
+</template>
+
+<script>
+import SpicyIcon from '@justeat/f-icons/src/img/icons/general/spicy.svg';
+
+export default {
+    components: {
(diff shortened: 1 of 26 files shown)


packages/f-vue-icons/src/components/HotIcon/HotIcon.vue
```diff
@@ -0,0 +1,13 @@
+<template>
+    <spicy-icon class="c-icon c-icon--vegetarian" />
```
xander-marjoram: How come we're using both hot and spicy?

DamianMullins: I feel like there was a reason, but not sure anymore. I'll make things consistent 🌶 

Also, just noticed that this component is using the wrong class.

sonnyprince (approved):

on-ye (approved):

PR #4, opened 2019-04-25
f-vue-icons@0.5.0 — Adding new icons.

DamianMullins:
### Added
- Alert, Basket, Collection, and Delivery icons.

### Changed
- Updated npm dependencies.

### DIFF
--- a/packages/f-vue-icons/package.json
+++ b/packages/f-vue-icons/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@justeat/f-vue-icons",
-  "version": "0.4.0",
+  "version": "0.5.0",
   "main": "dist/f-vue-icons.umd.min.js",
   "files": [
     "dist"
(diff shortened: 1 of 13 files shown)


dandel10n (approved):

kevinrodrigues (approved):

packages/f-vue-icons/package.json
```diff
@@ -35,14 +35,14 @@
...
-    "@vue/cli-service": "3.5.3",
+    "@vue/cli-plugin-babel": "3.6.0",
+    "@vue/cli-plugin-eslint": "3.6.0",
+    "@vue/cli-plugin-unit-jest": "3.6.3",
+    "@vue/cli-service": "3.6.0",
     "@vue/test-utils": "1.0.0-beta.29",
     "babel-core": "7.0.0-bridge.0",
     "babel-jest": "24.7.1",
-    "lerna": "3.13.2",
+    "lerna": "3.13.4",
```
ashleynolan, line 45: Does lerna need to be a dependency of the package?  Would have thought it only needed to be defined at the root of the repo

DamianMullins: I would have thought so too, not sure how it ended up there!

PR #6, opened 2019-05-31
v0.6.0 - created new components for additional icons

sonnyprince:
Addition of:
Cash, Clock, Local Legend, Map pin, Padlock and Stopwatch icons.

into f-vue-icons package.

### DIFF
--- a/packages/f-vue-icons/src/components/LocalLegendIcon/LocalLegendIcon.vue
+++ b/packages/f-vue-icons/src/components/LocalLegendIcon/LocalLegendIcon.vue
@@ -1,0 +1,16 @@
+<template>
+    <local-legend-icon :class="[iconClasses]" />
+</template>
+
+<script>
+import LocalLegendIcon from '@justeat/f-icons/src/img/icons/general/localLegend.svg';
+import iconsMixin from '../../mixins/icons.mixin';
+
+export default {
+    components: {
+        LocalLegendIcon
+    },
+
+    mixins: [iconsMixin]
+};
+</script>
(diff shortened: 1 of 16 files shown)


packages/f-vue-icons/src/components/LocalLegendIcon/LocalLegendIcon.vue
```diff
@@ -0,0 +1,16 @@
+<template>
+    <local-legend-icon :class="[iconClasses]" />
```
dandel10n: There are 'c-icon c-icon--smth' classes for all other icons, are they not in use here?

sonnyprince: actually - we should be using `c-icon` - i'll update!

dandel10n (approved):

packages/f-vue-icons/src/components/MapPinIcon/MapPinIcon.vue
```diff
@@ -0,0 +1,16 @@
+<template>
+    <map-pin-icon :class="['c-icon c-icon--mappin', iconClasses]" />
```
DamianMullins: Should be `c-icon--mapPin`.

packages/f-vue-icons/src/components/index.js
```diff
@@ -10,6 +10,12 @@ import OfferIcon from './OfferIcon';
 import PlusIcon from './PlusIcon';
 import SpicyIcon from './SpicyIcon';
 import VegetarianIcon from './VegetarianIcon';
+import PadlockIcon from './PadlockIcon';
```
DamianMullins: This is mega nit-picky and OCD, but if you feel like it could you arrange these alphabetically?

sonnyprince: done to all imports / exports 👍

DamianMullins (approved):

PR #9, opened 2019-06-21
f-footer@2.0.0-beta.0 Initial project setup

dandel10n:
`estling.js`, `babel.config.js` and `jest.config.js` will be moved from the package level later.


packages/f-footer/src/components/Base.test.js
```diff
@@ -0,0 +1,17 @@
+import { shallowMount } from '@vue/test-utils';
+import Footer from './Base.vue';
+
+describe('ChevronIcon', () => {
```
DamianMullins: Need to update description.

packages/f-footer/src/index.js
```diff
@@ -0,0 +1,28 @@
+// Import vue component
+import Footer from './components/Base.vue';
```
DamianMullins: Not sure we need all of the functions here? The `f-vue-icons` package just exposes the components and the build script add all of the required setup for each type of export.

DamianMullins: I could be wrong :)

packages/f-footer/src/components/Base/Base.vue
```diff
@@ -0,0 +1,10 @@
+<template>
```
DamianMullins: Any reason for the `Base` convention here? Could just be `Footer.vue` in the root of the project?

dandel10n: No, just added first component as a base one. Happy to change it

packages/f-footer/src/index.js
dandel10n: @DamianMullins I took it from here: https://vuejs.org/v2/cookbook/packaging-sfc-for-npm.html#What-does-my-packaged-component-look-like The same approach is used in f-searchbox. But I am not sure what is the best way actually.

packages/f-footer/src/components/Base.test.js
dandel10n: :white_check_mark:

packages/f-footer/src/components/Footer/Footer.vue
```diff
@@ -0,0 +1,10 @@
+<template>
```
ashleynolan: Is it convention to add the component so that it's in a folder (so `Footer/Footer.vue`).

Unless there's a reason for it, would make sense to me to just have the Footer.vue at the root of the `components` folder?

ashleynolan: Also, I'd be tempted to have a convention for the root `.vue` entry file for the component (`index.vue`/ or check Vue docs to see if there's a convention for this?)

ashleynolan (commented):
Looks good – just a couple of small convention/folder comments

packages/f-footer/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@justeat/f-footer",
+  "version": "0.1.0",
```
xander-marjoram: If we're publishing this package to npm, this version will fail as it already exists. Should we be going for `v2.0.0`?

dandel10n: Oh true! We can't use `@justeat/f-footer` name because we already have f-footer package

dandel10n: Actually we can add it as beta version

ashleynolan: Good catch @xander-marjoram – I'd definitely make this a `2.0.0` beta version until we're happy with it to fully be released

packages/f-footer/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
+# Changelog
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+v0.1.0
```
ashleynolan: Will need to update the `CHANGELOG`.  I'd copy over the old one and use the beta versions for now for updates until 2.0.0 releases, as that'll make sense I think.

ashleynolan: Can you update the commit as well (so it says `2.0.0-beta.0` rather than `0.1.0`

dandel10n: > Will need to update the `CHANGELOG`. I'd copy over the old one and use the beta versions for now for updates until 2.0.0 releases, as that'll make sense I think.

What do you mean "copy over the old one", original f-footer changelog?

packages/f-footer/src/components/Footer/Footer.vue
dandel10n: Vue styleguide docs [strongly recommended](https://vuejs.org/v2/style-guide/#Tightly-coupled-component-names-strongly-recommended) tightly coupled component names, like this:
```
components/
|- Footer.vue
|- FooterList.vue
|- FooterListLink.vue
```
Not sure if it will be useful to have `index.vue` in the root. We can change it later if the structure becomes messy or it is hard to find main component file.

DamianMullins: I'm happy for the root component to be named after what the thing is i.e. `Footer.vue` or `Header.vue`. Makes it easy to find, especially in a project like this where there could be multiple `Index.vue` components.

packages/f-footer/src/index.js
```diff
@@ -0,0 +1,28 @@
...
+    GlobalVue = window.Vue;
+} else if (typeof global !== 'undefined') {
+    GlobalVue = global.Vue;
+}
+if (GlobalVue) {
+    GlobalVue.use(plugin);
+}
+
+// To allow use as module (npm/webpack/etc.) export component
+export default Footer;
```
on-ye, line 28: Just a comment but this file is highly reusable and we should consider making it a part of the build process that gets written into every component at build time, as well as configs (babel, eslint, jest etc)

on-ye (approved):

packages/f-footer/src/index.js
dandel10n: That is a future plan, to move all you've mentioned to the `f-build` component

DamianMullins (approved):

PR #10, opened 2019-07-04
f-footer@2.0.0-beta.1 footerList child component with resource files

dandel10n:
PR contains 20 files but 8 of them are basic resource files :)

### Added
- `FooterList` container component
- `TestComponentForFooter` wrapper component for local testing
- tenants resource files
- logic to pick up tenants locales either from component's props or vue i18n plugin
- `@justeat/f-footer` to mono repo dependencies, was missed from initial setup

For visibility footer component for now looks like this:
Screen Shot 2019-07-04 at 16 36 40


packages/f-footer/.eslintrc.js
```diff
@@ -5,13 +5,18 @@ module.exports = {
         node: true
     },
 
+    globals: {
+        Vue: true
+    },
+
     extends: [
         'plugin:vue/recommended',
```
DamianMullins, line 13: I think we should pull in `eslint-config-fozzie` soon, what do you think?

packages/f-footer/babel.config.js
```diff
@@ -1,11 +1,24 @@
...
+    const isSettings = api.env('settings');
+    const presets = [];
+    const plugins = [];
+
+    if (!isTest) {
+        if (!isSettings) {
+            api.cache(true);
+            presets.push(['@vue/app']);
+        }
+        plugins.push('@babel/plugin-proposal-object-rest-spread');
```
DamianMullins: Just wondering why we need to do this?

packages/f-footer/package.json
```diff
@@ -25,9 +25,10 @@
   ],
   "scripts": {
     "prepare": "yarn lint && yarn test && yarn build",
-    "build": "vue-cli-service build --target lib --name f-footer ./src/index.js",
+    "build:settings": "cross-env BABEL_ENV=settings babel src/tenants -d dist/config",
+    "build": "vue-cli-service build --target lib --name f-footer ./src/index.js && yarn build:settings",
     "lint": "vue-cli-service lint --fix",
-    "test": "vue-cli-service test:unit"
+    "test": "vue-cli-service test:unit --no-cache"
```
DamianMullins: Turning off caching can increase the test run time significantly, were you seeing issues with caching enabled?

packages/f-footer/src/components/TestComponentForFooter.vue
```diff
@@ -0,0 +1,12 @@
+<template>
```
DamianMullins, line 1: What is this component used for?

packages/f-footer/babel.config.js
dandel10n: Good spot, I forgot to remove it after testing. Will do

packages/f-footer/src/components/TestComponentForFooter.vue
dandel10n: I made it to test footer component locally with passing different locales using `vue serve --open src/components/TestComponentForFooter.vue`.

packages/f-footer/package.json
dandel10n: Had an issue yesterday with tests failing with `SyntaxError: Unexpected token {` error. But now I found that running `./node_modules/jest/bin/jest.js --clearCache` fixes it so will remove `--no-cache` flag

packages/f-footer/.eslintrc.js
ashleynolan: Yep – I've got a separate PR for some config stuff, so I can switch the footer to use this too?

packages/f-footer/src/components/Footer.vue
```diff
@@ -1,10 +1,53 @@
 <template>
-    <p>Hello world</p>
+    <div class="f-footer">
```
ashleynolan: Could use the `footer` HTML tag as the outer wrapper?

ashleynolan: Also, best to keep the component naming for classes.

So this would be `c-footer`

packages/f-footer/src/tenants/index.js
```diff
@@ -0,0 +1,25 @@
+import uk from './en-GB';
+import au from './en-AU';
+import nz from './en-NZ';
+import ca from './en-CA';
```
ashleynolan: We shouldn't need config for `en-CA` for now?

packages/f-footer/src/tenants/index.js
```diff
@@ -0,0 +1,25 @@
...
+import dk from './da-DK';
+import es from './es-ES';
+import ie from './en-IE';
+import it from './it-IT';
+import no from './nb-NO';
+import frca from './fr-CA';
```
ashleynolan: Same for `fr-CA` (just as there aren't plans to use this in Canada yet)

packages/f-footer/src/tenants/index.js
dandel10n: I transfered all the tenants from current f-footer components but happy to remove CA

packages/f-footer/package.json
```diff
@@ -25,7 +25,8 @@
   ],
   "scripts": {
     "prepare": "yarn lint && yarn test && yarn build",
-    "build": "vue-cli-service build --target lib --name f-footer ./src/index.js",
+    "build:settings": "cross-env BABEL_ENV=settings babel src/tenants -d dist/config",
```
ashleynolan: I don't think you'll need this line for now – it was only in searchbox because we needed to import the config/copy separately but for now we're just pulling the config into the main `vue-cli` build here (until we try out dynamic imports).

ashleynolan (approved):

packages/f-footer/.eslintrc.js
DamianMullins: Sounds good.

packages/f-footer/src/components/TestComponentForFooter.vue
DamianMullins: Can I suggest that we move this into a `/demos` folder or similar? Just keeps the example stuff separate from the actual components.

DamianMullins (approved):

PR #11, opened 2019-07-05
f-footer@2.0.0-beta.2 transfer all footer data from the original f-footer module

dandel10n:
Just recourse files transferring

### DIFF
--- a/packages/f-footer/src/tenants/da-DK.js
+++ b/packages/f-footer/src/tenants/da-DK.js
@@ -25,5 +25,242 @@ export default {
             url: 'http://justeat-premium.dk/',
             text: 'Just Eat Premium'
         }
-    ]
+    ],
+    cuisines: 'Hvad har du lyst til?',
+    cuisineLinks: [
+        {
+            url: '/takeaway',
+            text: 'Takeaway'
+        },
+        {
+            url: '/takeaway/italiensk',
+            text: 'Pizza'
+        },
+        {
+            url: '/takeaway/sushi',
+            text: 'Sushi'
+        },
+        {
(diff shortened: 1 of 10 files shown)


packages/f-footer/src/tenants/da-DK.js
```diff
@@ -25,5 +25,242 @@ export default {
             url: 'http://justeat-premium.dk/',
             text: 'Just Eat Premium'
         }
-    ]
+    ],
+    cuisines: 'Hvad har du lyst til?',
+    cuisineLinks: [
```
DamianMullins, line 30: Couple of questions about the transaltions in this project:

- Does it use Vue i18n? I couldn't see anything when I had a quick look but could be missing something.
- If it's not using Vue i18n, how do the translations work?
- As the translations files grow do we need to think about ways to reduce the amount of code included in the component? We code split in menu web to avoid the end users having to download translations for tenants they don't need.

dandel10n: @DamianMullins It doesn't use Vue i18n, the plan is to use the translations the same way as on f-searchbox. The locale can be passed [via components props or be picked up from the parent's level](https://github.com/justeat/fozzie-components/blob/26e2606ad0fd8965fc59b36ef3dcb3a32b6c26a0/packages/f-footer/src/components/Footer.vue#L32). The reason is to not include Vue i18n plugin multiple times (on reusable components levels and on projects level). And yes, we still have an open question with dynamic loading of this data. I haven't got a chance to dig into it yet, but we will definitely do it later.

ashleynolan (approved):

DamianMullins (approved):

PR #12, opened 2019-07-05
root@v1.1.0 – Moving ESLint configs up to the root and adding browserslist shareable config module

ashleynolan:
### Added
- Shareable browserslist config

### Changed
- ESLint configs moved to root of monorepo


.eslintrc.js
```diff
@@ -1,17 +1,13 @@
 module.exports = {
     root: true,
-
```
ashleynolan, line 3: These are removed as they're part of `eslint-config-fozzie`

DamianMullins (approved):

dandel10n (approved):

PR #16, opened 2019-07-12
f-vue-icons@0.10.0 - Refactor payment and social icons, add app store icons

xander-marjoram:


package.json
```diff
@@ -55,6 +55,8 @@
   },
   "dependencies": {
     "@justeat/f-footer": "file:packages/f-footer",
-    "@justeat/f-vue-icons": "file:packages/f-vue-icons"
+    "@justeat/f-vue-icons": "file:packages/f-vue-icons",
+    "node-sass": "4.12.0",
```
ashleynolan: Can these just be `devDependencies` as they will just be used on build?

packages/f-vue-icons/src/components/AppStoreIcon/AppStoreIcon.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+    <component :is="iconComponent" />
+</template>
+
+<script>
+import IosDkIcon from '@justeat/f-icons/src/img/icons/appstore/ios.dk.svg';
```
ashleynolan: NIT – I'd probably name these as `IosIconDk` etc, to keep the format of the names easier to scan down

packages/f-vue-icons/src/components/AppStoreIcon/AppStoreIcon.vue
```diff
@@ -0,0 +1,48 @@
...
+import IosEnIcon from '@justeat/f-icons/src/img/icons/appstore/ios.svg';
+import IosEsIcon from '@justeat/f-icons/src/img/icons/appstore/ios.es.svg';
+import IosItIcon from '@justeat/f-icons/src/img/icons/appstore/ios.it.svg';
+import IosNoIcon from '@justeat/f-icons/src/img/icons/appstore/ios.no.svg';
+import AndroidDkIcon from '@justeat/f-icons/src/img/icons/appstore/android.dk.svg';
```
ashleynolan: Same as above with Android names

package.json
dandel10n: Should be in dev deps. @xander-marjoram you can just remove it from your PR, because I have these covered in my next PR for f-footer

packages/f-vue-icons/src/components/BaseIcon/BaseIcon.vue
```diff
@@ -0,0 +1,60 @@
+<template>
```
ashleynolan: Is BaseIcon just used for social and payment icons? Just wondering if the naming might be a bit confusing as `BaseIcon` could apply to any icon in this repo so may be worth making it a bit more specific to what they have in common that makes them different to the other icons (that doesn't make them a part of `BaseIcon.vue`)

packages/f-vue-icons/src/components/BaseProviderIcon/BaseProviderIcon.vue
```diff
@@ -0,0 +1,60 @@
...
+        MaestroIcon,
+        MastercardIcon,
+        MastercardSecurecodeIcon,
+        PaypalIcon,
+        VisaIcon,
+        VisaVerifiedIcon
+    },
+    props: {
+        icon: {
+            type: Object,
```
DamianMullins: Looks like this can be a `string` type as you're only referencing `icon.url` in the template?

xander-marjoram: @DamianMullins I'm also referring to `icon.key` in the computed property below

DamianMullins: What is the benefit if it being an object? I'm just wondering if it's worth splitting them out into separate props?

xander-marjoram: It's a good point and it's certainly up for discussion.

I personally prefer only passing in one prop rather than ending up having to pass in 3 or 4. It makes the parent component template a lot more cluttered, but passing in each prop individually does have the advantage of the child not knowing or caring about the structure of the data.

I'm not sure whose call this is, perhaps it's something we can discuss across the UI Guild.

DamianMullins: Yeah, I've always avoided objects as you have to know what shape/data is required to pass in, whereas with explicit props it's very clear what is going in. 

Also, if you have an object with all of the props on it you can do `v-bind="obj"` and it'll pass the object properties as individual props. Here's an example from Menu Web where MenuItem.vue [expects a bunch of props](https://github.je-labs.com/menu/menuweb/blob/541bef5f907c571f6722dbdf4ae52439e5964a65/src/MenuWeb/Assets/vue/components/menu/MenuItem.vue#L79-L104) and we satisfy them all [by passing along the entire`item` object](https://github.je-labs.com/menu/menuweb/blob/541bef5f907c571f6722dbdf4ae52439e5964a65/src/MenuWeb/Assets/vue/components/menu/MenuCategory.vue#L47) rather than declaring each one individually.

xander-marjoram: That's tidy. I'll switch over to that.

packages/f-vue-icons/src/components/BaseProviderIcon/BaseProviderIcon.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+    <component
+        :is="url ? 'a' : 'span'"
+        :href="url">
```
dandel10n: I think we should add `aria-label` here for a link with explanation where this link goes. Before we had an imd with alt text like "Just Eat on Facebook".

dandel10n: Actually we will need it for both `a` and `span` versions because we will render just svgs in both scenarios without text version of what is shown

DamianMullins: Wouldn't we add that in the project which is consuming the icon so that it can be overridden in different contexts?

packages/f-vue-icons/src/components/BaseProviderIcon/BaseProviderIcon.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+    <component
+        :is="url ? 'a' : 'span'"
```
DamianMullins: Why are we wrapping these icons with an element by the way? What if you just want to embed SVGs into the page with no wrapper?

packages/f-vue-icons/src/components/BaseProviderIcon/BaseProviderIcon.vue
DamianMullins: If this isn't a URL type then you're going to have an empty `href` attribute on a span?

packages/f-vue-icons/src/components/BaseProviderIcon/BaseProviderIcon.vue
xander-marjoram: Should be solved now. If the icon element doesn't have a `url` attribute then it will just be rendered, if it does, then it'll be wrapped in an anchor.

packages/f-vue-icons/src/components/BaseProviderIcon/BaseProviderIcon.vue
xander-marjoram: I've not used `aria-label` for now, I've added `alt` as another prop to be passed through and stuck that on the svg element.

DamianMullins: I've just done some reading up on `alt` attributes on SVGs and the recommendation is to use the `<title>` element. I don't think `alt` on embedded SVGs has any meaning.

DamianMullins:
One more question, what does base provider mean?

packages/f-vue-icons/src/components/BaseProviderIcon/BaseProviderIcon.vue
dandel10n: I think Damian is right. Here we just need to render svg, without link wrapper. And when we use it in the project we will wrap it into a link if needed. And `alt` tag won't work on svg level, svgs have `title` instead. But we will add it to a wrapper on a project level, I think

xander-marjoram:
@DamianMullins Ash asked me to come up with a less generic name than `BaseIcon` to cover payments and social icons, and exclude the other types of icon (like chevrons, etc.) to reduce confusion. In other repos we've called them "social providers" and "payment providers" so that was the best I could come up with.

packages/f-vue-icons/src/components/BaseProviderIcon/BaseProviderIcon.vue
xander-marjoram: @dandel10n @DamianMullins ✅

dandel10n (approved):

kevinrodrigues (approved):

PR #19, opened 2019-07-17
f-vue-icons@0.11.0 - Pass locale into AppStoreIcon

xander-marjoram:
And fix icon component names for Danish and Norwegian

### DIFF
--- a/packages/f-vue-icons/src/components/AppStoreIcon/AppStoreIcon.vue
+++ b/packages/f-vue-icons/src/components/AppStoreIcon/AppStoreIcon.vue
@@ -3,31 +3,43 @@
 </template>
 
 <script>
-import AndroidIconDa from '@justeat/f-icons/src/img/icons/appstore/android.dk.svg';
-import AndroidIconEn from '@justeat/f-icons/src/img/icons/appstore/android.svg';
-import AndroidIconEs from '@justeat/f-icons/src/img/icons/appstore/android.es.svg';
-import AndroidIconIt from '@justeat/f-icons/src/img/icons/appstore/android.it.svg';
-import AndroidIconNb from '@justeat/f-icons/src/img/icons/appstore/android.no.svg';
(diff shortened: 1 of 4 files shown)


dandel10n (approved):

packages/f-vue-icons/src/components/AppStoreIcon/AppStoreIcon.vue
```diff
@@ -3,45 +3,49 @@
...
+            required: false,
+            default: 'en-GB'
         }
     },
     computed: {
         iconComponent () {
             return `${this.name}-icon-${this.language}`;
+        },
+        language () {
+            return this.locale.split('-')[0];
```
DamianMullins: Do you not want the second part of the locale? What happens now if you want a different icon for Ireland (`en-IE`)?

xander-marjoram: Yeah I see your point. I've added imports for each of the countries, hopefully once we're tree-shaking it shouldn't be a problem.

packages/f-vue-icons/src/components/AppStoreIcon/AppStoreIcon.vue
```diff
@@ -3,45 +3,58 @@
...
             type: String,
-            default: 'en'
+            required: false,
+            default: 'en-GB'
         }
     },
     computed: {
         iconComponent () {
-            return `${this.name}-icon-${this.language}`;
+            return `${this.name}-icon-${this.locale.toLowerCase()}`;
```
DamianMullins, line 57: Doesn't look like the values generated here match the imported component names above? So for the `IosIconNbNo` component this would generate `IosIcon-nbno `.

Unless I'm missing something?

xander-marjoram: @DamianMullins It should come back as `ios-icon-nb-no`

DamianMullins (approved):

PR #20, opened 2019-07-18
f-footer@2.0.0-beta.6 - Use f-vue-icons

xander-marjoram:
### Added
- `yarn demo` command for easier prototyping
- Dependency on `f-vue-icons`
- Make use of `AppStoreIcon`, `BaseProviderIcon` and `ChevronIcon` from `f-vue-icons`
- Generic `IconList` component
- Component styles
- Extend tests for `LinkList` component

 ### Changed
- Remove `Footer` prefix from most components
- Remove unused fields from resource files and add locale


packages/f-footer/src/components/LinkList.vue
```diff
@@ -94,27 +158,9 @@ export default {
...
-            content: none;
-        }
-    }
-
-    @include media('>=wide') {
-        margin: spacing(x2) 0 0 0;
-    }
+.c-icon--chevron--small {
+    width: 16px;
+    display: inline-block;
```
dandel10n: Do we need it to be `inline-block` here?

packages/f-footer/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,19 @@
...
+</template>
+
+<script>
+export default {
+
+};
+</script>
+
+<style lang="scss">
+.c-footer-country-selector {
```
ashleynolan: Should be `c-footer-countrySelector`

packages/f-footer/src/components/FeedbackBlock.vue
```diff
@@ -29,6 +31,12 @@ export default {
 };
 </script>
 
-<style>
+<style lang="scss">
+
+.c-footer-feedback-block {
```
ashleynolan: Same as above – `c-footer-feedbackBlock`

packages/f-footer/src/components/Footer.vue
```diff
@@ -1,57 +1,57 @@
 <template>
     <footer class="c-footer">
-        <div class="c-footer-container c-footer-row c-footer-linkList">
+        <div class="c-footer-container c-footer-row c-footer-row--no-pad-below-wide">
```
ashleynolan: Dashes should be camel case (dashes are to signify child element only).

So `c-footer-row--noPadBelowWide`

packages/f-footer/src/components/Footer.vue
```diff
@@ -1,57 +1,57 @@
...
                 :key="index"
                 :link-list="linkList" />
         </div>
-        <div class="c-footer-row c-footer-row--shaded">
-            <div class="c-footer-container c-footer-row">
-                <footer-app-list
+        <div class="c-footer-row--light">
+            <div class="c-footer-row c-footer-container">
+                <icon-list
+                    class="c-icon-list c-icon-list--apps"
```
ashleynolan: `c-iconList` and `c-iconList--apps`

packages/f-footer/src/components/Footer.vue
```diff
@@ -114,41 +114,86 @@ export default {
...
     @include font-size(mid);
     padding: spacing(x2);
+    padding-left: 0;
 
-    @include media('>=wide') {
-        padding-left: 0;
-        padding-top: 0;
+}
+
+.c-footer-heading--short-below-wide {
```
ashleynolan: `c-footer-heading--shortBelowWide`

packages/f-footer/src/components/Footer.vue
```diff
@@ -114,41 +114,86 @@ export default {
...
+        flex-basis: 42%;
+    }
+
+    .c-icon-list--social {
+        flex-basis: 25%;
     }
 }
 
-.c-footer-row--shaded {
+.c-footer-row--light {
```
ashleynolan: It doesn't look like the default `c-footer-row` has a background, so makes sense for this to be called `shaded` rather than light? (as white is lighter than `$footer-bgLight` technically)

packages/f-footer/src/components/IconList.vue
```diff
@@ -0,0 +1,86 @@
...
+            type: String,
+            required: false,
+            default: 'en-GB'
+        }
+    }
+};
+</script>
+
+<style lang="scss">
+.c-icon-list {
```
ashleynolan: `c-iconList`

packages/f-footer/src/components/LinkList.vue
```diff
@@ -1,16 +1,31 @@
...
             data-js-test="linkList-header"
+            :tabindex="isBelowWide ? 0 : -1"
+            :disabled="!isBelowWide"
+            :aria-disabled="!isBelowWide"
+            :aria-expanded="!panelCollapsed ? 'true' : 'false'"
+            :aria-controls="listId"
             @click="onPanelClick">
-            {{ linkList.title }}
-        </h2>
+            <h2>
```
ashleynolan: Is it semantically valid to put heading tags inside a button? (I'm not sure myself, but doesn't look quite right)

packages/f-footer/src/components/LinkList.vue
```diff
@@ -54,24 +88,54 @@ export default {
 
 .c-footer-panel {
     flex: 1 0 0;
+
+    .c-icon--chevron {
+        display: none;
+    }
+
+    &.is-collapsed {
```
ashleynolan: This could probably just be c-footer-panel `.is-collapsed .c-footer-list` (as it would lower the specificity of the selector)

packages/f-footer/src/components/LinkList.vue
```diff
@@ -54,24 +88,54 @@ export default {
...
         &:last-of-type {
             border-bottom: none;
         }
 
-        &.is-collapsed {
-            .c-footer-list {
-                display: none;
-            }
+        .c-icon--chevron {
+            display: block;
```
ashleynolan: Better to put these media queries inline with the selector (so this would be on line 93 with the other chevron styles). Sass compresses the media queries into one statement later so it's better to make these more verbose here so that it's clear what states each selector has.

ashleynolan (changes requested):
Few naming and styling comments

packages/f-vue-icons/src/components/AppStoreIcon/AppStoreIcon.vue
```diff
@@ -3,6 +3,7 @@
 </template>
 
 <script>
+/* eslint-disable import/no-duplicates */
 import AndroidIconDaDk from '@justeat/f-icons/src/img/icons/appstore/android.dk.svg';
 import AndroidIconEnGb from '@justeat/f-icons/src/img/icons/appstore/android.svg';
```
ashleynolan: Would it not be worth only importing this once, and then using that import variable in the export:

```js
import AndroidIconEnGb from '@justeat/f-icons/src/img/icons/appstore/android.svg';

export default {
    components: {
        AndroidIconDaDk,
        AndroidIconEnGb,
        AndroidIconEnIe: AndroidIconEnGb,
        AndroidIconEnAu: AndroidIconEnGb,
        …
    }
}
```

Same for the ios SVGs?

packages/f-footer/src/components/tests/LinkList.test.js
```diff
@@ -16,28 +16,51 @@ const wrapper = shallowMount(LinkList, {
...
+
 describe('LinkList component', () => {
     it('should be defined', () => {
         expect(wrapper).toBeDefined();
     });
 
-    it('should contain the class "is-collapsed"', () => {
-        // Arrange & Act
-        const linkListWrapper = wrapper.find('[data-js-test="linkList-wrapper"]');
+    describe('for mobile widths', () => {
```
ashleynolan: Best to keep these tests referencing viewport width rather than device.

`for narrow viewport widths`

packages/f-footer/src/components/tests/LinkList.test.js
```diff
@@ -16,28 +16,51 @@ const wrapper = shallowMount(LinkList, {
...
+        beforeEach(() => {
+            window.innerWidth = 414;
+            window.dispatchEvent(new Event('resize'));
+        });
+
+        it('should contain the class "is-collapsed"', () => {
```
ashleynolan: Would make more sense (to me) to reference the behaviour rather than the classname. So:

`should be in a collapsed state`

packages/f-footer/src/components/tests/LinkList.test.js
```diff
@@ -16,28 +16,51 @@ const wrapper = shallowMount(LinkList, {
...
+            // Arrange & Act
+            const linkListWrapper = wrapper.find('[data-js-test="linkList-wrapper"]');
+
+            // Assert
+            expect(linkListWrapper.classes()).toContain('is-collapsed');
+        });
+
+        it('shouldn\'t contain class "is-collapsed" when linkList title is clicked', () => {
```
ashleynolan: `should be in an open state (not collapsed) when linkList title is clicked`

packages/f-footer/src/components/tests/LinkList.test.js
```diff
@@ -16,28 +16,51 @@ const wrapper = shallowMount(LinkList, {
...
+            // Assert
+            expect(linkListWrapper.classes()).not.toContain('is-collapsed');
+        });
     });
 
-    it('shouldn\'t contain class "is-collapsed" when linkList title is clicked', () => {
-        // Arrange
-        const linkListWrapper = wrapper.find('[data-js-test="linkList-wrapper"]');
-        const linkListHeader = wrapper.find('[data-js-test="linkList-header"]');
+    describe('for desktop widths', () => {
```
ashleynolan: `for wide viewport widths`

packages/f-footer/src/components/tests/LinkList.test.js
```diff
@@ -16,28 +16,51 @@ const wrapper = shallowMount(LinkList, {
...
+        beforeEach(() => {
+            window.innerWidth = 1200;
+            window.dispatchEvent(new Event('resize'));
+        });
 
-        // Act
-        linkListHeader.trigger('click');
+        it('should not contain the class "is-collapsed"', () => {
```
ashleynolan: `Should not be in a collapsed state`

packages/f-vue-icons/src/components/AppStoreIcon/AppStoreIcon.vue
xander-marjoram: Didn't realise that was possible, I'll give it a go

packages/f-footer/src/components/LinkList.vue
xander-marjoram: You're right, it's not valid. We'll try switching it around.

packages/f-footer/src/components/Footer.vue
```diff
@@ -1,57 +1,57 @@
 <template>
     <footer class="c-footer">
-        <div class="c-footer-container c-footer-row c-footer-linkList">
+        <div class="c-footer-container c-footer-row c-footer-row--noPadBelowWide">
```
ashleynolan, line 3: Are the `c-footer-container` and `c-footer-row` classes always used together (as it looks like they are here?

Seems like they have some styles that conflict with one another (padding for instance). If they can be merged into one `c-footer-container` class, that might make sense?

packages/f-footer/src/components/Footer.vue
```diff
@@ -114,41 +118,78 @@ export default {
...
     }
 }
 
 .c-footer-row {
     padding: 0;
     display: flex;
-    flex-flow: row nowrap;
+    flex-flow: column nowrap;
     justify-content: space-between;
+    padding: spacing(x2);
```
ashleynolan, line 123: Looks like a duplicate padding definition to that on line 132.

packages/f-vue-icons/src/components/AppStoreIcon/AppStoreIcon.vue
```diff
@@ -3,6 +3,7 @@
 </template>
 
 <script>
+/* eslint-disable import/no-duplicates */
```
ashleynolan: Can remove this now as this shouldn't fail the linting now after the import changes.

packages/f-footer/src/components/Footer.vue
dandel10n: I've removed padding from the container class, but I think it is better to have separate `f-footer-row` and `c-footer-container` class. Container is responsible for max-width of the component and row has rows specific styles.

packages/f-footer/src/components/Footer.vue
dandel10n: ✅

packages/f-footer/src/components/IconList.vue
```diff
@@ -0,0 +1,113 @@
...
+            <li
+                v-for="icon in icons"
+                :key="icon.key"
+                class="c-footer-listItem">
+                <a
+                    v-if="icon.url"
+                    :href="icon.url"
+                    :title="icon.alt">
+                    <component
+                        :is="isApps ? 'app-store-icon' : 'base-provider-icon'"
```
DamianMullins: As this logic is duplicated in two places it would be better to pull it out into a computed property and reference that here instead.

packages/f-footer/src/components/LinkList.vue
```diff
@@ -34,17 +53,32 @@ export default {
...
+        listHeadingId () {
+            return `${this.listId}-heading`;
+        },
+        isBelowWide () {
+            return this.currentScreenWidth <= 1024;
+        }
+    },
+    mounted () {
+        this.currentScreenWidth = window.innerWidth;
+        window.addEventListener('resize', () => {
```
DamianMullins: Should probably be throttling this, [have a look here for an example](https://github.je-labs.com/menu/menuweb/blob/87551c5c11f01d6afd7e473bd127cefd9539b45c/src/MenuWeb/Assets/vue/components/nav/CategoryNav.vue#L86-L103).

DamianMullins (approved):

packages/f-footer/src/components/IconList.vue
```diff
@@ -0,0 +1,118 @@
...
+    computed: {
+        iconChoice () {
+            return this.isApps ? 'app-store-icon' : 'base-provider-icon';
+        }
+    }
+};
+</script>
+
+<style lang="scss">
+.c-iconList {
```
ashleynolan, line 74: More of a question than anything else – where does this class (`c-iconList`) actually get applied to the markup?

ashleynolan (approved):

packages/f-footer/src/components/IconList.vue
dandel10n: this class is added to the "icon-list" component call in Footer.vue like this:
```
<icon-list
      :title="copy.downloadOurApps"
      :icons="copy.appStoreIcons"
      :is-apps="true"
      :locale="copy.locale"
      class="c-iconList c-iconList--apps" />
```

ashleynolan: Should `c-iconList` not just be a default class for the `iconList` component? Makes sense to extend it with modifier's for the apps etc though

PR #24, opened 2019-07-29
f-footer@2.0.0-beta.7 - Country selector component

xander-marjoram:
### Added
- `CountrySelector` component logic and styles
- `CountrySelector` tests

### DIFF
--- a/packages/f-footer/src/components/CountrySelector.vue
+++ b/packages/f-footer/src/components/CountrySelector.vue
@@ -72,11 +72,10 @@ export default {
             this.showCountryList = !this.showCountryList;
         },
         hideCountryList () {
-            if (this.showCountryList) {
-                this.toggleCountryList();
-            }
-            var button = document.querySelector('#countrySelector-button');
-            button.focus();
+            this.showCountryList = false;
+
+            document.querySelector('#countrySelector-button')
+                .focus();
         }
     }
 };
--- a/packages/f-footer/src/components/CountrySelector.vue
+++ b/packages/f-footer/src/components/CountrySelector.vue
@@ -7,7 +7,9 @@
             class="c-countrySelector-link c-countrySelector-button"
             type="button"
             @click="toggleCountryList"
-            :aria-expanded="showCountryList ? 'true' : 'false'">
+            :aria-expanded="showCountryList ? 'true' : 'false'"
+            :aria-label="changeCountryText"
+            aria-controls="countrySelector-countries">
             <flag-icon 
                 :country-code="currentCountryKey" />
             {{ currentCountryName }}
@@ -18,7 +20,9 @@
         </button>
         <ul
             v-show="showCountryList"
-            class="c-footer-countrySelector-list">
+            id="countrySelector-countries"
+            class="c-footer-countrySelector-list"
+            role="region">
             <li
                 v-for="country in countries"
                 :key="country.key">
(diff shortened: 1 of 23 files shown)


packages/f-footer/src/components/LinkList.vue
```diff
@@ -151,8 +151,7 @@ export default {
             width: auto;
         }
 
-        &:hover,
-        &:focus {
```
dandel10n, line 155: Why you removed focus from here?

packages/f-footer/src/components/CountrySelector.vue
```diff
@@ -1,19 +1,151 @@
...
-        <p>Country selector component</p>
+    <div 
+        class="c-footer-countrySelector"
+        @keyup.esc="hideCountryList">
+        <button
+            id="countrySelector-button"
+            class="c-countrySelector-link c-countrySelector-button"
+            type="button"
+            @click="toggleCountryList"
+            :aria-expanded="showCountryList ? 'true' : 'false'">
```
dandel10n: I would recommend to add `:aria-controls` attribute to show relationship of the button with the expandable region. The list of flags will need an id.

packages/f-footer/src/components/CountrySelector.vue
```diff
@@ -1,19 +1,151 @@
...
+            <flag-icon 
+                :country-code="currentCountryKey" />
+            {{ currentCountryName }}
+            <chevron-icon
+                v-if="!showCountryList"
+                :is-small="true" />
+            <cross-icon v-else />
+        </button>
+        <ul
```
dandel10n, line 23: and `:aria-labelledby="countrySelector-button"` and `role="region"` here

dandel10n:
Will jest tests come in separate PR for this component?

packages/f-footer/src/components/LinkList.vue
xander-marjoram: On focus, we already have the outline, we don't need the underline as well. This matches what we currently have in production (although the outline colour is different)

packages/f-footer/src/components/CountrySelector.vue
```diff
@@ -1,19 +1,150 @@
 <template>
-    <div class="c-footer-countrySelector">
-        <p>Country selector component</p>
+    <div 
+        class="c-footer-countrySelector"
```
ashleynolan: Could probably just call this `c-countrySelector`?  There's likely no dependency on it being a part of the footer and that way we could pull it out of the footer later if we ever wanted to.

packages/f-footer/src/components/CountrySelector.vue
```diff
@@ -1,19 +1,150 @@
...
+    padding: spacing(x2);
+    display: flex;
+    align-items: center;
+    justify-content: space-around;
+    background-color: $grey--lightest;
+    border: none;
+    cursor: pointer;
+}
+
+.c-footer-countrySelector-list {
```
ashleynolan: `c-countrySelector-list`?

ashleynolan (changes requested):
Couple of small updates from me

ashleynolan:
@xander-marjoram Needs a rebase and update to the description.

packages/f-footer/src/components/CountrySelector.vue
dandel10n: ✅

ashleynolan (approved):

dandel10n (approved):

PR #27, opened 2019-07-31
f-footer@2.0.0-beta.8 ButtonList component

dandel10n:
### Added
- `ButtonList` component
- Small styles fixes for IE11
- Some css global variables for MenuLog styles
- `theme` mixin for picking up MenuLog styles
- `@vue/cli-plugin-babel` to the package dev dependencies, it wasn't picked up from the root monorepo level

### Changed
- Restructured AU resource file for new button list block + added empty `linkButtonList` array for other tenants

### DIFF
--- a/packages/f-footer/src/assets/scss/common.scss
+++ b/packages/f-footer/src/assets/scss/common.scss
@@ -21,6 +21,20 @@ $theme: 'je' !default;
  $footer-chevronColor    : $grey--dark !default;
  $footer-socialIconColor : $grey--dark !default;
 
+ /**
+ * Global variables which need to be changed depending on the theme (TODO: move to separate component)
+ */
+ $font-family-headings       : 'Ubuntu', sans-serif;
+ $font-family-headings--ml   : 'Aspira', 'Times New Roman', Times, serif;
+
+ $font-weight-headings       : 300; // instead of browser default, bold
+ $font-weight-headings--ml   : 800; // Heading style is Aspira Heavy, which is 800 font-weight
+
+ $color-link-default         : $blue;
+ $color-link-default--ml     : $green;
+
+
+
 @mixin theme($type) {
     [data-theme="#{$type}"] & {
         @content;
--- a/packages/f-footer/src/assets/scss/common.scss
+++ b/packages/f-footer/src/assets/scss/common.scss
@@ -24,13 +24,8 @@ $theme: 'je' !default;
  /**
  * Global variables which need to be changed depending on the theme (TODO: move to separate component)
  */
- $font-family-headings       : 'Ubuntu', sans-serif;
  $font-family-headings--ml   : 'Aspira', 'Times New Roman', Times, serif;
-
- $font-weight-headings       : 300; // instead of browser default, bold
- $font-weight-headings--ml   : 800; // Heading style is Aspira Heavy, which is 800 font-weight
-
- $color-link-default         : $blue;
+ $font-weight-headings--ml   : 800;
  $color-link-default--ml     : $green;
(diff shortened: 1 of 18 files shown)


packages/f-footer/src/assets/scss/common.scss
```diff
@@ -21,6 +21,20 @@ $theme: 'je' !default;
  $footer-chevronColor    : $grey--dark !default;
  $footer-socialIconColor : $grey--dark !default;
 
+ /**
+ * Global variables which need to be changed depending on the theme (TODO: move to separate component)
+ */
+ $font-family-headings       : 'Ubuntu', sans-serif;
```
ashleynolan: This variable should already be available as part of fozzie's variables that you include? [here](https://github.com/justeat/fozzie/blob/master/src/scss/settings/_variables.scss#L60-L62)

packages/f-footer/src/assets/scss/common.scss
```diff
@@ -21,6 +21,20 @@ $theme: 'je' !default;
...
+ $font-family-headings--ml   : 'Aspira', 'Times New Roman', Times, serif;
+
+ $font-weight-headings       : 300; // instead of browser default, bold
```
ashleynolan: Same for this. Should only be the menulog override variables that you need to specify – the others should already exist.

packages/f-footer/src/components/ButtonList.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+    <div>
+        <h2 class="c-footer-heading">
+            {{ buttonList.title }}
+        </h2>
+        <div class="c-buttonsList">
```
ashleynolan: Would call this `c-buttonList` to match the JS variable on line 8.

ashleynolan (commented):
Couple of small comments

packages/f-footer/src/assets/scss/common.scss
dandel10n: @ashleynolan Yes, I decided to add it for better visibility of what we have for `je` theme and what we rewrite for `ml`. Happy to remove if it makes more sense not to duplicate them here

ashleynolan: Yeh, should be ok not to duplicate. Most (if not all) of the current fozzie variables will likely get pulled over into the new module when it's created anyway.

packages/f-footer/src/components/tests/__snapshots__/Footer.test.js.snap
```diff
@@ -1,7 +1,7 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Footer should render default component markup 1`] = `
-<footer class="c-footer">
+<footer data-theme="je" class="c-footer">
```
ashleynolan, line 4: Is it worth adding a snapshot to test the `ml` theme?

ashleynolan (approved):
Approved – could add a test to check the theme method produces the expected snapshot?

DamianMullins (approved):

packages/f-footer/src/components/tests/__snapshots__/Footer.test.js.snap
dandel10n: I am not sure about it. With the theme there are changes in css, all html markup stays the same.

ashleynolan: The data-attribute should change to `ml` though in the markup? As long as the method is tested, that should be enough tbh.

PR #28, opened 2019-08-02
f-footer@2.0.0-beta.9 legal field component + some styles fixes

dandel10n:
### Added
- `LegalField` component
- new metaLegalField object in resource files

### Changed
- Some styles fixes for footer rows

Example with Confianza icon for ES:
Screen Shot 2019-08-02 at 16 14 02
Screen Shot 2019-08-02 at 16 13 48
Screen Shot 2019-08-02 at 16 13 39

Example with VAT icon for IT:
Screen Shot 2019-08-02 at 16 12 41
Screen Shot 2019-08-02 at 16 12 52
Screen Shot 2019-08-02 at 16 13 23


packages/f-footer/src/components/Footer.vue
```diff
@@ -89,6 +89,11 @@ export default {
             theme
         };
     },
+    computed: {
+        metaLegalFieldEnabled () {
+            return Object.keys(this.copy.metaLegalField).length !== 0;
```
xander-marjoram: I think it might be slightly more readable to have `> 0` instead of `!== 0`

dandel10n: ✅

packages/f-footer/src/components/LegalField.vue
```diff
@@ -0,0 +1,71 @@
...
+            return `c-footer-certificates-icons--${this.info.icon.name}`;
+        },
+        iconUrl () {
+            return this.getImgUrl(this.info.icon.name);
+        }
+    },
+    methods: {
+        getImgUrl (name) {
+            // eslint-disable-next-line global-require, import/no-dynamic-require
+            return require(`../assets/img/${name}.png`);
```
DamianMullins, line 35: How is this going to work with regards to loading images from a CDN? In Menu Web we load _all_ assets from CDN to reduce some of the load on the web servers.

Also, any reason you didn't use dynamic imports here?

packages/f-footer/CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v2.0.0-beta.9
+------------------------------
+*August 02, 2019*
```
ashleynolan: Needs a bump

packages/f-footer/src/components/Footer.vue
```diff
@@ -172,6 +177,19 @@ export default {
     }
 }
 
+.c-footer-row--combined {
```
ashleynolan, line 183: Would be good to have a comment as to what this modifier does (in terms of how it affects the look display of the footer row).

packages/f-footer/src/components/Footer.vue
```diff
@@ -172,6 +177,19 @@ export default {
...
+    flex-flow: row wrap;
+
+    @include media('>=wide') {
+        flex-flow: row nowrap;
+    }
+}
+
+.c-footer-row--doubledBottomPad {
```
ashleynolan: I'd make this less specific in-case the padding amounts were to ever change (so that then the name isn't tied to being `double`.

Also – where do the `c-footer-row` styles get defined? It may make sense to define these next to those styles as they are potentially usable by any `c-footer-row`.

ashleynolan (commented):
Just a few comments from me

packages/f-footer/src/components/LegalField.vue
dandel10n: @DamianMullins at the moment we have only one image in the assets folder for ES specific footer. We don't use vuex in footer and I am not sure if it will be beneficial to add it for maintaining CDN path for one asset now. 

About dynamic imports, is it possible to import assets dynamically with name like `${name}.png`?

packages/f-footer/src/components/Footer.vue
dandel10n: Changed `.c-footer-row--doubledBottomPad` to `.c-footer-row--notEqualTopAndBottomPad`

c-footer-row lives on line 160 in the same file :)

packages/f-footer/src/components/Footer.vue
dandel10n: ✅

xander-marjoram (approved):

markedup (approved):

PR #30, opened 2019-08-05
f-footer@2.0.0-beta.10 feedback block styles and data-attr for gtm

dandel10n:
### Added
- Styles and `data-gtm-feedback` attr for feedback component

### DIFF
--- a/packages/f-footer/src/components/FeedbackBlock.vue
+++ b/packages/f-footer/src/components/FeedbackBlock.vue
@@ -1,12 +1,12 @@
 <template>
-    <div class="c-feedbackBlock" data-gtm-feedback>
+    <div class="c-feedback" data-gtm-feedback>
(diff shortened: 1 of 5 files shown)


packages/f-footer/src/components/FeedbackBlock.vue
```diff
@@ -1,12 +1,12 @@
 <template>
-    <div class="c-footer-feedbackBlock">
+    <div class="c-feedbackBlock" data-gtm-feedback>
```
ashleynolan: Would probably just call this component `c-feedback` (could probably append the word `Block` to a lot of our components otherwise, so don't think it's needed here).

dandel10n: :+1: Good point

markedup (approved):

xander-marjoram (approved):

PR #31, opened 2019-08-05
f-footer@2.0.0-beta.11 data-gtm attr for all the links

dandel10n:
### Added
- `data-trak` attributes for all the footer links
- `gtm` data in resource files for all the tenants
- `@justeat/f-trak` as a peer dependency

### Changed
- Some AU resource links according to the latest changes in live version of the footer

### DIFF
--- a/packages/f-footer/CHANGELOG.md
+++ b/packages/f-footer/CHANGELOG.md
@@ -9,8 +9,9 @@ v2.0.0-beta.11
 *August 6, 2019*
 
 ### Added
-- `data-gtm` attributes for all the footer links
+- `data-trak` attributes for all the footer links
 - `gtm` data in resource files for all the tenants
+- `@justeat/f-trak` as a peer dependency
 
 ### Changed
 - Some AU resource links according to the latest changes in live version of the footer 
--- a/packages/f-footer/CHANGELOG.md
+++ b/packages/f-footer/CHANGELOG.md
@@ -4,6 +4,18 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v2.0.0-beta.11
+------------------------------
+*August 6, 2019*
+
+### Added
+- `data-gtm` attributes for all the footer links
+- `gtm` data in resource files for all the tenants
+
+### Changed
+- Some AU resource links according to the latest changes in live version of the footer 
+
+
 v2.0.0-beta.10
 ------------------------------
 *August 6, 2019*
--- a/packages/f-footer/src/components/ButtonList.vue
+++ b/packages/f-footer/src/components/ButtonList.vue
@@ -9,7 +9,8 @@
                 :key="index"
                 :href="button.url"
                 target="_blank"
-                class="c-buttonList-button">
+                class="c-buttonList-button"
+                :data-gtm="['engagement|footer|' + button.gtm]">
                 {{ button.title }}
             </a>
         </div>
--- a/packages/f-footer/src/components/ButtonList.vue
+++ b/packages/f-footer/src/components/ButtonList.vue
@@ -10,7 +10,12 @@
                 :href="button.url"
                 target="_blank"
                 class="c-buttonList-button"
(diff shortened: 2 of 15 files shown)


markedup (approved):

packages/f-footer/src/components/ButtonList.vue
```diff
@@ -9,7 +9,8 @@
                 :key="index"
                 :href="button.url"
                 target="_blank"
-                class="c-buttonList-button">
+                class="c-buttonList-button"
+                :data-gtm="['engagement|footer|' + button.gtm]">
```
kevinrodrigues: [`engagement|footer|${button.gtm}`] would be better here.

packages/f-footer/src/components/IconList.vue
```diff
@@ -13,7 +13,8 @@
                 <a
                     v-if="icon.url"
                     :href="icon.url"
-                    :title="icon.alt">
+                    :title="icon.alt"
+                    :data-gtm="['engagement|footer|' + icon.gtm]">
```
kevinrodrigues: Similar to above.

packages/f-footer/src/components/LinkList.vue
```diff
@@ -30,7 +30,8 @@
                 :key="index">
                 <a
                     :href="link.url"
-                    :rel="link.rel">
+                    :rel="link.rel"
+                    :data-gtm="['engagement|footer|' + link.gtm]">
```
kevinrodrigues: ^

packages/f-footer/src/components/FeedbackBlock.vue
```diff
@@ -1,7 +1,8 @@
 <template>
     <div
         class="c-feedback"
-        data-gtm-feedback>
+        data-gtm-feedback
+        data-gtm="engagement|footer|click_feedback">
```
ashleynolan: I think we should be using `f-trak` to fire the dataLayer calls rather than `data-gtm` attributes for these?  It works in a similar way to this – just has a different data-attribute hook.

packages/f-footer/src/components/ButtonList.vue
dandel10n: Doesn't work this way...

kevinrodrigues: How comes? did you include the template strings: `` ?

Screen Shot 2019-08-06 at 10 26 20 AM

packages/f-footer/src/components/FeedbackBlock.vue
dandel10n: ✅

packages/f-footer/src/components/ButtonList.vue
dandel10n: Changed to use `data-track` instead

packages/f-footer/src/components/CountrySelector.vue
```diff
@@ -34,6 +34,12 @@
                     <a
                         class="c-countrySelector-link"
                         data-js-test="countrySelector-countryLink"
+                        :data-trak="`{
+                            event: 'click'
+                            category: 'engagement'
+                            action: 'footer'
+                            label: '${country.gtm}'
```
DamianMullins, line 41: Don't need the template string here just `label: country.gtm`.

DamianMullins: Just realised these are inside strings 🤦‍♂

DamianMullins (approved):

packages/f-footer/CHANGELOG.md
```diff
@@ -4,6 +4,19 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
 
 
+v2.0.0-beta.11
+------------------------------
+*August 6, 2019*
+
+### Added
+- `data-gtm` attributes for all the footer links
+- `gtm` data in resource files for all the tenants
+- `@justeat-f-trak` as a peer dependency
```
xander-marjoram: Nit: `@justeat/f-trak`

dandel10n: ✅

kevinrodrigues (approved):

PR #39, opened 2019-08-30
f-footer@2.0.0-beta.15 demo f-vue-icons package

dandel10n:
### Added
- `AppStoreIcon`, `BaseProviderIcon` and `FlagIcon` components

### Changed
- Replace f-vue-icons package with it's demo version
- Added Readme

### DIFF
--- a/packages/f-footer/src/components/FlagIcon.vue
+++ b/packages/f-footer/src/components/FlagIcon.vue
@@ -1,0 +1,48 @@
+<template>
+    <component :is="iconComponent" />
+</template>
+
+<script>
+import {
+    FlagAuIcon as FlagIconAu,
+    FlagBrIcon as FlagIconBr,
+    FlagCaIcon as FlagIconCa,
+    FlagChIcon as FlagIconCh,
+    FlagDkIcon as FlagIconDk,
+    FlagEsIcon as FlagIconEs,
+    FlagFrIcon as FlagIconFr,
+    FlagGbIcon as FlagIconGb,
+    FlagIeIcon as FlagIconIe,
+    FlagItIcon as FlagIconIt,
+    FlagNoIcon as FlagIconNo,
+    FlagNzIcon as FlagIconNz
+} from '@justeat/f-vue-icons';
+
+export default {
+    components: {
+        FlagIconAu,
+        FlagIconBr,
+        FlagIconCa,
+        FlagIconCh,
+        FlagIconDk,
+        FlagIconEs,
+        FlagIconFr,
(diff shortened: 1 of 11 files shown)


packages/f-footer/src/components/FlagIcon.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+    <component :is="iconComponent" />
+</template>
+
+<script>
+import {
+    FlagAuIcon as FlagIconAu,
```
DamianMullins, line 7: Was there a reason this format was used in f-vue-icons? The format you've used makes more sense to me?

packages/f-footer/src/components/tests/CountrySelector.test.js
```diff
@@ -61,22 +60,4 @@ describe('CountrySelector', () => {
         const link = wrapper.find('[data-js-test="countrySelector-countryLink"]');
         expect(link.attributes('href')).toBe(siteUrl);
     });
-
-    it('chevron icon should change to cross when clicked', () => {
```
DamianMullins, line 65: Does the icon no longer change to a cross?

simonsandersje (approved):

packages/f-footer/src/components/FlagIcon.vue
dandel10n: In beta f-icons component icon names build as file name + "Icon", so for `flagAU.svg` we receive `FlagAuIcon.js`. We can rewrite it later if needed, for now I just used f-icons as it is and changed the name here just for saving time with releasing footer.

packages/f-footer/src/components/tests/CountrySelector.test.js
dandel10n: It still changes, just with new `f-vue-icons` test started to fail. but manual test showed no issues. Had no time to properly had a look and fix it :(

packages/f-footer/src/components/FlagIcon.vue
DamianMullins: OK, let's get this out and refactor later.

DamianMullins (approved):

packages/f-footer/src/components/FlagIcon.vue
ashleynolan: Yeh, this is because the icon names are autogenerated to always have Icon on the name of them in f-icons in the build script. We could try to change this functionality if there was a country code present potentially, may take a bit of effort to do though.

Issue #42, opened 2019-09-04
Imported f-footer CSS overrides project CSS


DamianMullins:
When importing the f-footer css into a project classes they can override any existing styles, such as [the c-icon--chevron class in the `LinkList.vue` component](https://github.com/justeat/fozzie-components/blob/a067f18a33d5a956c470403befe20a1ea5767e1e/packages/f-footer/src/components/LinkList.vue#L170-L173), which overrides the style set within Fozzie and causes _all_ chevron icons to become 16x9px.

Could we look at using CSS modules in our shared components in order to prevent styles from affecting others around them? CSS modules can be enabled by adding the `module` attribute to the `style` tag (`<style module>`).

I think we've briefly discussed scoped CSS vs CSS modules in the past and were leaning more towards CSS modules as the styles don't allow styles to leak down to child components. 

[This blog post](https://www.netguru.com/codestories/vue.js-scoped-styles-vs-css-modules) describes the two approaches really well.

I can put together a PR to test with if that's OK?

dandel10n:
@DamianMullins modules solution sounds great :+1:

DamianMullins:
I'll have a look at that this afternoon.

PR #43, opened 2019-09-05
f-header@2.0.0-beta.1 logo component + tenants resource files

dandel10n:
### Added
- Logo component
- Tenants resource files
- Vue.config.js for rewriting default vue cli webpack config
- common.scss base styles imports
- theme and locale logic for Header component (later will be moved to independant package because it repeats in footer and searchbox already)

### DIFF
--- a/packages/f-header/src/assets/scss/common.scss
+++ b/packages/f-header/src/assets/scss/common.scss
@@ -1,0 +1,50 @@
+// Set $theme variable so that fozzie doesn't complain when vars are pulled in
+// n.b. This isn't used currently (unless there are theme specific vars from fozzie)
+// The theme is handled via a data-attribute on the component
+$theme: 'je' !default;
+
+@import '~@justeat/f-utils/src/scss/functions/index';
+@import '~@justeat/f-utils/src/scss/mixins/index';
+@import '~include-media/dist/include-media'; // Cleaner media query declarations – http://include-media.com
+@import '~@justeat/fozzie/src/scss/settings/variables';
+@import '~@justeat/fozzie/src/scss/base/reset';
+@import '~@justeat/fozzie-colour-palette/src/scss/';
+@import '~@justeat/fozzie/src/scss/base/typography';
+
+/**
+ * Component Variables
+ */
+$header-logo-color                 : $brand--red;
+$header-logo-color--transparent    : $white;
+
+$header-height--narrow              : 56px;
+$header-button--height              : $header-height--narrow;
+$header-button--width               : 56px;
+$header-height                      : 76px;
+// $header-height--narrow--multiLang   : $header-height--narrow + $languageSwitcher-height;
+$header-bg                          : $white;
+$header-separator                   : 1px;
+$header-separator                   : 4px;
+$header-border-color                : $grey--lightest;
+$header-buttonIcon-color            : $blue;
+$header-buttonCount-bg              : $blue;
+$header-buttonCount-color           : $white;
+$header-buttonCount-borderColor     : $white;
+
+
+$header--transparent-gradient-color : $black;
+$header--transparent-gradient       : 115px;
+$header--transparent-opacity        : 0.7;
+
+/**
+ * Global variables which need to be changed depending on the theme (TODO: move to separate component)
+//  */
+
+$header-buttonIcon-color--ml: $green;
+$header-buttonCount-bg--ml: $green;
+
+@mixin theme($type) {
+    [data-theme="#{$type}"] & {
+        @content;
+    }
+}
--- a/packages/f-header/src/components/Demo.vue
+++ b/packages/f-header/src/components/Demo.vue
@@ -7,8 +7,8 @@
             name="viewport"
             content="width=device-width, initial-scale=1">
         <vue-header
-            :is-transparent=true
-            locale="en-AU" />
+            :is-transparent="true"
+            locale="en-GB" />
     </div>
 </template>
 
--- a/packages/f-header/src/components/Demo.vue
+++ b/packages/f-header/src/components/Demo.vue
@@ -7,7 +7,8 @@
             name="viewport"
             content="width=device-width, initial-scale=1">
         <vue-header
-            locale="en-GB" />
+            :is-transparent=true
+            locale="en-AU" />
     </div>
 </template>
 
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -4,7 +4,7 @@
         class="c-header">
         <logo
             :theme="theme"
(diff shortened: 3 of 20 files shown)


packages/f-header/src/components/Header.vue
```diff
@@ -1,15 +1,70 @@
...
+            default: ''
+        },
+        isTransparent: {
+            type: Boolean,
+            required: true
+        }
+    },
+    data () {
+        const locale = this.getLocale();
+        const localeConfig = tenantConfigs[locale];
```
markedup, line 34: Could this be set with the fallback inline here?
```
const localeConfig = tenantConfigs[locale] || 'en-GB';
```
…and therefore remove the necessity for the check on line 50?

BanksySan (approved):

packages/f-header/src/components/Logo.vue
```diff
@@ -0,0 +1,113 @@
...
+            return `${this.theme}-icon`;
+        },
+        linkAltText () {
+            return `Go to ${this.companyName} homepage`;
+        }
+    }
+};
+</script>
+
+<style lang="scss" scoped>
```
zlatinivanov-je: Does it make sense to have a separate SCSS file here?

packages/f-header/src/tenants/index.js
```diff
@@ -0,0 +1,21 @@
+import uk from './en-GB';
```
zlatinivanov-je, line 1: Is this file going to include all tenant translations in the component?

zlatinivanov-je (approved):
A couple of minor questions. But really happy to see work done on this... will try on my project tomorrow. 👍

packages/f-header/src/components/Logo.vue
dandel10n: It is more visible to have component's styles in the same file as component's logic.  Actually I had a plan to change it to module styles according to this [discussion](https://github.com/justeat/fozzie-components/issues/42#issuecomment-528323259)

packages/f-header/src/tenants/index.js
dandel10n: Yes, but they are only around 400 bytes each. I am open to any suggestions on how we can optimize it

dandel10n:
> A couple of minor questions. But really happy to see work done on this... will try on my project tomorrow. 👍

@zlatinivanov-je thanks! to early to try, because there is only logo in the header component at the moment, just started working on it yesterday. But hope to publish it very soon!

markedup (commented):
Minor query, otherwise looks good!

packages/f-header/src/components/Header.vue
dandel10n: `tenantConfigs[locale]` returns whole resource object for specific tenant, which then written to `copy` vue data, not just locale.

packages/f-header/src/components/Header.vue
```diff
@@ -1,15 +1,70 @@
 <template>
-    <header class="c-header">
-        I am Header
+    <header
+        :data-theme="theme"
+        class="c-header">
+        <logo
+            :theme="theme"
+            :is-transparent=isTransparent
```
DamianMullins: `isTransparent` should be in quotes.

packages/f-header/src/components/Demo.vue
```diff
@@ -7,6 +7,7 @@
             name="viewport"
             content="width=device-width, initial-scale=1">
         <vue-header
+            :is-transparent="true"
```
DamianMullins, line 10: Can just do `is-transparent` here.

dandel10n: Yes, but it is only demo file where you can change it to false for testing purposes. So just faster to change true to false

packages/f-header/src/components/Logo.vue
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+        }
+    }
+
+    .c-icon--je g {
+        fill: $header-logo-color;
+    }
+
+    .c-icon-je--transparentBckgrnd g {
```
DamianMullins: I think `c-icon-je--transparentBg` might be a better class name?

packages/f-header/src/components/tests/Header.test.js
```diff
@@ -3,15 +3,58 @@ import Header from '../Header.vue';
...
+        };
+        const wrapper = shallowMount(Header, { propsData });
 
         // Assert
         expect(wrapper).toMatchSnapshot();
     });
+
+
+    it('should render ml themed component if AU local passed', () => {
+        // Arrange & Act
```
DamianMullins: Not a big deal but these joined "Arrange & Act" sections can be split out into separate "Arrange" and "Act" sections.

packages/f-header/src/components/Logo.vue
dandel10n: ✅

packages/f-header/src/components/tests/Logo.test.js
```diff
@@ -0,0 +1,88 @@
+import { shallowMount } from '@vue/test-utils';
+import Logo from '../Logo.vue';
+// import {
```
DamianMullins: Can this comment block be removed?

dandel10n: ✅

DamianMullins (approved):

packages/f-header/src/assets/scss/common.scss
```diff
@@ -0,0 +1,50 @@
+// Set $theme variable so that fozzie doesn't complain when vars are pulled in
+// n.b. This isn't used currently (unless there are theme specific vars from fozzie)
+// The theme is handled via a data-attribute on the component
+$theme: 'je' !default;
+
+@import '~@justeat/f-utils/src/scss/functions/index';
```
ashleynolan, line 6: Late comment (just a small optimization) – you should just be able to import `f-utils` base `index.scss` here, as it includes the functions and mixins. Will save you 1 whole import line 😉

dandel10n: Will add this change in my next PR :+1:

dandel10n: It looks like `@import`s inside `@justeat/f-utils/src/scss/index` are not recognised on the projects level. Receive an error:

```
@import 'mixins';
^
File to import not found or unreadable: mixins.
```

ashleynolan: I think I know why this is – I think it expects it to be `mixins/` annoyingly or something like that (if I remember right).

Could change this locally in your `node_modules` and see if that's right – if it is, can change it in `f-utils`…

PR #44, opened 2019-09-09
f-footer@v2.0.0-beta.16 — Adding CSS Modules.

### LINKED ISSUE #42: Imported f-footer CSS overrides project CSS

DamianMullins:
This PR introduces CSS Modules to the footer so that we can avoid styles affecting parts of web pages the module is applied to.

For some background [have a read through this blog post](https://www.netguru.com/codestories/vue.js-scoped-styles-vs-css-modules) on using CSS scoped and CSS Modules in Vue projects and the positives and negatives of each.

Re-usable styles (footer & icons) have been extracted into their own SCSS files and are then imported into the `styles` section so that the correct generated class name can be added into the markup.

### Changed
- Using CSS modules to apply styles.
- Applied ESLint auto-fix to components.
- Moved legal field image import out of method and inline in the component.
- `v-for` loop keys more robust by updating the key names.

Resolves #42

### DIFF
--- a/packages/f-footer/src/components/Footer.vue
+++ b/packages/f-footer/src/components/Footer.vue
@@ -2,10 +2,11 @@
     <footer
         :data-theme="theme"
         class="c-footer">
-        <div class="c-footer-container c-footer-row c-footer-row--noPadBelowWide">
+        <div
+            class="c-footer-container c-footer-row c-footer-row--noPadBelowWide">
             <link-list
-                v-for="(linkList, index) in copy.linkLists"
-                :key="index"
+                v-for="(linkList, i) in copy.linkLists"
+                :key="i + '_ButtonList'"
                 :link-list="linkList" />
         </div>
 
@@ -15,40 +16,46 @@
                     v-if="copy.linkButtonList.length"
                     class="c-footer-row c-footer-row--noBottomPad">
                     <button-list
-                        v-for="(buttonList, index) in copy.linkButtonList"
-                        :key="index"
+                        v-for="(buttonList, i) in copy.linkButtonList"
+                        :key="i + '_ButtonList'"
                         :button-list="buttonList" />
                 </div>
+
                 <div class="c-footer-row">
                     <icon-list
                         :title="copy.downloadOurApps"
                         :icons="copy.appStoreIcons"
-                        :is-apps="true"
                         :locale="copy.locale"
-                        class="c-iconList c-iconList--apps" />
+                        is-apps />
+
                     <feedback-block
                         :title="copy.feedback"
                         :text="copy.improveOurWebsite"
                         :button-text="copy.sendFeedback" />
+
                     <icon-list
                         :icons="copy.socialIcons"
                         :title="copy.followUs"
-                        class="c-iconList c-iconList--social" />
+                        is-social />
                 </div>
             </div>
         </div>
-        <div class="c-footer-container c-footer-row c-footer-row--combined c-footer-row--notEqualTopAndBottomPad c-footer-row--noPadBelowWide">
+
+        <div
+            class="c-footer-container c-footer-row c-footer-row--combined c-footer-row--notEqualTopAndBottomPad c-footer-row--noPadBelowWide">
             <country-selector
                 :current-country-name="copy.currentCountryName"
                 :current-country-key="copy.currentCountryKey"
                 :countries="copy.countries"
                 :change-country-text="copy.changeCurrentCountry" />
+
             <legal-field
                 v-if="metaLegalFieldEnabled"
                 :info="copy.metaLegalField" />
+
             <icon-list
                 :icons="copy.paymentIcons"
-                class="c-iconList c-iconList--payments" />
+                is-payments />
         </div>
     </footer>
 </template>
@@ -63,7 +70,6 @@ import LinkList from './LinkList.vue';
 import tenantConfigs from '../tenants';
 
 export default {
-    name: 'VueFooter',
     components: {
         ButtonList,
         CountrySelector,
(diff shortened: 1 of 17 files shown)


xander-marjoram:
Looks like you're missing a changelog entry, but otherwise LGTM

packages/f-footer/src/components/Footer.vue
```diff
@@ -15,40 +16,46 @@
...
                     <feedback-block
                         :title="copy.feedback"
                         :text="copy.improveOurWebsite"
                         :button-text="copy.sendFeedback" />
+
                     <icon-list
                         :icons="copy.socialIcons"
                         :title="copy.followUs"
-                        class="c-iconList c-iconList--social" />
+                        is-social />
```
zlatinivanov-je, line 39: I don't see other references to `is-social`. How is this prop used?

packages/f-footer/src/components/Footer.vue
```diff
@@ -15,40 +16,46 @@
...
                 :change-country-text="copy.changeCurrentCountry" />
+
             <legal-field
                 v-if="metaLegalFieldEnabled"
                 :info="copy.metaLegalField" />
+
             <icon-list
                 :icons="copy.paymentIcons"
-                class="c-iconList c-iconList--payments" />
+                is-payments />
```
zlatinivanov-je, line 58: As above

packages/f-footer/src/components/tests/__snapshots__/Footer.test.js.snap
```diff
@@ -13,16 +13,16 @@ exports[`Footer should render default component markup 1`] = `
     <div class="c-footer-container">
       <!---->
       <div class="c-footer-row">
-        <icon-list-stub icons="[object Object],[object Object]" title="Download our apps" isapps="true" locale="en-GB" class="c-iconList c-iconList--apps"></icon-list-stub>
+        <icon-list-stub icons="[object Object],[object Object]" title="Download our apps" isapps="true" locale="en-GB"></icon-list-stub>
         <feedback-block-stub title="Feedback" text="Help us improve our website" buttontext="Send feedback"></feedback-block-stub>
-        <icon-list-stub icons="[object Object],[object Object],[object Object]" title="Follow us" locale="en-GB" class="c-iconList c-iconList--social"></icon-list-stub>
+        <icon-list-stub icons="[object Object],[object Object],[object Object]" title="Follow us" issocial="true" locale="en-GB"></icon-list-stub>
```
zlatinivanov-je, line 18: Found that missing reference... :)

packages/f-footer/src/components/LinkList.vue
```diff
@@ -52,124 +61,116 @@ export default {
...
-        @include theme(ml) {
-            font-family: $font-family-headings--ml;
-            font-weight: $font-weight-headings--ml;
-        }
-
-        @include media('<wide') {
-            cursor: pointer;
-        }
+.c-footer-heading--button {
+    align-items: center;
```
zlatinivanov-je, line 132: Is there a preference to arrange the properties in alphabetical order?

zlatinivanov-je (approved):
Looks good to me. 

I experienced issues with App styles leaking down, affecting the footer. This appears to address those problems.

packages/f-footer/src/components/LinkList.vue
DamianMullins: No hard rule around it but most of our codebase tends to follow that style, plus it makes it easier to scan styles and find any properties you are looking for (also matches the order in devtools).

packages/f-footer/src/components/Footer.vue
```diff
@@ -122,7 +136,6 @@ export default {
 </script>
 
 <style lang="scss">
```
ashleynolan, line 138: Any reason the styles here wouldn't use `modules`?

DamianMullins: We wanted these styles to be used across the entire module, it was either add them here without the `modules` attribute or create another SCSS file which is imported into all components which need the classes.

Open to other thoughts/suggestions on this.

ashleynolan (approved):
Looks good. Does this change anything in terms of the relative size of the compiled CSS when it's imported?

ashleynolan:
In essence – totally happy with this. I think the main thing that would be good to think about would be how we use CSS Module syntax alongside styles that can be shared throughout the codebase (so for components that have a list of links for instance, do we scope these styles inside each component like we have done here in footer, or do we break this out as a separate component so it can be re-used, or do we have a separate small SCSS utility library – a bit like what fozzie base SCSS is now – where those styles can get pulled in and shared).

Doesn't affect this PR, but would be good to get people's thoughts about it, as I'm not sure we've thought too much about this as the number of components gets built out.

DamianMullins:
@ashleynolan the compiled CSS size is slightly larger due to the longer generated class names, but the change in size is minimal and after being GZipped is essentially the same.

PR #46, opened 2019-09-10
f-header@2.0.0-beta.2 skipToMain component

dandel10n:
### Added
- SkipToMain component

### Changed
- Made `is-transparent` property to be not required with default `false` value

### DIFF
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -31,7 +31,8 @@ export default {
         },
         isTransparent: {
             type: Boolean,
-            required: true
+            required: false,
+            default: false
         }
     },
     data () {
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -2,6 +2,9 @@
     <header
         :data-theme="theme"
         class="c-header">
+        <skip-to-main
+            :skip-to-main-content-text="copy.skipToMainContentText"
+            :extra-classes="[isTransparent ? 'c-skipTo--whiteLink' : '']" />
         <logo
             :theme="theme"
             :is-transparent="isTransparent"
@@ -12,11 +15,13 @@
 <script>
 import Logo from './Logo.vue';
 import tenantConfigs from '../tenants';
+import SkipToMain from './SkipToMain.vue';
 
 export default {
     name: 'VueHeader',
     components: {
-        Logo
+        Logo,
+        SkipToMain
     },
     props: {
         locale: {
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -3,8 +3,8 @@
         :data-theme="theme"
         class="c-header">
         <skip-to-main
-            :skip-to-main-content-text="copy.skipToMainContentText"
-            :extra-classes="[isTransparent ? 'c-skipTo--whiteLink' : '']" />
+            :text="copy.skipToMainContentText"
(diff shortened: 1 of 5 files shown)


packages/f-header/src/components/Header.vue
```diff
@@ -2,6 +2,9 @@
     <header
         :data-theme="theme"
         class="c-header">
+        <skip-to-main
+            :skip-to-main-content-text="copy.skipToMainContentText"
+            :extra-classes="[isTransparent ? 'c-skipTo--whiteLink' : '']" />
```
DamianMullins: Could pass a prop like `is-transparent` here instead and set the `c-skipTo--whiteLink` class inside the skip to main component instead?

That way the skip to main component is responsible for applying styles rather than the parent.

packages/f-header/src/components/SkipToMain.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+    <div :class="['c-skipTo', extraClasses]">
+        <a
+            class="is-visuallyHidden focusable"
```
DamianMullins: `is-focusable`?

packages/f-header/src/components/SkipToMain.vue
```diff
@@ -0,0 +1,68 @@
...
+            href="#skipToMain">{{ skipToMainContentText }}</a>
+    </div>
+</template>
+
+<script>
+export default {
+    props: {
+        skipToMainContentText: {
```
xander-marjoram: If we're already inside the `SkipToMain` component, I don't think the prop needs to be called `skipToMainContentText`, just `text` would probably suffice

xander-marjoram: (It's fine to keep the same name for the `copy` property)

packages/f-header/src/components/SkipToMain.vue
```diff
@@ -1,22 +1,21 @@
 <template>
-    <div :class="['c-skipTo', extraClasses]">
+    <div :class="['c-skipTo', {'c-skipTo--whiteLink' : trasparentBg}]">
```
DamianMullins: I think the linter will warn about having no space between the object key and value here? i.e. ` { 'c-skipTo--whiteLink' : trasparentBg }`

packages/f-header/src/components/Header.vue
```diff
@@ -2,6 +2,9 @@
     <header
         :data-theme="theme"
         class="c-header">
+        <skip-to-main
+            :text="copy.skipToMainContentText"
+            :trasparent-bg="isTransparent" />
```
xander-marjoram: Typo: `trasparent` -> `transparent`

dandel10n: ✅

xander-marjoram (approved):

DamianMullins (approved):

packages/f-header/src/components/SkipToMain.vue
```diff
@@ -0,0 +1,67 @@
...
+<style lang="scss">
+
+.c-skipTo {
+    position: absolute;
+    top: 0;
+    left: 0;
+    text-align: center;
+    z-index: zIndex(high);
+
+    .is-visuallyHidden {
```
ashleynolan, line 34: (catching up on old PRs still…) – is this not part of [helper functons](https://github.com/justeat/fozzie/blob/master/src/scss/trumps/_utilities.scss#L58) in `fozzie` that we could pull in and use?

PR #49, opened 2019-09-12
fozzie-components@1.2.0 – Consoildate utility methods in f-header and f-footer

smilburn1990:
### Changed
- Service has been made to handle duplicate code in header and footer.
- Gitignore updated
- Vue for header and footer changed and methods removed

## TODO
- Write tests for services

👍


BanksySan (dismissed):

dandel10n (changes requested):
@smilburn1990 hey!

LGTM only few small things:

- I think that `@babel/plugin-proposal-optional-chaining` should be added as a dev dependency on mono repo root level package.json to avoid duplication in children packages.
- Maybe we need to make a folder for shared services? Now we have `packages/f-header`,  `packages/f-footer` etc and a js file with shared service.
- And a conflict needs to be resolved

services/shared.service.js
```diff
@@ -0,0 +1,37 @@
+
```
ashleynolan: Looks like there's a duplicate of this file here and at the base of `/packages`?

services/shared.service.js
```diff
@@ -0,0 +1,37 @@
...
+/**
```
ashleynolan: Also, would be good to have some tests for these service functions (might be worth putting in a TODO, unless you fancy making these as realise you're just moving them).

services/shared.service.js
smilburn1990: Still WIP @ashleynolan - sorry my bad. Will let you know when it needs review 👍

services/shared.service.js
smilburn1990: Can we get a ticket created? Who is the best person to speak to? @ashleynolan

ashleynolan: We usually just create the tickets ourselves as we come across them – would add this to the [header epic](https://jira.just-eat.net/browse/WCP-107).  If you don't have access, let Darren know and he can make sure you have access.

smilburn1990: @ashleynolan - i will add one now, can you approve this when you get a moment, would be nice to move some stuff over today :)

ashleynolan (dismissed):
LGTM – think it just needs @dandel10n to re-review.

DamianMullins (dismissed):
Needs a version bump and changelog entry.

dandel10n:
@DamianMullins @ashleynolan  this change is related more to monorepo structure, moving repeated logic out of component level, than to one of the packages. Do you think it is worth to update both `f-footer` and `f-header` versions and changelogs?

ashleynolan:
@dandel10n If it's more of a mono-repo change, then the CHANGELOG in the root should be updated and the package.json there updated. If there's changes to header/footer though too, they'll also need a CHANGELOG entry and package bump as it should be a versioned release for both of them.

We probably need to document the process for changes like this but in essence, all packages that change should have a `version` and `CHANGELOG` entry so that it's traceable (mainly for if a bug occurs and we need to look back at version history).

dandel10n (approved):

ashleynolan (approved):

.gitignore
```diff
@@ -24,6 +24,7 @@ logs
 *.log
 .yalc
 yalc.lock
+package-lock.json
```
SachaZvetelman, line 27: We shouldn't be ignoring this file, in my opinion.

ashleynolan: We should have a `yarn.lock`, so it shouldn't be an issue to ignore the `package-lock.json`.

PR #50, opened 2019-09-13
WCP-149: Added the new link to header and updated content and tests

smilburn1990:
### Changed
- Translation files to add new content and links for 'Deliver with Just Eat' link.
- Test snapshot to accommodate new changes in vue

### Added
- New variables for SVG
- Toggle to only show on Homepage and UK
- Tracking to the new link

👍

### DIFF
--- a/packages/f-header/package.json
+++ b/packages/f-header/package.json
@@ -35,7 +35,7 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-vue-icons": "1.0.0-beta.3"
+    "@justeat/f-vue-icons": "1.0.0-beta.4"
   },
   "peerDependencies": {
     "@justeat/f-trak": "0.6.0"
--- a/packages/f-header/package.json
+++ b/packages/f-header/package.json
@@ -35,7 +35,7 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-vue-icons": "1.0.0-beta.4"
+    "@justeat/f-vue-icons": "1.0.0-beta.3"
   },
   "peerDependencies": {
     "@justeat/f-trak": "0.6.0"
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -17,8 +17,8 @@
             :account-logout="copy.accountLogout"
             :account-login="copy.accountLogin"
             :open-menu-text="copy.openMenuText"
-            :delivery-enquiry="showDeliveryEnquiry && copy.deliveryEnquiry"
-            :show-delivery-enquiry="showDeliveryEnquiry" />
+            :delivery-enquiry="showDeliveryEnquiryWithContent && copy.deliveryEnquiry"
+            :show-delivery-enquiry="showDeliveryEnquiryWithContent" />
     </header>
 </template>
 
@@ -66,6 +66,11 @@ export default {
             theme,
             user
         };
+    },
+    computed: {
+        showDeliveryEnquiryWithContent () {
+            return this.copy.deliveryEnquiry && this.showDeliveryEnquiry;
+        }
     },
     methods: {
         getLocale () {
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -18,7 +18,7 @@
             :account-login="copy.accountLogin"
             :open-menu-text="copy.openMenuText"
             :delivery-enquiry="copy.deliveryEnquiry"
-            :locale="'en-UK'"
+            :locale="'locale'"
             :is-homepage="isHomepage" />
     </header>
 </template>
@@ -45,7 +45,7 @@ export default {
         isHomepage: {
             type: Boolean,
             required: false,
-            default: true
+            default: false
         },
         isTransparent: {
             type: Boolean,
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -17,7 +17,7 @@
             :account-logout="copy.accountLogout"
             :account-login="copy.accountLogin"
             :open-menu-text="copy.openMenuText"
-            :delivery-enquiry="showDeliveryEnquiryWithContent && copy.deliveryEnquiry"
+            :delivery-enquiry="showDeliveryEnquiryWithContent ? copy.deliveryEnquiry : {}"
             :show-delivery-enquiry="showDeliveryEnquiryWithContent" />
     </header>
 </template>
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -16,7 +16,10 @@
             :help="copy.help"
             :account-logout="copy.accountLogout"
             :account-login="copy.accountLogin"
-            :open-menu-text="copy.openMenuText" />
+            :open-menu-text="copy.openMenuText"
+            :delivery-enquiry="copy.deliveryEnquiry"
+            :locale="'en-UK'"
+            :is-homepage="isHomepage" />
     </header>
 </template>
 
@@ -38,6 +41,11 @@ export default {
             type: String,
             required: false,
             default: ''
+        },
+        isHomepage: {
+            type: Boolean,
+            required: false,
+            default: true
         },
         isTransparent: {
             type: Boolean,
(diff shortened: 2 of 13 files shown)


packages/f-header/package.json
```diff
@@ -35,7 +35,7 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-vue-icons": "1.0.0-beta.4"
+    "@justeat/f-vue-icons": "1.0.0-beta.3"
```
pikax: downgraded?

packages/f-header/src/components/Navigation.vue
```diff
@@ -24,6 +24,24 @@
         <div
             :class="['c-nav-container', { 'is-visible': navIsOpen }]">
             <ul class="c-nav-list">
+                <li
+                    v-if="locale === 'en-UK' && isHomepage"
```
pikax: can you move to a computed or resolve it elsewhere?

pikax (changes requested):

packages/f-header/package.json
smilburn1990: @dandel10n - Has this changed now?

packages/f-header/src/components/Navigation.vue
smilburn1990: As opposed to from `data()`?

smilburn1990: Done 👍

packages/f-header/package.json
dandel10n: Should be 1.0.0-beta.4

smilburn1990: Updated @pikax 👍

pikax (approved):

packages/f-header/src/components/Header.vue
```diff
@@ -16,7 +16,10 @@
             :help="copy.help"
             :account-logout="copy.accountLogout"
             :account-login="copy.accountLogin"
-            :open-menu-text="copy.openMenuText" />
+            :open-menu-text="copy.openMenuText"
+            :delivery-enquiry="copy.deliveryEnquiry"
+            :locale="'locale'"
```
dandel10n: Does it pass a string 'locale'?

smilburn1990: @dandel10n - Yep e.g. 'en-UK'

packages/f-header/src/components/Navigation.vue
```diff
@@ -218,6 +257,22 @@ $nav-trigger-focus-bg--ml          : $green--offWhite;
...
 }
 
+// Hides icon on mobile devices
+.delivery-icon {
+    display: none;
+}
+
+// Show the delivery icon if desktop
+@include media('>=mid') {
+    .delivery-icon {
```
dandel10n: There is c-nav-icon class which can have a modifier like `.c-nav-icon--delivery `. There is one for `.c-nav-icon--profile` already. And I don't think that we need default width and height here because we can't say that 27px is default, it is in use only for this specific icon

packages/f-header/src/components/Header.vue
dandel10n: At the moment it pass just word 'locale', doesn't it?

smilburn1990: left that in from test. Changed 👍

DamianMullins: Yeah, this should be `:locale="locale"`

packages/f-header/src/components/Navigation.vue
smilburn1990: Using your class now and removed width and height vars 👍

packages/f-header/src/components/Header.vue
smilburn1990: is now 💃

packages/f-header/src/components/Navigation.vue
```diff
@@ -134,13 +154,30 @@ export default {
...
         }
     },
     data () {
         return {
             navIsOpen: false
         };
     },
+    computed: {
+        showDeliveryEnquiry () {
+            return this.locale === 'en-UK' && this.isHomepage;
```
DamianMullins: Shouldn't this be `'en-GB'`?

packages/f-header/src/components/Header.vue
dandel10n: You can add 'locale' to data attr to be returned (at the moment it doesn't) and then call it like `:locale="locale"`

packages/f-header/src/components/Navigation.vue
smilburn1990: In the changes but they aren't up yet. My internet is playing up 👍

packages/f-header/src/components/Navigation.vue
```diff
@@ -134,13 +154,30 @@ export default {
...
+        },
+        deliveryEnquiry: {
+            type: Object,
+            default: () => ({})
+        },
+        locale: {
+            type: String,
+            default: ''
+        },
+        isHomepage: {
```
ashleynolan: I think it may work better if this is structured slightly differently to move the logic out of the component.

So this prop could be `showDeliveryEnquiry` to remove the hard link to a page (like Home) in the props.

Then the application using the component (like HomeWeb) should determine if the value passed in is true/false – so in this case the homepage would do the test for if the locale is UK and as it's the homepage, if it is, then it could set the `showDeliveryEnquiry` prop to true). The default value for `showDeliveryEnquiry` would be false then so it doesn't show on other pages.

Just think it makes the logic a bit cleaner in the component itself that way.

smilburn1990: I'll remove all the logic then 👍

packages/f-header/src/components/Navigation.vue
```diff
@@ -218,6 +257,22 @@ $nav-trigger-focus-bg--ml          : $green--offWhite;
                 z-index 0s linear;
 }
 
+// Hides icon on mobile devices
+.delivery-icon {
+    display: none;
+}
+
+// Show the delivery icon if desktop
+@include media('>=mid') {
```
ashleynolan: Can also just write this media query so it's inline with the above selector:

```scss
.delivery-icon {
    display: none;

  @include media('>=mid') {
    display: block;
    …
  }
}

Comment (smilburn1990) [reply packages/f-header/src/components/Navigation.vue]: Doesn't exist anymore

Comment (smilburn1990) [reply packages/f-header/src/components/Navigation.vue]: Uses class @dandel10n suggested earlier

File: packages/f-header/src/components/Navigation.vue (hunk 6 above)
Comment (smilburn1990) [reply packages/f-header/src/components/Navigation.vue]: Shall we remove all the locale logic in the component too then? Same principle applies I suppose @ashleynolan

Comment (dandel10n) [reply packages/f-header/src/components/Navigation.vue]: If you leave empty "deliveryEnquiry" objects in resource files for all the tenants and the one with text, url, gtm for UK, then you can do the check if copy.deliveryEnquiry is not empty - show the link. And no need for locale check at all

File: packages/f-header/src/components/Header.vue [hunk 8]
```diff
@@ -43,6 +45,10 @@ export default {
             type: Boolean,
             required: false,
             default: false
+        },
+        showDeliveryEnquiry: {
+            type: Boolean,
+            default: true
```
dandel10n: Will it be safer to have false by default?

packages/f-header/src/components/Navigation.vue
smilburn1990: Not sure I like the idea of relying on whether content is there to be honest. Lets just handle logic from the outside and send on prop like @ashleynolan said. Keep the components as logic free as possible 👍

packages/f-header/src/components/Navigation.vue
```diff
@@ -422,7 +450,8 @@ $nav-trigger-focus-bg--ml          : $green--offWhite;
             }
         }
     }
-    .c-nav-icon--profile {
+    .c-nav-icon--profile,
+    .c-nav-icon--delivery {
```
dandel10n: Does the delivery icon look good in 20X22 size? On live site it is 22px X 27px.

packages/f-header/src/components/Navigation.vue
dandel10n: But what happen now if you switch local to es-ES for example with the `showDeliveryEnquiry=true`?

smilburn1990: The parent will handle it and the prop value will change

dandel10n: What I am trying to say is that there will be errors because there will no be no `deliveryEnquiry.text`, `deliveryEnquiry.url` and `deliveryEnquiry.gtm` for other tenants.

dandel10n (approved):

PR #52, opened 2019-09-13
f-header@2.0.0 beta.6 navigation tests

smilburn1990:
Added some tests for navigation component


packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -0,0 +1,161 @@
...
+        };
+
+        // Act
+        const wrapper = shallowMount(Navigation, { propsData });
+
+        // Assert
+        expect(wrapper.find('[data-js-test="delivery-enquiry"]').exists()).toBe(true);
+    });
+
+    it('should NOT show "Deliver for Just Eat" in nav if prop true and content there', () => {
```
dandel10n: if prop _false_

dandel10n (approved):

packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -0,0 +1,161 @@
...
+const resizeWindow = (x, y) => {
+    window.innerWidth = x;
+    window.innerHeight = y;
+    window.dispatchEvent(new Event('resize'));
+};
+
+describe('Navigation', () => {
+    it('should be defined', () => {
+        const propsData = {
+            ...defaultPropsData
```
DamianMullins: No need to spread this, you can assign it as is: `const propsData = defaultPropsData;`.

packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -0,0 +1,161 @@
...
+            userInfo: {
+                isAuthenticated: true
+            }
+        };
+
+        // Act
+        const wrapper = shallowMount(Navigation, { propsData });
+
+        // Assert
+        expect(wrapper.find('[data-js-test="logout"]').exists()).toBe(true);
```
DamianMullins, line 79: This looks like it could split out into two tests?

packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -0,0 +1,161 @@
...
```
DamianMullins: This looks like it could split out into two tests?

packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -0,0 +1,161 @@
...
+                isAuthenticated: true
+            }
+        };
+
+        const navData = {
+            navIsOpen: true
+        };
+
+        // Act
+        resizeWindow(767, 1800);
```
DamianMullins: Can you assign these values to variables so that it is clear from the name what the numbers refer to, please?

DamianMullins (dismissed):
Needs a version bump and changelog entry.

smilburn1990:
@ashleynolan @DamianMullins - Does this really need a version bump considering @dandel10n has already done one today? I could just add some stuff to her change log?

ashleynolan:
@smilburn1990 Yeh – any changes to packages need to be tagged and versioned (mainly so the commits line up with the `CHANGELOG`).

Ideally the releases would be automated to take some of the pain out of this, but atm we haven't had chance to work on getting that setup so agree it's a little bit of overhead atm to release new versions (but just one we've been sucking up for the time being)

ashleynolan (approved):

dandel10n (approved):

PR #54, opened 2019-09-16
f-header@2.0.0-beta.4 Styles for the header component

dandel10n:
### Added
- Styles for the header component
- Some ml css variable overrides
- User email rendering for mobile view

ml theme:
Screen Shot 2019-09-16 at 17 48 13

Screen Shot 2019-09-16 at 17 49 02

je theme:
Screen Shot 2019-09-16 at 17 42 46
Screen Shot 2019-09-16 at 17 49 31

### DIFF
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -1,24 +1,26 @@
 <template>
     <header
         :data-theme="theme"
-        class="c-header">
+        :class="['c-header', { 'c-header--transparent c-header--gradient': isTransparent }]">
         <skip-to-main
             :text="copy.skipToMainContentText"
             :transparent-bg="isTransparent" />
-        <logo
-            :theme="theme"
-            :is-transparent="isTransparent"
-            :company-name="copy.companyName"
-            :logo-gtm-label="copy.logo.gtm" />
-        <navigation
-            :user-info="userInfo"
-            :nav-links="copy.navLinks"
-            :help="copy.help"
-            :account-logout="copy.accountLogout"
-            :account-login="copy.accountLogin"
-            :open-menu-text="copy.openMenuText"
-            :delivery-enquiry="copy.deliveryEnquiry"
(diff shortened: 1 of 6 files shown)


BanksySan (approved):

simonsandersje (approved):

packages/f-header/src/components/Header.vue
```diff
@@ -101,4 +103,154 @@ export default {
...
+        color: $header-buttonCount-color;
+        background: $header-buttonCount-bg;
+        border: 1px solid $header-buttonCount-borderColor;
+
+        @include theme(ml) {
+            background: $header-buttonCount-bg--ml;
+        }
+    }
+
+    .c-header-button--primary {
```
ashleynolan, line 246: Would be good to have a comment about the type of button `c-header-button--primary` and `c-header-button--secondary` to give some context as to what these buttons do.

packages/f-header/src/components/Navigation.vue
```diff
@@ -469,10 +480,20 @@ $nav-trigger-focus-bg--ml          : $green--offWhite;
             }
         }
     }
-    .c-nav-icon--profile{
+    .c-nav-icon--profile {
         width: 20px;
         height: 22px;
+
+    // mobile menulog user profile icon to be green, not default blue
```
ashleynolan, line 487: NIT: Looks like this needs indenting?

packages/f-header/src/components/Navigation.vue
```diff
@@ -469,10 +480,20 @@ $nav-trigger-focus-bg--ml          : $green--offWhite;
...
+    & path {
+        @include media('<mid') {
+            @include theme(ml) {
```
ashleynolan, line 490: Worth checking the compiled output for this, as you may end up with an empty media query if the theme isn't `ml`. If you do, then worth switching the order of these media queries around so that the check for `ml` is first.

PR #56, opened 2019-09-17
f-header@2.0.0-beta.9 - header authentication service

smilburn1990:
### Changed
- Unit tests to reflect changed in component
- Snapshot for test

#### Added
- New logic to create return URL
- computed for `showDeliveryEnquiryWithContent`
- New snapshot for tests
- Axios for new logic
- New logic for retrieving user info 

### TODO
- Update docs

👍


package.json
```diff
@@ -65,6 +65,7 @@
   "dependencies": {
     "@justeat/f-footer": "file:packages/f-footer",
     "@justeat/f-header": "file:packages/f-header",
+    "@justeat/f-logger": "^0.7.1",
```
pikax: Can you remove the `^`, just to target a specific version.

packages/f-header/src/components/Navigation.vue
```diff
@@ -114,11 +114,11 @@
                 </li>
 
                 <li
-                    v-if="!userInfo.isAuthenticated"
+                    v-if="!userInfo"
```
pikax, line 115:
```suggestion
                    v-else
```

packages/f-header/src/components/Navigation.vue
```diff
@@ -114,11 +114,11 @@
...
                     class="c-nav-list-item"
                     data-js-test="login">
                     <a
-                        :href="accountLogin.url"
+                        :href="`${accountLogin.url}${returnUrl}`"
```
pikax: can you create a `computed` for this link?

pikax (changes requested):

packages/f-header/src/components/Navigation.vue
smilburn1990: @pikax - The logout is show on the else but it is embedded inside elements, so this won't work

pikax: you right, it seemed to be the next `li`, my bad

packages/f-header/src/components/Navigation.vue
smilburn1990: I think it's a bit overkill considering on `login` and `logout` use it but I'm happy to if anyone else agrees

smilburn1990: `returnUrl` is already a computed property

pikax: but your `href` = `accountLogin.url + returnUrl`, personally I like to add almost everything as computed or data.

DamianMullins (commented):
Should remove the `package-lock.json` file as everyone should be using Yarn.

packages/f-header/src/components/Navigation.vue
smilburn1990: All done pushing now 👍

package.json
DamianMullins: One issue with adding this as a dependency is that we won't be able to run any CI tasks on Travis as it is an internal package.

One way around this is to inject the logger via a prop?

smilburn1990: I'll do that 👍

smilburn1990: Or is it easier to just use `console.log`? Using it as a prop opens a lot of pitfalls when the prop doesn't get passed @DamianMullins

DamianMullins: `console.log` won't give us any visibility of errors occurring, `just-log` pushes logs to an endpoint which allows us to view those logs in Kibana.

smilburn1990: Ahh gotcha. So shall I pass props as a `[Function, Boolean]` and set the deafault as `false` with required `true`? @DamianMullins

pikax: I would create a prop `Function`, where you can send any number of arguments, the default would be an dummy function:
```js
props:{
  log: {type: Function, default: ()=>({})}
}
```

packages/f-header/package.json
```diff
@@ -36,6 +36,7 @@
   ],
   "dependencies": {
     "@justeat/f-vue-icons": "1.0.0-beta.4",
+    "axios": "^0.19.0",
```
pikax: shouldn't be version specific? `without ^`

packages/f-header/CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ v2.0.0-beta-8
 *September 20, 2019*
 
 ### Changed
+ - Props in `Navigation.vue`
```
pikax, line 12: Do you mind add what props you added? 
Also add them to [readme.md](https://github.com/justeat/fozzie-components/tree/master/packages/f-header)

smilburn1990: We are updating documentation to another ticket 💃 Will updated the Todo on the PR

packages/f-header/src/components/Header.vue
```diff
@@ -53,13 +53,10 @@ export default {
...
-            type: Object,
-            default: () => ({
-                isAuthenticated: false,
-                friendlyName: '',
-                email: ''
-            })
+        justLog: {
+            type: Function,
+            default: () => ({}),
+            required: true
```
pikax: does it need to be required?

smilburn1990: Changed

pikax (dismissed):

packages/f-header/src/components/Header.vue
```diff
@@ -19,7 +19,8 @@
                 :account-login="copy.accountLogin"
                 :open-menu-text="copy.openMenuText"
                 :delivery-enquiry="copy.deliveryEnquiry"
-                :show-delivery-enquiry="showDeliveryEnquiryWithContent" />
+                :show-delivery-enquiry="showDeliveryEnquiryWithContent"
+                :just-log="justLog" />
```
dandel10n, line 22: You still have `:user-info="userInfo"` on line 15, should it be removed?

smilburn1990: Nice spot. Changed 👍 @dandel10n

pikax (dismissed):

DamianMullins:
`package-lock.json` is still in the PR.

packages/f-header/src/components/Navigation.vue
```diff
@@ -231,20 +226,40 @@ export default {
...
+            const { name } = this.$route;
+            const { href } = this.$router.resolve({ name });
+
+            return encodeURIComponent(href);
+        },
+        returnLoginUrl () {
+            return `${this.accountLogin.url}${this.returnUrl}`;
+        },
+        returnLogoutUrl () {
+            return `${this.accountLogout.url}${this.returnUrl}`;
```
dandel10n, line 254: Could you please explain what the behaviour will be if returnUrl is false (when there is no `this.$route`)
As I understand `accountLogout.url = '/account/logout?returnurl='` will come from the resource files. Will it be enough?

packages/f-header/package.json
```diff
@@ -35,7 +35,8 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-vue-icons": "1.0.0-beta.4",
+    "@justeat/f-vue-icons": "^1.0.0-beta.4",
```
DamianMullins: Remove the `^` here.

packages/f-footer/package.json
```diff
@@ -34,7 +34,7 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-vue-icons": "1.0.0-beta.2",
+    "@justeat/f-vue-icons": "^1.0.0-beta.4",
```
DamianMullins: Remove the `^` here.

packages/f-header/src/components/Header.vue
dandel10n: Could you change the props list in README for the f-header please

packages/f-header/src/components/Navigation.vue
```diff
@@ -263,6 +278,18 @@ export default {
...
             this.currentScreenWidth = window.innerWidth;
+        },
+        async returnUserDetails () {
+            try {
+                const { data } = await axios.get('/api/account/details', {
+                    headers: {
+                        credentials: 'same-origin'
+                    }
+                });
+                if (data) this.userInfo = data;
```
DamianMullins: Can you wrap this in parathesis to keep it consistent with the code style used in our codebases, please?

packages/f-header/src/components/Navigation.vue
```diff
@@ -263,6 +278,18 @@ export default {
         },
         onResize () {
             this.currentScreenWidth = window.innerWidth;
+        },
+        async returnUserDetails () {
```
DamianMullins: Maybe `setUserDetails` might better describe what this method is doing?

yarn.lock
```diff
@@ -305,6 +305,14 @@
     "@babel/helper-plugin-utils" "^7.0.0"
     "@babel/plugin-syntax-optional-catch-binding" "^7.2.0"
 
+"@babel/plugin-proposal-optional-chaining@^7.6.0":
+  version "7.6.0"
+  resolved "http://packages.je-labs.com/npm/private-npm/@babel/plugin-proposal-optional-chaining/-/plugin-proposal-optional-chaining-7.6.0.tgz#e9bf1f9b9ba10c77c033082da75f068389041af8"
```
DamianMullins, line 310: You need to update this file by running `yarn install` or `yarn upgrade` to remove the references to our internal npm feed.

packages/f-header/src/components/Navigation.vue
```diff
@@ -46,11 +45,12 @@
                 </li>
 
                 <li
-                    :class="['c-nav-list-item has-sublist', { 'is-hidden': !userInfo.isAuthenticated, 'open': navIsOpen }]"
+                    :class="['c-nav-list-item has-sublist', { 'is-hidden': !userInfo, 'open': navIsOpen }]"
                     @mouseover="openNav"
                     @mouseleave="closeNav"
                     @keyup.esc="closeNav">
                     <a
+                        v-if="userInfo"
```
dandel10n: You don't need this check because parent li element will have display none with `'is-hidden': !userInfo` on line 48

packages/f-header/src/components/Navigation.vue
```diff
@@ -148,13 +148,13 @@
                     </li>
 
                     <li
+                        v-if="userInfo.isAuthenticated"
```
dandel10n: Should here `userInfo.isAuthenticated` be changed to just `userInfo` as in other places

packages/f-header/CHANGELOG.md
dandel10n, line 12: You've added your changes as `v2.0.0-beta-8` which is already published. Should become `v2.0.0-beta-9` now

packages/f-header/src/components/Navigation.vue
smilburn1990: It was carried over from the current static header in menuweb. To be honest I don't think we need it anymore either. let me get rid of it. @DamianMullins - Was there any particular reason it was used in the current header?

packages/f-header/CHANGELOG.md
smilburn1990: I have pulled from master and can't see an updated changelog @dandel10n

packages/f-header/src/components/Navigation.vue
DamianMullins: Without it you'll always be redirected back to the homepage after signing in/out rather than the page where you clicked the link.

pikax (dismissed):

packages/f-header/src/components/Navigation.vue
DamianMullins: I'm not sure what happens if you don't have vue-router configured though (I'm guessing it'll throw an error), so you'll probably want to add this back in and set the return URL using standard JS techniques.

pikax (dismissed):

dandel10n (dismissed):

packages/f-header/src/components/Navigation.vue
smilburn1990: Could you pass `$route` through as a prop? @DamianMullins

smilburn1990: Actually won't `$route` be accessible from the component?

pikax (dismissed):

packages/f-header/src/components/Navigation.vue
DamianMullins: Not sure, `$route` might be, `$router` won't.

DamianMullins: Either way, you need the return URL part otherwise you'd have to find your way back to the menu page after logging in (very annoying). 

You should be able to set the value using `document.location.pathname`? Making sure to encode it using `encodeURIComponent` like in the original code.

dandel10n (dismissed):

dandel10n (approved):

pikax (approved):

zlatinivanov-je:
Are we happy to have an internal menuweb link in the description?

smilburn1990:
@zlatinivanov-je  - No one can access but I have removed anyway

packages/f-header/src/components/Navigation.vue
```diff
@@ -231,20 +225,37 @@ export default {
...
-            currentScreenWidth: 0
+            currentScreenWidth: 0,
+            userInfo: false
         };
     },
     computed: {
         isBelowMid () {
             return this.currentScreenWidth <= 767;
+        },
+        returnUrl () {
```
zlatinivanov-je, line 246: Can we have the option to provide a return Url value. In our project we have specific requirement for Query string not matching the `document.location.pathname` at all times.

packages/f-header/src/components/Header.vue
```diff
@@ -12,14 +12,14 @@
                 :company-name="copy.companyName"
                 :logo-gtm-label="copy.logo.gtm" />
             <navigation
-                :user-info="userInfo"
```
zlatinivanov-je, line 15: I liked that approach since my app already has the user info Object available

packages/f-header/src/components/Header.vue
```diff
@@ -12,14 +12,14 @@
...
                 :nav-links="copy.navLinks"
```
zlatinivanov-je, line 15: Does it no make sense to have a single `copy` prop to pass, following a convention, instead of so many...

packages/f-header/src/components/Navigation.vue
```diff
@@ -263,6 +274,22 @@ export default {
         },
         onResize () {
             this.currentScreenWidth = window.innerWidth;
+        },
+        async setUserDetails () {
+            try {
```
zlatinivanov-je, line 279: In my case, the app already does that request in a separate occasion and can provide the object, instead. Duplicate request can be avoided and saves on `Axios` added.

zlatinivanov-je (commented):
Not sure if adding `Axios` and making the auth call here is necessary. I would assume the host app will already use a service(s) to get page data and could provide the necessary to the component.

DamianMullins:
@zlatinivanov-je Menu Web has no awareness of user details so we rely on the `api/account/details` endpoint to provide that. We don't plan on adding the user middleware to Menu Web, it was considered but ultimately we decided against as we wanted the back-end to be as light as possible.

pikax:
couldn't we use `fetch` and enforce a polyfill if that functionality gets used, IMHO is a bit more elegant than bundle axios.

packages/f-header/src/components/Navigation.vue
ashleynolan: Would be good to have the option to pass this in and not make the call if it's present. Think that's a valid use-case (like in Help)

ashleynolan:
@DamianMullins @zlatinivanov-je We could make this like the locale though, whereby having an optional prop that can accept the user details and would mean the header then doesn't make the call?

zlatinivanov-je:
@DamianMullins in my case it is not the backend that provides the user data. The app utilises user object for LiveChat, BotChat, and Order Help Automation components. The Vue app is the one that calls the same endpoint `/api/account/details` and has the user object already.

packages/f-header/src/components/Navigation.vue
smilburn1990: @zlatinivanov-je - I did raise this in `guild-ui` meeting last week and have raised it multiple times in other channels. It was a unanimous decision to go with this approach in the end, but I agree with @ashleynolan that we could toggle it off with a prop if the app already has the functionality.

smilburn1990: @zlatinivanov-je - I did raise this in `guild-ui` meeting last week and have raised it multiple times in other channels. It was a unanimous decision to go with this approach in the end, but I agree with @ashleynolan that we could toggle it off with a prop if the app already has the functionality.

zlatinivanov-je: @smilburn1990 Sorry I was away last week. But I am happy with @ashleynolan's proposal.

PR #58, opened 2019-09-18
f-header@2.0.0-beta.5 Remove Deliver with JE link from mobile screen size

dandel10n:
### Changed
 - Add a logic to hide "Deliver with Just Eat" link from mobile screen size
 - DeliveryEnquiry gtm label changed to be 'click_courier_signup'

### Removed
- Cleaned up unused css styles

### DIFF
--- a/packages/f-header/package.json
+++ b/packages/f-header/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-header",
   "description": "Fozzie Header – Globalised Header Component",
-  "version": "v2.0.0-beta.4",
+  "version": "v2.0.0-beta.5",
   "main": "dist/f-header.umd.min.js",
   "files": [
     "dist"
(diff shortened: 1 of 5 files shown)


simonsandersje (approved):

packages/f-header/package.json
```diff
@@ -35,7 +35,8 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-vue-icons": "1.0.0-beta.4"
+    "@justeat/f-vue-icons": "1.0.0-beta.4",
+    "lodash-es": "4.17.15"
```
smilburn1990, line 39: Will this only be used in this component? @dandel10n

packages/f-header/src/components/Navigation.vue
```diff
@@ -184,9 +184,22 @@ export default {
...
     data () {
         return {
-            navIsOpen: false
+            navIsOpen: false,
+            currentScreenWidth: 0
         };
     },
+    computed: {
+        isBelowMid () {
+            return this.currentScreenWidth <= 767;
```
smilburn1990, line 193: Should we store screen widths globally somewhere in fozzie? @dandel10n

DamianMullins (approved):

packages/f-header/package.json
dandel10n: It is in use in `f-footer`too but with monorepo structure, as I know, only dev dependencies can live in the root (https://github.com/lerna/lerna#common-devdependencies)

smilburn1990 (approved):

PR #60, opened 2019-09-19
f-header@2.0.0-beta.8 accessibility improvements

dandel10n:
### Changed
 - Added some accessibility improvements to the navigation component

### DIFF
--- a/packages/f-header/CHANGELOG.md
+++ b/packages/f-header/CHANGELOG.md
@@ -3,6 +3,15 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+v2.0.0-beta-8
+------------------------------
+*September 19, 2019*
+
+### Changed
+ - Added some accessibility improvements to the navigation component
+
+
 v2.0.0-beta-7
 ------------------------------
 *September 18, 2019*
--- a/packages/f-header/CHANGELOG.md
+++ b/packages/f-header/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v2.0.0-beta-8
 ------------------------------
-*September 19, 2019*
+*September 20, 2019*
 
 ### Changed
  - Added some accessibility improvements to the navigation component
--- a/packages/f-header/src/components/Navigation.vue
+++ b/packages/f-header/src/components/Navigation.vue
@@ -130,22 +130,6 @@
                     </a>
                 </li>
 
-                <li
-                    v-if="!isBelowMid"
-                    class="c-nav-list-item c-nav-list-item--support">
-                    <a
-                        :href="help.url"
-                        class="c-nav-list-link"
-                        :data-trak='`{
-                            "trakEvent": "click",
-                            "category": "engagement",
-                            "action": "header",
-                            "label": "${help.gtm}"
-                        }`'>
-                        {{ help.text }}
-                    </a>
-                </li>
-
                 <template v-if="isBelowMid">
                     <li class="c-nav-list-item c-nav-list-item--support">
                         <a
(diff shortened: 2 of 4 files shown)


packages/f-header/src/components/Navigation.vue
```diff
@@ -56,21 +53,22 @@
                     <a
                         class="c-nav-list-text"
                         href="/"
-                        :aria-expanded="navIsOpen ? 'true' : 'false'"
-                        aria-haspopup="true"
+                        :tabindex="isBelowMid ? '-1' : '0'"
```
DamianMullins: The values can just be numbers rather than strings.

packages/f-header/src/components/Navigation.vue
```diff
@@ -1,11 +1,11 @@
 <template>
-    <nav
```
BanksySan, line 2: Is this reformatting to bring it inline with the rest of the formatting conventions or is it a mistake?

BanksySan (changes requested):
Just confirm that the whitespace changes are on purpose and not incidental.

packages/f-header/src/components/Navigation.vue
dandel10n: ✅

packages/f-header/src/components/Navigation.vue
dandel10n: If it is only clean html markup like here that it should be on one line. If there are vue attributes than for better readability it should be splitted onto new lines - it is done by eslint check

BanksySan (dismissed):

packages/f-header/src/components/Navigation.vue
```diff
@@ -144,6 +145,44 @@
                         {{ help.text }}
                     </a>
                 </li>
+
+                <template v-if="isBelowMid">
```
pikax, line 133:
```suggestion
                <template v-else>
```
`v-if` and `v-else` makes easier to read

packages/f-header/CHANGELOG.md
```diff
@@ -3,6 +3,15 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+v2.0.0-beta-8
+------------------------------
+*September 19, 2019*
```
pikax:
```suggestion
*September 20, 2019*
```

packages/f-header/src/components/Navigation.vue
```diff
@@ -144,6 +145,44 @@
...
+                                "label": "${help.gtm}"
+                            }`'
+                            @blur="closeNav"
+                            @focus="openNav">
+                            {{ help.text }}
+                        </a>
+                    </li>
+
+                    <li
+
```
pikax: NIT: remove space

packages/f-header/CHANGELOG.md
ashleynolan: I didn't know you could make code suggestions like this inline…

Alt Text

packages/f-header/src/components/Navigation.vue
DamianMullins: If you're going to use if/else could you move this block to the top so the if check is against `isBelowMid` and not `!isBelowMid`, please.

pikax (dismissed):

packages/f-header/src/components/Navigation.vue
dandel10n: ✅

packages/f-header/src/components/Navigation.vue
```diff
@@ -93,10 +91,11 @@
                         </li>
 
                         <li
-                            class="c-nav-list-item c-nav-list-item--forceLast"
+                            v-if="!isBelowMid"
```
pikax, line 94: rollback?

pikax (changes requested):

packages/f-header/src/components/Navigation.vue
dandel10n: No, this one is shown only for wider screens and can't be a part of if/else template because here logout link is a part of `c-nav-popoverList` when help is not, and for mobile view they both are inside a dropdown.

DamianMullins (approved):

packages/f-header/src/components/Navigation.vue
pikax: My bad, seemed a rollback, sorry about that

pikax (approved):

PR #61, opened 2019-09-23
f-header@2.0.0-beta.11 - mobile transparent menu styles for open state

dandel10n:
### Added
- `no-js` css for mobile hamburger menu

### Changed
- Styling for mobile open menu state header when transparent

[Image from Gyazo](https://gyazo.com/f5a9460c16083e7271ee88f6605b9899)

### DIFF
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -1,14 +1,14 @@
 <template>
     <header
         :data-theme="theme"
-        :class="['c-header', { 'c-header--transparent c-header--gradient': isTransparent }]">
+        :class="['c-header', { 'c-header--transparent c-header--gradient': showTransparentHeader }]">
         <skip-to-main
             :text="copy.skipToMainContentText"
-            :transparent-bg="isTransparent" />
+            :transparent-bg="showTransparentHeader" />
         <div class="c-header-container">
             <logo
                 :theme="theme"
-                :is-transparent="isTransparent"
+                :is-transparent="showTransparentHeader"
                 :company-name="copy.companyName"
                 :logo-gtm-label="copy.logo.gtm" />
             <navigation
@@ -20,7 +20,8 @@
                 :delivery-enquiry="copy.deliveryEnquiry"
                 :show-delivery-enquiry="showDeliveryEnquiryWithContent"
                 :user-info-prop="userInfoProp"
-                :just-log="justLog" />
+                :just-log="justLog"
+                @onMobileNavToggle="mobileNavToggled" />
         </div>
     </header>
 </template>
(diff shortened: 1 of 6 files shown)


packages/f-header/src/components/Header.vue
```diff
@@ -1,14 +1,14 @@
 <template>
     <header
         :data-theme="theme"
-        :class="['c-header', { 'c-header--transparent c-header--gradient': isTransparent }]">
+        :class="['c-header', { 'c-header--transparent c-header--gradient': isTransparent && !mobileNavIsOpen }]">
```
pikax: moving `isTransparent && !mobileNavIsOpen` to a computed might be a good idea because you're repeating that for a few times in the template

packages/f-header/src/components/Navigation.vue
```diff
@@ -286,10 +291,14 @@ export default {
                     this.userInfo = data;
                 }
             } catch (err) {
-                if (this.justLog) {
+                if (this.justLog.error) {
```
pikax, line 300: `justLog` is describe as a function in the props on line 235, please update to an Object and validate if justLog is not `null` or `undefined`

packages/f-header/src/components/Navigation.vue
```diff
@@ -286,10 +291,14 @@ export default {
...
                     this.justLog.error('Error handling "setUserDetails" action', err);
                 }
             }
+        },
+        onHamburgerMenuClick () {
+            this.onNavToggle();
```
pikax, line 309: is onNavToggle being used elsewhere? Wouldn't be better to drop it, since `onHamburgerMenuClick` is the same as `onNavToggle`

pikax (changes requested):

packages/f-header/src/components/Navigation.vue
ashleynolan: I think they're slightly different (might be worth commenting the function tom explain this difference though).

`onNavToggle` applies to all screen widths whereas `onHamburgerClick` is only related to when the hamburger icon shows on mobile. Is that right @dandel10n ?

packages/f-header/src/components/Header.vue
```diff
@@ -63,16 +64,23 @@ export default {
...
+            mobileNavIsOpen
         };
     },
     computed: {
         showDeliveryEnquiryWithContent () {
             return this.copy.deliveryEnquiry && this.showDeliveryEnquiry;
         }
+    },
+    methods: {
+        mobileNavToggled (value) {
```
ashleynolan, line 88: Would be good to have some comments around these methods (and the functions/methods throughout the `.vue` files).

packages/f-header/src/components/Navigation.vue
dandel10n: @ashleynolan @pikax yes, Ash explained correctly, `onNavToggle` is only for toggling open/closed nav state and `onHamburgerMenuClick` activates onNavToggle and then emits the `onMobileNavToggle` for header components itself only for Hamburger menu click.

pikax (dismissed):

pikax (dismissed):

pikax (approved):

simonsandersje (approved):

packages/f-header/src/components/Header.vue
```diff
@@ -42,42 +43,50 @@ export default {
...
         showDeliveryEnquiryWithContent () {
             return this.copy.deliveryEnquiry && this.showDeliveryEnquiry;
+        },
+        showTransparentHeader () {
+            return this.isTransparent && !this.mobileNavIsOpen;
+        }
+    },
+    methods: {
+        // This method emits `navIsOpen` state from the navigation component
+        // to be able to deside when to show transparent header styles on mobile view
```
ashleynolan, line 87: Typo – `deside` > `decide`

PR #63, opened 2019-09-24
v2.0.0-beta-10 add-toggle-for-user-info

smilburn1990:
### Added
- New prop for `userInfo` coming from outside the component
- Logic to call the data if prop is `false`

### Changed
- Name of method

### DIFF
--- a/packages/f-header/CHANGELOG.md
+++ b/packages/f-header/CHANGELOG.md
@@ -9,7 +9,7 @@ v2.0.0-beta-10
 *September 20, 2019*
 
 ### Changed
-- Name of method from 'setUserDetails' to 'setUserInfo'
+- Name of method from `setUserDetails` to `setUserInfo`
 
 ### Added
 - New prop for `userInfo` coming from outside the component
--- a/packages/f-header/CHANGELOG.md
+++ b/packages/f-header/CHANGELOG.md
@@ -4,6 +4,18 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v2.0.0-beta-10
+------------------------------
+*September 20, 2019*
+
+### Changed
+- Name of method
+
+### Added
+- New prop for `userInfo` coming from outside the component
+- Logic to call the data if prop is `false`
+
+
 v2.0.0-beta-9
 ------------------------------
 *September 20, 2019*
--- a/packages/f-header/CHANGELOG.md
+++ b/packages/f-header/CHANGELOG.md
@@ -9,7 +9,7 @@ v2.0.0-beta-10
 *September 20, 2019*
 
 ### Changed
-- Name of method
+- Name of method from 'setUserDetails' to 'setUserInfo'
 
 ### Added
 - New prop for `userInfo` coming from outside the component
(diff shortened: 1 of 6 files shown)


packages/f-header/src/components/Header.vue
```diff
@@ -19,6 +19,7 @@
                 :open-menu-text="copy.openMenuText"
                 :delivery-enquiry="copy.deliveryEnquiry"
                 :show-delivery-enquiry="showDeliveryEnquiryWithContent"
+                :user-info-prop="userInfoProp"
```
kevinrodrigues, line 22: Do we need the `prop` part considering it's a `prop`? `user-info` should be sufficient.

packages/f-header/src/components/Navigation.vue
```diff
@@ -230,13 +230,18 @@ export default {
             type: Function,
             default: () => ({}),
             required: false
+        },
+        userInfoProp: {
```
kevinrodrigues, line 234: Similar to above.

packages/f-header/src/components/Navigation.vue
```diff
@@ -230,13 +230,18 @@ export default {
...
+            default: false,
+            required: false
         }
     },
     data () {
         return {
             navIsOpen: false,
             currentScreenWidth: 0,
-            userInfo: false
+            userInfo: this.userInfoProp
```
kevinrodrigues, line 244: Same as above.

packages/f-header/src/components/Navigation.vue
```diff
@@ -275,7 +282,7 @@ export default {
         onResize () {
             this.currentScreenWidth = window.innerWidth;
         },
-        async setUserDetails () {
+        async setUserInfo () {
```
kevinrodrigues, line 285: The catch statement would need updating from `setUserDetails` > `setUserInfo`. Plus the readme.

pikax (dismissed):

kevinrodrigues (commented):
Minor comments. You'd need the changelog and package updated too.

packages/f-header/src/components/Header.vue
smilburn1990: Will update changelog but cannot call in `user-info` as it already has data with the same name 👍

simonsandersje (dismissed):

packages/f-header/src/components/Header.vue
ashleynolan: Where does the `user-info` prop get set then? Is it just in the `Navigation.vue` file? 

If so, you could call it the same and just pass this straight through as a prop onto that component? (I remember doing something similar in the searchbox)

I think if you pass it straight through you wouldn't need to set it as data in the `Navigation.vue` file then.

packages/f-header/CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v2.0.0-beta-10
+------------------------------
+*September 20, 2019*
+
+### Changed
+- Name of method
```
kevinrodrigues: Sorry bit picky here but could we be more specific of the change.

kevinrodrigues (approved):

dandel10n (approved):

PR #64, opened 2019-09-26
f-header@2.0.0-beta.13-update-to-header-logic

smilburn1990:
### Changed
- Logic in the `getUserInfo` method inside `Navigation.vue`
- Versions for `f-services`, `f-header` and `f-footer`
- Conditionals to point at `isAuthenticated` as `userData` will always be returned
- Webpack on `f-services` to babel polyfill
 - Reference to `$i18n` localization library in `f-header` and `f-footer`

### Added
 - Prop for `userInfoUrl`

### DIFF
--- a/packages/f-header/package.json
+++ b/packages/f-header/package.json
@@ -35,7 +35,7 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-services": "0.5.0",
+    "@justeat/f-services": "0.6.0",
     "@justeat/f-vue-icons": "1.0.0-beta.4",
     "axios": "0.19.0",
     "lodash-es": "4.17.15"
--- a/packages/f-header/package.json
+++ b/packages/f-header/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-header",
   "description": "Fozzie Header – Globalised Header Component",
-  "version": "v2.0.0-beta.14",
+  "version": "v2.0.0-beta.15",
   "main": "dist/f-header.umd.min.js",
   "files": [
     "dist"
--- a/packages/f-header/package.json
+++ b/packages/f-header/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-header",
   "description": "Fozzie Header – Globalised Header Component",
-  "version": "v2.0.0-beta.12",
+  "version": "v2.0.0-beta.13",
   "main": "dist/f-header.umd.min.js",
   "files": [
     "dist"
@@ -36,7 +36,7 @@
   ],
   "dependencies": {
     "@justeat/f-services": "0.4.0",
-    "@justeat/f-vue-icons": "1.0.0-beta.4",
+    "@justeat/f-vue-icons": "^1.0.0-beta.4",
     "axios": "0.19.0",
     "lodash-es": "4.17.15"
   },
--- a/packages/f-header/package.json
+++ b/packages/f-header/package.json
@@ -35,7 +35,7 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-services": "0.6.0",
+    "@justeat/f-services": "^0.6.0",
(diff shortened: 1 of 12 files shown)


packages/f-header/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-header",
   "description": "Fozzie Header – Globalised Header Component",
-  "version": "v2.0.0-beta.12",
```
ashleynolan, line 4: Was beta.13 never used?

packages/f-header/src/components/Navigation.vue
```diff
@@ -288,7 +288,7 @@ export default {
         // If userInfoProp wasn't passed we make a call for userInfo on mounted hook
         async setUserInfo () {
             try {
-                const { data } = await axios.get('/api/account/details', {
+                const { data } = await axios.get('https://uk-publicweb-haproxy-staging-uk.staging-uk.je-labs.com/api/account/details', {
```
simonsandersje: This should go into a config file, as it'll be too easy to leave a hard-coded staging URL in the production version!

smilburn1990: Oh wow, sorry that was me testing. Good save 👍

packages/f-header/package.json
smilburn1990: Very briefly locally but we had to rebuild because there were problems with the `f-services` dependency it had :( @ashleynolan

packages/f-services/src/services/shared.service.js
```diff
@@ -13,18 +20,15 @@
...
-    // makes sure the locale is the same format as the tenants - 'en-GB'
-    let transformedLocale = `${locale.slice(0, -3)}${locale.slice(locale.lastIndexOf('-')).toUpperCase()}`;
+    let locale = transformLocale(tenantString === '' ? globalTenant?.locale : tenantString);
 
     // if the locale is either
     // a) not set
     // or b) set to a country code that this component does not recognise
     // it will be set to 'en-GB'
-    if (!tenantConfigs[transformedLocale]) transformedLocale = 'en-GB';
+    if (!tenantConfigs[locale]) locale = 'en-GB';
```
BanksySan, line 30: Do we have any code conventions around single line if statements?

BanksySan (changes requested):
Just a question about code conventions.

packages/f-services/src/services/shared.service.js
smilburn1990: I'm using the convention in the services from Menu that I have seen.

DamianMullins: They should be on a new line and wrapped in parenthesis, although we don't strictly enforce this.

smilburn1990: Want me to change it @DamianMullins ?

packages/f-services/src/services/shared.service.js
```diff
@@ -4,6 +4,14 @@
  * @module shared/service
  */
 
+/**
+ * Returns the locale in correct format 'en-GB'
```
DamianMullins: Why wouldn't a string be in the correct format?

packages/f-services/src/services/shared.service.js
DamianMullins: I think it's fine as it is here.

packages/f-services/src/services/shared.service.js
smilburn1990: Capitalization format. As opposed to `en-gb` @DamianMullins

smilburn1990: Changed it to "case format"

BanksySan (dismissed):

packages/f-services/src/services/shared.service.js
DamianMullins: Why would it be lower case?

smilburn1990: Because that's what `$i18n.locale` gives you

simonsandersje (dismissed):

PR #65, opened 2019-09-26
f-services@0.7.0-remove-method

smilburn1990:
### Removed 
- `transformLocale` method as it was not working

### DIFF
--- a/packages/f-services/CHANGELOG.md
+++ b/packages/f-services/CHANGELOG.md
@@ -25,7 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 *September 26, 2019*
 
  ### Added
-- `transformLocale` method for transforming the locale
+- `transformLocale` method for changing the case format of the locale
 
 
 0.4.0
--- a/packages/f-services/CHANGELOG.md
+++ b/packages/f-services/CHANGELOG.md
@@ -25,7 +25,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 *September 26, 2019*
 
  ### Added
-- Made a new method for transforming the locale
+- `transformLocale` method for transforming the locale
 
 
 0.4.0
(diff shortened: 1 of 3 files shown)


dandel10n (dismissed):

simonsandersje (dismissed):

packages/f-services/src/services/shared.service.js
```diff
@@ -13,18 +13,15 @@
  * @returns {String}
  */
 const getLocale = (tenantConfigs, tenantString, globalTenant) => {
-    const locale = tenantString === '' ? globalTenant?.locale : tenantString;
-
-    // makes sure the locale is the same format as the tenants - 'en-GB'
-    let transformedLocale = `${locale.slice(0, -3)}${locale.slice(locale.lastIndexOf('-')).toUpperCase()}`;
+    let locale = tenantString === '' ? globalTenant?.locale : tenantString;
```
kevinrodrigues, line 16: Do the tests need updating?

packages/f-services/CHANGELOG.md
```diff
@@ -4,6 +4,30 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+0.7.0
+------------------------------
+*September 26, 2019*
```
kevinrodrigues: Date needs bumping. 27th 📅

packages/f-services/src/services/shared.service.js
smilburn1990: Nope - Didn't get a chance to update them when I added it in the first place 👍

kevinrodrigues (approved):

packages/f-services/CHANGELOG.md
dandel10n: Technically changes were made yesterday and PR can wait days to be approved so I vote for not bumping the date if it is the latest change and there are no conflicts with other PRs

dandel10n (approved):

packages/f-services/CHANGELOG.md
kevinrodrigues: I agree but this is what we have done for all fozzie components in the past.

PR #75, opened 2019-10-02
f-services@0.10.0-add-tests-for-new-services

smilburn1990:
### Added
- Tests for new services that have been added (`getWindowWidth`, `getWindowHeight`, `addEvent`, `removeEvent`)

### Changed
- `webpack.config.js` and packages for a better build
- Name of the export in the `index.js` to `sharedServices` for consistency

### DIFF
--- a/packages/f-services/src/services/tests/Services.test.js
+++ b/packages/f-services/src/services/tests/Services.test.js
@@ -1,3 +1,4 @@
+import { throttle } from 'lodash-es';
 import sharedServices from '../shared.service';
 
 describe('sharedServices', () => {
@@ -98,7 +99,6 @@ describe('sharedServices', () => {
             // Arrange
             const locale = 'en-GB';
 
-
             // Act
             const result = sharedServices.getTheme(locale);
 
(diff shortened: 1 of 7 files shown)


simonsandersje (approved):

cak8 (approved):

packages/f-services/src/services/tests/Services.test.js
```diff
@@ -104,12 +99,89 @@ describe('sharedServices', () => {
...
-
             // Act
             const result = sharedServices.getTheme(locale);
 
             // Assert
             expect(result).toBe('je');
         });
     });
+
+    const windowWidth = 667;
```
kevinrodrigues, line 110: This should sit inside a before each or somewhere within a  describe.

packages/f-services/src/services/tests/Services.test.js
```diff
@@ -104,12 +99,89 @@ describe('sharedServices', () => {
...
+            // Assert
+            expect(result).toBe(375);
+        });
+    });
+
+    describe('addEvent', () => {
+        const eventName = 'resize';
+        const callBackFunction = jest.fn();
+
+        it('returns the correct event to be listened to WITHOUT being throttled', () => {
```
kevinrodrigues, line 148: What's the main difference between this test and the next? Looks like they're the same apart from the throttle value, not sure what it's really testing? I'd expect these two tests to test something like the throttle time to have a delay between calls?

smilburn1990: It just tests that the `throttle` toggle still runs as expected. I.e. if there is a `throtteTime` the method used lodash throttle on the event listener @kevinrodrigues 👍

packages/f-services/src/services/tests/Services.test.js
smilburn1990: I will add to next PR 👍

packages/f-services/src/services/tests/Services.test.js
kevinrodrigues: So I think in that case then one of these tests should check that when you pass in a `throttleTime` of greater than `0` `lodash's throttle` should be called and the second test should test that when you don't pass in a `throttleTime` or the time is `0` `lodash throttle` isn't called? Because even if the `throttleTime` is not greater than `0` the `callBackFunction` is always called anyway. If it's greater than zero then the `throttle(callBackFunction, throttleTime)` is called with two arguments. (in both cases the callback function always fires)

PR #79, opened 2019-10-04
f-services@0.11.0-add-package-for-ssr-issue

smilburn1990:
### Added
- `window-or-global` package for SSR issue
- `globalObject` in webpack config output


tommcclean (approved):
Nice find

ashleynolan (approved):

PR #81, opened 2019-10-04
f-metadata@v0.4.0 – Changing Braze to be a dynamic import

ashleynolan:
### Changed
- Changing Braze to be a dynamic import (to save on bundle size of `f-metadata`)

### DIFF
--- a/packages/f-metadata/CHANGELOG.md
+++ b/packages/f-metadata/CHANGELOG.md
@@ -6,17 +6,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.4.0
 ------------------------------
-*October 07, 2019*
+*November 15, 2019*
 
- ### Changed
-- Changing Braze to be a dynamic import (to save on bundle size of `f-metadata`)
+### Added
+- Support for handling content cards through SDK methods
+
+### Changed
+- Changing Braze to be a dynamic import
+- Don't bundle `f-metadata`, leave that to the consuming component.
 
 
 v0.3.0
 ------------------------------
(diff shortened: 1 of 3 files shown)


ashleynolan:
@xander-marjoram @chrisjtownsend Would be good to get your thoughts on this. I haven't fully tested it in something like HomeWeb yet, which would need to be done, but would mean we don't bundle in all of Braze when including the `f-metadata` package potentially.

DamianMullins:
@ashleynolan I'm very interested to see if this works, if so I'm guessing we could code split the config for components like f-searchbox, right?

DamianMullins (dismissed):

ashleynolan:
@DamianMullins It doesn't in its current state atm. Mainly as the build step removes the webpack chunk comments and so it then can't find the file when it's imported into HomeWeb.

I think in this case, I'm going to change the build of this bundle to not be packaged up as a library, so that it can be code-split by the consuming package (atm the code-split happened when the package is built and deployed to NPM), and I can't find any info online with respect to how this could work otherwise…

DamianMullins:
@ashleynolan Is the dynamic import working now?

ashleynolan:
@DamianMullins Yep, but only because the babel transpilation gets handled on the consuming application side now, rather than in the component itself.  So you have to tell your build to compile this module – which would be a bit of a nightmare if we had to do it for library components likethe searchbox/header 😞

kevinrodrigues:
@ashleynolan Is this one OK to close now?

ashleynolan:
@kevinrodrigues Yeh – this went in as part of v1.0.0

PR #85, opened 2019-10-10
f-header@2.0.0-beta.21-add-tests-for-new-methods

smilburn1990:
### Added
- Tests for the new methods in `Navigation.vue`

### Changed
- Logic that gets and sets the order count
- Comments around logic


packages/f-header/yarn.lock
```diff
@@ -0,0 +1,7240 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
xander-marjoram: I'm not sure we want to add this file

packages/f-header/CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v2.0.0-beta-21
+------------------------------
+*October 10, 2019*
+
+### Changed
+ - The way data is passed around in the order count logic
+ - Version number
```
xander-marjoram: I think that's a given 😄

packages/f-header/src/components/Navigation.vue
```diff
@@ -55,11 +55,11 @@
                     @mouseleave="closeNav"
                     @keyup.esc="closeNav">
                     <a
-                        class="c-nav-list-text"
-                        href="/"
                         :tabindex="isBelowMid ? -1 : 0"
                         :aria-expanded="!isBelowMid && navIsOpen ? 'true' : 'false'"
                         :aria-haspopup="isBelowMid ? false : true"
+                        class="c-nav-list-text"
```
ashleynolan, line 61: Comment out of interest than anything else – is this an auto-fix thing that put's certain attributes higher than others on an element?

smilburn1990: Yeah linter/IDE does it on save @ashleynolan

packages/f-header/src/components/Navigation.vue
```diff
@@ -290,14 +291,25 @@ export default {
 
         returnLogoutUrl () {
             return `${this.accountLogout.url}${this.returnUrl}`;
+        },
+
+        isOrderCountValid () {
+            return this.isOrderCountSupported && !this.getAnalyticsBlob;
```
xander-marjoram, line 297: If `getAnalyticsBlob` returns an object rather than `false`, will this become `false`?

smilburn1990: Yeah. if there is already a blob then the order count should not be updated @xander-marjoram

xander-marjoram: Cool, thanks. That might be a valuable code comment

packages/f-header/src/components/Navigation.vue
```diff
@@ -270,7 +270,8 @@ export default {
             navIsOpen: false,
             currentScreenWidth: sharedServices.getWindowWidth(),
             userInfo: this.userInfoProp,
-            orderCountInfo: false
+            currentTime: new Date().getTime(),
```
ashleynolan: I'm not sure if we should store current time, as by definition, it may not then be the `current` time when it's evaluated.

I know it's an edge case, but would rather we just get the time when we need to evaluate it.

packages/f-header/src/components/Navigation.vue
smilburn1990: Will add now 👍

packages/f-header/src/components/Navigation.vue
```diff
@@ -372,42 +386,18 @@ export default {
...
 
         // Sets the order count info in local storage
-        setAnalyticsBlob () {
-            window.localStorage.setItem('je-analytics', JSON.stringify(this.orderCountInfo));
-        },
-
-        // Gets the order count info in local storage
-        getLocalAnalyticsBlob () {
-            return window.localStorage.getItem('je-analytics');
+        setAnalyticsBlob (data) {
```
ashleynolan, line 398: Would be good to add some JSDoc comments to these functions.

packages/f-header/src/components/Navigation.vue
```diff
@@ -339,7 +351,9 @@ export default {
                 });
                 if (data.isAuthenticated) {
                     this.userInfo = data;
-                    this.saveUserData();
+                    if (this.isOrderCountValid || this.isOrderCountOutOfDate) {
+                        this.fetchOrderCountAndSave();
```
xander-marjoram, line 360: No change required now, but I've always thought this method breaks the single responsibility principle...
Something like this:

```
var data = this.fetchOrderCountData();
this.saveOrderCountData(data);
```

packages/f-header/src/components/Navigation.vue
smilburn1990: Should I create a computed property for it? @ashleynolan

packages/f-header/src/components/Navigation.vue
```diff
@@ -372,42 +386,18 @@ export default {
...
-            return window.localStorage.getItem('je-analytics');
+        setAnalyticsBlob (data) {
+            window.localStorage.setItem('je-analytics', JSON.stringify(data));
         },
 
         // Updates the user information with the count
-        enrichUserDataWithCount () {
-            this.userInfo.orderCount = this.orderCountInfo.Count;
+        enrichUserDataWithCount (data) {
+            this.userInfo.orderCount = data.count.Count;
```
ashleynolan: If this is the only parameter of data used, would it be better not to just pass in the `count` to the function rather than a whole JSON object?

packages/f-header/src/components/Navigation.vue
smilburn1990: Will bare in mind for the future

packages/f-header/src/components/Navigation.vue
```diff
@@ -290,14 +291,26 @@ export default {
...
+            return this.isOrderCountSupported && !this.getAnalyticsBlob;
+        },
+
+        isOrderCountOutOfDate () {
+            return this.localOrderCountExpires < this.currentTime;
+        },
+
+        // Gets the order count info in local storage
+        getAnalyticsBlob () {
+            return window.localStorage.getItem('je-analytics') || false;
```
xander-marjoram, line 311: Are we likely to run into any problems with SSR, using `window` here?

packages/f-header/src/components/Navigation.vue
```diff
@@ -372,42 +386,18 @@ export default {
...
-            if (!this.getLocalAnalyticsBlob) {
-                this.fetchOrderCountAndSave();
-            }
-            if (localOrderCountExpires < currentTime) {
-                this.fetchOrderCountAndSave();
-            }
-            this.enrichUserDataWithCount(localOrderCount);
-            this.pushUserData();
+        pushUserData (data) {
+            window.dataLayer.push(data);
```
ashleynolan, line 409: I think it makes more sense to push `this.userInfo` here (if that is exactly the same as `data`. Otherwise, this function shouldn't really be called `pushUserData`, as it could be any data that's passed in that is getting pushed.

packages/f-header/src/components/Navigation.vue
smilburn1990: No. Only with components used by components from my experience 👍 @xander-marjoram

packages/f-header/src/components/Navigation.vue
ashleynolan: I would just call it in the function when it gets used for now (as you're only checking the current time in one place).  So just making it a local variable in that function would make the most sense.  That way it'll be checking against the most accurate time (as potentially if it's stored in data or computed, the time could be out of date when the comparison is made).

packages/f-header/src/components/Navigation.vue
ashleynolan: It would only be an SSR issue if this get's called by the server. So as long as this function is called (or the function that calls this function) through `mounted` then it should be ok.

Could always add a check for `window` if we ever think that may be a problem though.

smilburn1990: That code had been on `ieng-staging` already and works fine 👍

smilburn1990: This code had been on `ieng-staging` and works fine 👍

packages/f-header/src/components/Navigation.vue
```diff
@@ -371,43 +390,23 @@ export default {
...
-            }
-            if (!this.getLocalAnalyticsBlob) {
-                this.fetchOrderCountAndSave();
-            }
-            if (localOrderCountExpires < currentTime) {
-                this.fetchOrderCountAndSave();
-            }
-            this.enrichUserDataWithCount(localOrderCount);
-            this.pushUserData();
+        pushUserData (userInfo) {
```
ashleynolan: Sorry - think wasn't very clear with my last comment.

This function should either push the userInfo directly `window.dataLayer.push(this. userInfo);` or the function name should be changed if it's just pushing an object (as there are no checks to ensure that it's user data being sent to the function, so it could just then be called `pushData`).

I think it probably makes sense just to push `this.userInfo ` unless there's a reason for not doing that?

packages/f-header/src/components/Navigation.vue
```diff
@@ -371,43 +390,23 @@ export default {
...
+         */
+        setAnalyticsBlob (data) {
+            window.localStorage.setItem('je-analytics', JSON.stringify(data));
         },
 
         // Updates the user information with the count
-        enrichUserDataWithCount () {
-            this.userInfo.orderCount = this.orderCountInfo.Count;
+        enrichUserDataWithCount (count) {
+            this.userInfo.orderCount = count;
```
ashleynolan, line 404: You're updating `this.userInfo` here, but pushing `this.userInfoProp` (line 384). Should it be `userInfo` that is pushed as that will have the orderCount as well?

packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -196,21 +188,205 @@ describe('Navigation', () => {
...
+            const spy = jest.spyOn(Storage.prototype, 'setItem');
+
+            // Assert
+            wrapper.vm.setAnalyticsBlob();
+
+            // Act
+            expect(spy).toHaveBeenCalled();
+        });
+    });
+
```
BanksySan, line 368: These tests could be refactored/simplified if they were data driven.  They all see to do the same action, just with different inputs/expectations.

BanksySan (changes requested):

packages/f-header/src/components/tests/Navigation.test.js
DamianMullins: The approach here is intentional; the benefit is that they are very easy to understand, especially when viewed individually, without needing to jump around the file to figure out what data is being used or how the unit under test is being called.

We do make use of `beforeEach` etc for some of our more complex unit tests.

Would be interesting to see an example of data-driven approach though for comparison and to consider using that instead.

BanksySan (approved):

simonsandersje (approved):

PR #90, opened 2019-10-21
f-header@2.0.0-beta.23-fix-mobile-bug

smilburn1990:
### Changed
 - Added missing class toggle back in
- Changed logic slightly for `ordercount` endpoint


packages/f-header/src/components/Header.vue
```diff
@@ -152,6 +140,17 @@ export default {
...
 }
 
+    // when the off-screen navigation is active (on mobile), it fixes to the top of the screen.
+    // this stops the content being forced upwards when this happens (preventing slight visual glitch)
+    .is-navInView {
+        @include media('<mid') {
+            overflow: hidden;
+            position: fixed;
+            width: 100%;
+            z-index: 9999;
```
DamianMullins: [You're looking for these](https://github.com/justeat/fozzie/blob/153c0b962a7c978ec0d7427cd012eae1b9eb0976/src/scss/settings/_variables.scss#L149-L155) 😃  `z-index: zIndex(high);`

tommcclean:
As a side note - I solve this problem in my projects using the head to add a class to the body that prevents overflow - https://nuxtjs.org/api/pages-head/
Even though visually you wouldn't notice the scroll; it will still be scrolling the body.
image
*The example is nuxt, but it just uses vue-meta I think

packages/f-header/src/components/Header.vue
smilburn1990: @DamianMullins - Added new variable `aboveHighest` to fozzie

smilburn1990: https://github.com/justeat/fozzie/pull/248

simonsandersje (dismissed):

packages/f-header/src/components/Header.vue
```diff
@@ -152,6 +140,17 @@ export default {
...
+    // this stops the content being forced upwards when this happens (preventing slight visual glitch)
+    .is-navInView {
+        @include media('<mid') {
+            overflow: hidden;
+            position: fixed;
+            width: 100%;
+            z-index: zIndex(aboveHighest);
+        }
+    }
+
```
markedup: Nit-pick: indenting looks off.

markedup (dismissed):
Small nit-pick, otherwise looks good.

packages/f-header/src/components/Header.vue
smilburn1990: Was already like that so I thought it might be one of the conventions I'm not used to 🤷‍♂ @markedup

markedup: @smilburn1990 Sorry, I meant this chunk looks indented a block more than the preceding closing brace—same as it was when it was inside that closing brace (so cut and paste symptom?).

packages/f-header/src/components/Header.vue
```diff
@@ -1,7 +1,7 @@
 <template>
     <header
         :data-theme="theme"
-        :class="['c-header', { 'c-header--transparent c-header--gradient': showTransparentHeader }]">
+        :class="['c-header', { 'c-header--transparent c-header--gradient': showTransparentHeader }, { 'is-navInView': mobileNavIsOpen }]">
```
DamianMullins: Dynamic classes should be inside a single object i.e.

```js
:class="['c-header', {
    'c-header--transparent c-header--gradient': showTransparentHeader,
    'is-navInView': mobileNavIsOpen
}]"
```

packages/f-header/src/components/Header.vue
```diff
@@ -152,6 +140,17 @@ export default {
     }
 }
 
+    // when the off-screen navigation is active (on mobile), it fixes to the top of the screen.
+    // this stops the content being forced upwards when this happens (preventing slight visual glitch)
+    .is-navInView {
```
DamianMullins: This style is now global and not scoped to the `c-header` class, you could put it back in its original place and update the selector to be `&.is-navInView`?

DamianMullins (dismissed):

packages/f-header/src/components/Header.vue
```diff
@@ -152,118 +155,118 @@ export default {
...
+// Adds a border to the header to separate it from the
+// main content at all widths
+.c-header--bordered {
+    border-bottom: $header-separator solid $header-border-color;
+}
 
-    .c-header--transparent {
-        background-color: transparent;
-        border: none;
+.c-header--transparent {
```
ashleynolan: Looks like a lot of indenting is being removed which is intended to be there (to indicate modifier/child classes).

Can you revert the indentation changes on these blocks?

smilburn1990: I was told to in a previous comment @ashleynolan

smilburn1990: Changed back 👍

packages/f-header/src/components/Header.vue
```diff
@@ -1,7 +1,11 @@
 <template>
     <header
         :data-theme="theme"
-        :class="['c-header', { 'c-header--transparent c-header--gradient': showTransparentHeader }]">
+        :class="['c-header', {
+            'c-header--transparent c-header--gradient': showTransparentHeader,
+            'is-navInView': mobileNavIsOpen
```
ashleynolan: Shouldn't this class be applied to the `html` element via JS (as it's also referenced in the `navigation.vue` to stop page scrolling)?

Not sure how you'd stop the page scrolling otherwise, as the overflow needs to be on the outmost element.

(that's how it currently works on live - might just have been missed when it got ported over to Vue)

packages/f-header/src/components/Header.vue
ashleynolan: 🐼  Definitely should be in there as it's convention we use across all components/styling

ashleynolan:
@tommcclean @smilburn1990 That's how we currently do it with the old component.  Looks like the styles are already there for this functionality, but the class just isn't being added anywhere to the HTML element.

Would probably be an easier fix to ensure this classname gets added ([here's how it's implemented in the old header for ref](https://github.com/justeat/f-header/blob/81cf4db6b6b0e4e32119c8655682a73bbfbf9b50/src/js/index.js#L36))

packages/f-header/src/components/Header.vue
smilburn1990: @ashleynolan - it is not used in `Navigation.vue`, do you want me to remove it? Also where are you seeing this in the old header?

ashleynolan: [Here in Navigation.vue](https://github.com/justeat/fozzie-components/blob/master/packages/f-header/src/components/Navigation.vue#L491-L511).

Might be worth pinging me on Slack, as I think just adding the classname via a method might solve your issue.

smilburn1990: @ashleynolan - Just reverted and pushed. Seems to work much better 👍

packages/f-header/src/components/Navigation.vue
```diff
@@ -326,6 +326,8 @@ export default {
     methods: {
         onNavToggle () {
             this.navIsOpen = !this.navIsOpen;
+            // This is added to remove the ability to scroll the page content when the mobile navigation is open
+            document.documentElement.classList.toggle('is-navInView');
```
ashleynolan: Looks good. Might need to [add a Vue equivalent for this too](https://github.com/justeat/f-header/blob/81cf4db6b6b0e4e32119c8655682a73bbfbf9b50/src/js/index.js#L38-L42) (I think on Home it may need this class).

simonsandersje (dismissed):

packages/f-header/src/components/Navigation.vue
smilburn1990: The `vue-meta` approach won't work on SSR @ashleynolan

ashleynolan: `vue-meta` approach?  Not sure if I'm getting confused – just meant that there should be another class added to the `<html>` if the header is transparent (could use the `data.isTransparent` value in `Header.vue` to check for this?

smilburn1990: No I mean for the other comment where you said I need to make it the vue version. But yeah I can toggle it on that 👍 @ashleynolan

ashleynolan: Gotcha – I just meant that the other class needed to be in this `.vue` file too (like the one added on this line is now)

packages/f-header/src/components/Navigation.vue
```diff
@@ -326,6 +331,8 @@ export default {
     methods: {
         onNavToggle () {
             this.navIsOpen = !this.navIsOpen;
+            // This is added to remove the ability to scroll the page content when the mobile navigation is open
+            document.documentElement.classList.toggle(this.isTransparent ? 'is-navInView--noPad' : 'is-navInView');
```
ashleynolan: Small one – it should add both classes if the nav is transparent.

ashleynolan (approved):

simonsandersje (approved):

PR #91, opened 2019-10-24
f-header@2.0.0-beta.24-update-dk-no-copy

smilburn1990:
### Changed
 - Copy on `NO` and `DK`

### DIFF
--- a/packages/f-header/package.json
+++ b/packages/f-header/package.json
@@ -35,7 +35,7 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-services": "^0.11.0",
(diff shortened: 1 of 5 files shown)


packages/f-header/package.json
```diff
@@ -35,7 +35,7 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-services": "0.11.0",
+    "@justeat/f-services": "^0.11.0",
```
simonsandersje: I think we'd agreed to stick to defined versions... in which case - lose the hat!

markedup (dismissed):

DamianMullins (approved):

markedup (approved):

PR #94, opened 2019-11-04
fix: improving packages.json and run prepare on all packages

pikax:
Had some issues on installing the packages when I clone the repo

The first time you run the dependencies they don't build, that forces you to manually build `f-services`, with this PR it will run prepare on all the projects

Also remove the dependencies from packages.json since we don't have any source code in the root, therefore no dependencies are needed.


smilburn1990 (dismissed):

package.json
```diff
@@ -61,12 +62,5 @@
     "sass-loader": "7.1.0",
     "vue-svg-loader": "0.12.0",
     "vue-template-compiler": "2.6.10"
-  },
-  "dependencies": {
-    "@justeat/f-footer": "file:packages/f-footer",
```
ashleynolan, line 66: I think these are here so that when a package imports another package from the monorepo, it uses the local package version (or something weird like that). @dandel10n set this up, so I'm not sure, but I'm pretty certain there's a reason for it being there! 😄

ashleynolan (changes requested):

package.json
pikax: Local packages should been pickup automatically using `lerna` + `yarn`

If you check other [monorepo project](https://github.com/vuejs/vue-next/blob/master/package.json) you can see the dependencies are not set.

How can I see if I broke something? 😄

ashleynolan: Just been trying to find where/why it was added. The first occurrence of it [was here](https://github.com/justeat/fozzie-components/commit/3d86abc0ce70d11d4360f643ed0e30221196355c#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR10-R15), and then [footer one was added here](https://github.com/justeat/fozzie-components/commit/c1529f560bd5cceb3501d16883c9abd0f815c553#diff-9cc4e14a8af45ccf473aefdb4cb63639R16).

I think it's used to effectively symlink the local package in the monorepo, but the documentation on it is very patchy. In terms of testing it, I'm not sure either – I guess from the other side of things, does having these lines in here affect anything negatively?  Seems strange it would've been added if it wasn't needed…

pikax: Not sure if affects anything having those lines, probably only if we moved the packages from **packages/** to other folder, I guess, but that's unlikely. My concern is having something in there that is not doing anything and adds unnecessary complexity.

**root/package.json** should be used for project dev dependencies (lerna, etc) or for the dependencies on the main project ( [vue](https://github.com/vuejs/vue), when you have a project in the root and also packages )

I spent quite a lot of time to get `f-header` to build, because `f-services` didn't build when using `yarn install` :/

My concern having this dependencies is causing confusion, IMHO if is not doing anything it shouldn't be there.

If this change was required to fix an error, I can try to reproduce on my machine, @dandel10n do you remember what was the error?

ashleynolan: Definitely agree that if it's not doing anything it shouldn't be there – just need to make sure it's not before removing it.

@dandel10n is away on maternity leave, so I think we're going to have to test this thoroughly and see if removing these lines has any effect or not. @xander-marjoram might be able to help what this was used for too, as he worked on the header/footer setup with Maryia too

DamianMullins: I remember there was an issue where incorrect packages were being picked up like @ashleynolan mentioned. It wasn't referencing the local copy. I'm guessing that what this fix is here for.

I believe @dandel10n is on maternity leave.

packages/f-footer/.eslintignore
```diff
@@ -0,0 +1,2 @@
+node_modules
```
ashleynolan, line 1: Do we need these duplicated in the package folders?

Would think the one in the root would get inherited (like with the base eslint config)

pikax: I added because when building the f-header, eslint was running rules against the `@justeat/f-services`, it isn't really necessary, might need to do some tests to check if the root is used by default

ashleynolan: I think `node_modules` inside the `/packages` folder may need to be added to the root eslint ignore specifically (as it's mentioned in the eslint docs), so may just need to add those paths in there if it doesn't ignore them otherwise.

package.json
dandel10n: Hi everyone! If my "mom brain" is still working I believe this structure relates to [the docs](https://github.com/lerna/lerna#common-devdependencies). If we want common dev dependencies to be stored in the root for easier updating them (only in one place not in every package) we need to use relative `file:` specifiers for children packages. The `lerna link convert` does this for you.

dandel10n: @pikax  but I agree with Ash, if you can properly test and see if your changes don't break things, than we can remove it. Try to remove all node_modules and build packages from the root without these relative paths. I think that the root will not know about children dependencies but I am not sure.

packages/f-footer/.eslintignore
pikax: Just double check seems the root folder doesn't get pickup at build.
Without it I get the error when building all: 
```log
             ^
@justeat/f-footer: 1170 errors and 38 warnings found.
@justeat/f-footer: 955 errors potentially fixable with the `--fix` option.
lerna ERR! MaxBufferError: stdout maxBuffer exceeded
lerna ERR!     at PassThrough.stream.on (C:\dev\pikax\github\fozzie-components\node_modules\get-stream\index.js:41:19)
lerna ERR!     at emitOne (events.js:121:20)
lerna ERR!     at PassThrough.emit (events.js:211:7)
lerna ERR!     at addChunk (_stream_readable.js:263:12)
lerna ERR!     at readableAddChunk (_stream_readable.js:246:13)
lerna ERR!     at PassThrough.Readable.push (_stream_readable.js:208:10)
lerna ERR!     at PassThrough.Transform.push (_stream_transform.js:147:32)
lerna ERR!     at PassThrough.afterTransform (_stream_transform.js:88:10)
lerna ERR!     at PassThrough._transform (_stream_passthrough.js:42:3)
lerna WARN complete Waiting for 1 child process to exit. CTRL-C to exit immediately.
lerna ERR! lerna stdout maxBuffer exceeded
error Command failed with exit code 1.
```

## Current Commit
using the current commit: 
```log
@justeat/f-header:  @ ./src/components/Header.vue?vue&type=script&lang=js&
@justeat/f-header:  @ ./src/components/Header.vue
@justeat/f-header:  @ ./src/index.js
@justeat/f-header:  @ C:/dev/pikax/github/fozzie-components/node_modules/@vue/cli-service/lib/commands/build/entry-lib.js
@justeat/f-header:   File                        Size                    Gzipped
@justeat/f-header:   dist\f-header.umd.min.js    105.81 KiB              32.66 KiB
@justeat/f-header:   dist\f-header.umd.js        505.10 KiB              119.54 KiB
@justeat/f-header:   dist\f-header.common.js     504.72 KiB              119.40 KiB
@justeat/f-header:   dist\f-header.css           12.70 KiB               2.98 KiB
@justeat/f-header:   Images and other types of assets omitted.
@justeat/f-header: Done in 135.31s.
lerna success run Ran npm script 'prepare' in 4 packages in 200.3s:
lerna success - @justeat/f-footer
lerna success - @justeat/f-header
lerna success - @justeat/f-services
lerna success - @justeat/f-vue-icons
Done in 206.68s.
```

Also I get this logs: 
```log
@justeat/f-header:   1:4320  error    Operator '=' must be spaced
                                                                                 space-infix-ops
@justeat/f-header:   1:4322  error    Requires a space before '}'
                                                                                 block-spacing
@justeat/f-header:   1:4322  error    Missing semicolon
                                                                                 semi
@justeat/f-header:   1:4325  error    Missing semicolon
                                                                                 semi
@justeat/f-header:   1:4325  error    Requires a space before '}'
                                                                                 block-spacing
@justeat/f-header:   1:4329  error    Newline required at end of file but not found
                                                                                 eol-last
@justeat/f-header: ✖ 1208 problems (1170 errors, 38 warnings)
```

By the line number `1` and col `4329` seems to be a `min` file, it still seems the eslint is still doing `min` files, not sure why.

package.json
pikax: I've actually removed and reinstall it a few times to confirm this PR is working, also done on my personal laptop.

I used node `8.16.1` and the current `LTS` `12`, my yarn version is `1.19.1`.

Can fetch the branch and test it please?

NOTE: 🎉 🎉 🎉 🎉 👶  @dandel10n congrats 👶  🎉 🎉 🎉 🎉

ashleynolan: I've done a quick test on that branch and it all seems to be working ok for me (on node `v10.14.2`, yarn `v1.15.2`).

I guess if we have any issues with these packages in the future, it should be easy enough to add this back in 😄

packages/f-footer/.eslintignore
ashleynolan: Was `yarn lint` failing for you when running it from the root then?

I've just tried to replicate this on master and `yarn lint` comes back fine for me with no errors…

pikax: this is doing `lerna run prepare` on the root

ashleynolan: Been trying to test this this afternoon and getting some really strange results.

So on master, running `lerna run lint` gives no errors, but running `lerna run prepare` returns a load of issues like you mentioned. The only difference (that I can see) is that by calling lint directly, `vue-cli-service lint` is called directly, rather than in the prepare call, `yarn run lint` is called, and then the `vue-cli-service` call.

Seems really strange how this results in one call paying attention to eslint ignore files and the other trying to lint the `dist` folders and coming up with thousands of errors…

Might try and investigate a little more tomorrow if I get time, but tbh if adding the `.eslintignore` files does fix this issue, maybe we just do that and forget about it! 😉

ashleynolan (approved):

DamianMullins (approved):

PR #98, opened 2019-11-08
f-header@2.0.0-beta.25-fix-logout-redirect

smilburn1990:
### Changed
- `returnUrl` method to return the correct redirect


packages/f-header/src/components/Navigation.vue
```diff
@@ -285,16 +285,23 @@ export default {
...
+            }
+            if (typeof document !== 'undefined') {
+                return encodeURIComponent(document.location.pathname);
+            }
+            return false;
         },
 
         returnLoginUrl () {
-            return `${this.accountLogin.url}${this.returnUrl}`;
+            return `${this.accountLogin.url}${this.returnUrl ? this.returnUrl : ''}`;
```
DamianMullins: Do you need to add the `?returnUrl=` part here and below?

DamianMullins: Also, I don't think you need the ternary, you can just pass an empty value to `returnUrl`.

packages/f-header/src/components/Navigation.vue
```diff
@@ -285,16 +285,15 @@ export default {
         },
 
         returnUrl () {
-            if (!this.$route && typeof document !== 'undefined') return encodeURIComponent(document.location.pathname);
-            return encodeURIComponent(this.$route ? this.$route.name : '');
-        },
-
-        returnLoginUrl () {
```
DamianMullins, line 292: Not sure why you've removed these? The login & logout URLs should look like this `/account/login?returnurl=%2Farea%2Fyo8-barlby` and `/account/logout?returnurl=%2Farea%2Fyo8-barlby`, right?

So you need the computed properties so that you can append the return URL.

What I'd recommend is to remove the `returnUrl` part from the config files as it doesn't need to be there 

```js
accountLogin: {
    ...
    url: '/account/login?returnurl='
    ....
}
````

To this 

```js
accountLogin: {
    ...
    url: '/account/login'
    ....
}
````

And then update the login link computed property to look like this 

```js
return ${this.accountLogin.url}?returnurl=${this.returnUrl}`;
```

DamianMullins (approved):

dandel10n (approved):

PR #99, opened 2019-11-08
f-services@0.12.0 - Updating build and packages

ashleynolan:
### Changed
- Switched over to rollup for bundling as webpack was giving strange results when importing as a library


packages/f-services/src/services/shared.service.js
```diff
@@ -16,7 +16,7 @@ import root from 'window-or-global';
  * @returns {string}
  */
 const getLocale = (tenantConfigs, tenantString, globalTenant) => {
-    let locale = tenantString === '' ? globalTenant?.locale : tenantString;
+    let locale = tenantString === '' ? globalTenant ?.locale : tenantString;
```
DamianMullins: Should remove the space in `globalTenant ?.locale`

ashleynolan: Have removed this commit – looks like ESLint is formatting it on save

BanksySan (approved):

kevinrodrigues (approved):

PR #101, opened 2019-11-08
f-services@0.13.0 - Bundling update

ashleynolan:
### Changed
- Making sure that necessary 3rd party dependencies are bundled as part of the package


smilburn1990 (approved):

BanksySan (commented):
What is `cjs` & `umd`?

PR #105, opened 2019-11-15
f-metadata@v1.0.0 - Dynamic bundling, content card support and adding userID

ashleynolan:
### Added
- Support for handling content cards through SDK methods
- User ID to `initialiseBraze` function
- Support for GTM with Braze

### Changed
- Changing Braze to be a dynamic import
- Don't bundle `f-metadata`, leave that to the consuming component.

### DIFF
--- a/packages/f-metadata/src/index.js
+++ b/packages/f-metadata/src/index.js
@@ -1,10 +1,40 @@
-const initialiseBraze = (apiKey, enableLogging = false, disableComponent = false) => {
+const initialiseBraze = (options = {
+    apiKey: null,
+    userId: null,
+    enableLogging: false,
+    disableComponent: false,
+    callbacks: {
+        handleContentCards: null
+    }
+}) => {
     if (typeof window !== 'undefined') {
-        if (!disableComponent) {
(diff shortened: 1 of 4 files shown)


cak8 (approved):

packages/f-metadata/src/index.js
```diff
@@ -1,10 +1,40 @@
...
+                        appboy.display.automaticallyShowNewInAppMessages();
+
+                        appboy.changeUser(options.userId, () => {
+                            window.dataLayer.push({
+                                event: 'appboyReady'
+                            });
+                        });
+
+                        window.appboy = appboy;
+                        appboy.openSession();
```
BanksySan, line 28: Should this bit be async?  i.e. is it making a network call?

ashleynolan: The braze docs don't indicate it should be (which I agree is a bit weird). It seems to work fine as it is though – we'll be testing this further once marketing start adding more content cards so I would guess we'll see if we hit any problems then that need us to tweak it based on that.

BanksySan (approved):

PR #106, opened 2019-11-18
Fixed Appboy ready event issue for GTM

cak8:
Moved `ChangeUser` after `openSession` to allow appboy to load before `appboyready` event is being sent to GTM

### DIFF
--- a/packages/f-metadata/src/index.js
+++ b/packages/f-metadata/src/index.js
@@ -18,14 +18,14 @@ const initialiseBraze = (options = {
 
                         appboy.display.automaticallyShowNewInAppMessages();
 
+                        appboy.openSession();
+                        window.appboy = appboy;
+
                         appboy.changeUser(options.userId, () => {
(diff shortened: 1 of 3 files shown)


packages/f-metadata/src/index.js
```diff
@@ -18,15 +18,15 @@ const initialiseBraze = (options = {
...
+
                         appboy.changeUser(options.userId, () => {
                             window.dataLayer.push({
                                 event: 'appboyReady'
                             });
                         });
 
-                        window.appboy = appboy;
-                        appboy.openSession();
-
```
BanksySan, line 29: Is there a reason to move the session opening?

cak8: Yes to allow appboy to load before the event is sent to GTM. I followed the same way it's done in the GTM script

DamianMullins (approved):

BanksySan (approved):

PR #107, opened 2019-11-19
f-footer@2.0.0-beta.31-toggle-off-ie-footer-links

smilburn1990:
### Added
- Toggle for `ie` courier links

### DIFF
--- a/packages/f-footer/src/components/Footer.vue
+++ b/packages/f-footer/src/components/Footer.vue
@@ -13,7 +13,7 @@
         <div class="c-footer-light">
             <div class="c-footer-container">
                 <div
-                    v-if="copy.linkButtonList.length && locale !== 'en-IE'"
+                    v-if="copy.linkButtonList.length && showCourierLinks"
                     data-test-id="courierLinks-wrapper"
                     class="c-footer-row c-footer-row--noBottomPad">
                     <button-list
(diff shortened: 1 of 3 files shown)


simonsandersje (dismissed):

packages/f-footer/src/components/Footer.vue
```diff
@@ -13,7 +13,7 @@
         <div class="c-footer-light">
             <div class="c-footer-container">
                 <div
-                    v-if="copy.linkButtonList.length && locale !== 'en-NZ'"
+                    v-if="copy.linkButtonList.length && locale !== 'en-IE'"
```
ashleynolan: Could we change this (or put it on our backlog to change) so that this is done via config, rather than as a check on the locale.

We should try not to have any locale specific `if/else`in the code – much better to do this with a prop that we can pass into the component settings.

smilburn1990: If we make it a prop, it would need to be an array so that we could toggle of multiple though right? I don't mind adding the logic now if you want? @ashleynolan

ashleynolan: So the prop could be `showCourierLinks` or something like that I'd guess, as that's the feature that's being toggled right?

smilburn1990: @ashleynolan - Updated 👍

ashleynolan (approved):

simonsandersje (approved):

PR #108, opened 2019-11-25
f-vue-icons@0.16.1-updated-class-toggle

smilburn1990:
### Added
- Prop and class for green `DeliveryIcon`

### DIFF
--- a/packages/f-vue-icons/src/components/DeliveryIcon/DeliveryIcon.vue
+++ b/packages/f-vue-icons/src/components/DeliveryIcon/DeliveryIcon.vue
@@ -1,7 +1,9 @@
 <template>
     <delivery-icon
-        :class="['c-icon c-icon--delivery', {
+        :class="['c-icon--delivery', {
             'c-icon--delivery--small': isSmall,
+            'c-icon': !isGreen,
+            'c-icon--green': isGreen,
             ...iconClasses
         }]" />
 </template>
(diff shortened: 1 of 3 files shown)


simonsandersje (approved):

markedup (approved):

packages/f-vue-icons/src/components/DeliveryIcon/DeliveryIcon.vue
```diff
@@ -1,7 +1,9 @@
 <template>
     <delivery-icon
-        :class="['c-icon c-icon--delivery', {
+        :class="['c-icon--delivery', {
```
DamianMullins, line 3: Why was the `c-icon` class removed? It should be present on _all_ icons?

packages/f-vue-icons/src/components/DeliveryIcon/DeliveryIcon.vue
```diff
@@ -1,7 +1,9 @@
...
             'c-icon--delivery--small': isSmall,
+            'c-icon': !isGreen,
```
DamianMullins, line 5: If you need to do something special with icons in a consuming application you can override the classes there.

packages/f-vue-icons/src/components/DeliveryIcon/DeliveryIcon.vue
```diff
@@ -1,7 +1,9 @@
...
+            'c-icon--green': isGreen,
```
DamianMullins, line 6: New colours [should be added to this mixin](https://github.com/justeat/fozzie-components/blob/master/packages/f-vue-icons/src/mixins/icons.mixin.js) so that they can be used in any icon

PR #110, opened 2019-11-28
f-vue-icons@v0.17.0-add-alchohol-icon

smilburn1990:
### Added
- Alcohol age icon

### DIFF
--- a/packages/f-vue-icons/src/components/index.js
+++ b/packages/f-vue-icons/src/components/index.js
@@ -21,6 +21,7 @@ import StopwatchIcon from './StopwatchIcon';
 import TickIcon from './TickIcon';
 import VegetarianIcon from './VegetarianIcon';
 import WarningIcon from './WarningIcon';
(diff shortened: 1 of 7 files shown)


packages/f-vue-icons/src/components/index.js
```diff
@@ -45,5 +46,6 @@ export {
     StopwatchIcon,
     TickIcon,
     VegetarianIcon,
-    WarningIcon
+    WarningIcon,
+    AlcoholAgeIcon
```
DamianMullins: This is real NIT, but could you add this alphabetically as it makes the long list easy to scan so you see if an icon is in the list really easily.

DamianMullins:
You'll need to [add the icon here as well](https://github.com/justeat/fozzie-components/blob/master/packages/f-vue-icons/src/index.js).

DamianMullins (approved):

BanksySan (approved):

PR #112, opened 2019-12-03
f-header@2.0.0-beta.32 - Offers inbox link

xander-marjoram:
Disabled by default, can be enabled by passing the `showOffersLink` prop into the header.

This will initially only be shown in UK, behind an experiment flag (TBD).

### DIFF
--- a/packages/f-header/src/components/Navigation.vue
+++ b/packages/f-header/src/components/Navigation.vue
@@ -28,7 +28,7 @@
         </label>
 
         <a
-            v-if="showOffers && isBelowMid"
+            v-if="showOffersLink && isBelowMid"
             data-js-test="offers-link-mobile"
             :data-trak='`{
                 "trakEvent": "click",
@@ -48,7 +48,7 @@
             :class="['c-nav-container', { 'is-visible': navIsOpen }]">
             <ul class="c-nav-list">
                 <li
-                    v-if="showOffers && !isBelowMid"
+                    v-if="showOffersLink && !isBelowMid"
                     class="c-nav-list-item">
                     <a
                         data-js-test="offers-link-desktop"
@@ -312,21 +312,9 @@ export default {
             return this.currentScreenWidth < 768;
         },
 
-        currentPath () {
-            return this.$route.path;
-        },
-
-        isHomepage () {
-            return this.currentPath === '/';
-        },
-
-        showOffers () {
-            return this.showOffersLink && this.isHomepage;
-        },
-
         returnUrl () {
             if (this.$route) {
-                return encodeURIComponent(this.currentPath);
+                return encodeURIComponent(this.$route.path);
             }
             if (typeof document !== 'undefined') {
                 return encodeURIComponent(document.location.pathname);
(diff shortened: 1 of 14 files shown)


packages/f-header/src/components/Navigation.vue
```diff
@@ -260,12 +307,24 @@ export default {
...
         isBelowMid () {
-            return this.currentScreenWidth <= 767;
+            return this.currentScreenWidth < 768;
+        },
+
+        currentPath () {
+            return this.$route.path;
+        },
+
+        isHomepage () {
```
smilburn1990: Think this is already handled with a prop called `isTransparent`. Not the greatest name but I'm pretty sure all it does is check the homepage 👍

xander-marjoram: I think it's good to have separation of concerns, because they're not mutually exclusive, and the extra overhead is pretty minimal.

xander-marjoram:
Just working on some tests. Should've written them before asking for review!

packages/f-header/src/components/Navigation.vue
smilburn1990: Yeah i agree, but maybe it should still be a prop coming in from the outside ?

smilburn1990: Just going on what I have been told on previous PR's

ashleynolan: Yeh, ideally the component should be as dumb as possible. So the consuming application (in this case HomeWeb) should be passing in whether to show the offers icon (it would default to false, but HomeWeb would pass in `true`), and then the component just uses that prop to show/hide the icon.

smilburn1990 (dismissed):
Looks good 👌

packages/f-header/src/components/Navigation.vue
```diff
@@ -669,6 +731,22 @@ $nav-trigger-focus-bg--ml          : $green--offWhite;
...
+    .c-nav-featureLink {
+        display: block;
+        width: 28px;
+        height: 28px;
+
+        @include media('<mid') {
+            position: absolute;
+            top: 0;
+            right: 0;
+            width: 28px + 2*spacing(x2);
```
kevinrodrigues: is the 2*spacing(x2) intentional?

xander-marjoram: Yep, I thought it was a bit clearer to see where the value came from rather than saying 60px, then you can see it's 28px padded on either side (`2*`) by `spacing(x2)`.

ashleynolan: Probably worth a comment to explain it, as I'm not sure it's immediately obvious why you wouldn't just use `spacing(x4)` (I think it makes sense now that you've explained it though)

xander-marjoram: Cool, I've put the width and height in variables, and hopefully made the reasoning clearer.

kevinrodrigues (approved):

ashleynolan (approved):

PR #121, opened 2020-01-29
f-metadata@v1.0.2 - Fixed content cards from sometimes not displaying on initial load

kevinrodrigues:


packages/f-metadata/src/index.js
```diff
@@ -28,10 +28,12 @@ const initialiseBraze = (options = {
                         });
 
                         appboy.requestContentCardsRefresh();
-                        const contentCards = appboy.getCachedContentCards();
-                        if (contentCards && contentCards.cards.length) {
-                            options.callbacks.handleContentCards(contentCards.cards);
-                        }
+                        
+                        appboy.subscribeToContentCardsUpdates(contentCards => {
```
ashleynolan, line 30: Could we ever get into a race condition here if the `requestContentCardsRefresh()`  returned before the subscription had been completed? Or is it better to request the refresh after the subscription call is setup?

kevinrodrigues: To be honest I don't think we need the `requestContentCardsRefresh()` call, because the subscription should handle showing updated cards automatically by waiting for the `sync` to happen, once that happens the callback will fire. I think the only time we'd need the  `requestContentCardsRefresh()` is if we had a button on the page that the user would click to pull through new content. Also by opening a session `app.openSession()` (line 21) would automatically invoke `requestContentCardsRefresh()` I believe. Will remove it and see what breaks :D

ashleynolan: Is it worth just removing it and checking if it all still works as expected?

kevinrodrigues: Removed it now. Looks like it still works ok still. Out of interest do you know why that method was introduced? I was hesitant to remove it as I couldn't see why it was added in the first place.

DamianMullins (approved):

simonsandersje (approved):

packages/f-metadata/src/index.js
```diff
@@ -26,12 +26,12 @@ const initialiseBraze = (options = {
...
-
-                        appboy.requestContentCardsRefresh();
-                        const contentCards = appboy.getCachedContentCards();
-                        if (contentCards && contentCards.cards.length) {
-                            options.callbacks.handleContentCards(contentCards.cards);
-                        }
+                        
+                        appboy.subscribeToContentCardsUpdates(contentCards => {
+                            if (contentCards && contentCards.cards.length) {
+                                options.callbacks.handleContentCards(contentCards.cards);
```
DamianMullins, line 32: Hi @kevinrodrigues I'm seeing the error "`Cannot read property 'handleContentCards' of undefined`" in staging on MW as we don't set a value for the option and `contentCards` comes through to us with a value. 

I think we need to add another guard here to make sure that `options.callbacks.handleContentCards` is actually set, what do you think?

DamianMullins: I've created #135 to fix this.

kevinrodrigues: That's my bad. Thanks for the fix 👍

PR #124, opened 2020-01-30
f-header@2.0.0-beta.37 - Changing the saved cards' URL to avoid an unnecessary redirects.

SachaZvetelman:
v2.0.0-beta.37
*Feb 11, 2020*

### Changed
- Saved cards' URL to avoid an unnecessary redirect.

### DIFF
--- a/packages/f-header/CHANGELOG.md
+++ b/packages/f-header/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v2.0.0-beta.35
+------------------------------
+*Feb 3, 2020*
+
+### Changed
+- Saved cards' URL to avoid an unnecessary redirect.
+
+
 v2.0.0-beta.34
 ------------------------------
 *January 17, 2020*
--- a/packages/f-header/package.json
+++ b/packages/f-header/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-header",
   "description": "Fozzie Header – Globalised Header Component",
-  "version": "v2.0.0-beta.34",
+  "version": "v2.0.0-beta.35",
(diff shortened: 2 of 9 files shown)


kevinrodrigues (dismissed):

simonsandersje (dismissed):

ashleynolan:
@SachaZvetelman Can you rename the commit and the title to be `f-header@2.0.0-beta.35`? Just makes it clear what it relates to

SachaZvetelman:
@ashleynolan, can certainly do that.

On a similar note, I think this strategy of including versions in commits has a major downside. If someone starts a new PR before mine gets merged, they will also name it `f-header@2.0.0-beta.35`, causing me to have to rename all my commits and my branch. Any opinions on this?

ashleynolan:
@SachaZvetelman I think it's just a case of whoever's PR gets merged first, then the next PR will need to update their version if they're also using the same version number.

They would need to update their `package.json` version as well anyway (as otherwise they can't release their package), so there's no way around having to update a PR if another one get's merged first. The only solution to this would be to somehow automate the version number change post-merge, but not sure if that's possible.

Appreciate it's not ideal though – we don't seem to hit that use-case a whole lot at least as our review/merge cycle seems to be much shorter for UI components than on something like HomeWeb.

SachaZvetelman:
@ashleynolan, agree it's not a big deal, but probably something a bit annoying. Updating `package.json` and the `CHANGELOG.md` is quite simple. Starting to edit past commits not so much, in general.

But again, not a big deal at this point.

ashleynolan:
@SachaZvetelman For reference – `git rebase -i HEAD~2` is the best method I've found for renaming commits (where `2` is the number of commits to change).

That then lets you rename commit msgs really quickly (as long as you have your default cmd line git editor set to VSCode or something that isn't VIM 😄). After renaming, just force-push to your branch and should all be sorted.

ashleynolan (dismissed):

kevinrodrigues (approved):

simonsandersje (approved):

PR #129, opened 2020-02-05
f-header@2.0.0 - Add new theme for brand-colour header

InsungMulumba:
Intended for [Opening Up Home on Mobile Web work](https://jira.just-eat.net/browse/CCS-218)

Sub-task [here](https://jira.just-eat.net/browse/CCS-330)

image

### DIFF
--- a/packages/f-header/package.json
+++ b/packages/f-header/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-header",
   "description": "Fozzie Header – Globalised Header Component",
-  "version": "v2.0.0-beta.35",
+  "version": "v2.0.0",
   "main": "dist/f-header.umd.min.js",
   "files": [
     "dist"
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -192,7 +192,7 @@ export default {
         }
     }
 
-    .c-header.c-header--red {
+.c-header--red {
         background-color: $red;
         min-height: 88px;
     }
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -1,8 +1,7 @@
 <template>
     <header
         :data-theme="theme"
-        :class="['c-header', {
-            'c-header--red': showRedBackgroundHeader,
+        :class="['c-header', headerBackgroundClass, {
             'c-header--transparent c-header--gradient': showTransparentHeader
         }]">
         <skip-to-main
@@ -114,8 +113,8 @@ export default {
         showTransparentHeader () {
             return this.headerBackgroundTheme === 'transparent' && !this.mobileNavIsOpen;
         },
-        showRedBackgroundHeader () {
-            return this.headerBackgroundTheme === 'red';
+        headerBackgroundClass () {
+            return this.headerBackgroundTheme === 'red' ? 'c-header--red' : '';
         }
     },
     methods: {
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -2,7 +2,7 @@
     <header
         :data-theme="theme"
         :class="['c-header', {
-            'c-header--brandColour': showBrandColourHeader,
+            'c-header--red': showRedBackgroundHeader,
             'c-header--transparent c-header--gradient': showTransparentHeader
         }]">
         <skip-to-main
@@ -113,6 +113,9 @@ export default {
         },
         showTransparentHeader () {
             return this.headerBackgroundTheme === 'transparent' && !this.mobileNavIsOpen;
+        },
+        showRedBackgroundHeader () {
+            return this.headerBackgroundTheme === 'red';
         }
     },
     methods: {
@@ -190,7 +193,7 @@ export default {
         }
     }
 
-    .c-header.c-header--brandColour {
+    .c-header.c-header--red {
         background-color: $red;
         min-height: 88px;
     }
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -1,16 +1,19 @@
 <template>
     <header
         :data-theme="theme"
-        :class="['c-header', { 'c-header--transparent c-header--gradient': showTransparentHeader }]">
+        :class="['c-header', {
+            'c-header--brandColour': showBrandColourHeader,
+            'c-header--transparent c-header--gradient': showTransparentHeader
+        }]">
         <skip-to-main
             :text="copy.skipToMainContentText"
             :transparent-bg="showTransparentHeader" />
         <div class="c-header-container">
             <logo
                 :theme="theme"
-                :is-transparent="showTransparentHeader"
                 :company-name="copy.companyName"
-                :logo-gtm-label="copy.logo.gtm" />
+                :logo-gtm-label="copy.logo.gtm"
+                :header-background-theme="headerBackgroundTheme" />
             <navigation
                 :nav-links="copy.navLinks"
                 :help="copy.help"
@@ -26,7 +29,7 @@
                 :user-info-url="userInfoUrl"
                 :order-count-url="orderCountUrl"
                 :is-order-count-supported="isOrderCountSupported"
-                :is-transparent="isTransparent"
+                :header-background-theme="headerBackgroundTheme"
                 @onMobileNavToggle="mobileNavToggled" />
         </div>
     </header>
@@ -50,11 +53,6 @@ export default {
         locale: {
             type: String,
             default: ''
-        },
-
-        isTransparent: {
-            type: Boolean,
-            default: false
         },
 
         showDeliveryEnquiry: {
@@ -90,6 +88,11 @@ export default {
         isOrderCountSupported: {
             type: Boolean,
             default: true
+        },
+
+        headerBackgroundTheme: {
+            type: String,
+            default: 'white'
         }
     },
     data () {
@@ -109,7 +112,7 @@ export default {
             return this.copy.deliveryEnquiry && this.showDeliveryEnquiry;
         },
         showTransparentHeader () {
-            return this.isTransparent && !this.mobileNavIsOpen;
+            return headerBackgroundTheme === 'transparent' && !this.mobileNavIsOpen;
         }
     },
     methods: {
@@ -185,6 +188,11 @@ export default {
             left: 0;
             width: 100%;
         }
+    }
+
+    .c-header.c-header--brandColour {
+        background-color: $red;
+        min-height: 88px;
     }
 
     .c-header-container {
--- a/packages/f-header/src/components/Header.vue
+++ b/packages/f-header/src/components/Header.vue
@@ -112,7 +112,7 @@ export default {
             return this.copy.deliveryEnquiry && this.showDeliveryEnquiry;
         },
         showTransparentHeader () {
-            return headerBackgroundTheme === 'transparent' && !this.mobileNavIsOpen;
+            return this.headerBackgroundTheme === 'transparent' && !this.mobileNavIsOpen;
         }
     },
     methods: {
--- a/packages/f-header/src/components/Logo.vue
+++ b/packages/f-header/src/components/Logo.vue
@@ -13,7 +13,7 @@
             :is="iconComponent"
             :class="['c-logo-img',
                      iconClassName,
-                     isLogoWhite]"
+                     logoColourModifier]"
             :data-js-test="iconClassName" />
     </a>
 </template>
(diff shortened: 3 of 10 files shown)


InsungMulumba:
Please do not review just yet, need to sanity check the work.

InsungMulumba:
Ready!

packages/f-header/src/components/Header.vue
```diff
@@ -1,7 +1,7 @@
 <template>
     <header
         :data-theme="theme"
-        :class="['c-header', { 'c-header--transparent c-header--gradient': showTransparentHeader }]">
+        :class="['c-header', { 'c-header--compressedPage': showCompressedHeader }, { 'c-header--transparent c-header--gradient': showTransparentHeader }]">
```
DamianMullins: The dynamic classes should all be within a single object

```js
:class="['c-header', {
    'c-header--compressedPage': showCompressedHeader,
    'c-header--transparent c-header--gradient': showTransparentHeader
}]">
```

packages/f-header/src/components/Header.vue
```diff
@@ -187,6 +192,15 @@ export default {
         }
     }
 
+    .c-header--compressedPage {
+        background-color: $red;
+        min-height: 88px;
+    }
+
+    .c-header--compressedPage.c-header--gradient::before {
```
DamianMullins: Do you need to do this? Is there a case where you'd have the transparent header and the red header at the same time?

packages/f-header/src/components/Header.vue
```diff
@@ -90,6 +90,11 @@ export default {
         isOrderCountSupported: {
             type: Boolean,
             default: true
+        },
+
+        showCompressedHeader: {
```
DamianMullins: Maybe something like `showBrandColourHeader` or similar might better describe what this prop does?

packages/f-header/src/components/Header.vue
InsungMulumba: is-Transparent also sets the navigation button to the same colour as the logo, so my thinking was to set it to `true`.
Otherwise it will default to blue as shown below
image

packages/f-header/src/components/Header.vue
InsungMulumba: I'm unsure about the naming myself to be honest, this was suggested by @kevinrodrigues but I'm happy to refactor

packages/f-header/src/components/Header.vue
InsungMulumba: will do

packages/f-header/src/components/Header.vue
DamianMullins: Can you not update so that new prop makes the menu icon white as well? Seems weird that you'd apply the transparent prop to make the icon white.

InsungMulumba: I've looked deeper into f-header and the transparent prop also [renders the logo white ](https://github.com/justeat/fozzie-components/blob/403537f37e24a795b3a3dcad5aeeff3da24d8661/packages/f-header/src/components/Logo.vue#L122)

@DamianMullins am I right to think the process from here is to add the `showCompressedHeader` prop to `Logo.vue` and `Navigation.vue`?

DamianMullins: Looks like it, and then add a style which makes the icon background white for a new modifier class related to the new prop.

packages/f-header/src/components/Header.vue
```diff
@@ -187,6 +195,15 @@ export default {
         }
     }
 
+    .c-header--brandColour {
+        background-color: $red;
+        min-height: 88px;
+    }
+
+    .c-header--brandColour.c-header--gradient::before {
```
ashleynolan: Will `showTransparentHeader` and `showBrandColourHeader` ever be used together like this? (seems like they shouldn't be?)

InsungMulumba: Damian raised the same point above, I'm working on adding the new prop to `Logo.vue` and `Navigation.vue`, and then I will revert this

packages/f-header/src/components/Logo.vue
```diff
@@ -11,7 +11,9 @@
         }`'>
         <component
             :is="iconComponent"
-            :class="[iconClassName, isTransparent && iconComponent === 'je-logo' ? 'c-icon-je--transparentBg' : '']"
+            :class="[iconClassName, isTransparent && iconComponent === 'je-logo' ? 'c-icon-je--transparentBg' : '', {
+                'c-icon-je--transparentBg': showBrandColourHeader
```
ashleynolan: Think `showBrandColourHeader` is a bit of a confusing name (I'm not quite sure what state this refers to – is it when we want the red background on the header?).

Would be good to put these class definitions across multi-lines too, as it's hard to see the different statements atm (which may be the issue).

Also, feels strange to have two props for what could essentially be controlled via one prop called something like `theme`. If we had a `theme` prop, we could pass in `transparent` or `default` for the different styles potentially, as I'm guessing we won't ever want them to be both switched on at once?

InsungMulumba: I've refactored the name a few times now, personally I'm unsure what the best name for this is. 
@DamianMullins @kevinrodrigues @SachaZvetelman any thoughts?

The class definition was on one line before this PR! But I agree it's a bit confusing, I can split it up.

I will need help for your last point about a new prop, can you send a small code snippet of what you mean?

ashleynolan: Yeh – I'd be happy to pair on this if it'd be easier?

So what I was thinking was rather than having `isTransparent` and `showBrandColourHeader` the header could just accept a `headerTheme` prop which takes a number of defined values. Otherwise everytime we want to change the theme style of the header, we'd need to add a new prop, which doesn't scale very well.

Like I say – happy to pair on this when you have some time tomorrow?

ashleynolan: Also see what you mean we class definition being on one line before, but as this is getting longer it's getting too verbose to see easily what's going on:

```js
:class="[
    iconClassName,
    isTransparent && iconComponent === 'je-logo' ? 'c-icon-je--transparentBg' : '',
    {
        'c-icon-je--transparentBg': showBrandColourHeader
    }]"
```

packages/f-header/src/components/Logo.vue
```diff
@@ -11,7 +11,12 @@
         }`'>
         <component
             :is="iconComponent"
-            :class="[iconClassName, isTransparent && iconComponent === 'je-logo' ? 'c-icon-je--transparentBg' : '']"
+            :class="[
+                iconClassName,
+                isTransparent && iconComponent === 'je-logo' ? 'c-icon-je--transparentBg' : '',
```
DamianMullins: Shouldn't really have ternaries in here as you're mixing them with the standard Vue way of applying classes. You should be able to do this?

```js
:class="[
    'c-logo-img',
    iconClassName,
    {
        'c-icon-je--transparentBg': isTransparent && iconComponent === 'je-logo'
        'c-icon-je--transparentBg': showBrandColourHeader
}]"
```

Also, probably best to move that `isTransparent && iconComponent === 'je-logo'` check into a computed prop/

packages/f-header/src/components/Logo.vue
```diff
@@ -61,8 +61,8 @@ export default {
         linkAltText () {
             return `Go to ${this.companyName} homepage`;
         },
-        showWhiteLogo () {
-            return this.isTransparent && iconComponent === 'je-logo' || this.showBrandColourHeader  ? 'c-icon-je--transparentBg':'' ;
+        whiteLogo () {
```
ashleynolan: If this is returning true/false, would be better named `isWhiteLogo`

InsungMulumba:
TO-DO: Refactor Tests

packages/f-header/src/components/Header.vue
```diff
@@ -2,7 +2,7 @@
     <header
         :data-theme="theme"
         :class="['c-header', {
-            'c-header--brandColour': showBrandColourHeader,
+            'c-header--red': showRedBackgroundHeader,
```
InsungMulumba: I believe this name is suitable as the computed prop is explicitly for **red** theme, similar to that for the transparent theme just above

packages/f-header/src/components/Header.vue
```diff
@@ -187,6 +193,11 @@ export default {
         }
     }
 
+    .c-header.c-header--red {
```
ashleynolan: Do you need the `.c-header` part of this selector? I would have thought it should still take affect without this (so just `.c-header--red {`

packages/f-header/src/components/Navigation.vue
```diff
@@ -2,7 +2,8 @@
     <nav class="c-nav c-nav--global">
         <button
             :class="['c-nav-trigger c-nav-toggle is-hidden--noJS', {
-                'is-open': navIsOpen
+                'is-open': navIsOpen,
+                'c-logo--brandColour': showBrandColourHeader
```
ashleynolan: Where does `showBrandColourHeader` come from? (not sure if I'm being dull, but I can't see it as a `prop` or `data` in this file?)

packages/f-header/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-header",
   "description": "Fozzie Header – Globalised Header Component",
-  "version": "v2.0.0-beta.36",
+  "version": "v2.0.0",
```
ashleynolan: Can you update this to just be "2.0.0"?

packages/f-header/src/components/Header.vue
```diff
@@ -109,7 +112,10 @@ export default {
             return this.copy.deliveryEnquiry && this.showDeliveryEnquiry;
         },
         showTransparentHeader () {
-            return this.isTransparent && !this.mobileNavIsOpen;
+            return this.headerBackgroundTheme === 'transparent' && !this.mobileNavIsOpen;
+        },
+        showRedBackgroundHeader () {
```
ashleynolan: Rather than have a function that returns `true`/`false` here for red, you could return the classname itself for the header background?

So that way you would return `c-header--red` (also makes it a bit more scalable should we need more different colours in future).

So something like this:

```js
headerBgClass () {
  if (this.headerBackgroundTheme === 'red') {
    return 'c-header--red';
  }
}
```

packages/f-header/src/components/Navigation.vue
InsungMulumba: Prop added :)

packages/f-header/src/components/Header.vue
InsungMulumba: c-header has it's own background-colour that takes precedence otherwise

packages/f-header/src/components/Header.vue
InsungMulumba: I will change this and do the same for the prop I just introduced in `Navigation.vue` (I committed that before seeing this comment)

simonsandersje (dismissed):

smilburn1990 (dismissed):

packages/f-header/src/components/Header.vue
ashleynolan: It should be the same specificity, and as this selector comes after `c-header` this style should take precedence.  If it's not, then might be something funky going on – does this definitely not work?

InsungMulumba: I'll try again

packages/f-header/src/components/Logo.vue
```diff
@@ -55,6 +56,9 @@ export default {
         },
         linkAltText () {
             return `Go to ${this.companyName} homepage`;
+        },
+        isLogoWhite () {
+            return (this.headerBackgroundTheme === 'transparent' || this.headerBackgroundTheme === 'red') && this.theme === 'je' ? 'c-icon-je--alt' : '';
```
ashleynolan, line 61: Should likely rename this function as it implies that it's returning a `Boolean` (when it's actually returning a string for the class?).

`logoColourModifier` might make sense?

packages/f-header/src/components/Navigation.vue
```diff
@@ -339,6 +339,9 @@ export default {
             const currentTime = new Date().getTime();
             return this.localOrderCountExpires < currentTime;
         },
+        NavToggleThemeClass () {
```
ashleynolan: should be `navToggleThemeClass` (so lowercase first letter)

ashleynolan (changes requested):
Couple of super small changes – then looks good

packages/f-header/src/components/Header.vue
InsungMulumba: Hey @ashleynolan  So I think the reason I can't get it locally is because I don't know how to access `packages/f-header/src/components/Header.vue` locally in Homeweb. (it's not in my node_modules) I'll revert this change, and it should propagate through. :)

packages/f-header/src/components/Logo.vue
InsungMulumba: This is done, not sure why it isn't flagging as `outdated`

ashleynolan (dismissed):

asherstoppard (approved):

smilburn1990 (approved):

PR #130, opened 2020-02-05
fozzie-components@v1.4.0 – Adding DangerJS checks to repo

ashleynolan:
### Added
- DangerJS Checks (via CircleCI)
- Pull Request Template
- `CODEOWNERS` file

### Changed
- #trivial changes to package scripts changing `prepare` hooks to `prepublishOnly` (so they only run on `npm publish` not on `npm install`)

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -1,0 +1,100 @@
+/* global danger, fail, message */
+const { pr } = danger.github;
+const bodyAndTitle = (pr.body + pr.title).toLowerCase();
+const isTrivial = bodyAndTitle.includes('#trivial');
+
+if (!isTrivial) {
+    const failedChangelogs = [];
+    const failedVersionBumps = [];
+    const modifiedFiles = danger.git.modified_files;
+
+    // Get an array of files in the root monorepo that have changed
+    const modifiedRootFiles = modifiedFiles.filter(filepath => !filepath.startsWith('packages/') && filepath !== 'yarn.lock');
+
+    // Get an array of packages that have changed, then make that array unique
+    const modifiedPackages = modifiedFiles.filter(filepath => filepath.startsWith('packages/'))
+        .map(filepath => filepath.split('/')[1]);
+    const uniqueModifiedPackages = new Set(modifiedPackages);
+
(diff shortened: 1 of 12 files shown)


packages/f-header/package.json
```diff
@@ -41,7 +41,8 @@
     "lodash-es": "4.17.15"
   },
   "peerDependencies": {
-    "@justeat/f-trak": "0.6.0"
+    "@justeat/f-trak": "0.6.0",
+    "@justeat/browserslist-config-fozzie": ">=1.1.1"
```
DamianMullins: This is already in the root package, do you need it here as well?

DamianMullins: Just noticed it's a peer dep. Is it needed in f-footer as well?

dangerfile.js
```diff
@@ -0,0 +1,100 @@
...
+
+    // Get an array of files in the root monorepo that have changed
+    const modifiedRootFiles = modifiedFiles.filter(filepath => !filepath.startsWith('packages/') && filepath !== 'yarn.lock');
+
+    // Get an array of packages that have changed, then make that array unique
+    const modifiedPackages = modifiedFiles.filter(filepath => filepath.startsWith('packages/'))
+        .map(filepath => filepath.split('/')[1]);
+    const uniqueModifiedPackages = new Set(modifiedPackages);
+
+    // Fail if the title of the PR isn't in the format of a version i.e. vX.X.X (such as v1.4.0)
```
xander-marjoram: Example in the comment (and the `fail`) should be something like `{package-name}@vX.X.X` to match the regex.

ashleynolan: Good catch – updated

packages/f-header/package.json
ashleynolan: I've removed this for now as not sure it's needed for this PR (and may need to bump package version separately). Will check if it's needed when I bump header/footer to v2 after this gets merged…

xander-marjoram:
@ashleynolan Can you bump the package versions and add a changelog entry for each package? :trollface:

xander-marjoram (dismissed):

xander-marjoram (approved):

PR #134, opened 2020-02-10
fozzie-components@v1.4.1 – Fix for dangerJS checks

ashleynolan:
### Fixed
- DangerJS fix for package paths and for false positive around reporting change in the root


ashleynolan:
Merging this as causing issues with other PRs

PR #143, opened 2020-02-19
f-footer@v2.0.0 – v2 release and small package.json update

ashleynolan:
### Changed
- Moving `f-footer` Vue component to be `v2.0.0`.
  If we need to update the old `f-footer` package, those changes can be released on the legacy `v1.x.x` release branch via the [legacy `f-footer` repo](https://github.com/justeat/f-footer).
- Separated out `lint` and `lint:fix` into two tasks (so CircleCI build can run lint task without fixing)


zlatinivanov-je:
Does that need an update to CircleCI

zlatinivanov-je (approved):
LGTM

ashleynolan:
@zlatinivanov-je It will do – but I need to get the script changes into each package first, so that'll be part of a separate PR 👍

PR #144, opened 2020-02-20
f-metadata@2.2.0: promisify and unit tests

asherstoppard:
Promisify f-metadata and add unit tests

## UI Review Checks

### DIFF
--- a/packages/f-metadata/package.json
+++ b/packages/f-metadata/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-metadata",
   "description": "Fozzie Metadata Component",
-  "version": "2.1.0",
+  "version": "2.2.0",
   "main": "src/index.js",
   "files": [
     "dist"
@@ -26,7 +26,8 @@
   ],
   "scripts": {
     "build": "yarn lint",
-    "lint": "vue-cli-service lint --fix"
+    "lint": "vue-cli-service lint --fix",
+    "test": "jest --coverage"
   },
   "browserslist": [
     "extends @justeat/browserslist-config-fozzie"
@@ -36,6 +37,7 @@
   },
   "devDependencies": {
     "@vue/cli-plugin-babel": "3.9.2",
-    "@vue/cli-plugin-eslint": "3.9.2"
+    "@vue/cli-plugin-eslint": "3.9.2",
+    "jest": "^25.1.0"
   }
 }
--- a/packages/f-metadata/package.json
+++ b/packages/f-metadata/package.json
@@ -25,6 +25,7 @@
     "fozzie"
   ],
   "scripts": {
+    "prepublish": "yarn test",
     "build": "yarn lint",
     "lint": "vue-cli-service lint --fix",
     "test": "jest --coverage"
--- a/packages/f-metadata/package.json
+++ b/packages/f-metadata/package.json
@@ -25,7 +25,7 @@
     "fozzie"
   ],
   "scripts": {
-    "prepublish": "yarn test",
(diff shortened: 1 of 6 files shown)


packages/f-metadata/package.json
```diff
@@ -25,8 +25,10 @@
     "fozzie"
   ],
   "scripts": {
+    "prepublish": "yarn test",
```
ashleynolan: Can you make this `prepublishOnly`? (otherwise it runs on `npm install` which clogs the build)

asherstoppard: Ah ok, yeah of course :+1:

packages/f-metadata/package.json
```diff
@@ -36,6 +38,7 @@
   },
   "devDependencies": {
     "@vue/cli-plugin-babel": "3.9.2",
-    "@vue/cli-plugin-eslint": "3.9.2"
+    "@vue/cli-plugin-eslint": "3.9.2",
+    "jest": "^25.1.0"
```
DamianMullins: Can you pin this one, please?

packages/f-metadata/package.json
```diff
@@ -25,8 +25,10 @@
     "fozzie"
   ],
   "scripts": {
+    "prepublish": "yarn test",
     "build": "yarn lint",
-    "lint": "vue-cli-service lint --fix"
+    "lint": "vue-cli-service lint --fix",
```
DamianMullins: Might be worth splitting this task out like we do in other repos into `lint` and `lint:fix` tasks?

packages/f-metadata/package.json
```diff
@@ -25,8 +25,10 @@
...
+    "test": "jest --coverage"
```
ashleynolan: Can you use the `vue-cli-service test:unit` here (to keep it consistent with other packages?).

Should be able to set the `collectCoverage` setting then to get coverage out of the tests

asherstoppard: Would we want to couple this module with `vue-cli-service` given that it's pure js? Appreciate there's commands in there already but I'm intrigued as to what it gives us over using the jest/eslint binaries directly?

ashleynolan: See what you mean – I'm not super precious about what's used (as it's all jest ultimately and it's only a dev dependency). I was mainly only thinking about it being an additional monorepo dependency (as `vue-cli-service` is effectively already available through the root depenencies), but happy to directly use jest if you think it fits better here.

kevinrodrigues (dismissed):

packages/f-metadata/package.json
asherstoppard: I guess it's more a question for the broader group in the UI guild. I'll change it over to `vue-cli-service` for the purpose of this PR and make note to raise it as a query in the next meeting :+1:

DamianMullins: I guess if `vue-cli-service` provides us the ability to with Jest already then it's one less dependecy to the project. On the other hand having Jest as a dependency means that we can stay up to date with the latest versions without having to wait for vue CLI to update. I'm happy either way. 

We do run tests using Jest directly in other modules as we're not always using the vue CLI.

ashleynolan: Yeh, it's much of a muchness I guess. As @asherstoppard mentioned, as long as we have clear guidance on this as a group it shouldn't really be an issue either way (so we're happy to use more direct dependencies if a module is more native JS, like this one is, but use `vue-cli` for vue-specific components).

ashleynolan (approved):

kevinrodrigues (approved):

PR #148, opened 2020-03-02
f-metadata@2.3.0: Change callback and return early

asherstoppard:
### Changed
- Callback with `null` if `apiKey` or `userId` is not defined.
- Return before SDK import if dependencies are not available.

## UI Review Checks

### DIFF
--- a/packages/f-metadata/src/index.js
+++ b/packages/f-metadata/src/index.js
@@ -20,27 +20,27 @@ const initialiseBraze = (options = {}) => new Promise((resolve, reject) => {
     window.dataLayer = window.dataLayer || [];
 
     return import(/* webpackChunkName: "appboy-web-sdk" */ 'appboy-web-sdk')
-            .then(({ default: appboy }) => {
-                appboy.initialize(apiKey, { enableLogging });
+        .then(({ default: appboy }) => {
+            appboy.initialize(apiKey, { enableLogging });
 
-                appboy.display.automaticallyShowNewInAppMessages();
+            appboy.display.automaticallyShowNewInAppMessages();
 
-                appboy.openSession();
-                window.appboy = appboy;
+            appboy.openSession();
+            window.appboy = appboy;
 
-                appboy.changeUser(userId, () => {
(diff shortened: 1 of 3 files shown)


DamianMullins (approved):

packages/f-metadata/src/index.js
```diff
@@ -12,32 +12,33 @@ const initialiseBraze = (options = {}) => new Promise((resolve, reject) => {
...
+        return resolve(null);
+    }
 
     window.dataLayer = window.dataLayer || [];
 
     return import(/* webpackChunkName: "appboy-web-sdk" */ 'appboy-web-sdk')
         .then(({ default: appboy }) => {
-            if (apiKey && apiKey.length && userId && userId.length) {
-                appboy.initialize(apiKey, { enableLogging });
+            appboy.initialize(apiKey, { enableLogging });
```
ashleynolan, line 24: What happens if `apiKey` is set to an erroneous value such as a blank string? (guessing that was what the `apiKey.length` was for previously?)

Not sure if there are any tests for this, but might be worth checking if there's one that covers that?

ashleynolan (commented):
Looks good – one small comment

packages/f-metadata/src/index.js
DamianMullins: A blank string value will be caught by `!apiKey` above, right?

ashleynolan: Not sure. I guess was just wondering what `apiKey.length` would be checking for when running `apiKey && apiKey.length` but if it's fine simplified then that's cool. Probably worth having a test for it though (unless we already have one and then I'll go away 😉)

asherstoppard: I think it was just a bit of confusion around the fact that `'' === false` :+1: They were both essentially checking the same thing. There's tests for both the userId and apiKey.

ashleynolan: Cool – I'm happy then 👍

ashleynolan (approved):

PR #153, opened 2020-03-05
v2.0.1: Update tenant file Ireland with new link URL.

simonsandersje:
Minor update to the link URL for the courier links in the footer for Ireland.

### DIFF
--- a/packages/f-footer/src/tenants/en-IE.js
+++ b/packages/f-footer/src/tenants/en-IE.js
@@ -152,7 +152,7 @@ export default {
                 },
                 {
                     title: 'Become a courier',
-                    url: 'https://www.fountain.com/just-eat/apply/dublin-courier?preview=true',
+                    url: 'https://couriers.just-eat.ie/application',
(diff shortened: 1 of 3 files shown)


packages/f-footer/src/tenants/en-IE.js
```diff
@@ -167,7 +167,7 @@ export default {
                 },
                 {
                     title: 'Courier portal',
-                    url: 'https://couriers.just-eat.ie/account/login',
+                    url: 'https://couriers.just-eat.ie/application',
```
xander-marjoram, line 170: Not sure I understand the need for the same link twice, there's a login link we could use for the portal: https://couriers.just-eat.ie/login

simonsandersje: @xander-marjoram It's on the Jira - I questioned it, but that's what they want :shrug

DamianMullins (approved):

minrwhite (approved):

ashleynolan (approved):

PR #155, opened 2020-03-31
fozzie-components@v1.7.0 – f-skeleton-component added for creating base components

ashleynolan:
### Added
- `f-skeleton-component` added for use when creating base component structure.


packages/f-skeleton-component/src/assets/scss/common.scss
```diff
@@ -0,0 +1,22 @@
+// Set $theme variable so that fozzie doesn't complain when vars are imported below
+// n.b. This variables isn't actually used (unless there are theme specific vars from fozzie)
+// Instead, the theme is handled via a data-attribute on the component (via the `@mixin theme()` below)
+$theme: 'je' !default;
+
+@import '~@justeat/f-utils/src/scss/functions/index';
```
DamianMullins, line 6: Should you include `f-utils` and `fozzie` in the project dependencies?

packages/f-skeleton-component/src/tenants/da-DK.js
```diff
@@ -0,0 +1,4 @@
+export default {
+    locale: 'da-DK',
```
DamianMullins, line 2: Do we need the `locale` property in these files? Aren't we already selecting the translations in code using a locale string?

packages/f-skeleton-component/src/assets/scss/common.scss
ashleynolan: Shouldn't need to, as these are hoisted in the mono-repo dependencies (in the base `package.json`)

packages/f-skeleton-component/src/tenants/da-DK.js
ashleynolan: It's a fallback for if the component is used when the `i18n` module isn't being used. In header/footer we use [this shared call](https://github.com/justeat/fozzie-components/blob/master/packages/f-header/src/components/Header.vue#L104) to check which locale setting to use.

ashleynolan: This is the shared service method that does that check too > https://github.com/justeat/fozzie-components/blob/master/packages/f-services/src/services/shared.service.js#L25-L28

DamianMullins (approved):

xander-marjoram (approved):

PR #156, opened 2020-03-31
f-services@1.0.0 - Split `sharedServices` module into `window` and `globalisation`, and add `axios` service.

xander-marjoram:
Tested by `yalc`ing to HW. It will only require a small update to consuming applications. They will now have to import the module they want, i.e.,
```
import { axiosServices, windowServices } from '@justeat/f-services';
```

Rather than including everything:
```
import sharedServices from '@justeat/f-services';
```

### Added
- Axios services module.

### Changed
- Split central "shared" module into separate "window" and "globalisation" modules.
- `jest` config updates. `moduleNameMapper` removed, `jsx` removed from checks (as we don't use) and `transformIgnorePatterns` updated.
- Rollup package updates.

## UI Review Checks

## Browsers Tested
- [ ] Chrome (latest)
- [ ] Internet Explorer 11


packages/f-services/tests/window.test.js
```diff
@@ -0,0 +1,87 @@
...
+    });
+
+    describe('removeEvent', () => {
+        it('removes given event listener', () => {
+            // Arrange
+            const eventName = 'resize';
+            const callBackFunction = jest.fn();
+
+            // Act
+            removeEvent(eventName, callBackFunction);
```
minrwhite: Shouldn't we have a corresponding `addEvent()` as part of the `Arrange`ment in order to test that `removeEvent()` is operating correctly?

packages/f-services/src/index.js
```diff
@@ -1,12 +1,16 @@
...
+import axios from './axios';
+import globalisation from './globalisation';
+import window from './window';
 
-// Import vue component
-import sharedServices from './services/shared.service';
+export {
+    axios,
+    globalisation,
+    window
```
ashleynolan: I'd probably look to avoid using `window` as an export name (simply as it clashes with the `window` global).

Maybe calling these explicitly `windowService` and `axiosService` might be better to avoid any name clashes?

packages/f-services/tests/globalisation.test.js
```diff
@@ -0,0 +1,109 @@
...
+            const locale = 'en-NZ';
+
+            // Act
+            const result = getTheme(locale);
+
+            // Assert
+            expect(result).toBe('ml');
+        });
+
+        it('returns "je" for Just Eat theme if locale is NOT "en-NZ"', () => {
```
ashleynolan: Should this description just be `returns "je" for Just Eat theme if locale is NOT "en-NZ" or "en-AU"`?

packages/f-services/src/index.js
xander-marjoram: I did have those originally but thought `Service` was a redundant suffix as we're already within f-services. The readme also recommends exporting with an alias but happy to change this.

packages/f-services/src/axios.js
```diff
@@ -0,0 +1,128 @@
+import axios from 'axios';
```
ashleynolan, line 1: Could you add some JSDoc style comments to this file and functions.

packages/f-services/tests/globalisation.test.js
```diff
@@ -0,0 +1,109 @@
...
+            };
+
+            // Act
+            const result = getLocale(tenantConfigs, tenantString, globalTenant);
+
+            // Assert
+            expect(result).toBe('en-GB');
+        });
+
+        it('returns the correct locale if tenantConfigs does contain the locale from tenantString', () => {
```
minrwhite: I understand the wording of the description here in that it matches up with the rest of the `it()`s in this `describe()`, but it scans a bit weirdly to read. I'd suggest 
```suggestion
        it('returns the correct locale if tenantConfigs contains the locale from tenantString', () => {
```

ashleynolan (commented):
Looks good – just a couple of super small comments from me.

packages/f-services/src/index.js
ashleynolan: Yeh, noticed the README did that, but would rather err on the safe side I think with export names. get what you mean about the redundancy, but I think it's fine if it's a convention to suffix all services with that.

packages/f-services/src/window.js
```diff
@@ -0,0 +1,57 @@
...
+export const addEvent = (eventName, callBackFunction, throttleTime = 0) => {
+    const callback =
+        throttleTime >= 0
+            ? throttle(callBackFunction, throttleTime)
+            : callBackFunction;
+    return root.addEventListener(eventName, callback);
+};
+
+/**
+ * Removes chosen event.
```
minrwhite:
```suggestion
 * Removes given event listener from specified event.
```

packages/f-services/src/index.js
```diff
@@ -1,12 +1,16 @@
 /**
- * @overview Fozzie services JavaScript behaviour.
+ * @overview Fozzie's collection of shared javascript modules
  *
  * @module f-services
  */
 
+import axiosServices from './axios';
+import globalisationServices from './globalisation';
+import windowServices from './window';
```
minrwhite: You could change these `import`s to `exports` and bypass having to define the export below entirely.

zlatinivanov-je:
I've read through and it looks good to me. 
It is possible to steer more confusion around this, but would people here consider using `windowService` instead of `windowServices`

packages/f-services/package.json
```diff
@@ -37,16 +37,18 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
+    "axios": "0.19.2",
```
DamianMullins: Do you think it's worth adding axios as a `peerDependency` so that it is up to the consuming application to provide the axios npm module?

packages/f-services/src/axios.js
```diff
@@ -0,0 +1,155 @@
...
+        ]
+    }
+});
+
+/**
+ * A Smart Gateway wrapper for createCamelCaseClient.
+ *
+ * @param {string} options
+ */
+const createSmartGatewayClient = ({
```
DamianMullins: I imagine you probably don't need this function here? It was specifically for smart gateway and expects object properties as we provide them from the backend of Menu Web.

packages/f-services/package.json
xander-marjoram: That's a good shout

packages/f-services/src/axios.js
xander-marjoram: Thanks, I was kinda hoping you'd say that. Will remove it now.

packages/f-services/package.json
```diff
@@ -41,12 +41,16 @@
...
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.0.5",
     "rollup": "1.26.3",
     "rollup-plugin-babel": "4.3.3",
-    "rollup-plugin-commonjs": "10.1.0",
-    "rollup-plugin-node-resolve": "5.2.0",
     "rollup-plugin-terser": "5.1.2"
+  },
+  "peerDependencies": {
+    "axios": "0.19.2"
```
DamianMullins: Best to prefix this one with `^` so that the consuming application can update within the major version without warnings.

DamianMullins (dismissed):

DamianMullins:
Looks like CI is failing is failing because there is a reference to the internal NPM registry in the `yarn.lock` file. 

@xander-marjoram Would you be able to update the lock file so that it uses the yarn registry URL instead? 

You can either remove the line in the lock file and run `yarn install`; run `yarn upgrade` which should bump all versions up to the latest matched; or delete the lock file entirely and run `yarn install` to recreate it. 

None of those are great, the upgrade one is probably the best option IMO. If anyone knows of a nicer way to update registry URLs which have already been written I'd like to know 😄

DamianMullins (dismissed):

minrwhite (dismissed):

minrwhite (approved):

DamianMullins (approved):

PR #163, opened 2020-04-16
storybook@v0.1.0

siggerzz:
This is still very much a WIP, however it's a start!

`cd packages/storybook`
`yarn`
`yarn storybook:serve`

This will start up the Storybook app. (We'll need to host this somewhere eventually)
From here, you'll be able to interact with the `f-header` and `f-footer` components in isolation.

**Note: I was having some webpack issues, which I was able to resolve by doing the various Sass imports, however this shouldn't be merged until it's resolved. A front-end engineer is probably better placed to resolve this :)**

image

### DIFF
--- a/packages/f-footer/CHANGELOG.md
+++ b/packages/f-footer/CHANGELOG.md
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Latest (roll into next release)
 ------------------------------
+*April 16, 2020*
+
+### Added
+- `stories` folder that'll contain `*.stories.js` files for use with Storybook
+
+v2.1.1
+------------------------------
 *March 30, 2020*
 
 ### Changed
--- a/packages/f-footer/CHANGELOG.md
+++ b/packages/f-footer/CHANGELOG.md
@@ -4,20 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-Latest (roll into next release)
-------------------------------
-*April 16, 2020*
-
-### Added
-- `stories` folder that'll contain `*.stories.js` files for use with Storybook
-
 v2.1.1
 ------------------------------
-*March 30, 2020*
+*April 20, 2020*
+
+### Added
+- `stories` folder that will contain `*.stories.js` files for use with Storybook
 
 ### Changed
 - Updated name of component in `index.js` from `Footer` > `VueFooter` (as `Footer` is a reserved word according to linting rules)
-
 
 v2.1.0
 ------------------------------
--- a/packages/f-footer/src/assets/scss/icons.scss
+++ b/packages/f-footer/src/assets/scss/icons.scss
@@ -1,5 +1,3 @@
-@import "../assets/scss/common.scss";
-
 .c-icon--chevron {
     display: none;
     height: 9px;
--- a/packages/f-footer/src/assets/scss/icons.scss
+++ b/packages/f-footer/src/assets/scss/icons.scss
@@ -1,3 +1,5 @@
+@import "../assets/scss/common.scss";
+
 .c-icon--chevron {
     display: none;
     height: 9px;
--- a/packages/f-footer/src/components/ButtonList.vue
+++ b/packages/f-footer/src/components/ButtonList.vue
@@ -35,7 +35,6 @@ export default {
 </script>
 
 <style lang="scss" module>
-
 .c-buttonList {
     align-items: center;
     display: flex;
--- a/packages/f-footer/src/components/ButtonList.vue
+++ b/packages/f-footer/src/components/ButtonList.vue
@@ -35,7 +35,6 @@ export default {
 </script>
 
 <style lang="scss" module>
-@import "../assets/scss/common.scss";
 
 .c-buttonList {
     align-items: center;
--- a/packages/f-footer/src/components/ButtonList.vue
+++ b/packages/f-footer/src/components/ButtonList.vue
@@ -35,6 +35,8 @@ export default {
 </script>
 
 <style lang="scss" module>
+@import "../assets/scss/common.scss";
+
 .c-buttonList {
     align-items: center;
     display: flex;
--- a/packages/f-footer/src/components/Footer.vue
+++ b/packages/f-footer/src/components/Footer.vue
@@ -33,7 +33,7 @@
                         :title="copy.feedback"
                         :text="copy.improveOurWebsite"
                         :button-text="copy.sendFeedback" />
-
+                        
                     <icon-list
                         :icons="copy.socialIcons"
                         :title="copy.followUs"
@@ -82,7 +82,7 @@ export default {
     },
 
     props: {
-        locale: {
+        localeProp: {
             type: String,
             default: ''
         },
@@ -92,17 +92,14 @@ export default {
             default: true
         }
     },
-
-    data () {
-        const locale = sharedServices.getLocale(tenantConfigs, this.locale, this.$i18n);
-        const localeConfig = tenantConfigs[locale];
-
-        return {
-            copy: { ...localeConfig }
-        };
-    },
-
     computed: {
+        locale () {
+            return sharedServices.getLocale(tenantConfigs, this.localeProp, this.$i18n);
+        },
+        copy () {
+            const localeConfig = tenantConfigs[this.locale];
+            return localeConfig;
+        },
         theme () {
             return sharedServices.getTheme(this.locale);
         },
--- a/packages/f-footer/src/components/Footer.vue
+++ b/packages/f-footer/src/components/Footer.vue
@@ -82,7 +82,7 @@ export default {
     },
 
     props: {
-        localeProp: {
+        locale: {
             type: String,
             default: ''
         },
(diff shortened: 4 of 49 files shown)


packages/f-footer/src/components/Footer.vue
```diff
@@ -96,15 +96,16 @@ export default {
...
 
         return {
-            copy: { ...localeConfig },
-            theme
+            copy: { ...localeConfig }
         };
     },
 
     computed: {
+        theme () {
```
siggerzz, line 103: Changed this so that we could dynamically change the theme within Storybook. Not sure what implications this could have though...

packages/f-header/src/assets/scss/common.scss
```diff
@@ -8,7 +8,7 @@ $theme: 'je' !default;
 @import '~include-media/dist/include-media'; // Cleaner media query declarations – http://include-media.com
 @import '~@justeat/fozzie/src/scss/settings/variables';
 @import '~@justeat/fozzie/src/scss/base/reset';
-@import '~@justeat/fozzie-colour-palette/src/scss/';
+@import '~@justeat/fozzie-colour-palette/src/scss/index';
```
siggerzz: Storybook was complaining for some reason about this...

packages/f-header/src/components/Header.vue
```diff
@@ -103,16 +103,17 @@ export default {
...
         const mobileNavIsOpen = false;
 
         return {
             copy: { ...localeConfig },
-            theme,
             mobileNavIsOpen
         };
     },
     computed: {
+        theme () {
```
siggerzz, line 108: ditto

packages/f-footer/src/components/Footer.vue
xander-marjoram: I think this is probably a good thing if we ever wanted to do multiple tenants/themes on a single page anyway, or experimentations with theming.

xander-marjoram:
Looks like you'll need to do a minor/patch version increment of f-header and f-footer, with changelog updates as well.

packages/f-footer/CHANGELOG.md
```diff
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Latest (roll into next release)
 ------------------------------
+*April 16, 2020*
```
xander-marjoram: If there's already a `Latest` section, those entries should be rolled into yours, like so:
```markdown
v2.1.1
*April 16, 2020*

### Added
- `stories` folder that will contain `*.stories.js` files for use with Storybook

### Changed
- Updated name of component in `index.js` from `Footer` > `VueFooter` (as `Footer` is a reserved word according to linting rules)
```

packages/f-header/CHANGELOG.md
```diff
@@ -3,9 +3,15 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-
 Latest (roll into next release)
 ------------------------------
+*April 16, 2020*
```
xander-marjoram: Ditto here

packages/f-footer/src/assets/scss/icons.scss
```diff
@@ -1,3 +1,5 @@
+@import "../assets/scss/common.scss";
```
ashleynolan: This file should get imported already [by this line](https://github.com/justeat/fozzie-components/blob/master/packages/f-footer/vue.config.js#L14)?

packages/f-footer/src/components/ButtonList.vue
```diff
@@ -35,6 +35,8 @@ export default {
 </script>
 
 <style lang="scss" module>
+@import "../assets/scss/common.scss";
```
ashleynolan: Same for all these references. Might be worth catching up about this, as don't really want to add references in for every component for this if possible.

ashleynolan (changes requested):
Might be worth catching up about the `common.scss` additions. Guessing this was a webpack issue that prompted the addition?

packages/f-footer/src/components/Footer.vue
ashleynolan: I don't think it would have too much of an implication tbh 👍

packages/f-footer/src/components/LinkList.vue
```diff
@@ -121,7 +121,7 @@ export default {
 </script>
 
 <style lang="scss" module>
-@import '@/assets/scss/icons.scss';
+@import '../assets/scss/icons.scss';
```
ashleynolan, line 124: Was the `@` failing build somewhere? This is webpack specific syntax so should work I would think (would be interested if it was failing).

packages/storybook/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@justeat/storybook",
+  "version": "0.1.0",
+  "scripts": {
+    "storybook:build": "vue-cli-service storybook:build -c config/storybook",
+    "storybook:serve": "vue-cli-service storybook:serve -p 6006 -c config/storybook"
+  },
+  "devDependencies": {
+    "@storybook/addon-actions": ">=5.3.0",
```
ashleynolan: Can you lock these versions (so just remove the characters before the number on each package?)

ashleynolan (dismissed):

packages/f-footer/src/components/LinkList.vue
siggerzz: Yeah was seeing some issue when compiling. Again, not sure if it's related to all the webpack related stuff or not 👍

siggerzz:
@ashleynolan Yep, all these `common.scss` imports are to get around the webpack related issues. One the webpack stuff is resolved, these can be removed 👍

DamianMullins (dismissed):

packages/f-footer/src/components/Footer.vue
```diff
@@ -82,7 +82,7 @@ export default {
     },
 
     props: {
-        locale: {
+        localeProp: {
```
DamianMullins: Why was the name of this prop changed? The `Prop` suffix is kind of redundant.

Also, if this is going to change then we should bump the major version of the package as it requires users to update their code.

DamianMullins: I see it was changed so you could introduce a computed property, I'd argue that the name of that property should be updated rather than the prop.

siggerzz: Because I had to make `locale` a computed in order for changes to take effect within Storybook if you changed the locale (it was part of the component data previously.)

Having the following code results in issues, because both the computed + the prop are called `locale`

```
        locale () {
            return sharedServices.getLocale(tenantConfigs, this.locale, this.$i18n);
        },
```

I can change the name of the computed instead if you like?

siggerzz: @DamianMullins Sure thing. I guess it reduces the changed file count too :)

DamianMullins (approved):

packages/f-footer/src/components/Footer.vue
tommcclean: Perfect for when we have a nuxt site and when you change from one country to another, it could be completely client side :D

packages/f-footer/CHANGELOG.md
```diff
@@ -4,14 +4,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
 
 
-Latest (roll into next release)
+v2.1.1
 ------------------------------
-*March 30, 2020*
+*April 20, 2020*
+
+### Added
+- `stories` folder that will contain `*.stories.js` files for use with Storybook
```
asherstoppard, line 12: Might be worth considering `__stories__` as a directory to ensure it's at the top and bring it inline with `__tests__`?

packages/f-header/src/components/Header.vue
```diff
@@ -23,7 +23,6 @@
                 :show-delivery-enquiry="showDeliveryEnquiryWithContent"
                 :offers-copy="copy.offers"
                 :show-offers-link="showOffersLink"
-                :show-for-you-copy="showForYouCopy"
```
tommcclean, line 26: Is removing this intended?

siggerzz: Yep. As far as I can tell, this prop wasn't actually doing anything. Although if you can prove me wrong, I'll add it back in 👀

tommcclean:
I think you should add a readme to the storybook folder - so people can read instructions, and what it is

siggerzz:
> I think you should add a readme to the storybook folder - so people can read instructions, and what it is

@tommcclean Does this suffice - https://lmgtfy.com/?q=storybookjs ? 😁 

In all seriousness though, what would you want me to add?

tommcclean:
> > I think you should add a readme to the storybook folder - so people can read instructions, and what it is
> 
> @tommcclean Does this suffice - https://lmgtfy.com/?q=storybookjs ? 😁
> 
> In all seriousness though, what would you want me to add?

Run commands, any troubleshooting tips, why we use it. In brief

siggerzz:
Spoke with @tommcclean about adding a readme / updating PR template. Will do this as part of my next ticket, which will be to host Storybook.

tommcclean (approved):

packages/f-footer/CHANGELOG.md
siggerzz: Are we doing this with tests? I couldn't see an example anywhere?

PR #164, opened 2020-04-17
f-registration@v0.2.0 - `f-registration` base HTML structure and styling

ashleynolan:
### Added
- `f-registration` base HTML structure and styling has been added.

### Changed
- Components separated out into separate `.vue` files.

### DIFF
--- a/packages/f-registration/src/components/Button.vue
+++ b/packages/f-registration/src/components/Button.vue
@@ -3,23 +3,20 @@
         type='submit'
         :class="[
             $style['o-btn'],
-            $style[buttonTypeClass],
+            $style[buttonStyleClass],
             (isFullWidth ? $style['o-btn--block'] : '')
         ]"
     >
-        {{ buttonText }}
+        <slot />
     </button>
 </template>
 
 <script>
 export default {
     name: 'ButtonComponent',
+
     props: {
-        buttonText: {
-            type: String,
-            default: ''
-        },
-        buttonType: {
+        buttonStyle: {
             type: String,
             default: ''
         },
@@ -29,11 +26,10 @@ export default {
         }
     },
     computed: {
-        buttonTypeClass () {
-            if (this.buttonType) {
-                return `o-btn--${this.buttonType}`;
-            }
-            return '';
+        buttonStyleClass () {
+            return (this.buttonStyle)
+                ? `o-btn--${this.buttonStyle}`
+                : '';
         }
     }
 };
@@ -61,6 +57,20 @@ $btn-primary-textColor              : $white;
 
 $btn-disabled-bgColor               : $grey--lighter;
 
+
+/**
+ * Base button styles – Based on csswizardry.com/beautons
+ *
+ * 1. Allow us to better style box model properties.
+ * 2. Line different sized buttons up a little nicer.
+ * 3. Make buttons inherit font styles.
+ * 4. Force all elements using beautons to appear clickable.
+ * 5. Normalise box model styles.
+ * 6. If the button’s text is 1em, and the button is (3 * font-size) tall, then
+ *    there is 1em of space above and below that text. We therefore apply 1em
+ *    of space to the left and right, as padding, to keep consistent spacing.
+ * 7. Fixes odd inner spacing in IE7.
+ */
 
 .o-btn {
     display: inline-block;                      /* [1] */
--- a/packages/f-registration/src/components/Button.vue
+++ b/packages/f-registration/src/components/Button.vue
@@ -1,0 +1,172 @@
+<template>
+    <button
+        type='submit'
+        :class="[
+            $style['o-btn'],
+            $style[buttonTypeClass],
+            (isFullWidth ? $style['o-btn--block'] : '')
+        ]"
+    >
+        {{ buttonText }}
+    </button>
+</template>
+
+<script>
+export default {
+    name: 'ButtonComponent',
+    props: {
+        buttonText: {
+            type: String,
+            default: ''
+        },
+        buttonType: {
+            type: String,
+            default: ''
+        },
+        isFullWidth: {
+            type: Boolean,
+            default: false
+        }
+    },
+    computed: {
+        buttonTypeClass () {
+            if (this.buttonType) {
+                return `o-btn--${this.buttonType}`;
+            }
+            return '';
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+
+// Button styling (should be pulled out as a separate component)
+
+$btn-default-bgColor                : $grey--lightest;
+$btn-default-text-colour            : $grey--dark;
+$btn-default-weight                 : 500;
+$btn-default-height                 : 2.6;
+$btn-default-font-size              : base--scaleUp;
+$btn-default-font-family            : $font-family-headings;
+$btn-default-bgColor--hover         : $grey--lighter;
+$btn-default-borderRadius           : 2px;
+$btn-default-hozPadding             : 1em;
+$btn-default-vertPadding            : 11px;
+
+$btn-primary-bgColor                : $blue;
+$btn-primary-bgColor--hover         : $blue--dark;
+$btn-primary-bgColor--focus         : $blue--darkest;
+$btn-primary-textColor              : $white;
+
+$btn-disabled-bgColor               : $grey--lighter;
+
+
+.o-btn {
+    display: inline-block;                      /* [1] */
+    vertical-align: middle;                     /* [2] */
+    font-family: $btn-default-font-family;      /* [3] */
+    @include font-size($btn-default-font-size); /* [3] */
+    cursor: pointer;                            /* [4] */
+    margin: 0;                                  /* [5] */
+    padding: $btn-default-vertPadding $btn-default-hozPadding; /* [5, 6] */
+    overflow: visible;                          /* [7] */
+    text-align: center;
+    font-weight: $btn-default-weight;
+    line-height: 1;
+
+    // You may want to change this
+    background-color: $btn-default-bgColor;
+    border-radius: $btn-default-borderRadius;
+    border: 1px solid transparent;
+    user-select: none;
+
+    color: $btn-default-text-colour;
+    text-decoration: none;
+
+    &:hover,
+    &:active,
+    &:focus {
+        background-color: $btn-default-bgColor--hover;
+
+        &:not(.o-btnLink) {
+            outline: none; // no need as already has a focus/active state
+        }
+    }
+
+    &,
+    &:hover,
+    &:active,
+    &:focus,
+    &:visited {
+        text-decoration: none;
+    }
+
+    // Disabled state
+    &.disabled,
+    &[disabled] {
+        cursor: default;
+
+        &,
+        &:hover {
+            background-color: $btn-disabled-bgColor;
+        }
+    }
+
+    &[type='submit'] {
+        margin-top: spacing(x2);
+    }
+}
+
+/**
+ * Modifier – .o-btn--primary
+ *
+ * Sets the btn colour to site primary colour
+ */
+
+.o-btn--primary {
+    background-color: $btn-primary-bgColor;
+
+    &,
+    &:link,
+    &:visited {
+        color: $btn-primary-textColor;
+    }
+
+    &:hover,
+    &:active,
+    &:focus {
+        color: $btn-primary-textColor;
+    }
+
+    &:hover {
+        background-color: $btn-primary-bgColor--hover;
+    }
+
+    &:active,
+    &:focus {
+        background-color: $btn-primary-bgColor--focus;
+    }
+}
+
+/**
+ * Modifier – btn-block
+ *
+ * Makes the btn full width
+ */
+
+.o-btn--block {
+    display: block;
+    width: 100%;
+    padding-left: 0;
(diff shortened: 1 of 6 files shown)


packages/f-registration/src/components/Button.vue
```diff
@@ -0,0 +1,172 @@
...
+            type: Boolean,
+            default: false
+        }
+    },
+    computed: {
+        buttonTypeClass () {
+            if (this.buttonType) {
+                return `o-btn--${this.buttonType}`;
+            }
+            return '';
```
tommcclean:
```suggestion
            return (this.buttonType)
                ? `o-btn--${this.buttonType}`
                : '';
```

packages/f-registration/src/components/Card.vue
```diff
@@ -0,0 +1,102 @@
+<template>
+    <div
+        :data-theme="theme"
```
tommcclean: I think `theme` needs to be defined

packages/f-registration/src/components/Button.vue
```diff
@@ -0,0 +1,172 @@
...
+
+<script>
+export default {
+    name: 'Button',
+    props: {
+        buttonText: {
+            type: String,
+            default: ''
+        },
+        buttonType: {
```
xander-marjoram: Could there be some confusion here around the `type` attribute? Would `buttonStyle` be a better name?

packages/f-registration/src/components/Button.vue
```diff
@@ -0,0 +1,172 @@
+<template>
+    <button
+        type='submit'
+        :class="[
+            $style['o-btn'],
+            $style[this.buttonTypeClass],
+            (isFullWidth ? $style['o-btn--block'] : '')
```
DamianMullins, line 7:
```suggestion
            { $style['o-btn--block']: isFullWidth }
```

packages/f-registration/src/components/Button.vue
```diff
@@ -0,0 +1,172 @@
...
+        ]"
+    >
+        {{ buttonText }}
+    </button>
+</template>
+
+<script>
+export default {
+    name: 'Button',
+    props: {
```
tommcclean, line 18: I notice we tend to put line breaks between items in the object normally, should we apply that here too? (Same for other component too)

packages/f-registration/src/components/Button.vue
```diff
@@ -0,0 +1,172 @@
...
+$btn-primary-bgColor                : $blue;
+$btn-primary-bgColor--hover         : $blue--dark;
+$btn-primary-bgColor--focus         : $blue--darkest;
+$btn-primary-textColor              : $white;
+
+$btn-disabled-bgColor               : $grey--lighter;
+
+
+.o-btn {
+    display: inline-block;                      /* [1] */
```
DamianMullins, line 76: What do the numbers relate to in the comments here?

packages/f-registration/src/components/Button.vue
```diff
@@ -0,0 +1,172 @@
+<template>
```
DamianMullins, line 1: This feels like something which could potentially be used in multiple projects, do you think it's worth considering pulling components like this out into another project in the future?

packages/f-registration/src/components/Button.vue
```diff
@@ -0,0 +1,172 @@
...
+    <button
+        type='submit'
+        :class="[
+            $style['o-btn'],
+            $style[this.buttonTypeClass],
+            (isFullWidth ? $style['o-btn--block'] : '')
+        ]"
+    >
+        {{ buttonText }}
```
DamianMullins: I think it would be better to use `<slot />` here which means that when the component is used you do this instead of passing a prop:


```js
<form-button>
    Create Account
</form-button>
```

packages/f-registration/src/components/FormField.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+    <div class="c-formField">
+        <label :class="$style['o-form-label', 'c-formField-label']">{{ labelText }}</label>
```
tommcclean: Wonder if we should use a `v-if` here given the default is blank (or make it required)

packages/f-registration/src/components/Card.vue
```diff
@@ -0,0 +1,102 @@
...
+        :class="[
+            $style['c-card'],
```
xander-marjoram, line 4: What does this syntax give us that `class="c-card c-card--rounded c-card--outline c-card--contentWrapper"` wouldn't?

packages/f-registration/src/components/Button.vue
```diff
@@ -0,0 +1,172 @@
...
+    </button>
+</template>
+
+<script>
+export default {
+    name: 'Button',
```
DamianMullins: Typically a Vue component name is made up of at least two parts in order to avoid conflicting with standard HTML elements, like this one would, maybe you could prefix button with something to differentiate it from a standard HTML button?

packages/f-registration/src/components/Card.vue
```diff
@@ -0,0 +1,102 @@
...
+        ]"
+    >
+        <h1
+
+            :class="[
+                'beta',
+                $style['c-card-heading']
+            ]"
+        >
+            {{ cardTitle }}
```
DamianMullins, line 17: You can use a named `slot` here to allow users to add content rather than using a `prop`.

packages/f-registration/src/components/Registration.vue
```diff
@@ -1,18 +1,49 @@
...
+                input-type="email" />
+
+            <form-field
+                label-text="Password"
+                input-type="password" />
+
+            <form-button
+                button-text="Create Account"
+                button-type="primary"
+                :is-full-width="true" />
```
DamianMullins: You can just do `is-full-width` here.

packages/f-registration/src/components/FormField.vue
xander-marjoram: Should we wrap the label element around the input? That gives us the implicit link between the two without needing `for` and `id`.

packages/f-registration/src/components/Button.vue
ashleynolan: Good shout – will update

packages/f-registration/src/components/Card.vue
ashleynolan: Shouldn't have moved this over – will put it back in the Registration component.

packages/f-registration/src/components/Button.vue
ashleynolan: Yeh, that's a better name. I'll switch it over 👍

packages/f-registration/src/components/Button.vue
ashleynolan: So between `name` and `props`? This should be part of our ESLint ruleset if we want to apply this everywhere tbh, as it'll be too hard to keep reviewing and could get inconsistent.

If we do want this, I'll raise a PR to add this to the vue eslint ruleset.

packages/f-registration/src/components/Card.vue
ashleynolan: It's syntax that's needed to connect the classes to the styling for CSS Modules. Without this syntax, the classnames don't get linked when the styles are run through CSS Modules.

packages/f-registration/src/components/Button.vue
ashleynolan: This errors when I try to switch this – I get `'v-bind' directives require an attribute value`. I think it's because the attributes are already inside an array that it doesn't parse this properly.

packages/f-registration/src/components/Button.vue
ashleynolan: I missed the comment out that should be included as part of this. Will add this In.

packages/f-registration/src/components/Button.vue
ashleynolan: Yep – that'll be my next PR 😄 

Going to do the card component first, then the form field and then button. Just wanted to get a whole component in first so we could demo it in the POC.

packages/f-registration/src/components/Button.vue
ashleynolan: Good idea – have changed

packages/f-registration/src/components/Button.vue
ashleynolan: I've left it like this for now, as for some reason this is meant to fail our ESLint rules, but isn't atm (so wanted to leave it like this as a test case).

Will look at why this isn't reporting though shortly and then will change it as part of that PR if that's ok?

packages/f-registration/src/components/Card.vue
ashleynolan: Is there a clear benefit to doing it here? I've just had a go at doing it, but it seems like essentially would need to change to:

```vue
<template>
    <card
        :data-theme="theme"
    >
        <template v-slot:title>
            Create Account
        </template>
        <form
            type="post"
            :class="$style['o-form']"
        >
```

To the `Registration.vue` and then the Card.vue would look like this:

```vue
<template>
    <div
        :class="[
            $style['c-card'],
            $style['c-card--rounded'],
            $style['c-card--outline'],
            $style['c-card--contentWrapper']
        ]"
    >
        <h1
            v-if="$slots.title"
            :class="[
                'beta',
                $style['c-card-heading']
            ]"
        >
            <slot name="title" />
        </h1>
        <slot />
    </div>
</template>
```

Just wondering if this is actually simpler to use – seems slightly more verbose for anyone wanting to add a title potentially, so was wondering if there a clear benefit otherwise?

packages/f-registration/src/components/Registration.vue
ashleynolan: Done

ashleynolan:
@DamianMullins @xander-marjoram @tommcclean Have actioned those comments – could you take another look?

packages/f-registration/src/components/Button.vue
tommcclean: Yeah all children seems to be the consensus, even between the things inside methods and props

DamianMullins: Agree, it makes the components much easier to scan by adding a little breathing space.

packages/f-registration/src/components/Button.vue
DamianMullins: OK, happy for this to remain as is, would be good to figure out what is happening at some point.

packages/f-registration/src/components/Button.vue
DamianMullins: Sure 👍

packages/f-registration/src/components/Card.vue
DamianMullins: So we use slots in the MegaModal component in this way as it allows you to also pass in some custom markup if required (like adding `span`'s etc). If you don't think you'd want to allow that then a prop is fine.

DamianMullins: Also, you can use the named slot shorthand like this `<template #title>` which is a little bit nicer.

packages/f-registration/src/components/FormField.vue
ashleynolan: The short answer is – it depends.

I've still got to style the label with respect to the inputs, so this may change dependent on that. Usually, the reason to wrap an input with a label (or not) is due to the limitations [this sometimes causes on styling](https://stackoverflow.com/questions/11992026/is-it-better-to-wrap-the-label-tag-around-a-form-item-or-use-the-for-attribute). I'll have a play around with this as part of the label styling separately.

packages/f-registration/src/components/Card.vue
ashleynolan: Ok,  makes sense when wanting to allow for `span`'s or other markup.

I'll probably leave it as it is for now, as I think the styling for the card heading shouldn't need extra markup for now looking at the current components. Can always change it to use slots then if needed if that sounds ok?

packages/f-registration/src/components/Button.vue
ashleynolan: I've added the line breaks for now. We should look to see if we can add this as part of the ESLint config though to save having to review styling like this ideally.

packages/f-registration/src/components/Button.vue
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * 1. Allow us to better style box model properties.
+ * 2. Line different sized buttons up a little nicer.
+ * 3. Make buttons inherit font styles.
+ * 4. Force all elements using beautons to appear clickable.
+ * 5. Normalise box model styles.
+ * 6. If the button’s text is 1em, and the button is (3 * font-size) tall, then
+ *    there is 1em of space above and below that text. We therefore apply 1em
+ *    of space to the left and right, as padding, to keep consistent spacing.
+ * 7. Fixes odd inner spacing in IE7.
```
DamianMullins, line 72: Do we care about this?

DamianMullins (approved):

xander-marjoram (approved):

packages/f-registration/src/components/Button.vue
ashleynolan: Maybe not – I'll test it with and without when I do more full-on browser testing and get rid of it if it doesn't seem needed

PR #165, opened 2020-04-21
generator-component@0.1.0 - Initial version

SachaZvetelman:
v0.1.0
*April 21, 2020*

### Added
- First version to be able to create new components.


packages/generator-component/generators/app/index.js
```diff
@@ -0,0 +1,54 @@
...
+
+        this.registerTransformStream(rename(path => {
+            path.basename = path.basename
+                                .replace(/(Skeleton)/g, nameTransformations.filename)
+                                .replace(/__/g, ''); // We don't want to have file names such as .test.js or .stories.js, otherwise Jest or Storybook will pick them up from the templates folder.
+        }));
+
+        this.fs.copyTpl(
+            this.templatePath('**/*'),
+            this.destinationPath(`../f-${this.answers.name}/`),
```
siggerzz: Can this just be `name` instead? Given all the other existing components are lowercase?

packages/generator-component/generators/app/index.js
```diff
@@ -0,0 +1,54 @@
...
+            {
+                description: this.answers.description,
+                name: nameTransformations
+            },
+            null,
```
siggerzz, line 46: What does this `null` represent?

packages/generator-component/generators/app/templates/stories/Skeleton.__stories__.js
```diff
@@ -0,0 +1,12 @@
+import { storiesOf } from '@storybook/vue';
+import { select, boolean } from '@storybook/addon-knobs';
```
siggerzz: This can be removed as it's not currently being used.

packages/generator-component/generators/app/index.js
SachaZvetelman: Good point. Missed that one!
Actually, gonna use `nameTransformations.default` to keep it consistent with other usages.

packages/generator-component/generators/app/index.js
SachaZvetelman: That parameter is for `templateOptions`, which we are not using. We don't want to specify any `templateOptions`, so we go for `null`.

packages/generator-component/generators/app/templates/stories/Skeleton.__stories__.js
```diff
@@ -0,0 +1,12 @@
...
+import <%= name.component %> from '../src/components/<%= name.filename %>.vue';
+
+storiesOf('Shared', module)
```
siggerzz: Might want to change this to `storiesOf('Components', module)` as this will determine the category the component is shown in within Storybook. As seen here -

image


I only put f-footer / f-header` in a 'Shared' category because of their usage across the site (Although on second thought, I should probably rename it to 'Global') 😁

packages/generator-component/generators/app/templates/stories/Skeleton.__stories__.js
```diff
@@ -0,0 +1,12 @@
...
+  .add('f-<%= name.default %>', () => ({
+    components: { <%= name.component %> },
+    props: {
```
siggerzz: `props: {}` can also be removed as it's not being used.

SachaZvetelman: I actually left this on purpose since most of our components will have `props`. But I can remove it.

siggerzz (changes requested):
Just a few comments!

packages/generator-component/generators/app/templates/stories/Skeleton.__stories__.js
siggerzz: Yeah that's fair enough, happy to leave it in :)

siggerzz (dismissed):
🎉

packages/generator-component/generators/app/templates/src/components/Skeleton.vue
```diff
@@ -1,17 +1,17 @@
...
-        :class="$style['c-skeleton']">
-        I am a Skeleton Component
+        :class="$style['c-<%= name.class %>']">
+        I am a <%= name.readme %> Component
     </div>
 </template>
 
 <script>
-import sharedServices from '@justeat/f-services';
+import { globalisationServices } from '@justeat/f-services';
```
xander-marjoram, line 10: ~This file is still called Skeleton.vue, is that correct?~

We should be referencing f-services@>1.0.0 in [this file](https://github.com/justeat/fozzie-components/pull/165/files#diff-64ebc12b31b14b18e9f14fca28a7d2c5) if we're importing globalisation services like this

packages/generator-component/generators/app/templates/src/components/tests/Skeleton.__test__.js
```diff
@@ -0,0 +1,10 @@
+import { shallowMount } from '@vue/test-utils';
```
xander-marjoram, line 1: ~Does this file also need renaming?~

packages/generator-component/generators/app/templates/src/components/Skeleton.vue
```diff
@@ -1,17 +1,17 @@
 <template>
     <div
         :data-theme="theme"
-        :class="$style['c-skeleton']">
-        I am a Skeleton Component
+        :class="$style['c-<%= name.class %>']">
+        I am a <%= name.readme %> Component
```
xander-marjoram: Can we get this to use i18n out-of-the-box? I think you could just change this to `$t('skeletonText')` (maybe rename that to just 'text' or something?)

packages/generator-component/generators/app/index.js
```diff
@@ -0,0 +1,54 @@
...
+            component: utils.getComponentName(name), // e.g VueHeader or UserMessage,
+            default: name, // e.g. header or user-message
+            filename: utils.getComponentFilename(name), // Header(.vue) or UserMessage(.vue)
+            readme: utils.getReadmeName(name), // Header or User Message
+            template: utils.getComponentTemplateName(name) // vue-header or user-message,
+        };
+
+        this.registerTransformStream(rename(path => {
+            path.basename = path.basename
+                                .replace(/(Skeleton)/g, nameTransformations.filename)
```
xander-marjoram, line 35: Ah I see, this is how we do the renaming?

SachaZvetelman: Yup, that's correct.

xander-marjoram (dismissed):

siggerzz (dismissed):

siggerzz (approved):

xander-marjoram (approved):

packages/generator-component/generators/app/templates/LICENSE
```diff
@@ -0,0 +1,17 @@
+Apache License
+Version 2.0, January 2004
+http://www.apache.org/licenses/
+
+Copyright 2020 Just Eat Holding Ltd
```
ashleynolan, line 5: Future improvement that would be cool would be to make sure the year here uses the current year (rather than always 2020)

ashleynolan (approved):
Looks good – one thing that would be a good addition, but shouldn't stop this merging.

Would be happy to test this when I create the form-field component btw

ashleynolan:
@SachaZvetelman Looks like the Danger fail is a false positive so let me know when you want this merged and will just push it in.

PR #168, opened 2020-04-27
f-card@v0.2.0 – Updating component with markup and core functionality

ashleynolan:
### Added
- Component markup and core functionality added.
- Storybook config added.

---

## UI Review Checks

Screen Shot 2020-04-27 at 14 29 30

## Browsers Tested

- [x] Chrome (latest)

### DIFF
--- a/packages/f-card/README.md
+++ b/packages/f-card/README.md
@@ -68,4 +68,62 @@ yarn install
 yarn demo
 ```
 
-## Documentation to be completed once module is in stable state.
+### Storybook
+
+The component is also available to demo through our storybook instance which can be served locally by running `yarn storybook:serve` from the mono-repo root.
+
+
+## Props
+
+`f-card` has a number of props that allow you to customise its functionality.
+
+The props that can be defined are as follows:
+
+<table class="js-options">
+    <tr>
+        <th align="left">Prop</th>
+        <th>Required</th>
+        <th>Type</th>
+        <th>Default</th>
+        <th align="left">Description</th>
+    </tr>
+    <tr>
+        <td><code>locale</code></td>
+        <td align="center">false</td>
+        <td align="center">string</td>
+        <td align="center"><code>en-GB</code></td>
+        <td>
+          <p>Sets the locale of the component (which determines what theme and translations to use.</p>
+          <p>If the application consuming the <code>f-card</code> component is using the vue <code>i18n</code> module, then the locale from that module will be used when this prop isn't defined. When this prop is defined, it takes precedence over the locale defined by the <code>i18n</code> module.</p>
+          <p>If not defined and the <code>i18n</code> module isn't present, the default locale used is <code>en-GB</code>.</p>
+        </td>
+    </tr>
+    <tr>
+        <td><code>isRounded</code></td>
+        <td align="center">false</td>
+        <td align="center">boolean</td>
+        <td align="center"><code>false</code></td>
+        <td>
+            When set to <code>true</code>, rounded corners are applied to the card component.
+        </td>
+    </tr>
+    <tr>
+        <td><code>hasOutline</code></td>
+        <td align="center">false</td>
+        <td align="center">boolean</td>
+        <td align="center"><code>false</code></td>
+        <td>
+            When set to <code>true</code>, an outline is applied to the card component.
+        </td>
+    </tr>
+    <tr>
+        <td><code>isPageContentWrapper</code></td>
+        <td align="center">false</td>
+        <td align="center">boolean</td>
+        <td align="center"><code>false</code></td>
+        <td>
+            <p>When set to <code>true</code>, applies styles to make the card act like a page content wrapper.</p>
+            <p>The card will be full width on narrow devices, and then a fixed width above a certain breakpoint width (about 480px), when the card will be centred on the page.</p>
+        </td>
+    </tr>
+</table>
--- a/packages/f-card/README.md
+++ b/packages/f-card/README.md
@@ -79,51 +79,9 @@ The component is also available to demo through our storybook instance which can
 
 The props that can be defined are as follows:
 
-<table class="js-options">
-    <tr>
-        <th align="left">Prop</th>
-        <th>Required</th>
-        <th>Type</th>
-        <th>Default</th>
-        <th align="left">Description</th>
-    </tr>
-    <tr>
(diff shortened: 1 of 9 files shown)


packages/f-card/src/components/tests/Card.test.js
```diff
@@ -3,8 +3,113 @@ import Card from '../Card.vue';
...
         const propsData = {};
+
+        // Act
         const wrapper = shallowMount(Card, { propsData });
+
+        // Assert
         expect(wrapper.exists()).toBe(true);
     });
+
+    it('cardTitle shouldn’t be visible if it’s not set in props', () => {
```
kevinrodrigues: Could you start these with `should`, also might make it more readable and easier to write if the item you're testing is wrapped in a describe. e.g: 

```javascript
describe('cardTitle', () => {
  it('should not be visible if it’s not set in props')

  it() ....
})

File: packages/f-card/src/components/Card.vue [hunk 2]
```diff
@@ -17,32 +32,93 @@ export default {
...
+
+        @include media('>=narrow') {
+            padding-left: 10%;
+            padding-right: 10%;
+        }
+
+        @include media('>=#{$card--pageContentWrapper-width}') {
+            width: $card--pageContentWrapper-width;
+            margin: 0 auto;
+            padding-left: spacing() * 10;
```
kevinrodrigues, line 113: Out of interest would be make more sense to just use 80px here and below instead of the calculation?

ashleynolan: Should really just add it into `fozzie` so I can use the spacing function directly. I'll get that added as a separate PR and then update this in a separate PR if that's ok?

packages/f-card/src/components/tests/Card.test.js
ashleynolan: 👍 Will get these updated

packages/f-card/src/components/Card.vue
ashleynolan: Opened [this PR](https://github.com/justeat/fozzie/pull/295), that'll help address this.

packages/f-card/src/demo/Index.vue
```diff
@@ -25,5 +38,6 @@ export default {
 
 body {
     margin: 0;
+    background: #f1f1f1;
```
InsungMulumba, line 41: Would it be worth adding a variable for this colour in fozzie?

packages/f-card/README.md
```diff
@@ -68,4 +68,62 @@ yarn install
...
+The component is also available to demo through our storybook instance which can be served locally by running `yarn storybook:serve` from the mono-repo root.
+
+
+## Props
+
+`f-card` has a number of props that allow you to customise its functionality.
+
+The props that can be defined are as follows:
+
+<table class="js-options">
```
DamianMullins: Any reason you didn't use markdown syntax here? I'm fine with HTM but the markdown purist in me wanted to know 😸

packages/f-card/src/components/tests/Card.test.js
```diff
@@ -3,8 +3,121 @@ import Card from '../Card.vue';
 
 describe('Card', () => {
     it('should be defined', () => {
+        // Arrange
```
DamianMullins, line 6: 🚀

packages/f-card/src/components/tests/Card.test.js
```diff
@@ -3,8 +3,121 @@ import Card from '../Card.vue';
...
+                const propsData = {};
+
+                // Act
+                const wrapper = shallowMount(Card, { propsData });
+
+                // Assert
+                expect(wrapper.vm.isRounded).toBe(false);
+            });
+
+            it('should be set to `true` if the `isRounded` prop is passed in as `true`', () => {
```
DamianMullins, line 57: Feels like we're testing that Vue is assigning props correctly with some of these `props` tests, do you think they are required?

packages/f-card/src/demo/Index.vue
```diff
@@ -7,7 +7,20 @@
             name="viewport"
             content="width=device-width, initial-scale=1">
         <card-component
-            locale="en-GB" />
+            locale="en-GB"
+            card-heading="I am a demo card"
+            is-rounded=true
```
DamianMullins: You can omit the `true` part of these props.

packages/f-card/src/demo/Index.vue
ashleynolan: The demo files will b deleted shortly in favour of using storybook for testing locally, so this'll disappear with that when I take those out. Would have been otherwise though 👍

packages/f-card/README.md
ashleynolan: Mainly as I'm not sure it was possible to do the styling I wanted all in markdown. I can have a go converting it over at some point though and let you know how I get on 😉

packages/f-card/src/components/tests/Card.test.js
ashleynolan: They're kind of placeholder tests atm. I want to test that the classes get assigned in the rendered markup when the prop is set, but for some reason `vue-jest` doesn't render the CSS module classes straight out the box (tried fixing but I think it's going to take a while to get this working).

I'll be coming back to these tests once I've finished the initial run of the registration component, so will change these then (as agree, testing the data assignment isn't that useful).

packages/f-card/README.md
DamianMullins: I wouldn't worry about it :)

packages/f-card/src/components/tests/Card.test.js
DamianMullins: Sure.

packages/f-card/src/demo/Index.vue
ashleynolan: Removed

packages/f-card/README.md
ashleynolan: As I had to update something else anyway, I had a go and it wasn't as hard as I thought. So have updated now 👍

ashleynolan:
@kevinrodrigues @DamianMullins @InsungMulumba Could I get another look – have updated based on the feedback now 👍

DamianMullins (approved):

kevinrodrigues (approved):

PR #177, opened 2020-05-06
f-services@1.0.1 - Change index.js syntax so that Storybook is compatible.

siggerzz:
- As of f-services 1.0.0, the `getLocale` function doesn't work with Storybook. This PR will fix that issue, thanks to @xander-marjoram :)

### DIFF
--- a/packages/f-services/src/index.js
+++ b/packages/f-services/src/index.js
@@ -1,9 +1,8 @@
-/**
- * @overview Fozzie's collection of shared javascript modules
- *
- * @module f-services
- */
-
-export { default as axiosServices } from './axios';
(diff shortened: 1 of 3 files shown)


packages/f-services/src/index.js
```diff
@@ -1,9 +1,8 @@
-/**
- * @overview Fozzie's collection of shared javascript modules
- *
- * @module f-services
- */
```
xander-marjoram, line 5: Please can you re-add the module description?

minrwhite (dismissed):

xander-marjoram (approved):

minrwhite (approved):

PR #179, opened 2020-05-11
fozzie-components@v1.12.0 – Updating vue-cli-unit-jest across all packages

ashleynolan:
** N.b. I would usually split these updates into separate PRs, but upgrading the test framework version meant putting them all into one PR that can go out together. **

---

### Changed
- Updated some base dependencies (all minor version bumps).

---

### Added
- Added some simple initial tests.

### Changed
- Updating `vue-test-utils` to v1 and `@vue/cli-plugin-unit-test` to v4.3.1.
- Use `node current` in unit test Babel config, so that it supports `async > await` properly.

### Removed
- `testMatch` from jest config, as not needed.

---

### Fixed
- Have added proper mocking for API data via `axios`. Quite a few of the tests were trying to call API endpoints and were passing without actually testing anything (but passed giving the impression they were). Have fixed this now so that these tests all work as expected with proper mocks.
- Fixed some tests that broke when upgrading to the latest `vue-test-utils`.

### Changed
- Updated some of the methods that handle API calls to return as promises, so that they are easier to unit test.
- Updating `vue-test-utils` to v1 and `@vue/cli-plugin-unit-test` to v4.3.1.
- Use `node current` in unit test Babel config, so that it supports `async > await` properly.

### Removed
- `testMatch` from jest config, as not needed.

---

### Changed
- Updated template to use `vue-test-utils` to v1 and `@vue/cli-plugin-unit-test` to v4.3.1.

### Removed
- `testMatch` from jest config, as not needed.

---

## UI Review Checks

### DIFF
--- a/packages/f-form-field/src/components/tests/FormField.test.js
+++ b/packages/f-form-field/src/components/tests/FormField.test.js
@@ -1,10 +1,74 @@
 import { shallowMount } from '@vue/test-utils';
 import FormField from '../FormField.vue';
+import { DEFAULT_INPUT_TYPE, VALID_INPUT_TYPES } from '../../constants';
+
 
 describe('FormField', () => {
     it('should be defined', () => {
         const propsData = {};
         const wrapper = shallowMount(FormField, { propsData });
         expect(wrapper.exists()).toBe(true);
+    });
+
+    describe('props ::', () => {
+        beforeEach(() => {
+            console.error = jest.fn();
+        });
+        describe('inputType ::', () => {
+            it('should be set to type `text` by default if no value is set', () => {
+                // Arrange
+                const propsData = {};
+
+                // Act
+                const wrapper = shallowMount(FormField, { propsData });
+                const formInput = wrapper.find('input'); // change to .c-formField when CSS Modules is working
+
+                // Assert
+                expect(formInput.attributes('type')).toBe(DEFAULT_INPUT_TYPE);
+            });
+
+            it('should set the type of the input element if the prop value passed in is valid', () => {
+                // Arrange
+                const propsData = {
+                    inputType: 'email'
+                };
+
+                // Act
+                const wrapper = shallowMount(FormField, { propsData });
+                const formInput = wrapper.find('input'); // change to .c-formField when CSS Modules is working
+
+                // Assert
+                expect(formInput.attributes('type')).toBe(propsData.inputType);
+            });
+
+            it('should fallback to type `text` if inputType is invalid', () => {
+                // Arrange
+                const propsData = {
+                    inputType: 'invalidType'
+                };
+
+                // Act
+                const wrapper = shallowMount(FormField, { propsData });
+                const formInput = wrapper.find('input'); // change to .c-formField when CSS Modules is working
+
+                // Assert
+                expect(console.error).toHaveBeenCalledTimes(1); // checks that vue is throwing an error because of the defined prop validation function
+                expect(formInput.attributes('type')).toBe(DEFAULT_INPUT_TYPE);
+            });
+
+            it.each(VALID_INPUT_TYPES)('should set the type of form input element as expected if inputType=%p is specified', definedType => {
+                // Arrange
+                const propsData = {
+                    inputType: definedType
+                };
+
+                // Act
+                const wrapper = shallowMount(FormField, { propsData });
+                const formInput = wrapper.find('input'); // change to .c-formField when CSS Modules is working
+
+                // Assert
+                expect(formInput.attributes('type')).toBe(definedType);
+            });
+        });
     });
 });
--- a/packages/f-form-field/src/components/tests/FormField.test.js
+++ b/packages/f-form-field/src/components/tests/FormField.test.js
@@ -11,9 +11,6 @@ describe('FormField', () => {
     });
 
     describe('props ::', () => {
-        beforeEach(() => {
-            console.error = jest.fn();
-        });
         describe('inputType ::', () => {
             it('should be set to type `text` by default if no value is set', () => {
                 // Arrange
@@ -46,6 +43,7 @@ describe('FormField', () => {
                 const propsData = {
                     inputType: 'invalidType'
                 };
+                console.error = jest.fn();
 
                 // Act
                 const wrapper = shallowMount(FormField, { propsData });
--- a/packages/f-header/CHANGELOG.md
+++ b/packages/f-header/CHANGELOG.md
@@ -4,7 +4,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-v2.4.9
+v2.4.0
 ------------------------------
 *May 12, 2020*
 
--- a/packages/f-header/CHANGELOG.md
+++ b/packages/f-header/CHANGELOG.md
@@ -3,6 +3,24 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+v2.4.9
+------------------------------
+*May 12, 2020*
+
+### Fixed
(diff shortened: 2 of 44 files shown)


packages/f-header/CHANGELOG.md
```diff
@@ -3,6 +3,24 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+v2.4.9
```
xander-marjoram: 2.4.0

packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -67,13 +74,13 @@ describe('Navigation', () => {
...
         const wrapper = shallowMount(Navigation, { propsData });
-        wrapper.vm.data = defaultData;
+        await wrapper.setData(defaultData); // need to await this for the state to fully update the DOM
 
         // Assert
         expect(wrapper.find('[data-js-test="delivery-enquiry"]').exists()).toBe(true);
     });
 
-    it('should NOT show "Deliver for Just Eat" in nav if prop false and the content is there', () => {
+    it('should NOT show "Deliver for Just Eat" in nav if prop false and the content is there', async () => {
```
xander-marjoram: "should NOT show delivery enquiry link ..." rather than the expected text, which may change, and could also be Menulog, etc

ashleynolan: Have updated

packages/f-header/src/components/tests/Logo.test.js
```diff
@@ -53,14 +53,14 @@ describe('Logo', () => {
         };
 
         // Act
-        const wrapper = shallowMount(Logo, { propsData });
+        const wrapper = mount(Logo, { propsData });
         const logo = wrapper.find('[data-js-test="c-icon--je"]');
 
         // Assert
         // with dynamically rendered components
```
asherstoppard: This comment can now be removed :+1:

packages/f-header/src/components/tests/Logo.test.js
```diff
@@ -72,14 +72,14 @@ describe('Logo', () => {
...
```
asherstoppard: This comment can now be removed 👍

packages/f-header/src/components/tests/__mocks__/axios.js
```diff
@@ -0,0 +1,22 @@
+/**
```
asherstoppard, line 1: I'd consider using AxiosMockAdapter. It gives you a little more control over error states and the interface is a lot less stringent

asherstoppard:
Might be worth re-considering the `__test__` directory name as it keeps it isolated from the production code and makes it easier to spot. Also works well as a pattern with `__stories__` etc.

packages/f-header/src/components/tests/__mocks__/axios.js
```diff
@@ -0,0 +1,22 @@
...
+            return Promise.resolve({
+                friendlyName: 'Bob',
+                isAuthenticated: true,
+                email: 'bob.magoo@just-eat.com'
+            });
+        } else if (url === '/analytics/ordercount') {
+            return Promise.resolve({
+                Count: 1,
+                UserId: '6342d07e',
+                Created: '2020-05-07T15:24:19.7875501Z',
```
zlatinivanov-je: Could those be created using `Date` and will it matter on what is displayed e.g. only orders not older than "x" time

zlatinivanov-je (dismissed):
LGTM 

One minor question

packages/f-header/src/components/tests/__mocks__/axios.js
ashleynolan: I tried adding this package, but hit a bit of a brick wall (and as this was a blocker to some other work, didn't want to put more time into changing the implementation).

Will make a separate ticket though if you think it's worth using over the base jest mocking.

xander-marjoram (dismissed):

packages/f-header/src/components/tests/__mocks__/axios.js
ashleynolan: I've updated so these use Date so it's a bit clearer what the dates being used are now.

ashleynolan:
@asherstoppard Would be good to take the suggestion around naming the test directory name somewhere to discuss (as this should be a separate change I think).

Could you post the suggestion in our [team discussion thread](https://github.com/orgs/justeat/teams/ui-developers-guild) and then once we've got agreement over what should change we can put in a ticket for it.

packages/f-header/src/components/tests/Logo.test.js
ashleynolan: Done

packages/f-header/src/components/tests/Logo.test.js
ashleynolan: Done

ashleynolan:
@asherstoppard @xander-marjoram @zlatinivanov-je Have updated now based on feedback

packages/f-form-field/src/components/tests/FormField.test.js
```diff
@@ -1,10 +1,74 @@
...
 describe('FormField', () => {
     it('should be defined', () => {
         const propsData = {};
         const wrapper = shallowMount(FormField, { propsData });
         expect(wrapper.exists()).toBe(true);
     });
+
+    describe('props ::', () => {
+        beforeEach(() => {
+            console.error = jest.fn();
```
kevinrodrigues: Does this one need to be run for all tests? If not might be worth moving it to the `IT` statement that requires it.

packages/f-form-field/src/components/tests/FormField.test.js
```diff
@@ -1,10 +1,74 @@
...
+            });
+
+            it('should set the type of the input element if the prop value passed in is valid', () => {
+                // Arrange
+                const propsData = {
+                    inputType: 'email'
+                };
+
+                // Act
+                const wrapper = shallowMount(FormField, { propsData });
```
kevinrodrigues, line 34: Picky one, but I think these two should come under `// Arrange`?

DamianMullins:
@ashleynolan @asherstoppard There is already [a discussion on the UI board around unit test folder naming](https://github.com/orgs/justeat/teams/ui-developers-guild/discussions/1), the agreed outcome was to use a folder called `_tests` to hold the tests.

packages/f-header/src/components/tests/__mocks__/axios.js
asherstoppard: Again, it's definitely worth a conversation as the overhead of maintaining a static mock far outweighs that of implementing the static mocks for each method etc.

It doesn't change the functionality at all so more than happy to put in another discussion or change it at a later time 🙂

ashleynolan: Yeh for sure – would be good to understand the benefits so we can prioritise it.

Will pick this up with you separately 😄

packages/f-form-field/src/components/tests/FormField.test.js
ashleynolan: This just copies the structure of the tests from all other components (the `mount`/`shallowMount`  happens under `// Act` for all the other tests.

I think it kind of makes sense there, but if we want to update, would probably be best to do it as part of a separate PR to do it everywhere if we think that's better for the structure.

ashleynolan:
@DamianMullins I've added a [ticket to look at this](https://jira.just-eat.net/browse/WCB-834) and added it to the web backlog.

packages/f-form-field/src/components/tests/FormField.test.js
ashleynolan: Have updated now

kevinrodrigues (approved):

asherstoppard (approved):

packages/f-form-field/src/components/tests/FormField.test.js
kevinrodrigues: Yeah not a biggy at all, I think across other verticals it's under the `Arrange` set up which for me makes sense as your basically 'arranging' your test for an assertion. For me `Act` is more about invoking something that directly has an impact on the `expect`.

PR #188, opened 2020-05-26
f-registration@v0.5.0 – Added validation and error messages

ashleynolan:
### Added
- `Vuelidate` package added for validation of form inputs. Validation currently set-up to show to form submission.
- Error messages for error states and styling for error messages (needs to be in the form rather than on the input because of the way CSS Modules works – the plan will be to move this styling to a more generic `f-form` component down the line).
- `name` attributes added to each form field component.
- Added base `typography` file to the demo component so that it shows how the component will look with the global fozzie typographic styles applied (global styles like this will be made available as a separate include that is imported into each component in the future).

### Changed
- Stubbed `.css` file module imports as was causing build error.

N.b. Unit tests and README info coming in future PR as these components are productionised

---

## UI Review Checks

Screen Shot 2020-05-26 at 17 31 57

## Browsers Tested

- [x] Chrome (latest)
- [x] Mobile (iPhone Xs)

### DIFF
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -8,31 +8,82 @@
         <form
             type="post"
             :class="$style['o-form']"
+            @submit="checkValidation"
         >
             <form-field
+                v-model="firstName"
+                name="firstName"
                 data-test-id="input-first-name"
                 label-text="First name"
                 input-type="text"
-                label-style="inline" />
-
+                label-style="inline">
+                <template #error>
+                    <p
+                        v-if="($v.firstName.$invalid && !$v.firstName.required) && $v.firstName.$dirty"
+                        :class="$style['o-form-error']">
+                        <warning-icon :class="$style['o-form-error-icon']" />
+                        Please enter your first name
+                    </p>
+                </template>
+            </form-field>
 
             <form-field
+                v-model="lastName"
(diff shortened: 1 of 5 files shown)


packages/f-registration/src/components/Registration.vue
```diff
@@ -8,31 +8,82 @@
...
+                name="firstName"
                 data-test-id="input-first-name"
                 label-text="First name"
                 input-type="text"
-                label-style="inline" />
-
+                label-style="inline">
+                <template #error>
+                    <p
+                        v-if="($v.firstName.$invalid && !$v.firstName.required) && $v.firstName.$dirty"
```
kevinrodrigues: I'd be tempted to move this logic into a computed property, same for the others below.

DamianMullins: Agree, giving names to these conditions would make it easier to follow what they are there for.

ashleynolan:
@DamianMullins @kevinrodrigues I've updated this to have computed props for those error messages now

packages/f-registration/src/components/Registration.vue
ashleynolan: Done

DamianMullins (approved):

kevinrodrigues (approved):

PR #192, opened 2020-06-02
f-registration@v0.6.0

adamashton:
Added create account API call when clicking create account button

## UI Review Checks

## Browsers Tested
- [X] Chrome (latest)
- [X] Internet Explorer 11

### API Error
image

### DIFF
--- a/packages/f-registration/babel.config.js
+++ b/packages/f-registration/babel.config.js
@@ -3,8 +3,7 @@ module.exports = api => {
     const isTest = api.env('test');
     const presets = [];
     const plugins = [
-        '@babel/plugin-proposal-optional-chaining', // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
-        '@babel/plugin-transform-regenerator',
+        '@babel/plugin-proposal-optional-chaining', // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining        
         '@babel/plugin-transform-runtime'
     ];
 
--- a/packages/f-registration/babel.config.js
+++ b/packages/f-registration/babel.config.js
@@ -3,8 +3,11 @@ module.exports = api => {
     const isTest = api.env('test');
     const presets = [];
     const plugins = [
-        '@babel/plugin-proposal-optional-chaining' // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
+        '@babel/plugin-proposal-optional-chaining', // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
+        '@babel/plugin-transform-regenerator',
+        '@babel/plugin-transform-runtime'
     ];
+
     const builtIns = (api.env('development') ? 'entry' : false);
 
     if (!isTest) {
--- a/packages/f-registration/babel.config.js
+++ b/packages/f-registration/babel.config.js
@@ -3,8 +3,7 @@ module.exports = api => {
     const isTest = api.env('test');
     const presets = [];
     const plugins = [
-        '@babel/plugin-proposal-optional-chaining', // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining        
-        '@babel/plugin-transform-runtime'
+        '@babel/plugin-proposal-optional-chaining' // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
     ];
 
     const builtIns = (api.env('development') ? 'entry' : false);
@@ -12,13 +11,15 @@ module.exports = api => {
     if (!isTest) {
         api.cache(true); // Caches the computed babel config function – https://babeljs.io/docs/en/config-files#apicache
         presets.push(['@vue/app', { useBuiltIns: builtIns }]);
+        // Alias for @babel/preset-env
+        // Hooks into browserslist to provide smart Babel transforms
+        // https://babeljs.io/docs/en/babel-preset-env
+        presets.push('@babel/env');
+    } else {
+        // use current node version for transpiling test files
+        presets.push(['@babel/env', { targets: { node: 'current' } }]);
     }
-
-    // Alias for @babel/preset-env
-    // Hooks into browserslist to provide smart Babel transforms
-    // https://babeljs.io/docs/en/babel-preset-env
-    presets.push('@babel/env');
-
+    
     return {
         presets,
         plugins
--- a/packages/f-registration/package.json
+++ b/packages/f-registration/package.json
@@ -48,7 +48,6 @@
     "@justeat/browserslist-config-fozzie": ">=1.1.1"
   },
   "devDependencies": {
-    "@babel/plugin-transform-runtime": "7.10.1",
     "@vue/cli-plugin-babel": "4.2.3",
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.3.1",
--- a/packages/f-registration/package.json
+++ b/packages/f-registration/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-registration",
   "description": "Fozzie Form Field Component",
-  "version": "0.4.0",
+  "version": "0.5.0",
   "main": "dist/f-registration.umd.min.js",
   "files": [
     "dist"
@@ -38,8 +38,10 @@
   ],
   "dependencies": {
     "@justeat/f-card": "0.2.0",
-    "@justeat/f-form-field": "0.3.0",
-    "@justeat/f-services": "1.0.0"
+    "@justeat/f-form-field": "0.4.0",
+    "@justeat/f-services": "1.0.0",
+    "@justeat/f-vue-icons": "1.0.0-beta.5",
+    "vuelidate": "0.7.5"
   },
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.1.1"
--- a/packages/f-registration/package.json
+++ b/packages/f-registration/package.json
@@ -38,7 +38,7 @@
   ],
   "dependencies": {
     "@justeat/f-card": "0.2.0",
-    "@justeat/f-form-field": "0.4.0",
+    "@justeat/f-form-field": "0.5.0",
     "@justeat/f-services": "1.0.0",
     "@justeat/f-vue-icons": "1.0.0-beta.5",
     "vuelidate": "0.7.5"
--- a/packages/f-registration/package.json
+++ b/packages/f-registration/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-registration",
   "description": "Fozzie Form Field Component",
-  "version": "0.5.0",
+  "version": "0.6.0",
   "main": "dist/f-registration.umd.min.js",
   "files": [
     "dist"
@@ -41,6 +41,7 @@
     "@justeat/f-form-field": "0.5.0",
     "@justeat/f-services": "1.0.0",
     "@justeat/f-vue-icons": "1.0.0-beta.5",
+    "axios": "^0.19.2",
     "vuelidate": "0.7.5"
   },
   "peerDependencies": {
--- a/packages/f-registration/package.json
+++ b/packages/f-registration/package.json
@@ -48,10 +48,13 @@
     "@justeat/browserslist-config-fozzie": ">=1.1.1"
   },
   "devDependencies": {
+    "@babel/plugin-transform-runtime": "^7.10.1",
     "@vue/cli-plugin-babel": "4.2.3",
     "@vue/cli-plugin-eslint": "3.9.2",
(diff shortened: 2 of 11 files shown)


packages/f-registration/src/components/Registration.vue
```diff
@@ -8,8 +8,14 @@
         <form
             type="post"
             :class="$style['o-form']"
-            @submit="checkValidation"
+            @submit.prevent="onFormSubmit"
         >
+            <p
+                v-if="shouldShowGenericError"
```
kevinrodrigues: I think you could just pass in `genericErrorMessage` instead as the computed prop isn't doing much but returning that value.

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -1,25 +1,101 @@
...
+            document.body.appendChild(div);
+            return mount(Registration, { propsData, sync: false, attachTo: div });
+        }
+
+        afterEach(() => {
+            axiosMock.reset();
+            axiosMock.resetHistory();
+        });
+
+        it('should post correct data and emit success event when API responds with 201 Created', async () => {
```
kevinrodrigues: There's quite a lot going on in here. For clarity would it not be better to use a spy and mock the data. Seems that this test and the next one is mostly testing `RegistrationServiceApi.createAccount`. Ideally this service should have it's own tests for these type of calls and in here you would mostly test that this method was called with some params and you can also add tests for some of the local data values that are changed.

adamashton: Yeh, done.

kevinrodrigues (dismissed):

packages/f-registration/babel.config.js
```diff
@@ -3,8 +3,11 @@ module.exports = api => {
     const isTest = api.env('test');
     const presets = [];
     const plugins = [
-        '@babel/plugin-proposal-optional-chaining' // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
+        '@babel/plugin-proposal-optional-chaining', // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
+        '@babel/plugin-transform-regenerator',
```
DamianMullins: Do you need to add this plugin?

packages/f-registration/src/event-names/index.js
```diff
@@ -0,0 +1,7 @@
+const CreateAccountSuccess = 'registration-create-account-success';
```
DamianMullins: You could move this file to the root of `/src` and rename it to `event-names`?

packages/f-registration/babel.config.js
adamashton: They're needed to allow async methods in unit testing.

DamianMullins: I know `plugin-transform-runtime` is but I'm not sure `plugin-transform-regenerator` is required? Also, it's not in the dev dependencies below?

adamashton: You're right. I did not need `plugin-transform-regenerator`. Where is the dev devependenvies for babel plugins?

DamianMullins: Looks good. I mean the `devDependencies` in the package.json file. You've already added `plugin-transform-runtime` so you're all good.

packages/f-registration/src/components/Registration.vue
```diff
@@ -189,13 +208,36 @@ export default {
...
+            this.shouldDisableCreateAccountButton = true;
+            try {
+                const registrationData = {
+                    firstName: this.firstName,
+                    lastName: this.lastName,
+                    email: this.email,
+                    password: this.password
+                };
+                await RegistrationServiceApi.createAccount(this.createAccountUrl, this.tenant, registrationData)
+                    .then(() => {
```
DamianMullins: If you're using await you shouldn't need the promise chain here? Everything in `.then()` can just go on the line below the awaited API call.

You should be able to move your error handling into a `catch` block before `finally` below.

packages/f-registration/src/services/RegistrationServiceApi.js
```diff
@@ -0,0 +1,15 @@
...
+    async createAccount (url, tenant, data) {
+        const config = {
+            method: 'post',
+            headers: {
+                'Content-Type': 'application/json',
+                'Accept-Tenant': tenant
+            }
+        };
+        return axios
+            .post(url, data, config);
```
adammorr, line 13: The tenant should be part of the URL as opposed to setting the `Accept-Tenant` header

adammorr (changes requested):

packages/f-registration/src/services/RegistrationServiceApi.js
adamashton: Not if we're calling AccountWeb, which we are at the moment.

packages/f-registration/src/components/Registration.vue
adamashton: I see, nice.

packages/f-registration/src/services/RegistrationServiceApi.js
adammorr: OK so we're releasing a version of the package with AccountWeb and then a follow up for AccountApi?

adamashton: Yup

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -1,25 +1,92 @@
...
             const wrapper = shallowMount(Registration, { propsData });
             expect(wrapper.exists()).toBe(true);
         });
     });
+
+    describe('when creating an account', () => {        
+        function mountComponentAndAttachToDocument () {
+            const div = document.createElement('div');
+            document.body.appendChild(div);
+            return mount(Registration, { propsData, sync: false, attachTo: div });
```
DamianMullins: What does `sync` do here? I can't see it in the vue-test-utils docs?

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -1,25 +1,92 @@
...
+        }
+
+        it('should post correct data and emit success event when service succeeds', async () => {
+            // Arrange
+            RegistrationServiceApi.createAccount.mockImplementation(async () => Promise.resolve());
+            const wrapper = mountComponentAndAttachToDocument();
```
DamianMullins, line 43: Why do you need to attach the component to the document?

packages/f-registration/src/components/tests/Registration.test.js
adamashton: Not needed actually

packages/f-registration/src/components/tests/Registration.test.js
adamashton: attachToDocument is deprecated, so I need to use attachTo to ensure that the trigger('click') actually triggers the click event for Vue

DamianMullins: So the call to `trigger` doesn't submit the form if you don't do this? Interested as I've not seen this syntax or hit this issue before.

adamashton: I was having that problem yes, and I followed this. https://stackoverflow.com/questions/53382235/trigger-form-submit-on-button-click-in-vue-unit-test

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -1,25 +1,92 @@
...
+            const div = document.createElement('div');
+            document.body.appendChild(div);
+            return mount(Registration, { propsData, attachTo: div });
+        }
+
+        it('should post correct data and emit success event when service succeeds', async () => {
+            // Arrange
+            RegistrationServiceApi.createAccount.mockImplementation(async () => Promise.resolve());
+            const wrapper = mountComponentAndAttachToDocument();
+            try {
```
DamianMullins, line 44: Do you need the try/catch blocks?

DamianMullins (dismissed):

packages/f-registration/babel.config.js
```diff
@@ -3,8 +3,10 @@ module.exports = api => {
     const isTest = api.env('test');
     const presets = [];
     const plugins = [
-        '@babel/plugin-proposal-optional-chaining' // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
+        '@babel/plugin-proposal-optional-chaining', // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining        
+        '@babel/plugin-transform-runtime'
```
ashleynolan: More for info than anything else – what did you need to add this for?

packages/f-registration/package.json
```diff
@@ -41,16 +41,20 @@
...
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.1.1"
   },
   "devDependencies": {
+    "@babel/plugin-transform-runtime": "^7.10.1",
     "@vue/cli-plugin-babel": "4.2.3",
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.3.1",
     "@vue/test-utils": "1.0.2",
+    "axios-mock-adapter": "^1.18.1",
```
ashleynolan: Can you lock these dependencies (so get rid of the `^` and then rerun yarn to update the lockfile

packages/f-registration/babel.config.js
adamashton: It's something to do with transpiling from ES6 to ES5 when you have async methods in your unit tests. I was really just googling the error message that came up.

packages/f-registration/src/components/Registration.vue
```diff
@@ -124,13 +133,21 @@ export default {
             type: String,
             default: ''
         },
+        tenant: {
```
ashleynolan: Would be good to handle this through locale, as we should be able to map locales to tenants.

We do this in [`f-searchbox` for instance](https://github.com/justeat/f-searchbox/blob/6367272646a21824449bb55df9e371497d40a3db/src/tenants/index.js#L16-L30) – just means one less prop to define as it implicit config

ashleynolan (changes requested):

packages/f-registration/package.json
adamashton: Sure, done.

packages/f-registration/babel.config.js
ashleynolan: Ah ok – we've hit that issue in a couple of other packages.

Best way to solve it is to change the babel config to [match this](https://github.com/justeat/fozzie-components/blob/master/packages/f-header/babel.config.js#L10-L17) – the main thing that does is run the tests using the latest version of Node so that you don't need to transpile async/await for the tests. If you still see the issues after adding that give me a shout as sometimes you then also need to manually clear the jest cache so that the new babel config comes through correctly.

packages/f-registration/src/components/Registration.vue
adamashton: Probably getting outside of scope for the ticket I'm working on. Perhaps for another ticket.

packages/f-registration/babel.config.js
adamashton: Thanks, fixed it with the updated babel config

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -1,25 +1,92 @@
...
+                const firstName = 'Adam',
+                    lastName = 'Ashton',
+                    email = 'adam.ashton+jetest@just-eat.com',
```
xander-marjoram: I would avoid checking personal info in (even if it's quite easy to guess/work out).

packages/f-registration/src/components/Registration.vue
```diff
@@ -124,13 +133,21 @@ export default {
             type: String,
             default: ''
         },
+        tenant: { // TODO Compute tenant value from local config instead of using this prop
```
xander-marjoram: We normally recommend adding a JIRA ID so that stuff like this doesn't get forgotten

packages/f-registration/src/components/tests/Registration.test.js
adamashton: Sure

xander-marjoram (approved):

ashleynolan (approved):

adammorr (approved):

packages/f-registration/babel.config.js
DamianMullins: Was it not needed to transpile the async code in the main codebase as well? Or does `present-env` handle that?

ashleynolan: `preset-env` should handle the transform (which is why there was a regenerator runtime error shown in Jest, as that's what it converts it to)

PR #195, opened 2020-06-04
f-metadata@2.6.0: Check for an existing appboy sdk

asherstoppard:
Check for an existing version of the appboy sdk on window and skip the module import and subsequent setup.

## UI Review Checks

- *N/A* This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist)


minrwhite (dismissed):

packages/f-metadata/CHANGELOG.md
```diff
@@ -3,12 +3,15 @@
...
+2.6.0
 ------------------------------
-*June 4, 2020*
+*June 5, 2020*
+
+### Added
+- Check for previously initialised Braze instances on window and skip import of the SDK
 
 ### Changed
+- Split codebase and add bundling through parcel
```
ashleynolan: Can you add some info to the README why the package is now available as both a bundle through rollup and without bundling (as a regular import)?

The reason we didn't bundle it before was so that we could codesplit the SDK bundle from Braze, so would be good to add that reason alongside what the new bundle is being used for.

ashleynolan: Also, should this say `rollup` rather than `parcel`?

ashleynolan (commented):
Looks good – would be good to know what the rollup stuff is being added to solve though (and get that added to the README so anyone new to the code can see how it's bundled)

packages/f-metadata/src/_tests/index.test.js
```diff
@@ -0,0 +1,100 @@
...
+    beforeEach(() => {
+        jest.resetAllMocks();
+    });
+
+    it('should resolve as null if disableComponent is provided', async () => {
+        // Arrange & Act
+        const instance = await initialiseBraze({ ...settings, disableComponent: true });
+
+        // Assert
+        expect(appboySDK.initialize).not.toHaveBeenCalled();
```
kevinrodrigues: Should this expect be here? Based on the description I think it's already covered in the next few tests?

asherstoppard: @kevinrodrigues Apologies on the delay getting back to you! Just to clarify: do you think this is covered in the other scenarios or in `expect(instance).toBe(null)`?

asherstoppard:
@ashleynolan Rollup was originally added to this lib to try and assist with the bundling of the dynamic import chunk. I think it's actually caused more issues than it's solved though so I've removed it as part of this PR with a view to look at patterns for chunks at a later point :+1:

minrwhite (dismissed):

packages/f-metadata/src/_tests/index.test.js
kevinrodrigues: Nps. So the description seems to suggest that we're only checking for the null output. Just wondering if you'd want to update this description to check for the `initialize` assert too? Although the next test does cover this so might not need that particular assert here.

kevinrodrigues (dismissed):
LGTM - minor comment added but not a biggy.

packages/f-metadata/src/_tests/index.test.js
asherstoppard: Ah, ok cool, gotcha 🙂 I'll remove this :+1:

kevinrodrigues (approved):

minrwhite (approved):

PR #197, opened 2020-06-05
f-content-cards@0.1.0

asherstoppard:
This PR introduces the baseline setup for the `f-content-cards` component.

## Added

- Initial `f-content-cards` component documentation and configuration
- Placeholder `ContentCards` component, unit test and storybook entry

## UI Review Checks

- *N/A* This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist)


packages/f-braze-content-cards/babel.config.js
```diff
@@ -0,0 +1,24 @@
+module.exports = api => {
+    // Use `isTest` to determine what presets and plugins to use with jest
+    const isTest = api.env('test');
+    const presets = [];
+    const plugins = [
+        '@babel/plugin-proposal-optional-chaining' // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
+    ];
+    const builtIns = (api.env('development') ? 'entry' : false);
+
+    if (!isTest) {
```
ashleynolan: I'd recommend updating this to match [the config here](https://github.com/justeat/fozzie-components/blob/master/packages/f-header/babel.config.js#L10-L17).

Just means if you add any `async` statements later then jest won't have a panic attack.

packages/f-braze-content-cards/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "node": ">=10.0.0"
+  },
+  "keywords": [
+    "fozzie"
+  ],
+  "scripts": {
+    "prepublishOnly": "yarn lint && yarn test && yarn build",
+    "build": "vue-cli-service build --target lib --name f-braze-content-cards ./src/index.js",
+    "demo": "vue serve --open ./src/demo/Index.vue",
+    "lint": "vue-cli-service lint",
```
ashleynolan: Small one that I need to update in the generator (as changed it yesterday). Can you add `--no-fix` to the lint script? Otherwise it autofixes when running the lint task in the CI.

packages/f-braze-content-cards/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@justeat/f-braze-content-cards",
```
ashleynolan: Should we just call this `f-content-cards`?

I know it's probably unlikely that this component wouldn't use braze in the future, but I've seen too many legacy names crop up like this to not mention it! 😉

packages/f-content-cards/stories/ContentCards.js
```diff
@@ -0,0 +1,8 @@
+import { storiesOf } from '@storybook/vue';
+import ContentCards from '../src/components/ContentCards.vue';
+
+storiesOf('Components', module)
```
ashleynolan: I'm currently updating this in the generator – this is the old storybook syntax which they updated in the new version.

Can you change it to [be in this format](https://github.com/justeat/fozzie-components/blob/master/packages/f-header/stories/header.stories.js)? (I'll update the generator today as been meaning to do this since I updated the other packages).

packages/f-content-cards/src/index.js
```diff
@@ -0,0 +1,33 @@
+import globalThis from 'core-js/features/global-this';
```
ashleynolan, line 1: More out of interest than anything else – what does this package do? (mainly should we look to add this to other packages later)

asherstoppard: It's just a polyfill for `globalThis` support in IE :+1:

Changing to `globalThis` means we can avoid typeof undefined checks and move to optional chaining:

```js
// Previously
let result = null
if (typeof window === 'undefined') result = window.someKey

// Now
const result = globalThis?.window?.someKey
```

ashleynolan (dismissed):

minrwhite (approved):

kevinrodrigues (approved):

ashleynolan (approved):

PR #200, opened 2020-06-09
f-content-cards@0.2.0: Add Promotion Card 1 and 2

asherstoppard:
**Apologies for the size of this PR but a lot of the components have just been ported from HomeWeb.**

Port existing functionality for `Promotion_Card_1` and `Promotion_Card_2` from HomeWeb.

## UI Review Checks


minrwhite (dismissed):

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -1,26 +1,385 @@
 <template>
     <div
-        :class="$style['c-content-cards']"
-        data-test-id="ContentCards">
-        Hello World
+        class="c-restaurantCards c-restaurantCards--wrap"
```
ashleynolan: Not sure if this class naming is a little confusing – as the component is called content card and the classnames being called `c-restaurantCards`. I think `c-ContentCards` could make more sense so that the classnames match up?

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -1,26 +1,385 @@
...
-    display: flex;
-    justify-content: center;
-    min-height: 80vh;
-    width: 80vw;
-    margin: auto;
-    border: 1px solid $red;
-    font-family: $font-family-base;
-    @include font-size(large);
-}
+<style lang="scss">
```
ashleynolan, line 81: Would be good to keep this as CSS modules so that the styles are better encapsulated (unless there's a reason for removing it?)

packages/f-content-cards/src/components/cardTemplates/CardContainer.vue
```diff
@@ -0,0 +1,101 @@
...
+                        .filter(key => key.indexOf('line_') !== -1)
+                        .map(key => this.extras[key]);
+        },
+
+        isAnniversaryCard () {
+            return this.type === 'Anniversary_Card_1';
+        }
+    }
+};
+</script>
```
ashleynolan, line 102: Would be good to pull in the relevant styles into this component file rather than have them as part of the outer component.

packages/f-content-cards/src/services/contentCard.service.js
```diff
@@ -0,0 +1,88 @@
...
+    'Anniversary_Card_1',
+    'Header_Card',
+    'Promotion_Card_1',
+    'Promotion_Card_2',
+    'Restaurant_FTC_Offer_Card',
+    'Terms_And_Conditions_Card',
+    'Voucher_Card_1'
+];
+
+class ContentCards {
```
ashleynolan, line 23: Would be good to have some comments for these methods to give an overview of what they do (JSDoc style)

packages/f-content-cards/stories/ContentCards.stories.js
```diff
@@ -0,0 +1,24 @@
...
+export default {
+    title: 'Components/Organisms',
+    decorators: [withKnobs]
+};
+
+export const ContentCardscomponent = () => ({
+    components: { ContentCards },
+    props: {
+        apiKey: {
+            default: text('API Key', 'f714b0fc-6de5-4460-908e-2d9930f31339')
```
ashleynolan: My normal question around API Keys - this is fine to be public potentially?

packages/f-content-cards/src/components/ContentCards.vue
asherstoppard: Completely agree, this is just a first pass to get in a working example, we'll be looking to introduce encapsulated styling on a component by component basis as part of `0.5.0` :+1:

packages/f-content-cards/src/components/cardTemplates/CardContainer.vue
asherstoppard: Completely agree, this is just a first pass to get in a working example, we'll be looking to introduce encapsulated styling on a component by component basis as part of 0.5.0 👍

packages/f-content-cards/stories/ContentCards.stories.js
asherstoppard: These aren't unsafe as they're output in plain text as part of localstorage in production, however, they shouldn't be applied explicitly here. I'll get them removed :+1:

minrwhite (dismissed):

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -1,26 +1,385 @@
...
+    .c-contentCards--centred {
+        justify-content: center;
+    }
+    .c-contentCards--wrap {
+        flex-wrap: wrap;
+
+        &.has-offers-carousel {
+            margin-top: 0;
+
+            .c-contentCard {
```
ashleynolan: Looks like this selector is defined twice – I'd move these styles down to line 130 where the others are.

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -1,26 +1,385 @@
...
+
+        .c-contentCards--wrap & {
+            display: flex;
+            flex-direction: column;
+            flex: 0 0 40%;
+            margin: 0 spacing() spacing(x3) 0;
+            width: 100%;
+        }
+
+        @include media('>=narrowMid') {
```
ashleynolan: Tend to group media queries as close to the base styles as possible – so I'd move these up to line 123 and then the `.c-contentCards--wrap & {` ones I'd put inside that selector as a separate media query.

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -1,26 +1,385 @@
...
+
+            .c-contentCards--wrap & {
+                margin: 0 spacing() spacing(x3);
+            }
+        }
+
+        .c-contentCard-banner {
+            position: relative;
+            display: inline-block;
+            background: #cd381f;
```
ashleynolan: We shouldn't be using any hex colours directly – is this colour not in the colour palette (and if it's not, it should either be added or changed to match one of the colours in there)

packages/f-content-cards/src/components/cardTemplates/CardContainer.vue
```diff
@@ -0,0 +1,101 @@
...
+        } = this.card;
+        const {
+            button_1: button,
+            icon_1: icon,
+            image_1: image,
+            order,
+            custom_card_type: type,
+            voucher_code: voucherCode
+        } = extras;
+        return {
```
ashleynolan, line 76: Super NIT – could you add a line-break between the return and the destructuring above. Just to make it a bit more scannnable when looking to see what the return data property is (same in the other files where this is done).

packages/f-content-cards/src/components/cardTemplates/utils/_tests/index.test.js
```diff
@@ -0,0 +1,38 @@
...
+                input: '1_2_3',
+                output: '123'
+            }
+        ];
+
+        // Act & Assert
+        await Promise.all(tests.map(({ input, output }) => expect(normaliseCardType(input)).toBe(output)));
+    });
+
+    it('Should return the input if type is not equal to string', async () => {
```
ashleynolan: More of a Q than a comment – should this also return some sort of error or are you happy it just outputting what comes in in this instance?

packages/f-content-cards/src/services/_tests/contentCard.service.test.js
```diff
@@ -0,0 +1,274 @@
...
+                title: 'Kepler‑186f',
+                extras: { order: '1' }
+            }];
+            const service = new ContentCardService({ cards });
+
+            // Act
+            const { cards: result } = service.orderCardsByOrderValue().output();
+
+            // Assert
+            expect(service.orderCardsByOrderValue().output().cards).toEqual(result);
```
ashleynolan: Isn't this just testing `service.orderCardsByOrderValue().output().cards = service.orderCardsByOrderValue().output().cards` essentially? Should it be testing against a hardcoded object result?

packages/f-content-cards/src/services/_tests/contentCard.service.test.js
```diff
@@ -0,0 +1,274 @@
...
+                title: 'Wasp-19b',
+                extras: { updated: '2020-02-17T18:23:58.000Z', customCardType: 'promotion_card_3' }
+            }];
+            const service = new ContentCardService({ cards });
+
+            // Act
+            const { cards: result } = service.orderCardsByUpdateValue().output();
+
+            // Assert
+            expect(service.orderCardsByUpdateValue().output().cards).toEqual(result);
```
ashleynolan: Same here as above (not sure if I'm missing something with the restructuring above and how this is checking the result)

packages/f-content-cards/src/services/_tests/contentCard.service.test.js
```diff
@@ -0,0 +1,274 @@
...
+                title: 'Jupiter Hot',
+                extras: { updated: '2020-02-17T18:23:58.000Z', custom_card_type: 'promotion_card_3' } // eslint-disable-line camelcase
+            }];
+            const service = new ContentCardService({ cards });
+
+            // Act
+            const { cards: result } = service.removeDuplicateContentCards().output();
+
+            // Assert
+            expect(service.removeDuplicateContentCards().output().cards).toEqual(result);
```
ashleynolan: Same here?

packages/f-content-cards/src/services/_tests/contentCard.service.test.js
```diff
@@ -0,0 +1,274 @@
...
+                    title: 'Wasp-19b',
+                    extras: { updated: '2020-02-17T18:23:58.000Z', customCardType: 'promotion_card_3' }
+                }];
+                const service = new ContentCardService({ cards });
+
+                // Act
+                const { cards: result } = service.filterCards().output();
+
+                // Assert
+                expect(service.filterCards().output().cards).toEqual(result);
```
ashleynolan: Same here as above

packages/f-content-cards/src/services/contentCard.service.js
ashleynolan: 👍

ashleynolan (commented):
Looks good – a few styling and test things I noticed, but otherwise the structure looks good

packages/f-content-cards/src/components/cardTemplates/utils/_tests/index.test.js
asherstoppard: I've always run formatting functions this way just so it doesn't snag but I'm happy to change it over to return null/log an error/throw if the type isn't correct?

packages/f-content-cards/src/services/_tests/contentCard.service.test.js
asherstoppard: Yeah you're absolutely right, this was just laziness on my part. I've made the tests more robust :+1:

minrwhite (approved):

ashleynolan (approved):

PR #203, opened 2020-06-10
f-card@0.3.0: Add card heading position prop

luke-andrew-parry:
Add new prop to control the alignment of the card heading. 
The prop can have a value of `left`, `center` or `right`.

### Left 
left

### Center
center

### Right
right

## Browsers Tested

- [X] Chrome (latest)
- [ ] Internet Explorer 11


packages/f-card/src/components/Card.vue
```diff
@@ -13,6 +13,7 @@
                 'beta',
                 $style['c-card-heading']
             ]"
+            :style="{ textAlign: cardHeadingPosition }"
```
ashleynolan: Could you do this by adding a class to the heading rather than applying the style directly?

(so `c-card-heading--centred` and `c-card-heading--right` – and then adding those classes to the styles below?

ashleynolan: Just for reference – the classes can be applied like above with `c-card--rounded` etc

luke-andrew-parry: Sure. I can align it with `c-card--rounded`.

packages/f-card/src/components/Card.vue
```diff
@@ -119,6 +127,15 @@ $card--pageContentWrapper-width           : 460px;
         margin-bottom: spacing(x2);
     }
 
+    .c-card--left {
```
ashleynolan: Not sure you need this one, as left is the default for `text-align` anyway. So if it's left, you can just do nothing and it will be aligned left.

packages/f-card/src/components/Card.vue
```diff
@@ -11,7 +11,8 @@
             v-if="cardHeading"
             :class="[
                 'beta',
-                $style['c-card-heading']
+                $style['c-card-heading'],
+                $style[`c-card--${cardHeadingPosition}`]
```
ashleynolan: Could make this `(cardHeadingPosition !== 'left' ? $style[`c-card--${cardHeadingPosition}`] : '').

(I assume that would be valid syntax, but hard to tell without trying it).

packages/f-card/src/components/tests/Card.test.js
```diff
@@ -119,5 +119,47 @@ describe('Card', () => {
...
+            });
+
+            it('should only allows `left`, `right` or `center` to be passed in.', () => {
+                // Arrange
+                const propsData = {};
+
+                // Act
+                const wrapper = shallowMount(Card, { propsData });
+
+                // Assert
```
kevinrodrigues: Minor but this comment should move to the expect block below.

packages/f-card/src/components/tests/Card.test.js
```diff
@@ -119,5 +119,47 @@ describe('Card', () => {
...
+                };
+
+                // Act
+                const wrapper = shallowMount(Card, { propsData });
+
+                // Assert
+                expect(wrapper.vm.cardHeadingPosition).toBe('center');
+            });
+
+            it('should only allows `left`, `right` or `center` to be passed in.', () => {
```
kevinrodrigues: `should only allow...`

packages/f-card/src/components/tests/Card.test.js
luke-andrew-parry: Corrected & pushed up.

packages/f-card/src/components/tests/Card.test.js
luke-andrew-parry: Corrected & pushed up.

kevinrodrigues (approved):

ashleynolan (approved):

PR #205, opened 2020-06-11
f-registration@v0.7.0: Add login link to registration form.

luke-andrew-parry:
Added a link to the top of the registration form to link off to the login page.

image

### DIFF
--- a/packages/f-registration/CHANGELOG.md
+++ b/packages/f-registration/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.7.0
 ------------------------------
-*June 1, 2020*
+*June 11, 2020*
 
 ### Changed
 - Storybook category updated to `Organisms`.
--- a/packages/f-registration/CHANGELOG.md
+++ b/packages/f-registration/CHANGELOG.md
@@ -10,10 +10,6 @@ v0.7.0
 ### Changed
 - Storybook category updated to `Organisms`.
 
-Latest (rolled into next release)
-------------------------------
-*June 9, 2020*
-
 ### Added
 - Login link to registration component.
 
--- a/packages/f-registration/CHANGELOG.md
+++ b/packages/f-registration/CHANGELOG.md
@@ -3,12 +3,19 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.7.0
+------------------------------
+*June 1, 2020*
+
+### Changed
+- Storybook category updated to `Organisms`.
+
 Latest (rolled into next release)
 ------------------------------
 *June 9, 2020*
 
-### Changed
-- Storybook category updated to `Organisms`.
+### Added
+- Login link to registration component.
 
 
 v0.6.0
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -4,7 +4,14 @@
         :card-heading="title"
         is-rounded
         has-outline
-        is-page-content-wrapper>
+        is-page-content-wrapper
+        card-heading-position="center">
+        <p
+            v-if="loginSettings"
+            :class="$style['login-link']"
+            data-test-id="create-account-login-link">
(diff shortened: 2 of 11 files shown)


packages/f-registration/CHANGELOG.md
```diff
@@ -3,13 +3,20 @@
...
-Latest (rolled into next release)
+v0.7.0
 ------------------------------
-*June 9, 2020*
+*June 11, 2020*
 
 ### Changed
 - Storybook category updated to `Organisms`.
 
+Latest (rolled into next release)
```
ashleynolan: You can delete this header and just add the `#Added` section under the `0.7.0` release

packages/f-registration/src/components/Registration.vue
```diff
@@ -4,7 +4,14 @@
...
         is-rounded
         has-outline
-        is-page-content-wrapper>
+        is-page-content-wrapper
+        card-heading-position="center">
+        <p
+            v-if="loginSettings"
+            :class="$style['login-link']"
+            data-test-id="create-account-login-link">
+            {{ loginSettings.preLinkText }}<a :href="loginSettings.url">{{ loginSettings.linkText }}</a>
```
ashleynolan, line 13: If `loginSettings` is defined but without a `loginSettings.url` does this link URL just return empty? Might be worth this checking that `loginSettings.url` and `loginSettings.linkText` exist (could do this in a computed property and then put a `v-if` on the `a` tag)

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -28,6 +33,14 @@ describe('Registration', () => {
             const wrapper = shallowMount(Registration, { propsData });
             expect(wrapper.exists()).toBe(true);
         });
+
+        it('should show login link if loginSettings prop set.', () => {
```
ashleynolan: Would be good to have some tests around what happens if these are not set (if a `URL` or `linkText` value are not passed in for instance.

packages/f-registration/src/components/Registration.vue
```diff
@@ -248,21 +262,29 @@ export default {
...
+}
 
 * + .o-form {
     margin-top: spacing(x2);
 }
 
+.login-link {
+    text-align: center;
+    a {
+        color: #266abd;
```
ashleynolan: Should use `$blue` here instead of the direct hex code.

packages/f-registration/src/components/Registration.vue
```diff
@@ -4,7 +4,14 @@
         :card-heading="title"
         is-rounded
         has-outline
-        is-page-content-wrapper>
+        is-page-content-wrapper
+        card-heading-position="center">
+        <p
+            v-if="shouldShowLoginLink"
+            :class="$style['login-link']"
```
ashleynolan: Sorry didn't spot this before – should name this to fit with our naming scheme. So would be `c-loginLink`

ashleynolan: You'll need to update this to `c-loginLink`

luke-andrew-parry: 🤦‍♂️

luke-andrew-parry: Done and pushed

ashleynolan (dismissed):

packages/f-registration/src/components/Registration.vue
```diff
@@ -148,6 +155,10 @@ export default {
         createAccountUrl: {
             type: String,
             required: true
+        },
+        loginSettings: {
+            type: Object,
+            default: undefined
```
kevinrodrigues: Would it not be better to set a default value here instead of undefined? e.g `() => ({})`?

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -28,6 +33,47 @@ describe('Registration', () => {
             const wrapper = shallowMount(Registration, { propsData });
             expect(wrapper.exists()).toBe(true);
         });
+
+        it('should show the login link if loginSettings prop set.', () => {
```
kevinrodrigues, line 32: I think you might need to cover one more test. `When 'loginSettings' prop does not exist` based on the computed prop you've just added.

kevinrodrigues (approved):

ashleynolan (approved):

PR #213, opened 2020-06-25
f-content-cards@0.7.0: Temporarily remove `vue-lazyload` lib

asherstoppard:
The `vue-lazyload` library is currently causing issues whereby images are intermittently failing. Investigation and mitigation is ongoing but this PR is to remove the lib until we've found a fix or other solution.

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [x] Internet Explorer 11

- [x] Firefox (latest)
- [x] Opera (latest)

### DIFF
--- a/packages/f-content-cards/CHANGELOG.md
+++ b/packages/f-content-cards/CHANGELOG.md
@@ -4,7 +4,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-v0.6.0
+v0.7.0
 ------------------------------
 *June 25th, 2020*
 
--- a/packages/f-content-cards/CHANGELOG.md
+++ b/packages/f-content-cards/CHANGELOG.md
@@ -8,6 +8,13 @@ v0.6.0
 ------------------------------
 *June 24th, 2020*
 
+### Removed
+- `vue-lazyload` while intermittent issues loading images are investigated and rectified.
+
+v0.6.0
+------------------------------
+*June 24th, 2020*
+
 ### Added
 - `@justeat/f-metadata` as a dependency
 
--- a/packages/f-content-cards/CHANGELOG.md
+++ b/packages/f-content-cards/CHANGELOG.md
@@ -6,7 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.6.0
 ------------------------------
-*June 24th, 2020*
+*June 25th, 2020*
+
+### Changed
+- `@justeat/f-metadata` from a dependency to a devDependency
 
 ### Removed
 - `vue-lazyload` while intermittent issues loading images are investigated and rectified.
--- a/packages/f-content-cards/package.json
+++ b/packages/f-content-cards/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-content-cards",
   "description": "Fozzie Content Cards",
-  "version": "0.6.0",
+  "version": "0.7.0",
   "main": "dist/f-content-cards.umd.min.js",
   "files": [
     "dist"
@@ -39,7 +39,6 @@
     "@justeat/f-services": "1.0.0",
     "@justeat/f-metadata": "2.7.0",
     "core-js": "3.6.5",
-    "vue-lazyload": "1.3.3",
     "lodash.findindex": "4.6.0",
     "lodash.orderby": "4.6.0"
   },
--- a/packages/f-content-cards/package.json
+++ b/packages/f-content-cards/package.json
@@ -37,7 +37,6 @@
   ],
   "dependencies": {
     "@justeat/f-services": "1.0.0",
-    "@justeat/f-metadata": "2.7.0",
     "core-js": "3.6.5",
     "lodash.findindex": "4.6.0",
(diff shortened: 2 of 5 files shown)


minrwhite (dismissed):

packages/f-content-cards/CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.6.0
```
minrwhite:
```suggestion
v0.7.0
```

minrwhite (approved):

packages/f-content-cards/CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.7.0
+------------------------------
+*June 25th, 2020*
+
+### Changed
+- `@justeat/f-metadata` from a dependency to a devDependency
```
xander-marjoram, line 12: Can I ask why, please?

packages/f-content-cards/package.json
```diff
@@ -37,16 +37,15 @@
   ],
   "dependencies": {
     "@justeat/f-services": "1.0.0",
-    "@justeat/f-metadata": "2.7.0",
     "core-js": "3.6.5",
```
DamianMullins, line 40: Not one for this PR but should we be pulling in `core-js` just to provide a polyfill for `globalThis`? Could we not write a small util within the project itself allowing us to remove a large dependency?

DamianMullins (approved):

packages/f-content-cards/CHANGELOG.md
asherstoppard: f-metadata is bundled in as part of the f-content-cards build. If it's left in dependencies it will also be pulled into the consuming application too.

DamianMullins: That's a good question, looks like it's being used in `ContentCards.vue`?

DamianMullins: If you are expecting the consumer of the package to provide `f-metadata` then I think it should be listed as a `peerDependency`?

packages/f-content-cards/package.json
asherstoppard: @DamianMullins We're pulling in the whole library but we're importing the feature directly `import globalThis from 'core-js/features/global-this'`. Admittedly this is a little wasteful in terms of bloating the dev environment but the actual bundled code will still be lean. I'm happy to raise a ticket to look into this as I don't believe core-js runs as a monorepo in a way that allows for features to be installed in isolation?

packages/f-content-cards/CHANGELOG.md
minrwhite: I requested it be moved:

> If you are expecting the consumer of the package to provide `f-metadata` ...

We're not, because the build task bundles it into the library file produced into the dist directory.

packages/f-content-cards/package.json
minrwhite: We possibly can move this to `devDependencies` also as per `f-metadata`, if we grant that the library code that pollyfills `globalThis` will be compiled into the js library build artifact.

DamianMullins: As long as the final build only includes the polyfill you imported I think that's OK.

packages/f-content-cards/CHANGELOG.md
DamianMullins: OK, I think I need to get my head around this a bit better! Could we not move most of the dependencies into `devDependencies` in that case?

asherstoppard: This shouldn't be a requirement from the consuming application as it's bundled into both `f-content-cards.common.js` and `f-content-cards.umd.js` as part of build :+1:

xander-marjoram: @DamianMullins I was just about to ask the same question. Without wanting to derail the PR, I'm not sure what the distinction is if we have a build step

minrwhite: If I understand the distinction correctly, this `package.json` should describe in `dependencies` (and `peerDependencies`) all packages which are required for use of the resulting build artifact by consuming applications. `devDependencies` should describe all packages which are required for testing and building the distributable in the first place. I'll see if I can find some documentation to support this.

minrwhite: The first paragraph [here](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file) makes the distinction between packages required for development and those required for production. I'm pretty certain that building the distributable comes under the 'development' umbrella.

minrwhite: So to come back to @DamianMullins point - we probably could move out most of the dependencies to be devDependencies. Agree that doing this as part of this PR would derail it a bit, we've got plenty more work to do on this component, so we can do that in a follow up pr if you're happy with that approach?

CC @xander-marjoram

DamianMullins: Thanks @benwhite-deltas

PR #214, opened 2020-06-25
f-metadata@2.8.0

minrwhite:
### Added

- Methods for logging card click and view events, with the view to presenting a consistent
  interface for sending information back to braze:
  - `logCardClick()`
  - `logCardImpressions()`

### DIFF
--- a/packages/f-metadata/CHANGELOG.md
+++ b/packages/f-metadata/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v2.8.0
 ------------------------------
-*June 24, 2020*
+*June 26, 2020*
 
 ### Added
 - Methods for logging card click and view events, with the view to presenting a consistent
--- a/packages/f-metadata/CHANGELOG.md
+++ b/packages/f-metadata/CHANGELOG.md
@@ -3,6 +3,17 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v2.8.0
+------------------------------
+*June 24, 2020*
+
+### Added
+- Methods for logging card click and view events, with the view to presenting a consistent
+  interface for sending information back to braze:
+  - `logCardClick()`
+  - `logCardImpressions()`
+
+
 v2.7.0
 ------------------------------
 *June 24, 2020*
--- a/packages/f-metadata/src/_tests/index.test.js
+++ b/packages/f-metadata/src/_tests/index.test.js
@@ -1,5 +1,4 @@
 import appboySDK from 'appboy-web-sdk';
-import initialiseBraze, { sessionTimeoutInSeconds } from '../index';
 import configureBraze from '../configureBraze';
 import isAppboyInitialised from '../utils/isAppboyInitialised';
 
@@ -16,7 +15,10 @@ jest.mock('appboy-web-sdk', () => ({
     changeUser: jest.fn(),
     ab: {
         InAppMessage: Object
-    }
+    },
+    logCardClick: jest.fn(),
+    logCardImpressions: jest.fn(),
+    requestImmediateDataFlush: jest.fn()
 }));
 
 const apiKey = '__API_KEY__';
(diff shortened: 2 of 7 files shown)


asherstoppard (dismissed):

minrwhite:
Rebased on top of latest master to resolve build issues

packages/f-metadata/src/index.js
```diff
@@ -58,4 +73,27 @@ const initialise = async (options = {}) => {
...
+    const appboy = await appBoyPromise;
+
+    const output = appboy[type](payload, true);
+    appboy.requestImmediateDataFlush();
+
+    return output;
+};
+
+/**
+ * @param {Object} card Content card for which to log a click event
```
ashleynolan: Should this just say `Content card` rather than `card Content card`?

ashleynolan: Actually it-s probably the name of the parameter right? If so, could you put a dash between it so it reads: 

`card - Content card for which to log a click event`

packages/f-metadata/CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v2.8.0
+------------------------------
+*June 24, 2020*
```
ashleynolan: Date bump

packages/f-metadata/src/index.js
DamianMullins: Either way is valid JSDoc syntax, but agree that adding the dash makes it more readable - especially when your IDE doesn't highlight the different parts.

packages/f-metadata/src/_tests/index.test.js
```diff
@@ -97,4 +106,103 @@ describe('f-metadata', () => {
...
+        it('should return different promises that resolve to the same value', async () => {
+            // Arrange
+            const appBoy1 = initialiseBraze(settings);
+
+            // Act
+            const appBoy2 = initialiseBraze(settings);
+
+            // Assert
+            expect(appBoy1).not.toBe(appBoy2);
+            await expect(appBoy1).resolves.toBe(await appBoy2);
```
DamianMullins, line 131: Do you need the `.resolves` part here?

packages/f-metadata/src/index.js
```diff
@@ -24,15 +30,23 @@ const initialise = async (options = {}) => {
 
     window.dataLayer = window.dataLayer || [];
 
+    if (appBoyPromise) {
```
ashleynolan, line 36: Is there an example of how this gets set so that it passes as true? Does the promise come through options or is it being set before initialise is called by one of the helper functions.

packages/f-metadata/src/index.js
minrwhite: The `card` here is the name of the variable, that IDEs can use as a basis for naming parameters.

packages/f-metadata/src/_tests/index.test.js
minrwhite: Yes - `appBoy1` is a promise, and `.resolves` unwraps it to it's resolved value. The `await` does the equivalent for `appboy2`

packages/f-metadata/src/index.js
minrwhite: This is a guard against `initialise()` being called more than once, and attempting to reboot the entire appboy sdk. If it has been called previously, then `appBoyPromise` will exist. I could add a comment to that effect, @ashleynolan ?

packages/f-metadata/src/index.js
minrwhite: Super - I will add the dash

packages/f-metadata/src/_tests/index.test.js
DamianMullins: OK, do you need the await keyword before `expect` as I don't think that returns a promise? Also, more out of curiosity, what happens if you move the `await` keyword before `appBoy1` inside the call to `expect`? Does that resolve in place allowing you to remove the the `.resolves` part? So `expect(await appBoy1).toBe(await appBoy2);`

minrwhite: > OK, do you need the await keyword before `expect` as I don't think that returns a promise?

It does return a promise [if the `.resolves` modifier is used](https://jestjs.io/docs/en/expect#resolves), as that turns any assertion into an asynchronous one. I could alternatively return the value returned from the assertion directly?

> Also, more out of curiosity, what happens if you move the `await` keyword before `appBoy1` inside the call to `expect`? Does that resolve in place allowing you to remove the the `.resolves` part? So `expect(await appBoy1).toBe(await appBoy2);`

It would do, but IMO I'd prefer using the `.resolves` modifer as I think that scans better semantically.

DamianMullins: Cool, thanks, good to know!

packages/f-metadata/src/index.js
ashleynolan: Ah, that makes sense. Adding a comment would be good  👍

ashleynolan (approved):

DamianMullins (approved):

PR #223, opened 2020-06-30
f-content-cards@1.0.0: Update README

asherstoppard:
Update the README with all relevant information for using and contributing to `f-content-cards`

### DIFF
--- a/packages/f-content-cards/CHANGELOG.md
+++ b/packages/f-content-cards/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v1.0.0
+------------------------------
+*July 1st, 2020*
+
+### Changed
+- README to include details on prerequisites, usage, development, testing and contributions
+
+
 v0.8.1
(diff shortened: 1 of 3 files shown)


DamianMullins (approved):

packages/f-content-cards/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v1.0.0
```
minrwhite, line 7: What's the justification for making this the first 'stable' version, rather than `v0.9.0`? 

I think we should at least make the styles scoped to individual components before we declare this a stable component, IMO.

ashleynolan: Yep I agree with @benwhite-deltas – I think v1 should be a component you'd be happy leaving as-is to some extent. I think the styles not being scoped should be a requirement for that.

asherstoppard: @ashleynolan @benwhite-deltas Fair points and I totally agree, I'll change this over to 0.9.0 :+1:

asherstoppard:
Re-opening as 0.9.0

PR #227, opened 2020-07-03
f-vue-icons@v1.1.0 - Adding colour props.

DamianMullins:
### Added
- Colour props which conditionally add CSS classes to components.


packages/f-vue-icons/build.js
```diff
@@ -9,13 +9,44 @@ export default {
     name: '${name}',
 
     props: {
+        isWhite: {
```
ashleynolan: Could we restructure this slightly to have one single `theme` prop? ( that then has `white`, `blue` etc passed into it)

Makes it a lot more scalable for future themes that way

packages/f-vue-icons/build.js
```diff
@@ -9,13 +9,44 @@ export default {
...
+            type: Boolean,
+            default: false
+        },
+
+        isOrange: {
+            type: Boolean,
+            default: false
+        },
+
+        pushLeft: {
```
ashleynolan: I think that layout like this should now move to the application component styling rather than be on the icon itself. Happy to discuss though.

ashleynolan (changes requested):

packages/f-vue-icons/build.js
DamianMullins: Agree, removing this one.

packages/f-vue-icons/build.js
```diff
@@ -37,7 +68,7 @@ Promise.all(icons.map(icon => {
         .then(() => fs.writeFile(filepath, component, 'utf8'));
 })).then(() => {
     const main = icons
-        .map(icon => `export { default as ${icon.pascalCasedComponentName} } from '../icons/${icon.pascalCasedComponentName}';`)
-        .join('\n\n');
+        .map(icon => `export { default as ${icon.pascalCasedComponentName} } from '../icons/${icon.pascalCasedComponentName}';\n`)
+        .join('');
```
ashleynolan, line 48: If you're removing the line breaks from the join, do you need the join anymore?

packages/f-vue-icons/build.js
DamianMullins: Maybe if we're thinking of going down that route we should hold off on adding these props. 

I only say that as the best way I can think of updating the props as you mention is to accept a string prop (which we can't easily validate) and the class setup becomes something like below which looks like it could easily be mis-used?

```js
ctx.data.class = {
    `c-ficon--${ctx.props.theme}`: ctx.props.theme !== ''
};
```

DamianMullins: Also, personal preference, I liked the way colours were set by adding an `is-white` rather than `theme="white"` 😸 

Happy to close this PR if adding this isn't actually going to be useful in a lot cases anyway.

packages/f-vue-icons/build.js
DamianMullins: Yes, if you don't set it then the result of the `.map()` call causes each item to added with a comma at the start of each line. It seems there's an implicit join happening.

PR #232, opened 2020-07-09
storybook@0.4.0 🌐 Updates sass-loader for contextual import

minrwhite:
### Changed
- Updated `sass-loader` to 7.3.1 to allow for contextual import of common.scss files from
  Single File Components in subdirectories as per [here](https://github.com/justeat/fozzie-components/compare/storybook@0.4.0...example/post-order-card-scoped-styles)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+Latest (roll into next release)
+------------------------------
+*July 9, 2020*
+
+### Changed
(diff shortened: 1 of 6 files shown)


asherstoppard (dismissed):

DamianMullins (approved):

kevinrodrigues (approved):

CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+Latest (roll into next release)
```
ashleynolan, line 7: Just for future reference – the root `package.json` is only incremented for our own documentation, it's never actually `released`. So this can just be bumped to the next minor version when changes in the root project are made.

PR #236, opened 2020-07-09
f-content-cards@1.0.0

minrwhite:
### Changed
- Styles separated out to individual components and scoped to avoid bleed-out


packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -179,14 +179,7 @@ export default {
...
 
-<style lang="scss">
-    // c-contentCard
-    // Styling for a card that holds data on a restaurant
-    // Currently consists of a main image, optional restaurant logo thumbnail
-    // And then info text below (header, tags, description)
-
-    $contentCardRadius: 8px;
-
+<style lang="scss" scoped>
```
ashleynolan: We have gone for CSS Modules (through the `module` attribute) rather than the default vue scoped attribute in the past – mainly as it lets you share the CSS classes and properties that you set in the actual component as well as in the styling (things like breakpoints can be handy to share for instance in some use cases).

Don't mind using scoped here as long as we're sure it won't be at odds with using CSS modules elsewhere.

packages/f-content-cards/src/assets/scss/common.scss
```diff
@@ -10,6 +10,8 @@ $theme: 'je' !default;
 @import '~@justeat/fozzie/src/scss/base/reset';
 @import '~@justeat/fozzie-colour-palette/src/scss';
 
+$contentCardRadius: 8px;
```
ashleynolan: Maybe a question to put to design rather than anything to hold up this PR, but we use a standard `border-radius` of `4px` (see [variables in Fozzie](https://github.com/justeat/fozzie/blob/master/src/scss/settings/_variables.scss#L137)). Would be interesting to know if this should be normalised (it may be done through PIE changes shortly anyway though)

packages/f-content-cards/src/components/cardTemplates/PostOrderCard.vue
```diff
@@ -83,3 +83,110 @@ export default {
...
+            @include font-size(large);
+
+            margin-bottom: spacing(x2);
+
+            @include media ('<mid') {
+                margin: spacing(x2);
+            }
+        }
+
+        /deep/ .c-contentCard-thumbnail {
```
ashleynolan: I think rather than changing the styles via `/deep/` here and below, it may be a bit more encapsulated to pass in a class onto the `CardContainer` (either directly or via a prop) that adds a modifier class to `c-contentCard` to change the styles inside that component (not sure what the difference in styles is so it's hard to suggest what that modifier would be named).

That way all the styles for the card component live there (and if you ever need to reuse the modifiers for a future content card you also could do too.

packages/f-content-cards/src/components/cardTemplates/PostOrderCard.vue
```diff
@@ -83,3 +83,110 @@ export default {
...
+        }
+
+        /deep/ .c-contentCard-img {
+            display: block;
+            max-width: 100%;
+        }
+    }
+
+    .c-postOrderCard--condensed {
+        /deep/ .c-contentCard-bgImg {
```
ashleynolan: Same here as above – you could pass a modifier class onto the `c-contentCard` component that could handle this change in styling so that all the styles for that component live inside that component SFC.

packages/f-content-cards/src/assets/scss/common.scss
minrwhite: I think this was to match the border radius set in the scss styles in OrderWeb IIRC - will confirm

minrwhite: Confirmed - the 8px border radius is for the UI blocks in the order confirmation/details page. I'll return the 8px border radius so it's only relevant to the post-order cards, and that we use the standard fozzie radius in other cards. FYI @asherstoppard

asherstoppard (dismissed):

packages/f-content-cards/src/components/cardTemplates/PostOrderCard.vue
```diff
@@ -24,7 +25,10 @@
 </template>
 
 <script>
+/* eslint-disable import/no-duplicates */
 import CardContainer from './CardContainer.vue';
+import CardContainerStyles from './CardContainer.vue?vue&type=style&index=0&lang=scss&module=true&';
```
DamianMullins, line 32: Haven't seen this before, I assume it imports the styles only from the other component? It might be worth adding a link in a comment so that future devs know what is going on?

asherstoppard (dismissed):

packages/f-content-cards/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v1.0.0
+------------------------------
+*July 14, 2020*
```
simonsandersje: Date needs updating already!

packages/f-content-cards/src/components/cardTemplates/CardContainer.vue
```diff
@@ -144,3 +145,212 @@ export default {
...
+    }
+
+    .c-contentCard-thumbnail {
+        border: 1px solid $grey--lighter;
+        margin-top: - (32px + spacing(x2)); // This offsets the thumbnail above the top of the info card
+        width: 48px;
+        min-height: 48px;
+    }
+
+    .c-postOrderCardContainer {
```
DamianMullins, line 272: Any reason why the post order card styles aren't set in the `PostOrderCard` component?

minrwhite: @DamianMullins - This was suggested by @ashleynolan - beacuse the elements that are being styled are represented in the template here, all styles that will apply to them should be defined here. The above class is applied to the `CardContainer` component when consumed by `PostOrderCard` in order to ensure that these styles are applied.

kevinrodrigues (dismissed):

simonsandersje (dismissed):

simonsandersje (approved):

kevinrodrigues (approved):

minrwhite:
To confirm - visual change on percy check above is because content card border radius should be fozzie default (4px) for non-post-order content cards, which wasn't the case before.

packages/f-content-cards/src/components/cardTemplates/PostOrderCard.vue
```diff
@@ -1,16 +1,17 @@
...
             v-if="title"
-            class="c-postOrderCard-title"
+            :class="[$style['c-postOrderCard-title']]"
             data-test-id="contentCard-postOrderCard-title"
         >
             {{ title }}
         </h2>
         <card-container
             :card="card"
+            :class="[cardContainerStyles['c-postOrderCardContainer'] , { [cardContainerStyles['c-postOrderCard--condensed']]: !image && icon }]"
```
ashleynolan, line 14: Sorry for the delay in reviewing (can go into a separate commit).

Would make sense to replicate the class naming of the `CardContainer` element here and just apply a modifier classname. So an explicit modifier name would be something like `c-contentCard--postOrder` but it may be that you can make the name generic to the layout/style changes that happen to be applied to a postOrder card?

Then you could use adjacent CSS selectors for `c-contentCard--postOrder.c-contentCard--condensed { … }` to apply a condensed set of styles to that card if that makes sense?

packages/f-content-cards/src/components/cardTemplates/CardContainer.vue
ashleynolan: Yep – if you can make the classname relate to the `c-contentCard` as a modifier (as mentioned below in my other comment) will make the styling here look more a part of this component.

packages/f-content-cards/src/components/cardTemplates/PostOrderCard.vue
minrwhite: Will fix this in a follow up commit as suggested. Thanks, @ashleynolan

PR #245, opened 2020-07-15
f-registration@v0.9.0: Update registration component to handle AccountWeb API endpoint

laurenrodney:
Update registration component to handle AccountWeb API endpoint.
---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/f-registration/src/components/Registration.vue
```diff
@@ -78,6 +78,12 @@
                         <warning-icon :class="$style['o-form-error-icon']" />
                         Please enter a valid email address
                     </p>
+                    <p
```
tommcclean, line 82: I think we should extract these three validation messages into a warning component (functional), and drive the message from a computed and passed as a prop which when blank the component uses not to display itself. We could get rid of a lot of markup and improve readability here if we did.

Could be a seperate ticket though if you add a JIRA and todo with the JIRA ID to the code

siggerzz (dismissed):

packages/f-registration/src/components/Registration.vue
```diff
@@ -239,8 +247,13 @@ export default {
...
             } catch (error) {
-                this.genericErrorMessage = error;
-                this.$emit(EventNames.CreateAccountFailure, error);
+                const thrownErrors = error.Errors ? error.Errors : error;
+                if (Array.isArray(thrownErrors) && thrownErrors.some(thrownError => thrownError.ErrorCode === '409')) {
+                    this.shouldShowEmailAlreadyExistsError = true;
+                } else {
+                    this.genericErrorMessage = error;
+                }
+                this.$emit(EventNames.CreateAccountFailure, thrownErrors);
```
tommcclean, line 261: Does this event serve a purpose? I imagine the success one is used to direct the parent to re-route, but onfailure I can't think of a reason the parent might want to act.

packages/f-registration/src/components/Registration.vue
```diff
@@ -239,8 +247,13 @@ export default {
                 await RegistrationServiceApi.createAccount(this.createAccountUrl, this.tenant, registrationData);
                 this.$emit(EventNames.CreateAccountSuccess);
             } catch (error) {
-                this.genericErrorMessage = error;
-                this.$emit(EventNames.CreateAccountFailure, error);
+                const thrownErrors = error.Errors ? error.Errors : error;
```
tommcclean: I think you can probably change this to `const thrownErrors = error.Errors || error`

packages/f-registration/src/components/Registration.vue
laurenrodney: This event was originally added [here](https://github.com/justeat/fozzie-components/pull/192/files#diff-6b4cb4fba4aead92cfe500f9deea38b9R229), but I'm not entirely sure what for

packages/f-registration/src/components/Registration.vue
```diff
@@ -239,8 +247,13 @@ export default {
...
+                if (Array.isArray(thrownErrors) && thrownErrors.some(thrownError => thrownError.ErrorCode === '409')) {
+                    this.shouldShowEmailAlreadyExistsError = true;
+                } else {
```
tommcclean, line 258: I think this could be challenging, because thrownErrors might or might not be an array, I am not sure you can assume because it isn't a 409; that it is not an array, so you might find it stores an object in the genericErrorMessage.

For example a 429 rate limit request, 404 if the component goes out of service, 500 if something breaks on the other end or a 400 if validation is not kept in sync.

I think the easiest thing to do might be to store a generic error message that is not sourced from the error object. "Something went wrong, please try again later..." while we would hopefully log the failure and have visibility of it somewhere else.

packages/f-registration/src/components/Registration.vue
tommcclean: Yeah looks like it was simply added for the test, which isn't ideal; but I guess you didn't add it in this PR. Mostly a curiousity tbf, maybe we can whack it out later when the component gets some use

packages/f-registration/src/components/Registration.vue
```diff
@@ -17,11 +17,12 @@
...
             @submit.prevent="onFormSubmit"
         >
+            <!-- TODO WCB-1031 - Extract error messages into a separate component -->
             <p
-                v-if="genericErrorMessage"
+                v-if="shouldShowGenericErrorMessage"
                 :class="$style['o-form-error']">
                 <warning-icon :class="$style['o-form-error-icon']" />
-                {{ genericErrorMessage }}
+                Something went wrong, please try again later
```
ashleynolan: It would be good to populate the generic error message with the result of `thrownErrors` (line 251), mainly as this gives the server a way of bubbling up an error message that the UI doesn't know about.

So for instance, if the server had more specific detail to give the user about an error from a server-side specific validation failing, this would be passed straight through (a bit like how the account exists message was passed through previously, until we made a specific case handling it).

packages/f-registration/src/components/Registration.vue
kevinrodrigues: It's worth checking if the consuming app is listening out for this event before removing it. If it's just for the tests then we should remove this and in the tests we should be mocking the catch response.

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -142,7 +142,36 @@ describe('Registration', () => {
                 await flushPromises();
 
                 // Assert
-                expect(wrapper.vm.genericErrorMessage).not.toBeNull();
+                expect(wrapper.vm.shouldShowGenericErrorMessage).toBe(true);
+                expect(wrapper.vm.shouldShowEmailAlreadyExistsError).toBe(false);
```
ashleynolan: I wouldn't test for `shouldShowEmailAlreadyExistsError` as part of this case (to keep each test more single purpose, if you wanted to make sure this was false when a generic error fires, I'd have a separate test for that specifically).

I'd be inclined to add an expect to check for what you expect a generic error message from the server to return though (so the default message that would appear).

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -142,7 +142,36 @@ describe('Registration', () => {
...
+                await wrapper.find("[data-test-id='input-last-name']").setValue(lastName);
+                await wrapper.find("[data-test-id='input-email']").setValue(email);
+                await wrapper.find("[data-test-id='input-password']").setValue(password);
+
+                // Act
+                wrapper.find("[data-test-id='create-account-submit-button']").trigger('click');
+                await flushPromises();
+
+                // Assert
+                expect(wrapper.vm.shouldShowGenericErrorMessage).toBe(false);
```
ashleynolan: Same here as above – I'd have a separate test to check this value when a `409` fires, as otherwise if this test fails because of this line the `it` statement won't make much sense.

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -142,7 +142,36 @@ describe('Registration', () => {
...
-                expect(wrapper.vm.genericErrorMessage).not.toBeNull();
+                expect(wrapper.vm.shouldShowGenericErrorMessage).toBe(true);
+                expect(wrapper.vm.shouldShowEmailAlreadyExistsError).toBe(false);
+                expect(wrapper.emitted(EventNames.CreateAccountFailure).length).toBe(1);
+            } finally {
+                wrapper.destroy();
+            }
+        });
+
+        it('should populate error message and emit failure event when service responds with a 409', async () => {
```
kevinrodrigues: Maybe for another PR but you can shorten this test to make it clearer. You could spy on the `RegistrationServiceApi.createAccount` method and mock the mockImplementation (similar to what you have already )and instead of setting up the values via the wrapper.find setup block you could invoke `onFormSubmit` and then the expect blocks would be as you have them already. Just reduces the need to find and set values.

packages/f-registration/src/components/Registration.vue
tommcclean: Yeah that was why I asked; cos there is no working consuming app yet :D I think it was just added for the test; but tbf it wasn't added in this PR

ashleynolan: My best guess would be that it's probably to fire a GTM analytics event in AccountWeb? (similar to how we fire tracking events with searchbox)

kevinrodrigues (approved):

packages/f-registration/src/components/Registration.vue
```diff
@@ -239,8 +248,17 @@ export default {
...
                 this.$emit(EventNames.CreateAccountSuccess);
             } catch (error) {
-                this.genericErrorMessage = error;
-                this.$emit(EventNames.CreateAccountFailure, error);
+                const thrownErrors = error.Errors || error;
+                if (Array.isArray(thrownErrors)) {
+                    if (thrownErrors.some(thrownError => thrownError.ErrorCode === '409')) {
+                        this.shouldShowEmailAlreadyExistsError = true;
+                    } else {
+                        this.genericErrorMessage = thrownErrors[0].Description || 'Something went wrong, please try again later';
```
ashleynolan, line 256: Doesn't need to be done as this PR, but I'd ask someone from copy team for a better generic error message here to add-in (as `Something went wrong, please try again later` isn't particularly helpful for a user trying to register – might be that it needs to link to the help centre or something).

ashleynolan (approved):
LGTM – would be good to update the copy for the fallback error message in a separate ticket.

packages/f-registration/src/components/Registration.vue
laurenrodney: New ticket has been created 👍

PR #244, opened 2020-07-14
f-header@3.4.0: Fix regression with account details response

zlatinivanov-je:
__*f-header regression fix with account details response*__

---

## v3.4.0
**Changed**
- Fix `f-header` regression with api response for user details.
- Added `?testuser` Demo mocks.
- Updated README.md with Demo mock details.
_

---

## UI Review Checks

Screenshot 2020-07-10 at 15 38 46

- [x] README and/or UI Documentation has been [created|updated]
- [x] Unit tests have been [created|updated]
- [x] This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist)

## Browsers Tested

- [x] Chrome (latest)
- [x] Internet Explorer 11
- [x] Mobile (Please list device/browser – Ideally one iPhone model and one Android)

### List any other browsers that this PR has been tested in:

- [View the Browser Support Checklist](http://fozzie.just-eat.com/documentation/general/browser-support)

PR #246, opened 2020-07-15
fozzie-components@1.23.0 - Remove demo scripts in favour of Storybook / Update README.md's

siggerzz:


packages/f-header/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-header",
   "description": "Fozzie Header – Globalised Header Component",
-  "version": "3.3.0",
+  "version": "3.5.0",
```
tommcclean, line 4: What happened to 3.4.0 🚀

siggerzz: https://github.com/justeat/fozzie-components/pull/244

tommcclean (commented):
LGTM - Except that I don't think we should bump the versions as we aren't going to release; happy to be pursuaded otherwise though.

ashleynolan (changes requested):
I actually think it's worth keeping the separate demo components – my main reason is that there's no other way of debugging them through Vue dev tools as this doesn't work with storybook (I think because the component gets iframed in).

Happy to hear if there's a big reason for getting rid of them, but I think there's a use-case for keeping them in addition to the main storybook way of viewing them.

siggerzz:
Not needed

PR #247, opened 2020-07-16
f-vue-icons@v1.0.0

ashleynolan:
## Changed
- A lot! Component has been modified to pull in the new f-icons v2 package and build vue components from those icons. The main reason for making these changes is to allow for tree-shaking of f-vue-icons (previously, all icons had to be loaded in, which isn't good for performance when an application may only use a couple of the icons).
- Icons are now given a class of c-ficon (changed from c-icon). This is to make it really clear which version of f-vue-icons an application is now using.
- Some logo names have changed between the last beta release and the full v1 release. This was just part of normalising the icon names and so will need to be updated in consuming packages (the main logo names have changed for instance).
- ESLint autofix turned off (so that tests don't pass due to --fix being applied, but then publish subsequently fails).


DamianMullins:
I've worked on this a fair bit already so won't add my tick, but the code looks good to me, plus I've added the beta 11 version to Menu Web which has been in production for almost a week now.

kevinrodrigues (approved):
LGTM :shipit:

DamianMullins:
@ashleynolan one question, which @tommcclean also raised at the UI Showcase last week; is there much value in keeping this and the f-icons projects separate going forward? As it stands, at the movement, in order to add a new icon, you have to first add it to the f-icons project and then update f-vue-icons in order to add it here as well. Feels like a single step in one project might be better in the future? 

One downside is that users could only consume the icons as Vue components, but do we know if anyone is actually doing that?

DamianMullins (approved):

DamianMullins:
OK, I give in, I've ticked it 😄

ashleynolan:
@DamianMullins @tommcclean So the reason the two projects have been kept separate is that it means that the `f-icons` package can be used by any framework in the future (as it essentially provides a base library for the icons). This will likely become more important as we start to standardise icons with Skip (which is happening through PIE), and then perhaps in future with Takeaway as well, as both of those platforms use React.

Agree it's two steps, but the `f-vue-icons` step is more of a bundling step as no extra code needs o be written manually (we could potentially automate the generation of a `f-vue-icons` PR if `f-icons` changed if we wanted to make this step easier).

tommcclean:
I am always a fan of as little friction as possible; I found it challenging to resolve accessibility issues with some SVG's because they go through so many imports that you have to trace everything through every level to see how they get manipulated.

I think the least friction might actually be to maintain the SVG's in f-icons and f-vue-icons seperately; and just update both when new icons are added, but understand that not everybody would agree with me on that.

Or even better, but slightly beyond my comprehension at the moment, why not export the f-vue-icons as web components so they can be imported by React and Vue, thus removing the need for f-icons, which would purely be legacy at that point. I am led to understand that exporting as web components is as easy as changing the vue cli libary step, then we could share components no matter what tech they were created in.

ashleynolan:
@tommcclean In terms of maintaining them separately, from experience that would be more friction I would think – trying to get people to update an icon in two places would just lead to them getting out of sync and could lead to more problems than it tries to solve.  We could do work to export the `f-vue-icons` as web components too, but then we'd still have to consolidate `f-icons` into `f-vue-icons` and that piece of work wouldn't be trivial.

I think part of the problem is likely a lack of documentation around where things sit between the two packages – if there was a clear set of info about how to add an icon in `f-vue-icons` and its relationship with `f-icons`, that would probably help as the README is pretty sparse atm with that kind of information. Would that help?

Would be good to understand the issue you had too as can make sure that's addressed in any README info.

tommcclean:
Cool, worth asking the question. The problem I had was I needed a yalc chain of four packages to surface errors in the SVG's where they actually showed up as they only got into the broken state through the various imports, hence it seemed logical to simplify by removing from the chain.

ashleynolan:
Yeh, I can see how that might annoying then. I think using yalc to debug the icons will be hard irrespective due to the fact they get compiled from an SVG as you say (wherever they sit) – but maybe a clearer process which is documented a bit better will help slightly. Happy to pick this up when I get some time.

PR #253, opened 2020-07-21
f-vue-icons@1.2.0, storybook@0.9.0

minrwhite:
List-display of Icons:

image

### Added
- Adds a story to show all icons in a list view for demonstration purposes.

### Fixed
- Ensures common and fozzie styles applied exclusively for scss fragments within `src`
  directory of a component only (allows stories themselves to be styled separately)


ashleynolan (dismissed):
Think this is a great idea.

Looks like there should be a way of adding custom styles to storybook to move these CSS rules out of the markup ([this suggestion may work](https://github.com/storybookjs/storybook/issues/6364#issuecomment-485651328)?).

Happy for that to go in as a separate PR though.

packages/f-vue-icons/stories/Icons.stories.js
```diff
@@ -0,0 +1,40 @@
...
+        <template v-for="(component, componentName) in icons">
+            <div style="border-bottom: solid 1px lightgrey; min-height: 32px;"><span>{{ componentName }}</span></div>
+            <div style="border-bottom: solid 1px lightgrey; min-height: 32px;"><component :is="component" style="max-width: 50%; max-height: 32px" /></div>
+        </template>
+    </div>`,
+    data: () => ({
+        icons
+    }),
+    methods: {
+        log: item => {
```
ashleynolan: QQ – where does this method get used?

minrwhite: Aha - this was left in. It was used when I was experimenting around with deconstructing the icons module into individual Icons. Will remove.

packages/f-vue-icons/stories/Icons.stories.js
```diff
@@ -0,0 +1,40 @@
...
+import * as IconsComponents from '../dist/f-vue-icons.es';
+
+const icons = {};
+
+Object.keys(IconsComponents).forEach(icon => {
+    icons[icon] = IconsComponents[icon];
+});
+
+export default {
+    title: 'Components/Atoms/Icons',
```
DamianMullins: Looks like you have an extra level which the other packages don't have? They only set "Components/Atoms/", was this intentional? You can see the result in your screenshot.

minrwhite: Had this in as it was as originally thought I might want a story per icon. Makes sense to keep as `Components/Atoms` with the list output we ended up on. Changed in latest commits.

minrwhite:
@ashleynolan Turned out the scss issue was due to the vue.config.js being unable to add the common.scss/fozzie imports, which we don't want for the story wrapper templates, anyhow! The latest change fixes this for storybook by restricting the scss additions to target items in the `src/` directory only.

packages/f-vue-icons/stories/Icons.scss
```diff
@@ -0,0 +1,15 @@
+.s-icons {
```
ashleynolan: Super NIT – can you name this file `_icons.scss`, just as that matches other scss filenames (underscore indicates it doesn't compile out to a CSS file but gets imported elsewhere).

ashleynolan (commented):
Looks good – just one tiny comment

DamianMullins (approved):

ashleynolan (approved):

PR #256, opened 2020-07-22
f-content-cards@1.4.0 🦆 Mocks out xhr requests for braze data

minrwhite:
Adds the capability to check the visual appearance of content cards within the final component wrapper, with faked data.

### Added
- `xhr-mock`, `crypto-js` and `faker` as dev dependencies
- Predictable faked card data for braze SDK responses

### Changed
- Story can multi-select different card types and then 'refresh'

### Example of mocked data (copied from Percy 😼)

image

### DIFF
--- a/packages/f-content-cards/package.json
+++ b/packages/f-content-cards/package.json
@@ -51,6 +51,7 @@
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.4.6",
     "@vue/test-utils": "1.0.3",
+    "crypto-js": "^4.0.0",
     "faker": "^4.1.0",
     "node-sass-magic-importer": "5.3.2",
     "xhr-mock": "^2.5.1"
--- a/packages/f-content-cards/package.json
+++ b/packages/f-content-cards/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-content-cards",
   "description": "Fozzie Content Cards",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "main": "dist/f-content-cards.umd.min.js",
   "files": [
     "dist"
@@ -51,6 +51,8 @@
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.4.6",
     "@vue/test-utils": "1.0.3",
-    "node-sass-magic-importer": "5.3.2"
+    "faker": "^4.1.0",
+    "node-sass-magic-importer": "5.3.2",
+    "xhr-mock": "^2.5.1"
   }
 }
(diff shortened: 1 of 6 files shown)


packages/f-content-cards/stories/mockData/cards.js
```diff
@@ -0,0 +1,197 @@
...
+};
+/* eslint-enable camelcase */
+
+/**
+ * A set of key types to label as expected by the 'options' knob in storybook
+ *
+ * @type {any}
+ */
+export const labelledMultiSelectAllowedValues = Object.fromEntries(Object.keys(cardTypes)
+    .map(key => [key, cardTypes[key].label]));
```
asherstoppard:
```suggestion
export const labelledMultiSelectAllowedValues = Object.fromEntries(Object.entries(cardTypes)
    .map([key, { label }] => [key, label]));
```

asherstoppard (dismissed):
Just one comment but doesn't change functionality so happy to approve :+1:

packages/f-content-cards/stories/mockData/data.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+export default () => {
+    const now = Math.floor(Date.now() / 1000);
+
+    /* eslint-disable camelcase */
+    return {
+        triggers: [],
+        config: {
+            time: now,
+            attributes_blacklist: [
```
xander-marjoram, line 27: Are these properties able to be renamed or is that outside of our control?

minrwhite: Unfortunately these property names are consumed as part of the braze SDK internals, so not modifiable from our side.

xander-marjoram (approved):

kevinrodrigues (approved):

packages/f-content-cards/package.json
```diff
@@ -51,6 +51,9 @@
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.4.6",
     "@vue/test-utils": "1.0.3",
-    "node-sass-magic-importer": "5.3.2"
+    "crypto-js": "^4.0.0",
```
ashleynolan, line 54: Super tiny NIT (can happen as any future PR) – can you remove the `^` from these dependencies so that they're properly locked to a version.

packages/f-content-cards/stories/ContentCards.stories.js
```diff
@@ -1,31 +1,109 @@
...
 import { defaultEnabledCardTypes } from '../src/services/contentCard.service';
-import { withA11y } from '@storybook/addon-a11y';
+
+import cards from './mockData/cards';
+import data from './mockData/data';
+
+/**
+ * Resets all locally stored braze data so that the stubbed data is always fresh on page load
+ */
+function resetBrazeData () {
```
ashleynolan, line 19: And another tiny one – may be good to pull out helper functions into a separate file, just to keep the `xx.stories.js` files as light as we can. As this is the first one that seems to have grown with functions and `const` data, may be good to discuss a bit more widely.

Any thoughts on how we may want to structure this in our stories? //cc @kevinrodrigues @DamianMullins

PR #260, opened 2020-07-28
f-content-cards@1.6.0: Add emitters

asherstoppard:
Add the following event emitters so the client can hook into certain aspects of f-metadata through the f-content-cards component:

- `@on-appboy-init`
- `@get-card-count`
- `@has-loaded`
- `@on-error`

## UI Review Checks


DamianMullins (dismissed):

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -178,9 +182,14 @@ export default {
                 .getTitleCard()
                 .arrangeCardsByTitles()
                 .output();
+
             this.cards = cards;
             this.titleCard = titleCard;
             this.hasLoaded = true;
+
+            this.$emit('on-appboy-init', appboy);
+            this.$emit('get-card-count', cards.length);
```
kevinrodrigues, line 191: Would card return as an empty array if it doesn't come back with anything?

kevinrodrigues (dismissed):

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -178,9 +182,14 @@ export default {
                 .getTitleCard()
                 .arrangeCardsByTitles()
                 .output();
+
             this.cards = cards;
             this.titleCard = titleCard;
             this.hasLoaded = true;
+
+            this.$emit('on-appboy-init', appboy);
```
minrwhite: 1. appboy is an old name for braze - would `on-braze-init` be better in an external (to the package) interface?

asherstoppard: That's a good point! I'll get it changed over :+1:

packages/f-content-cards/src/components/ContentCards.vue
asherstoppard: That's correct, it defaults to an empty array in the content card constructor :+1:

DamianMullins (approved):

minrwhite (approved):

PR #269, opened 2020-08-03
f-registration@0.10.0 - Handle 409 response to show custom error message

xander-marjoram:
Update the expected structure of the axios error response object, and update the logic so that we can show a custom error message for 409 response (an account with this email address already exists).

---

## UI Review Checks

## Browsers Tested
- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -248,12 +248,16 @@ export default {
                 await RegistrationServiceApi.createAccount(this.createAccountUrl, this.tenant, registrationData);
                 this.$emit(EventNames.CreateAccountSuccess);
             } catch (error) {
-                const thrownErrors = error.Errors || error;
+                let thrownErrors = error;
+                if (error && error.response && error.response.data && error.response.data.errors) {
+                    thrownErrors = error.response.data.errors;
+                }
+
                 if (Array.isArray(thrownErrors)) {
-                    if (thrownErrors.some(thrownError => thrownError.ErrorCode === '409')) {
+                    if (thrownErrors.some(thrownError => thrownError.errorCode === '409')) {
                         this.shouldShowEmailAlreadyExistsError = true;
                     } else {
-                        this.genericErrorMessage = thrownErrors[0].Description || 'Something went wrong, please try again later';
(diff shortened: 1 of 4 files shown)


packages/f-registration/src/components/Registration.vue
```diff
@@ -248,12 +248,16 @@ export default {
                 await RegistrationServiceApi.createAccount(this.createAccountUrl, this.tenant, registrationData);
                 this.$emit(EventNames.CreateAccountSuccess);
             } catch (error) {
-                const thrownErrors = error.Errors || error;
+                let thrownErrors = error;
+                if (error && error.response && error.response.data && error.response.data.errors) {
+                    thrownErrors = error.response.data.errors;
+                }
+
```
laurenrodney, line 255: Would this work/be any more efficient: `thrownErrors = error.response.data.errors || error` ?

simonsandersje (approved):

laurenrodney (approved):

packages/f-registration/src/components/Registration.vue
xander-marjoram: So I think if were to do
```
const thrownErrors = error.response.data.errors || error;
```
we would get "Cannot read property 'response' of undefined" or something similar

And then if we added `|| error` I don't think that could ever be reached, because if `error.response.data.errors` is falsy (which is would need to be to hit the second part of the `||`) then we wouldn't enter the if-block in the first place

ashleynolan: You can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here though to make this a lot cleaner.

```
if (error?.response?.data?.errors) {
```

ashleynolan: I think you could do:

```js
const thrownErrors = error?.response?.data?.errors || error;
```

DamianMullins: That'd work as long as we're using the required plugin to enable the syntax or using the latest version of `@babel/preset-env`.

PR #270, opened 2020-08-05
f-content-cards@1.8.0: Transform data model at source

asherstoppard:
Applies data transformation's to the incoming data, avoiding the need to do it in each individual component.

## UI Review Checks

- N/A This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist)


packages/f-content-cards/stories/VoucherCard.stories.js
```diff
@@ -52,8 +50,8 @@ export const VoucherCardcomponent = () => ({
         cardTitle: {
             default: text('Card Title', ' Aliquam et aliquam et.')
         },
-        description: {
-            default: text('Card Description', 'Voluptas sint id pariatur.')
+        subtitle: {
```
minrwhite, line 53: You'll need to change the last parameter in the call to `voucherCard` on line 79 also, this is causing the difference in Percy.

packages/f-content-cards/src/services/utils/getCardUrlTarget.js
```diff
@@ -0,0 +1,20 @@
+export const internalDomains = ['just-eat', 'justeat', 'menulog'];
+
+const getCardUrlTarget = cardUrl => {
+    try {
+        const url = new URL(cardUrl);
+        const openInNewWindow = !internalDomains.some(partial => url.hostname.indexOf(partial) > -1);
+        return openInNewWindow ? {
+            attribute: '_blank',
+            rel: 'noopener noreferrer'
```
minrwhite: I'm aware this is functionality that is already in there that has been moved, but it might be worth removing the [`noreferrer`](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noreferrer) here, as presumably we want things that we are advertising to know when we are sending traffic to them (granted, this can be done through analytics, but selective cookie/tracking blocking can prevent this from working correctly). Definitely good to keep `noopener` though as we don't want the opened tab to have any introspection into the js environment of our site.

minrwhite (approved):

DamianMullins (approved):

PR #272, opened 2020-08-05
f-registration@0.11.0 - Add validation for first and last name

adamashton:
Improved validation for first name and last name.

## Browsers Tested

- [X] Chrome (latest)
- [X] Edge

### DIFF
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -36,7 +36,19 @@
                         v-if="shouldShowFirstNameRequiredError"
                         :class="$style['o-form-error']">
                         <warning-icon :class="$style['o-form-error-icon']" />
-                        Please enter your first name
+                        Please include your first name
+                    </p>
+                    <p
+                        v-if="shouldShowFirstNameMaxLengthError"
+                        :class="$style['o-form-error']">
+                        <warning-icon :class="$style['o-form-error-icon']" />
+                        First name exceeds 50 chars
+                    </p>
+                    <p
+                        v-if="shouldShowFirstNameInvalidCharError"
+                        :class="$style['o-form-error']">
+                        <warning-icon :class="$style['o-form-error-icon']" />
+                        Name should only contain letters, hyphens or apostrophes
                     </p>
                 </template>
             </form-field>
@@ -53,7 +65,13 @@
                         v-if="shouldShowLastNameRequiredError"
                         :class="$style['o-form-error']">
                         <warning-icon :class="$style['o-form-error-icon']" />
-                        Please enter your last name
+                        Please include your last name
+                    </p>
+                    <p
+                        v-if="shouldShowLastNameMaxLengthError"
+                        :class="$style['o-form-error']">
+                        <warning-icon :class="$style['o-form-error-icon']" />
+                        Last name exceeds 50 chars
                     </p>
                 </template>
             </form-field>
@@ -119,7 +137,7 @@
 <script>
 import { globalisationServices } from '@justeat/f-services';
 import { validationMixin } from 'vuelidate';
-import { required, email } from 'vuelidate/lib/validators';
+import { required, email, maxLength } from 'vuelidate/lib/validators';
 import { WarningIcon } from '@justeat/f-vue-icons';
 import Card from '@justeat/f-card';
 import '@justeat/f-card/dist/f-card.css';
@@ -129,6 +147,14 @@ import FormButton from './Button.vue';
 import tenantConfigs from '../tenants';
 import RegistrationServiceApi from '../services/RegistrationServiceApi';
 import EventNames from '../event-names';
+
+// Returns true if there are no invalid chars in value. Valid chars are: a-z, A-Z, apostrophe, hyphen
+const validCharsInName = value => {
+    if (typeof value === 'undefined' || value === null || value === '') {
+        return true;
+    }
+    return /^[a-zA-Z'-]*$/.test(value);
+};
 
 export default {
     name: 'Registration',
@@ -190,7 +216,13 @@ export default {
     computed: {
         // Returns true if required validation conditions are not met and if the field has been `touched` by a user
         shouldShowFirstNameRequiredError () {
-            return (this.$v.firstName.$invalid && !this.$v.firstName.required) && this.$v.firstName.$dirty;
+            return this.$v.firstName.$invalid && !this.$v.firstName.required && this.$v.firstName.$dirty;
+        },
+        shouldShowFirstNameMaxLengthError () {
+            return this.$v.firstName.$invalid && !this.$v.firstName.maxLength && this.$v.firstName.$dirty;
+        },
+        shouldShowFirstNameInvalidCharError () {
+            return this.$v.firstName.$invalid && !this.$v.firstName.validCharsInName && this.$v.firstName.$dirty;
         },
         // Returns true if required validation conditions are not met and if the field has been `touched` by a user
         shouldShowLastNameRequiredError () {
@@ -215,10 +247,13 @@ export default {
 
     validations: {
         firstName: {
-            required
+            required,
+            maxLength: maxLength(50),
+            validCharsInName
         },
         lastName: {
-            required
+            required,
+            maxLength: maxLength(50)
         },
         email: {
             required,
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -48,7 +48,7 @@
                         v-if="shouldShowFirstNameInvalidCharError"
                         :class="$style['o-form-error']">
                         <warning-icon :class="$style['o-form-error-icon']" />
-                        Name should only contain letters, hyphens or apostrophes
+                        First name should only contain letters, hyphens or apostrophes
                     </p>
                 </template>
             </form-field>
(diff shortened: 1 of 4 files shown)


packages/f-registration/src/components/Registration.vue
```diff
@@ -36,7 +36,19 @@
...
                         :class="$style['o-form-error']">
                         <warning-icon :class="$style['o-form-error-icon']" />
-                        Please enter your first name
+                        Please include your first name
+                    </p>
+                    <p
+                        v-if="shouldShowFirstNameMaxLengthError"
+                        :class="$style['o-form-error']">
+                        <warning-icon :class="$style['o-form-error-icon']" />
+                        First name exceeds 50 chars
```
xander-marjoram, line 45: Should we stick with "characters"? Its meaning is slightly more obvious to non-technical users or those using screen readers.

packages/f-registration/src/components/Registration.vue
```diff
@@ -53,7 +65,19 @@
...
                         :class="$style['o-form-error']">
                         <warning-icon :class="$style['o-form-error-icon']" />
-                        Please enter your last name
+                        Please include your last name
+                    </p>
+                    <p
+                        v-if="shouldShowLastNameMaxLengthError"
+                        :class="$style['o-form-error']">
+                        <warning-icon :class="$style['o-form-error-icon']" />
+                        Last name exceeds 50 chars
```
xander-marjoram, line 74: Same here

packages/f-registration/src/components/Registration.vue
```diff
@@ -53,7 +65,19 @@
...
+                    </p>
+                    <p
+                        v-if="shouldShowLastNameInvalidCharError"
+                        :class="$style['o-form-error']">
+                        <warning-icon :class="$style['o-form-error-icon']" />
+                        Last name should only contain letters, hyphens or apostrophes
```
xander-marjoram, line 80: What about spaces?

packages/f-registration/src/components/Registration.vue
adamashton: I'm going off the ticket here, which comes from the UI designers...

packages/f-registration/src/components/Registration.vue
adamashton: Again, see ticket

packages/f-registration/src/components/Registration.vue
adamashton: Hm, gonna go back to Aaron Wilson about this one because I reckon we should allow spaces.

packages/f-registration/src/components/Registration.vue
```diff
@@ -130,6 +154,14 @@ import tenantConfigs from '../tenants';
 import RegistrationServiceApi from '../services/RegistrationServiceApi';
 import EventNames from '../event-names';
 
+// Returns true if there are no invalid chars in value. Valid chars are: a-z, A-Z, apostrophe, hyphen
+const validCharsInName = value => {
+    if (typeof value === 'undefined' || value === null || value === '') {
+        return true;
+    }
+    return /^[a-zA-Z'-]*$/.test(value);
```
xander-marjoram, line 162: This also excludes names with accented characters, or those using different alphabets

adamashton: Again, this is from the ticket.

packages/f-registration/src/components/Registration.vue
```diff
@@ -188,23 +220,32 @@ export default {
...
     computed: {
-        // Returns true if required validation conditions are not met and if the field has been `touched` by a user
+        // Validation methods return true if the validation conditions
+        // have not been met and the field has been `touched` by a user.
         shouldShowFirstNameRequiredError () {
-            return (this.$v.firstName.$invalid && !this.$v.firstName.required) && this.$v.firstName.$dirty;
+            return this.$v.firstName.$invalid && !this.$v.firstName.required && this.$v.firstName.$dirty;
+        },
+        shouldShowFirstNameMaxLengthError () {
+            return this.$v.firstName.$invalid && !this.$v.firstName.maxLength && this.$v.firstName.$dirty;
```
xander-marjoram, line 229: I _think_ I see how this works, but perhaps we could make it a little clearer and easy to maintain by extracting some of these behind some more computed properties? For example:
```
firstNameIsPresent () {
    return this.$v.firstName.required;
},
firstNameTooLong () {
    return !this.$v.firstName.maxLength;
}
```
What do you think?

adamashton: Initially I thought I agreed with this, it's definitely much cleaner. However, it's doing slightly more in that it's testing to see if the input has been touched as well. The test for `this.$v.firstName.$invalid` is unneeded actually.
It would look like this,
```
<p v-if="!firstNameIsPresent() && this.$v.firstName.$dirty"

...

firstNameIsPresent () {
    return this.$v.firstName.required;
}
```

packages/f-registration/src/components/Registration.vue
philgoff: yes, I would say use Characters. Chars is a developer term.

packages/f-registration/src/components/Registration.vue
philgoff: yes, I would say use Characters. Chars is a developer term.

packages/f-registration/src/components/Registration.vue
xander-marjoram: I appreciate you're following the AC provided on the ticket, but I think we need to make this regex more inclusive

xander-marjoram (changes requested):
There are a few questions that need answering around accessibility/inclusiveness. I'd appreciate some input from your PM.

packages/f-registration/src/components/Registration.vue
philgoff: @adamashton One to take back to Aaron and mention other accented characters or those using different alphabets.

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -189,5 +189,45 @@ describe('Registration', () => {
...
+            const wrapper = mountComponentAndAttachToDocument();
+            Object.defineProperty(wrapper.vm.$v, '$invalid', { get: jest.fn(() => false) });
+            try {
+                wrapper.find('[data-test-id="input-first-name"]').setValue('wh4t @ w3!rd |\\|ame');
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+
+                // Assert
```
smarkey, line 223: Should there be an assertion to ensure that the field either gets cleared or remains populated (depending on requirements) on failure event emission and presentation of the validation error message?

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -189,5 +189,45 @@ describe('Registration', () => {
                 wrapper.destroy();
             }
         });
+
```
philgoff, line 192: Where is the NameMaxLengthError check?

philgoff: Could we switch up the checks between first name and last name fields, please?

packages/f-registration/src/components/Registration.vue
smarkey: Non-blocking, but I've brought this to Aaron:

```
It came up on the PR that the use of the word "chars" isn't particularly user friendly. Additionally since it's not a word in the dictionary it may be not particularly accessible when you consider screen readers will be reading that out.
Do you know who provided the designs/copy so I can just query it with them?
```

packages/f-registration/src/components/Registration.vue
xander-marjoram: I think a better approach may be to instead _exclude_ certain characters, like `/^[^\[\]{}()#\/\\]$/`?

packages/f-registration/src/components/Registration.vue
```diff
@@ -130,6 +154,14 @@ import tenantConfigs from '../tenants';
 import RegistrationServiceApi from '../services/RegistrationServiceApi';
 import EventNames from '../event-names';
 
+// Returns true if there are no invalid chars in value. Valid chars are: a-z, A-Z, apostrophe, hyphen
+const validCharsInName = value => {
```
ashleynolan, line 158: Probably tidier to put this under a `methods` section of the default export below, as that's what we tend to do in other components

packages/f-registration/src/components/Registration.vue
```diff
@@ -130,6 +154,14 @@ import tenantConfigs from '../tenants';
 import RegistrationServiceApi from '../services/RegistrationServiceApi';
 import EventNames from '../event-names';
 
+// Returns true if there are no invalid chars in value. Valid chars are: a-z, A-Z, apostrophe, hyphen
```
ashleynolan, line 157: Can you make sure this comment is in JSDoc format (we're starting to try and standardise our comments a bit more).

Usually can get a JSDoc extension that helps with this too.

adamashton:
Closing PR for now as we work out the final validation rules.

packages/f-registration/src/components/Registration.vue
minrwhite: Accented characters will be disallowed from an allowlist unless the unicode `u` flag is used anyhow, in which case IMO the list of excluded characters would be of similar length to the list of included characters (for example, would you want to allow block-drawing characters? emoji?).

It may be worth looking into using [Unicode property escapes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes), though given as those have only been possible in the last couple of versions of most browsers, and not at all in IE11, a fallback to a series of allowed unicode value ranges (e.g. `[\uXXXX-\uXXXX]`) should perhaps be used instead.

PR #273, opened 2020-08-06
f-registration@0.11.0 - Add validation for first and last name

adamashton:
Improved validation for first name and last name.

The validation rules are to be aligned with [current production](https://www.just-eat.co.uk/account/register) so that we can compare drop-off rates when deployed before relaxing the validation later on.

## Browsers Tested

- [X] Chrome (latest)
- [X] Edge

### DIFF
--- a/packages/f-registration/CHANGELOG.md
+++ b/packages/f-registration/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.11.0
+------------------------------
+*August 5, 2020*
+
+### Added
+- Validation on First name and Last name regarding max length and invalid characters
+
+
 v0.10.0
 ------------------------------
 *August 3, 2020*
--- a/packages/f-registration/CHANGELOG.md
+++ b/packages/f-registration/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.11.0
 ------------------------------
-*August 5, 2020*
+*August 6, 2020*
 
 ### Added
 - Validation on First name and Last name regarding max length and invalid characters
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -36,7 +36,19 @@
                         v-if="shouldShowFirstNameRequiredError"
                         :class="$style['o-form-error']">
                         <warning-icon :class="$style['o-form-error-icon']" />
-                        Please enter your first name
+                        Please include your first name
+                    </p>
+                    <p
+                        v-if="shouldShowFirstNameMaxLengthError"
+                        :class="$style['o-form-error']">
+                        <warning-icon :class="$style['o-form-error-icon']" />
+                        First name exceeds 50 chars
+                    </p>
+                    <p
+                        v-if="shouldShowFirstNameInvalidCharError"
+                        :class="$style['o-form-error']">
+                        <warning-icon :class="$style['o-form-error-icon']" />
+                        Name should only contain letters, hyphens or apostrophes
                     </p>
                 </template>
             </form-field>
@@ -53,7 +65,13 @@
                         v-if="shouldShowLastNameRequiredError"
                         :class="$style['o-form-error']">
                         <warning-icon :class="$style['o-form-error-icon']" />
-                        Please enter your last name
+                        Please include your last name
+                    </p>
+                    <p
+                        v-if="shouldShowLastNameMaxLengthError"
+                        :class="$style['o-form-error']">
+                        <warning-icon :class="$style['o-form-error-icon']" />
+                        Last name exceeds 50 chars
                     </p>
                 </template>
             </form-field>
@@ -119,7 +137,7 @@
 <script>
 import { globalisationServices } from '@justeat/f-services';
 import { validationMixin } from 'vuelidate';
-import { required, email } from 'vuelidate/lib/validators';
+import { required, email, maxLength } from 'vuelidate/lib/validators';
 import { WarningIcon } from '@justeat/f-vue-icons';
 import Card from '@justeat/f-card';
 import '@justeat/f-card/dist/f-card.css';
@@ -129,6 +147,14 @@ import FormButton from './Button.vue';
 import tenantConfigs from '../tenants';
 import RegistrationServiceApi from '../services/RegistrationServiceApi';
 import EventNames from '../event-names';
+
+// Returns true if there are no invalid chars in value. Valid chars are: a-z, A-Z, apostrophe, hyphen
+const validCharsInName = value => {
+    if (typeof value === 'undefined' || value === null || value === '') {
+        return true;
+    }
+    return /^[a-zA-Z'-]*$/.test(value);
+};
 
 export default {
     name: 'Registration',
@@ -190,7 +216,13 @@ export default {
     computed: {
         // Returns true if required validation conditions are not met and if the field has been `touched` by a user
         shouldShowFirstNameRequiredError () {
-            return (this.$v.firstName.$invalid && !this.$v.firstName.required) && this.$v.firstName.$dirty;
+            return this.$v.firstName.$invalid && !this.$v.firstName.required && this.$v.firstName.$dirty;
+        },
+        shouldShowFirstNameMaxLengthError () {
+            return this.$v.firstName.$invalid && !this.$v.firstName.maxLength && this.$v.firstName.$dirty;
+        },
+        shouldShowFirstNameInvalidCharError () {
+            return this.$v.firstName.$invalid && !this.$v.firstName.validCharsInName && this.$v.firstName.$dirty;
         },
         // Returns true if required validation conditions are not met and if the field has been `touched` by a user
         shouldShowLastNameRequiredError () {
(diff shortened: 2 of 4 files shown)


packages/f-registration/src/components/Registration.vue
```diff
@@ -130,6 +154,19 @@ import tenantConfigs from '../tenants';
...
+ * Tests for existence of valid chars only in a string.
+ *
+ * @param {string} value The string to test.
+ * @return {boolean} True if there are no invalid chars in value, false otherwise.
+ */
+const validCharsInName = value => {
+    if (typeof value === 'undefined' || value === null || value === '') {
+        return true;
+    }
+    return /^[\u0060\u00C0-\u00F6\u00F8-\u017Fa-zA-Z-' ]*$/.test(value);
```
minrwhite: Probably should exclude the multiplication sign ('×' = `\u00D7`) here if we're disallowing the divide sign at `\u00f7`.

`/^[\u0060\u00C0-\u00F6\u00F8-\u017Fa-zA-Z-' ]*$/.test('×')` => `true`

```suggestion
    return /^[\u0060\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u017Fa-zA-Z-' ]*$/.test(value);
```

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -189,5 +189,67 @@ describe('Registration', () => {
                 wrapper.destroy();
             }
         });
+
+        it('should show error message and emit failure event when the first name field is not populated', async () => {
```
minrwhite, line 193: We should have an equivalent test for last name field.

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -189,5 +189,67 @@ describe('Registration', () => {
...
+                expect(wrapper.vm.shouldShowFirstNameRequiredError).toBe(true);
+                expect(wrapper.vm.shouldShowFirstNameMaxLengthError).toBe(false);
+                expect(wrapper.vm.shouldShowFirstNameInvalidCharError).toBe(false);
+                expect(wrapper.emitted(EventNames.CreateAccountFailure).length).toBe(1);
+            } finally {
+                wrapper.destroy();
+            }
+        });
+
+        it('should show error message and emit failure event when the first name field is populated with invalid input', async () => {
```
minrwhite, line 212: We should have an equivalent test for last name field.

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -189,5 +189,67 @@ describe('Registration', () => {
...
+                expect(wrapper.vm.shouldShowFirstNameRequiredError).toBe(false);
+                expect(wrapper.vm.shouldShowFirstNameMaxLengthError).toBe(false);
+                expect(wrapper.vm.shouldShowFirstNameInvalidCharError).toBe(true);
+                expect(wrapper.emitted(EventNames.CreateAccountFailure).length).toBe(1);
+            } finally {
+                wrapper.destroy();
+            }
+        });
+
+        it('should show error message and emit failure event when the last name field is populated with too long an input', async () => {
```
minrwhite, line 297: We should have an equivalent test for first name field.

packages/f-registration/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.11.0
+------------------------------
+*August 5, 2020*
```
minrwhite: Date bump required

packages/f-registration/src/components/Registration.vue
```diff
@@ -130,6 +154,19 @@ import tenantConfigs from '../tenants';
...
 import EventNames from '../event-names';
 
+/**
+ * Tests for existence of valid chars only in a string.
+ *
+ * @param {string} value The string to test.
+ * @return {boolean} True if there are no invalid chars in value, false otherwise.
+ */
+const validCharsInName = value => {
+    if (typeof value === 'undefined' || value === null || value === '') {
```
minrwhite: I don't think you need this check - `null` or `undefined` values are coerced to `''` and `''` will match the test. In which case, the entire function could just be:

```js
const validCharsInName = value => /^[\u0060\u00C0-\u00F6\u00F8-\u017Fa-zA-Z-' ]*$/.test(value);
```

packages/f-registration/src/components/Registration.vue
```diff
@@ -130,6 +154,19 @@ import tenantConfigs from '../tenants';
 import RegistrationServiceApi from '../services/RegistrationServiceApi';
 import EventNames from '../event-names';
 
+/**
```
adamashton, line 157: Vuelidate requires a public/static method here for it's validation to work.

packages/f-registration/src/components/Registration.vue
adamashton: I am using feature parity with ConsumerWeb production so we can do like for like testing.

minrwhite:
Also, please watch the private jira reference leakage. Please name your branches `{component}@version` or `{component}/{feature}` as the repo should ideally be in a state where it could be made public if desired (following sign-off, of course).

packages/f-registration/src/components/Registration.vue
minrwhite: Cool, will resolve this for now and query in our team channel.

xander-marjoram (approved):

minrwhite (approved):

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -189,5 +189,159 @@ describe('Registration', () => {
...
+            const wrapper = mountComponentAndAttachToDocument();
+            Object.defineProperty(wrapper.vm.$v, '$invalid', { get: jest.fn(() => false) });
+            try {
+                // Act
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+
+                // Assert
+                expect(wrapper.vm.shouldShowFirstNameRequiredError).toBe(true);
+                expect(wrapper.vm.shouldShowFirstNameMaxLengthError).toBe(false);
```
ashleynolan, line 204: Late to the reviews – I'm not sure checking `shouldShowFirstNameMaxLengthError` and `shouldShowFirstNameInvalidCharError` are that relevant here?

Would perhaps be more useful to check the string that is shown as the error to see that the right error message is being displayed, rather than that other `firstName` validations come back as false?

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -189,5 +189,159 @@ describe('Registration', () => {
...
+            Object.defineProperty(wrapper.vm.$v, '$invalid', { get: jest.fn(() => false) });
+            try {
+                wrapper.find('[data-test-id="input-first-name"]').setValue('wh4t @ w3!rd |\\|ame');
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+
+                // Assert
+                expect(wrapper.vm.shouldShowFirstNameRequiredError).toBe(false);
```
ashleynolan, line 224: Same here – seems strange to check that these other `computedProps` are false. If we really want to check the false-iness of these, it's better to have a separate test that describes why, but I think the main thing to test in this block is that `shouldShowFirstNameInvalidCharError` is `true` and that the relevant error message string is shown?

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -189,5 +189,159 @@ describe('Registration', () => {
...
+            try {
+                const longValue = 'abcdefghijabcdefghijabcdefghijabcdefghijabcdefghijabcdefghij';
+                wrapper.find('[data-test-id="input-first-name"]').setValue(longValue);
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+
+                // Assert
+                expect(wrapper.vm.shouldShowFirstNameRequiredError).toBe(false);
```
ashleynolan, line 246: Same here as above.

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -189,5 +189,159 @@ describe('Registration', () => {
...
+            try {
+                wrapper.find('[data-test-id="input-first-name"]').setValue('Adam');
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+
+                // Assert
+                expect(wrapper.vm.shouldShowLastNameRequiredError).toBe(true);
+                expect(wrapper.vm.shouldShowLastNameMaxLengthError).toBe(false);
```
ashleynolan, line 268: And the same for the `lastName` checks as above.

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -189,5 +189,159 @@ describe('Registration', () => {
...
+                wrapper.find('[data-test-id="input-last-name"]').setValue('O\'Neil-Wight');
+                wrapper.find('[data-test-id="input-email"]').setValue('ashton.adamms+jetest@just-eat.com');
+                wrapper.find('[data-test-id="input-password"]').setValue('Secure123');
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+
+                // Assert
+                expect(wrapper.vm.shouldShowFirstNameRequiredError).toBe(false);
```
ashleynolan, line 335: Checking that the error message isn't shown should be enough for this test rather than listing all of these expects to be `false`?

packages/f-registration/src/components/tests/Registration.test.js
adamashton: I'll modify these tests in a new release today.

PR #271, opened 2020-08-05
f-mega-modal@0.1.0 - Component structure and core functionality

DamianMullins:
### Added
- Component structure and basic configuration (created using `generator-component`).
- Component markup and core functionality.
- Storybook config.
- Unit tests.
- Readme content.

## UI Review Checks

- [x] README and/or UI Documentation has been created
- [x] Unit tests have been created
- [x] This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist)

## Browsers Tested

- [x] Chrome (latest)
- [x] Internet Explorer 11
- [x] Safari iPhone 11
- [x] Chrome iPhone 11
- [x] Chrome Pixel 3a
- [x] Chrome Pixel 4

PR #274, opened 2020-08-06
f-mega-modal@0.2.0 - Minor fixes and updates.

DamianMullins:
### Added
- CSS classes which can be targeted in a consuming application..

### Changed
- Tidied up close button styles.
- Use Fozzie colour aliases where possible.

### Fixed
- Spelling mistaxe.
- Moved config comment to correct prop.

### DIFF
--- a/packages/f-mega-modal/README.md
+++ b/packages/f-mega-modal/README.md
@@ -109,8 +109,8 @@ The props that can be defined are as follows:
 | `is-narrow` | `Boolean` | `false` | Use the narrow visual style. |
 | `is-wide` | `Boolean` | `false` | Use the wide visual style. |
 | `is-flush` | `Boolean` | `false` | Removes passing around the modal content. |
-| `is-full-height` | `Boolean` | `false` | Sets the modal content to full height of the screen. |
-| `is-scrollable` | `Boolean` | `false` | Makes the modal content scrollable.<br/><br/><blockquote>Note this only applies to small screen devices.</blockquote> |
+| `is-full-height` | `Boolean` | `false` | Sets the modal content to full height of the screen.<br/><br/><blockquote>Note this only applies to small screen devices.</blockquote> |
+| `is-scrollable` | `Boolean` | `false` | Makes the modal content scrollable. |
 | `is-close-fixed` | `Boolean` | `false` | Sets the modal close button position to `fixed`. |
 | `is-close-rounded` | `Boolean` | `false` | Sets the modal close button to display in a rounded style. |
 | `has-overlay` | `Boolean` | `true` | Controls whether or not to display an overlay behind the modal. |
--- a/packages/f-mega-modal/README.md
+++ b/packages/f-mega-modal/README.md
@@ -120,7 +120,7 @@ The props that can be defined are as follows:
 
 ### CSS Classes
 
-The modal has it's own styles which are scoped to the component using CSS modules to prevent conflicts with existing styles on the page.
+The modal has its own styles which are scoped to the component using CSS modules to prevent conflicts with existing styles on the page.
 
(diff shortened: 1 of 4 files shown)


packages/f-mega-modal/README.md
```diff
@@ -109,15 +109,30 @@ The props that can be defined are as follows:
...
 | `is-close-fixed` | `Boolean` | `false` | Sets the modal close button position to `fixed`. |
 | `is-close-rounded` | `Boolean` | `false` | Sets the modal close button to display in a rounded style. |
 | `has-overlay` | `Boolean` | `true` | Controls whether or not to display an overlay behind the modal. |
 | `has-close-button` | `Boolean` | `true` | Controls whether or not to display the modal close button. |
 | `close-on-blur` | `Boolean` | `true` | Controls whether or not to close the modal when the user clicks outside of the modal. |
 | `close-button-copy` | `String` | `"Close modal"` | Sets the hidden text value for the close button which is used by screen-readers. |
 
+### CSS Classes
+
+The modal has it's own styles which are scoped to the component using CSS modules to prevent conflicts with existing styles on the page.
```
xander-marjoram: Nit: `its`

xander-marjoram (approved):

asherstoppard (approved):

packages/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -1,14 +1,15 @@
 <template>
     <div
         ref="megaModal"
-        :class="[$style['c-megaModal'], {
+        :class="['c-megaModal', $style['c-megaModal'], {
```
ashleynolan, line 4: Out of interest – what does repeating this classname allow for?

xander-marjoram: Damian mentioned it in his talk earlier, and I asked him same question from looking at #271 😄 

The second is for scoping with CSS modules. All of the classes in the SFC defined below (`<style lang="scss" module>`) will have generated names to prevent the styles from bleeding into other components.

The first one is just a standard CSS class for hooking onto in other ways, either on a per-project basis or with fozzie, etc, or for testing (although we should be using attributes for this).

DamianMullins: It's there to hook custom styles onto as CSS modules will generate a class like `Modal MegaModal_c-megaModal_3w2Fg` which if you want to target you have to do something like `[class*='c-megaModal']`.

Adding the standard `c-megaModal` class (with no styles targeting it in this component) means that in consuming projects you can simply target `.c-megaModal`.

DamianMullins: Thanks @xander-marjoram, you commented whilst I was responding but your comment explains better than mine!

ashleynolan: Ah ok – I guess we probably need to balance how much we use this, as this kind of goes against using CSS modules in the first place (that the styles for that component should be packaged in with the component and it doesn't get affected by the consuming application). Is this mostly for extending the functionality of the modal specifically for menu in this case?

This may also actually help with an issue we're seeing in jest where theCSS module classes don't get rendered in the tests so it's impossible to check for the presence of a class (such as when a prop is `true`, display this classname).

DamianMullins: So the styles for the modal are still scoped to this component - like you said the new classes are purely there to hook onto to extend the styles in the consuming application. The styles I added for the modal are purposely very minimal to allow for a good base to build on top of.

So in menu we target these selectors to style the modals to look different on small screen devices and for the custom headings in the item selector, adding that into the component felt a bit messy, and also the custom styles are very menu specific.

PR #282, opened 2020-08-18
f-services@1.1.0 - Add axios interceptors for adding response time to response object

xander-marjoram:
Slight change of approach: this leaves the publishing of stats up to the consuming application, so f-services doesn't need to know how/where you choose to do this. A callback hook has been provided that should be used for pushing the stats. This callback function should expect an Axios response object, extended to include a new `responseTimeMs` property.

### Added
- Request/response interceptors to add response times to response object.
- `axios-mock-adapter` to help with testing.

---

## UI Review Checks

## Browsers Tested
- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/f-services/src/responseTimes.js
```diff
@@ -0,0 +1,56 @@
...
+        config.meta = config.meta || {};
+        config.meta.requestStartedAt = new Date().getTime();
+        return config;
+    });
+
+    interceptors.response.use(response => {
+        const timeTakenMs = new Date().getTime() - response.config.meta.requestStartedAt;
+
+        if (process.env.NODE_ENV === 'development') {
+            console.log(`Executed (${response.config.url}) in ${timeTakenMs} ms`);
```
DamianMullins: Will the linter fail on this line?

xander-marjoram: It didn't when running `yarn lint` inside f-services, but I suspect it may have inside a consuming component. I still need to test it inside Magikarp 👍

simonsandersje (approved):

DamianMullins (approved):

PR #284, opened 2020-08-18
storybook@v0.9.0 – Updating to latest storybook version and adding typography docs #trivial

ashleynolan:
### Added
- Config for parsing `.mdx` files which can be used for markdown documentation with components.

### Changed
- Updated to Storybook `v6` (and updated stories across other packages as part of the migration).


minrwhite (dismissed):

packages/f-footer/stories/footer.stories.js
```diff
@@ -31,6 +31,4 @@ export const FooterComponent = () => ({
     template: '<vue-footer :showCourierLinks="showCourierLinks" :locale="locale" />'
 });
 
-FooterComponent.story = {
-    name: 'f-footer'
-};
+FooterComponent.storyName = 'f-footer';
```
xander-marjoram, line 34: Nit: It does irk me that some of these have `Component` and some have `component`

packages/storybook/package.json
```diff
@@ -9,15 +9,15 @@
...
   "devDependencies": {
     "@percy/storybook": "3.3.0",
-    "@storybook/addon-a11y": "^5.3.19",
-    "@storybook/addon-actions": "5.3.19",
-    "@storybook/addon-knobs": "5.3.19",
-    "@storybook/addon-links": "5.3.19",
-    "@storybook/addon-notes": "5.3.19",
-    "@storybook/vue": "5.3.19",
+    "@storybook/addon-a11y": "6.0.5",
+    "@storybook/addon-actions": "6.0.5",
```
xander-marjoram: Do we still need this as a dependency if we're not calling it in main.js?

ashleynolan: Nope - it's now part of the essentials package, so I'll delete

packages/f-footer/stories/footer.stories.js
ashleynolan: Git blame whoever named the component 😉  Guessing it's just been copy-pasted from the first content card that was created?

// cc @benwhite-deltas @asherstoppard  One for your next PR with any content card updates? 😆

packages/storybook/package.json
ashleynolan: Have just updated

packages/storybook/config/storybook/main.js
```diff
@@ -1,10 +1,12 @@
...
-    '@storybook/addon-knobs',
-    '@storybook/addon-links',
-    '@storybook/addon-notes',
-    '@storybook/addon-a11y/register'
-  ]
-};
\ No newline at end of file
+    stories: [
+        '../../../**/*.stories.@(js|mdx)',
+        '../../../../stories/*.stories.@(js|mdx)'
```
DamianMullins, line 4: Does the previous glob not pick this path up?

DamianMullins (approved):

DamianMullins:
@ashleynolan I assume the Percy checks are OK as it looks like a padding adjustment has shifted them all slightly?

packages/storybook/config/storybook/main.js
ashleynolan: Weirdly, I don't think it did when I was playing with the patterns.

This was the recommended pattern from the docs as apparently the old version of storybook didn't implement `proper globs` 🤷‍♂️ .  Apparently `@` matches exactly one of the patterns in the brackets according to the spec. I'd never seen it used before though…!

ashleynolan:
@DamianMullins Yeh, I think this must be a storybook v6 padding addition

ashleynolan:
@benwhite-deltas @xander-marjoram Any chance you could give this a second look?

minrwhite (approved):

PR #287, opened 2020-08-21
f-header@v3.7.0 - Dependency updates.

DamianMullins:
### Changed
- Updated f-vue-icons version.
- Axios is now a peer dependency.

### DIFF
--- a/packages/f-header/CHANGELOG.md
+++ b/packages/f-header/CHANGELOG.md
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v3.7.0
+------------------------------
+*August 21, 2020*
+
+### Changed
+- Updated f-vue-icons version.
+- Axios is now a peer dependency.
+
+
 v3.6.0
 ------------------------------
 *August 19, 2020*
--- a/packages/f-header/package.json
+++ b/packages/f-header/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-header",
   "description": "Fozzie Header – Globalised Header Component",
-  "version": "3.6.0",
+  "version": "3.7.0",
   "main": "dist/f-header.umd.min.js",
   "files": [
     "dist",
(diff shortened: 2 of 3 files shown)


packages/f-header/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v3.7.0
+------------------------------
+*August 21, 2020*
+
+### Changed
+- Updated f-vue-icons version.
+- Axios is now a peer dependency.
```
ashleynolan, line 13: Is this moved to a peer dependency so that we don't include it twice when an application is already using axios?

Just want to make sure I understand why this has moved.

DamianMullins: That was the intention. Looking at the bundle output now I can see that Axios is _still always_ bundled regardless. I think the solution is to also mark it as an external dependency in the Vue config file - so it adds a `require` hook for `axios` into the bundle rather than including the entire library - can you think of any downsides to doing that?

DamianMullins: I can see it breaks the demo script.

DamianMullins: One benefit is that the package size is reduced as `axios` is no longer bundled up.

I feel like maybe there's a bigger question here of whether or not we should be using vue-cli to bundle our libraries. Most libraries out there use things like rollup as they offer ESM outputs which allow to do things like tree-shaking etc - vue-cli doesn't offer that as it uses webpack under the hood which can't generate ESM bundles.

ashleynolan: Yeh, for sure. IMO, I don't mind what we use for bundling (we use a variety across different projects). I think we mainly use `vue-cli` because it gives an easy setup for compiling out to a vue lib. I guess if we were to switch to rollup or somethign similar for packages, we'd just need to have a recommended setup for using it to mak it as simple to use.

DamianMullins: Sure, makes sense. We use rollup in a few places, as well as [`bili`](https://github.com/egoist/bili) which provides sensible defaults for rollup, in a few places in our Fozzie packages so I'm sure we could put a recommended config together. I have actually played around with doing that in the past for the mega-modal but never had time to take it very far. I'll try to pencil in some time soon to do that.

PR #292, opened 2020-08-26
f-header@v3.7.0

luke-andrew-parry:
This PR adds a new prop to the header and navigation components to allow the login & user info icon to be hidden if necessary.


packages/f-header/src/components/Navigation.vue
```diff
@@ -271,6 +272,11 @@ export default {
             default: false
         },
 
+        showLoginOrUserInfo: {
```
DamianMullins: If these are grouped together should it be `showLoginAndUserInfo`?

packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -529,4 +530,48 @@ describe('Navigation', () => {
...
+            const propsData = {
+                ...defaultPropsData,
+                showLoginOrUserInfo: false
+            };
+
+            // Act
+            const wrapper = shallowMount(Navigation, {
+                propsData,
+                data () {
+                    return {
```
DamianMullins: Shouldn't need to spread the object here, you can simply return the value i.e. `return defaultData;`.

packages/f-header/src/components/Header.vue
```diff
@@ -29,6 +29,7 @@
                 :order-count-url="orderCountUrl"
                 :is-order-count-supported="isOrderCountSupported"
                 :header-background-theme="headerBackgroundTheme"
+                :show-login-or-user-info="showLoginOrUserInfo"
```
ashleynolan: I'd probably just call this `showLoginInfo`. As long as we comment the prop to explain what it does, I think that's still obvious enough that it'll hide the user info too when it's false.

I can add a props tabel to the `README` shortly too so we can get it documented properly there.

packages/f-header/src/components/Navigation.vue
luke-andrew-parry: It depends on whether the user is logged in or not. If they're not logged in we want to hide the login button, if they are we want to hide the user icon. I don't want to hide the whole nav container as we probably still want the `help` link displayed.

luke-andrew-parry: As per @ashleynolan's suggestion below I'll rename the prop to `showLoginInfo`.

ashleynolan (commented):
Looks good – can you add the new prop to the [storybook config](https://github.com/justeat/fozzie-components/tree/master/packages/f-header/stories) as well so that it can be toggled on/off.

packages/f-header/src/components/Navigation.vue
```diff
@@ -86,7 +86,7 @@
                 </li>
 
                 <li
-                    :class="['c-nav-list-item has-sublist', { 'is-hidden': !userInfo || !showLoginOrUserInfo, 'open': navIsOpen }]"
+                    :class="['c-nav-list-item has-sublist', { 'is-hidden': !userInfo || !showLoginInfo, 'open': navIsOpen }]"
```
DamianMullins: NIT: this is a little hard to read, I'd suggest wrapping the object properties onto new lines.

```suggestion
                    :class="['c-nav-list-item has-sublist', {
                        'is-hidden': !userInfo || !showLoginInfo,
                        'open': navIsOpen
                    }]"
```

packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -529,4 +530,45 @@ describe('Navigation', () => {
...
+            // Arrange
+            const propsData = {
+                ...defaultPropsData,
+                showLoginInfo: false
+            };
+
+            // Act
+            const wrapper = shallowMount(Navigation, { propsData });
+            await wrapper.setData({
+                ...propsData,
```
DamianMullins: Should this be `defaultData`?

luke-andrew-parry: No. We need to set the `showLoginInfo` to `false` for this test. It defaults to true.

ashleynolan (dismissed):

packages/f-header/src/components/tests/Navigation.test.js
DamianMullins: Passing `props` data into `data` doesn't seem right though? You're already passing the `props` in on line 562?

luke-andrew-parry: ah yes. Sorry, I see what the typo now! I'll fix and push up.

DamianMullins (approved):

packages/f-header/src/components/Navigation.vue
```diff
@@ -86,7 +86,11 @@
                 </li>
 
                 <li
-                    :class="['c-nav-list-item has-sublist', { 'is-hidden': !userInfo, 'open': navIsOpen }]"
+                    :class="['c-nav-list-item has-sublist', {
```
xander-marjoram, line 89: (This is a question for another PR, I don't expect you to do this @luke-andrew-parry)

 @ashleynolan Shouldn't this element be wrapped in a `v-if`? It seems a lot simpler than having to hide it with CSS and set `tabindex=-1`, etc, and then we will also get the performance improvements of not having to write as much to the DOM.

xander-marjoram: And I know that if we do choose to render the list, it won't always be open, so we would have a `v-if` which is `true` if the user is logged in, and a `v-show` (or stick with CSS classes) for toggling on-hover, etc.

luke-andrew-parry: @xander-marjoram / @ashleynolan are you happy for me to merge this PR and move this discussion /changes into a separate PR?

xander-marjoram: @luke-andrew-parry Yes that's fine with me

ashleynolan: Yep – happy with that. I think we probably could use a v-if for the `is-hidden` part of this element for sure. I think the `open` class would still need to be done as it is though. Happy for this to go in as a separate update by someone though, as agree it shouldn't be on @luke-andrew-parry to update.

ashleynolan (approved):

PR #296, opened 2020-09-01
fozzie-components@1.25.0 - Add WebdriverIO to Fozzie-components for UI based component tests

siggerzz:


packages/f-header/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v3.8.0
+------------------------------
+*September 1, 2020*
+
+### Added
+- WebdriverIO tests
```
SachaZvetelman: Nit: WebDriverIO 👍

packages/f-header/package.json
```diff
@@ -28,10 +28,19 @@
   "scripts": {
     "prepublishOnly": "yarn && yarn lint && yarn test && yarn build",
     "build": "vue-cli-service build --target lib --name f-header ./src/index.js",
-    "demo": "vue serve --open ./src/demo/Index.vue",
+    "demo": "vue serve --port 8080 ./src/demo/Index.vue",
```
SachaZvetelman, line 31: I realise you added this so you can browse it from the UI tests, but I believe `8080` is used by default when you do `yarn dev`, for example. Should we choose a different port to avoid clashes?

packages/f-header/package.json
```diff
@@ -46,6 +55,7 @@
     "@justeat/f-trak": ">=0.6.0"
   },
   "devDependencies": {
+    "@babel/register": "^7.10.5",
```
SachaZvetelman: Remove `^` please :)

packages/f-header/test/specs/f-header.component.spec.js
```diff
@@ -0,0 +1,10 @@
+import HeaderComponent from '../../test-utils/component-objects/f-header.component';
+
+describe('f-header component tests', () => {
+    it('should display the f-header component', () => {
+        browser.url('http://localhost:8080');
+
+        HeaderComponent.logoIsDisplayed();
```
SachaZvetelman: We normally put the `is`, `can`, `has`, etc, at the beginning of the name: `isLogoDisplayed();`

packages/f-registration/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.17.0
+------------------------------
+*September 1, 2020*
+
+### Added
+- WebdriverIO component tests
```
SachaZvetelman: Same, hehe.

packages/f-registration/package.json
```diff
@@ -28,11 +28,15 @@
   "scripts": {
     "prepublishOnly": "yarn lint && yarn test && yarn build",
     "build": "vue-cli-service build --target lib --name f-registration ./src/index.js",
-    "demo": "vue serve --open ./src/demo/Index.vue",
+    "demo": "vue serve --port 8081 ./src/demo/Index.vue",
```
SachaZvetelman, line 31: Why `8081` in this case? :)

packages/f-registration/test/specs/f-registration.component.spec.js
```diff
@@ -0,0 +1,22 @@
...
+
+        const userInfo = {
+            firstName: '',
+            lastName: '',
+            email: '',
+            password: ''
+        };
+
+        RegistrationComponent.submitRegistrationForm(userInfo);
+        assert.ok(RegistrationComponent.firstNameEmptyErrorIsDisplayed());
```
SachaZvetelman: Same re naming as above.

SachaZvetelman (changes requested):

packages/f-header/package.json
siggerzz: The tests utilise the `yarn dev` script, so there shouldn't be any conflicts 👍

test/infrastructure/healthcheck.js
```diff
@@ -0,0 +1,29 @@
...
+
+        if (status === 200) {
+            clearInterval(intervalId);
+        }
+    } catch (error) {
+        console.log('Attempt', attempts, error.code); // eslint-disable-line
+    }
+
+    if (attempts >= 15) {
+        console.error('Too many attempts'); // eslint-disable-line
```
BanksySan: Could this message be more descriptive?  Something like:

    `Failed to get ${address} after {numberOfAttempts} attempts.  Server returned {HhhpStatusCode}_?`

It might actually be better to throw an exception as that will cause the process to exit with a non-zero code.

packages/f-registration/package.json
siggerzz: By having each component load on a unique port, it means we can run `test-component:chrome` script at the root level and have the tests run in parallel.

Previously this wasn't possible because all demo scripts got started in `8080`.

I'll look into a better approach to dynamically pick a free port at some point (as you may have something else on your machine running on 8081 for example)


I think this is okay for now though, given I've only added tests to 2 packages

BanksySan (changes requested):

SachaZvetelman (dismissed):

packages/f-header/test/specs/f-header.component.spec.js
```diff
@@ -0,0 +1,10 @@
+import HeaderComponent from '../../test-utils/component-objects/f-header.component';
+
+describe('f-header component tests', () => {
+    it('should display the f-header component', () => {
+        browser.url('http://localhost:8080');
+
+        HeaderComponent.isLogoDisplayed();
```
kevinrodrigues: There's no `expect` here so this test won't be doing anything?

siggerzz: good spot!

packages/f-registration/test/specs/f-registration.component.spec.js
```diff
@@ -0,0 +1,22 @@
...
+
+        const userInfo = {
+            firstName: '',
+            lastName: '',
+            email: '',
+            password: ''
+        };
+
+        RegistrationComponent.submitRegistrationForm(userInfo);
+        assert.ok(RegistrationComponent.isFirstNameEmptyErrorDisplayed());
```
kevinrodrigues: Should these be using the `expect` then asserting the result? e.g `expect(foo).toBe(true)` Happy with it as is but i feel it's easier to understand when reading it back.

packages/f-registration/test/specs/f-registration.component.spec.js
```diff
@@ -0,0 +1,22 @@
+import RegistrationComponent from '../../test-utils/component-objects/f-registration.component';
+import assert from 'assert';
+
+describe('f-header component tests', () => {
+    it('should display error if first name field is empty', () => {
+        browser.url('http://localhost:8081');
```
kevinrodrigues, line 7: Could you add the AAA comments please?

`// Arrange`

packages/f-registration/test/specs/f-registration.component.spec.js
```diff
@@ -0,0 +1,22 @@
...
+
+        const userInfo = {
+            firstName: '',
+            lastName: '',
+            email: '',
+            password: ''
+        };
+
+        RegistrationComponent.submitRegistrationForm(userInfo);
```
kevinrodrigues, line 17: `// Act`

packages/f-registration/test/specs/f-registration.component.spec.js
kevinrodrigues: `// Assert`

siggerzz: Was just personal preference, but happy to change as it's the syntax that matches the jest tests 👍

kevinrodrigues (approved):

SachaZvetelman (approved):

PR #297, opened 2020-09-02
f-services@1.2.0 - Handle missing navigator API

xander-marjoram:
### Changed
- Handle missing navigator API so that the default axios timeout is returned when used on the server.

---

## UI Review Checks

## Browsers Tested
- [ ] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/f-services/tests/axios.test.js
+++ b/packages/f-services/tests/axios.test.js
@@ -37,9 +37,11 @@ describe('getNetworkDetails', () => {
         expect(connection).toBe(null);
     });
 
-    it('should return navigation.connection if it is defined', () => {
+    it('should return navigator.connection if it is available', () => {
         // Arrange
-        windowSpy.mockImplementation(() => ({ connection: 'CONNECTION' }));
+        windowSpy.mockImplementation(() => ({
+            connection: 'CONNECTION'
+        }));
 
         // Act
         const connection = getNetworkDetails();
@@ -48,7 +50,7 @@ describe('getNetworkDetails', () => {
         expect(connection).toBe('CONNECTION');
     });
 
-    it('should return navigation.mozConnection if it is defined and navigator.connection is not', () => {
+    it('should return navigator.mozConnection if it is available', () => {
         // Arrange
         windowSpy.mockImplementation(() => ({
             mozConnection: 'MOZ-CONNECTION'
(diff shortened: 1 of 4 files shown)


packages/f-services/tests/axios.test.js
```diff
@@ -14,13 +17,86 @@ beforeEach(() => {
...
+        // Act
+        const connection = getNetworkDetails();
+
+        // Assert
+        expect(connection).toBe('MOZ-CONNECTION');
+    });
+
+    it('should return navigation.webkitConnection if it is defined and navigator.connection and navigator.mozConnection are not', () => {
+        // Arrange
+        windowSpy.mockImplementation(() => ({
```
DamianMullins, line 68: I feel like the test descriptions should just mention the type of navigator under test without needing to mention the others? So, for this test, I only care that `webkitConnection` is returned when it is available.

packages/f-services/tests/axios.test.js
```diff
@@ -14,13 +17,86 @@ beforeEach(() => {
...
+        }));
+
+        // Act
+        const connection = getNetworkDetails();
+
+        // Assert
+        expect(connection).toBe('MOZ-CONNECTION');
+    });
+
+    it('should return navigation.mozConnection preferentially over navigator.webkitConnection', () => {
```
DamianMullins: Do we care about the preference? There will never be an instance where more than one is available, and even if there was we should only care that we get a`connection`object back.

packages/f-services/tests/axios.test.js
xander-marjoram: Ah I see, I suppose you would never have more than 1 available at once...

packages/f-services/tests/axios.test.js
```diff
@@ -14,13 +17,86 @@ beforeEach(() => {
     callback = jest.fn();
 });
 
+afterEach(() => {
+    jest.resetAllMocks();
+});
+
+describe('getNetworkDetails', () => {
+    beforeEach(() => {
+        windowSpy = jest.spyOn(window, 'navigator', 'get');
```
kevinrodrigues, line 26: Is this correct? I would have thought it should be as below and only like this if you want to return an object value (getter/setter) from the navigator?

`jest.spyOn(window.navigator, 'connection');`

xander-marjoram: I think they're pretty equivalent. In this case it means I can mock the different connection types and the case where navigator is `undefined` all with one mock.

xander-marjoram: Rather than needing all of the following:
```
jest.spyOn(window.navigator, 'connection', 'get');
jest.spyOn(window.navigator, 'mozConnection', 'get');
jest.spyOn(window.navigator, 'webkitConnection', 'get');
jest.spyOn(window, 'navigator', 'get');
```

kevinrodrigues: I was mostly thinking about just setting the spy on the `navigator` itself and then in each block you can return the mocked value, happy as it is tho.

kevinrodrigues (approved):

ashleynolan (approved):

PR #298, opened 2020-09-04
f-registration@0.18.0 Adding legal text and relevant links

adamashton:
### Added
- Legal text and relevant hyperlinks

### Chrome
image

### Mobile
image

## UI Review Checks

## Browsers Tested

- [X] Chrome (latest)
- [X] Mobile Android

### DIFF
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -152,15 +152,18 @@
             {{ copy.navLinks.termsAndConditions.prefix }}
             <a
                 data-test-id="ts-and-cs-link"
-                :href="copy.navLinks.termsAndConditions.url">{{ copy.navLinks.termsAndConditions.text }}</a>{{ copy.navLinks.termsAndConditions.suffix }}
+                :href="copy.navLinks.termsAndConditions.url"
+                target="_blank">{{ copy.navLinks.termsAndConditions.text }}</a>{{ copy.navLinks.termsAndConditions.suffix }}
             {{ copy.navLinks.privacyPolicy.prefix }}
             <a
                 data-test-id="privacy-policy-link"
-                :href="copy.navLinks.privacyPolicy.url">{{ copy.navLinks.privacyPolicy.text }}</a>
+                :href="copy.navLinks.privacyPolicy.url"
+                target="_blank">{{ copy.navLinks.privacyPolicy.text }}</a>
             {{ copy.navLinks.cookiesPolicy.prefix }}
             <a
                 data-test-id="cookies-policy-link"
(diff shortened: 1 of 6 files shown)


packages/f-registration/src/tenants/en-GB.js
```diff
@@ -1,3 +1,23 @@
...
+        termsAndConditions: {
+            prefix: 'By creating an account you agree to our ',
+            suffix: '.',
+            text: 'Terms and Conditions',
+            url: '/info/terms-and-conditions'
+        },
+        privacyPolicy: {
+            prefix: 'Please read our ',
+            text: 'Privacy Policy',
+            url: '/privacypolicy'
```
simonsandersje: /privacypolicy is a redirect, so better to use /info/privacy-policy

packages/f-registration/src/tenants/en-GB.js
```diff
@@ -1,3 +1,23 @@
...
+        },
+        cookiesPolicy: {
+            prefix: ' and ',
+            text: 'Cookie Policy',
+            url: '/cookies-policy',
```
simonsandersje: Same as above - /info/cookies-policy

simonsandersje (approved):

luke-andrew-parry (approved):

packages/f-registration/src/components/Registration.vue
```diff
@@ -378,4 +395,11 @@ export default {
         text-decoration: none;
     }
 }
+
+.c-legal-hyperlinks {
+    a {
+        color: $blue;
+        text-decoration: none;
```
ashleynolan, line 402: Sorry for the delay in replying to this – we need to raise this with design, as any links in a paragraph of text like this have to be underlined for AA accessibility (as the lack of colour contrast means people with colour blindness won't see these as links otherwise).

I'll flag it with Bruno and then once I've got confirmation will put a PR in the update this potentially.

adamashton: The ticket ACC2-344 shows styling without under lining for links. Please coordinate efforts with Aaron Wilson as there may be other UI designers that need to know this. The work can come in through the accounts team.

ashleynolan: Will do – I've just spoken to Bruno Molena and he confirmed that these should be underlined. I'll speak to Aaron about getting these changes in.

ashleynolan: Just for reference too, the `log in` link should be underlined too (that one looks like it is underlined in the current design, but not in the component currently). Can probably update this at the same time.

PR #305, opened 2020-09-09
fozzie-components@1.30.0 - Publish npm packages as part of CI

siggerzz:


DamianMullins:
@siggerzz will this script try to publish every package?

siggerzz:
@DamianMullins  Currently, yeah. Yet to find an intelligent way to only publish changed packages

DamianMullins:
OK, so I imagine the `npm publish` task will fail, currently, for any packages with versions already published.

siggerzz:
@DamianMullins I usually publish at the root level on my machine and it works fine. From what I can tell, whilst it attempts to publish all packages, it'll skip any that're already up-to-date 👍

DamianMullins:
@siggerzz Nice, I'll have to try that next time I publish a package.

ashleynolan (approved):
LGTM – out of interest, where does it get the auth token? Is that a key set in Circle's config?

siggerzz:
@ashleynolan Yep, it's in the project config within CircleCI. Currently it's using an npm token that's linked to my account, so it'd be good to create a 'CircleCI' account for NPM and add it to org, that way it doesn't look like I'm the only one publishing packages 😁 .

As it's picked up via env var, we can merge this ticket as is though and then change the token after 👍

DamianMullins (approved):

PR #306, opened 2020-09-10
f-registration@0.21.0 - Add registrationSource and marketingPreferences

xander-marjoram:
... to account creation POST body.

### Added
- `registrationSource` and `marketingPreferences` to account creation post request body.

### Changed
- Rename `shouldDisableCreateAccountButton` data variable to `disableCreateAccountButton`.

---

## UI Review Checks

## Browsers Tested
- [ ] Chrome (latest)
- [ ] Internet Explorer 11


packages/f-registration/src/components/Registration.vue
```diff
@@ -144,7 +144,7 @@
                 data-test-id="create-account-submit-button"
                 button-style="primary"
                 is-full-width
-                :disabled="shouldDisableCreateAccountButton">
+                :disabled="disableCreateAccountButton">
```
adamashton: A lot of the other variables/properties are prefixed with `should`. _Should_ we be keeping them consistent?

xander-marjoram: @adamashton I didn't notice this actually, I think this is one of the only packages where we do that but I'm happy to undo it for now.

minrwhite (approved):

tommcclean (approved):

adamashton (approved):

PR #310, opened 2020-09-14
fozzie-components@v1.31.0 – Make registration form WCAG AA.

BanksySan:
---

The registration form needs to support our screen reader using users.
This implementation is not AA, but is a much better experience than previously.  To get the page to AA standard will require us involving the design team as there would be significant considerations for them, as such this is a _best efforts_ to be improved on in the future.

There are minimal visual changes in this implementation, the only one being that only one error will be reported per input field.  This is unlikely to be noticeable to any user, and if it is it's a trivial inconvenience.
---

## UI Review Checks

## Browsers Tested

- [X] Chrome (latest)
- [ ] Internet Explorer 11


packages/f-form-field/src/components/FormField.vue
```diff
@@ -111,6 +116,9 @@ export default {
     methods: {
         updateValue (event) {
             this.$emit('input', event.target.value);
+        },
+        focus () {
```
kevinrodrigues: Is this method being used somewhere?

packages/f-form-field/src/components/FormField.vue
```diff
@@ -70,6 +71,10 @@ export default {
         dataTestId: {
             type: String,
             default: ''
+        },
+        invalidReason: {
+            type: String,
+            default: null
```
kevinrodrigues: Should be `''`

packages/f-registration/src/components/Registration.vue
```diff
@@ -250,33 +236,56 @@ export default {
...
          * focus on the input field.
          */
+        describeFirstnameErrorMessage () {
+            if (this.$v.firstName.$dirty)
+            {
+                if (!this.$v.firstName.required) {
+                    return 'Please include your first name';
+                }
 
-        shouldShowFirstNameRequiredError () {
```
kevinrodrigues, line 302: Personally i prefer this approach better, feels easier to read, did the changes bring any other benefit?

packages/f-form-field/src/components/FormField.vue
BanksySan: It's not being called yet. normally I would delete uncalled methods as they can break but this one seems so trivial I left it.  I shouldn't have though, just dodgy reasoning on my part.

packages/f-registration/src/components/Registration.vue
BanksySan: It was mostly to reduce the number of computed methods, partly because I prefered the look of the new version.

packages/f-form-field/src/components/FormField.vue
```diff
@@ -125,27 +125,27 @@ $form-input-borderColour                  : $grey--light;
 $form-input-borderColour--focus           : $grey--dark;
 
 .c-formField {
-    & + & {
-        margin-top: spacing(x2);
-    }
+  & + & {
```
ashleynolan: Can you revert any of the indentation changes? (should be spacing of 4 – not sure if this is missing from the editorconfig?)

packages/f-registration/src/components/Registration.vue
ashleynolan: I think when we add in locale translations (for AU for example), this new way may get quite messy potentially. I think I'm with Kev on preferring the old style.

I think when the error-message gets pulled ut as a separate component too, the repetition in the view should disappear and tidy it up.

tommcclean: It did bring another benefit; it was easier to consolidate the large number of computeds; not just for readability. 

But because he wanted one element, so that you would have a consistent ID for the accessibility attributes; in the broken down form there was three different elements; and so you would have needed to figure out which one to dynamically assign to the attribute.

There was also an edge case where you could get two errors to show up at once, which is undesirable.

packages/f-registration/src/components/Registration.vue
```diff
@@ -359,7 +369,37 @@ export default {
...
+                    errorCount++;
+                }
+                if (this.$v.firstName.$anyError) {
+                    firstInvalidField = 'firstName';
+                    errorCount++;
+                }
+
+                this.genericErrorMessage = `There are ${errorCount} errors in the form.`;
+            } else {
+                this.genericErrorMessage = null;
```
xander-marjoram: Should this be `''` for consistency with the other error messages?

packages/f-registration/src/components/Registration.vue
xander-marjoram: Nothing to add other than that I prefer the new style, I find it more logical and readable.

packages/f-registration/src/components/Registration.vue
laurenrodney: Does it need to be `null` for the `v-if="genericErrorMessage"` check at the top of the form?

laurenrodney (dismissed):

packages/f-registration/src/components/Registration.vue
xander-marjoram: I think not as `Boolean('')` evalutates to `false`, but I'm not sure if that's categorical proof

laurenrodney: Ok that's what I was wondering

packages/f-registration/src/components/Registration.vue
```diff
@@ -359,7 +369,37 @@ export default {
 
         isFormInvalid () {
             this.$v.$touch();
-            return this.$v.$invalid;
+            const isInvalid = this.$v.$invalid;
+            let firstInvalidField = null;
+
+            if (isInvalid) {
```
ashleynolan: It feels like this should be a separate method rather than inside the `isFormInvalid` method (as it means this method isn't as simple as it probably should be otherwise).

Perhaps having the actual errorCount as a computed property that gets updated when `isFormInvalid` is true would separate this out better?

xander-marjoram: Something like this?
```
errorCount () {
    const {
        firstName, lastName, email, password
    } = this.$v;

    return [
        firstName.$anyError,
        lastName.$anyError,
        email.$anyError,
        password.$anyError
    ]
    .filter(x => x)
    .length;
}
```

packages/f-registration/src/components/Registration.vue
BanksySan: `null` is more correct, as it's the absence of an error message rather that an instance of one of zero length.  That being said, if we have a convention of using `''` then this should conform to that.

packages/f-registration/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.30.0
```
xander-marjoram: Looks like you still need to bump the version in f-registration/package.json

packages/f-registration/test/specs/component/f-registration.component.spec.js
```diff
@@ -23,6 +29,38 @@ describe('f-registration component tests', () => {
         expect(RegistrationComponent.isLastNameEmptyErrorDisplayed()).toBe(true);
         expect(RegistrationComponent.isEmailEmptyErrorDisplayed()).toBe(true);
         expect(RegistrationComponent.isPasswordEmptyErrorDisplayed()).toBe(true);
+        expect(RegistrationComponent.errorSummaryContainerMessageCount()).toBe(1);
+        expect(RegistrationComponent.errorSummaryContainerMessageText()).toBe('There are 4 errors in the form.');
+    });
+
+    it.only('should not display errors if all fields are valid', () => {
```
xander-marjoram: Should this `only` be removed?

BanksySan: Yes.

packages/f-registration/test/specs/component/f-registration.component.spec.js
```diff
@@ -23,6 +29,38 @@ describe('f-registration component tests', () => {
...
+            password: 'dummy-password'
+        };
+
+        expect(RegistrationComponent.errorSummaryContainerRole()).toBe('alert');
+        expect(RegistrationComponent.errorSummaryContainerMessageCount()).toBe(0);
+        expect(RegistrationComponent.errorSummaryContainerDisplayed()).toBe(false);
+
+        // Act
+        RegistrationComponent.submitRegistrationForm(userInfo);
+        // browser.debug();
```
xander-marjoram: Comment to be removed?

packages/f-registration/test-utils/component-objects/f-registration.component.js
```diff
@@ -24,6 +25,8 @@ const emailExistsError = () => $('[data-test-id="error-email-exists"]');
 
 const passwordEmptyError = () => $('[data-test-id="error-password-empty"]');
 
+
```
xander-marjoram: Can't seem to comment on the exact line, but do the email error selectors a few lines above here need updating? How will we apply the correct data-test-id so we know what the error is without reading the error text?

packages/f-registration/src/components/Registration.vue
```diff
@@ -13,139 +13,118 @@
...
-                <template #error>
+                <!-- For when we want to add validation on blur of input - @blur="$v.email.$touch" -->
+                <template
+                    v-if="describeEmailErrorMessage"
+                    #error>
                     <p
-                        v-if="shouldShowEmailRequiredError"
+                        id="error-message-email"
                         :class="$style['o-form-error']"
                         data-test-id='error-email-empty'>
```
xander-marjoram: Sorry I keep noticing little things, should this `data-test-id` be updated because it no longer just applies to empty emails, but to invalid ones too?

packages/f-form-field/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-form-field",
   "description": "Fozzie Form Field – Fozzie Form Field Component",
-  "version": "0.6.0",
+  "version": "0.7.0",
```
ashleynolan: Version needs a bump now I think as other changes have been merged

packages/f-form-field/src/components/FormField.vue
```diff
@@ -126,26 +126,26 @@ $form-input-borderColour--focus           : $grey--dark;
...
 .c-formField {
     & + & {
-        margin-top: spacing(x2);
+      margin-top: spacing(x2);
     }
 }
-    .c-formField-inputWrapper {
-        position: relative;
-    }
+.c-formField-inputWrapper {
```
ashleynolan: Can you keep this indentation? We tend to indent by the class depth, so it's clear the HTML hierarchy it applies to.

packages/f-registration/src/components/Registration.vue
```diff
@@ -396,8 +394,24 @@ export default {
...
+                    v.lastName.$anyError,
+                    v.email.$anyError,
+                    v.password.$anyError
+                ].filter(x => x)
+                .length;
+            }
+
+            v.$touch();
+
+            if (v.$invalid) {
```
ashleynolan, line 454: Feel like this should be in a `watch` method, that reacts to this value updating, rather than as part of the `isFormInvalid` method. Can keep this method small then and abstract the rest out to that watch method.

packages/f-registration/package.json
```diff
@@ -47,7 +47,7 @@
     "@justeat/f-card": "0.7.0",
     "@justeat/f-form-field": "0.7.1",
     "@justeat/f-services": "1.0.0",
-    "@justeat/f-vue-icons": "1.3.0",
+    "@justeat/f-vue-icons": "^1.4.0",
```
BanksySan: Did you mean to use a `^` here?

ross-gratton: Nope I think this came in automatically, I'll remove.

ashleynolan:
Upgrade note – I've had to clean up some of our git history. If you could rebase master onto your branch and push back up to this PR, that'd be appreciated 🙌

xander-marjoram:
Hey @BanksySan, what are your plans for this PR?

ashleynolan:
Closing this PR, as it'll need to be re-worked off master to get merged as it's v out-of-date with master. I'd suggest copy/pasting to a new branch if we want to make these changes in current master.

PR #313, opened 2020-09-15
f-registration@0.23.0 Include Registration API service as part of npm f-registration

siggerzz:
**Description**
- We need to move the consumer contract tests from fozzie-components to Magikarp because of potential security concerns exposing our endpoints / request body within these tests.

Before we can do this, we need to make the RegistrationApi service available in the f-registration package, so that when we move the tests, we're able to invoke the call.


siggerzz:
An alternative to this would be to do the following inside Magikarp:

- Use jest to mount f-registration component > invoke the method that calls the service.

Whilst I prefer the approach in this PR, I'd be interested to know if there are any major downsides of my approach, or if the one mentioned is preferable?

SachaZvetelman (approved):

tommcclean (approved):

PR #321, opened 2020-09-17
V0.4.1: Add tests for f-card computed properties

FayeCarter:
Tests added for f-card computed properties.

### DIFF
--- a/packages/f-card/CHANGELOG.md
+++ b/packages/f-card/CHANGELOG.md
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.4.1
+------------------------------
+*September 17 2020*
+
+### Added
+- Tests for f-card computed properties.
+
 
 v0.4.0
 ------------------------------
--- a/packages/f-card/CHANGELOG.md
+++ b/packages/f-card/CHANGELOG.md
@@ -3,12 +3,12 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-v0.4.1
+v0.5.0
 ------------------------------
 *September 17 2020*
 
 ### Added
-- Tests for f-card computed properties.
+- Tests for `f-card` computed properties.
 
 
 v0.4.0
--- a/packages/f-card/CHANGELOG.md
+++ b/packages/f-card/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.5.0
 ------------------------------
-*September 17 2020*
+*September 18 2020*
 
 ### Added
 - Tests for `f-card` computed properties.
--- a/packages/f-card/package.json
+++ b/packages/f-card/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-card",
   "description": "Fozzie Card Component – Used for providing wrapper card styling to an element (or group of elements)",
-  "version": "0.4.1",
+  "version": "0.5.0",
   "main": "dist/f-card.umd.min.js",
   "files": [
     "dist"
--- a/packages/f-card/package.json
+++ b/packages/f-card/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-card",
   "description": "Fozzie Card Component – Used for providing wrapper card styling to an element (or group of elements)",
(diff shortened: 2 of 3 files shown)


packages/f-card/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.4.1
+------------------------------
+*September 17 2020*
+
+### Added
+- Tests for f-card computed properties.
```
SachaZvetelman: Nit: let's write `f-card` (wrap in "`").

packages/f-card/src/components/tests/Card.test.js
```diff
@@ -1,6 +1,18 @@
 import { shallowMount } from '@vue/test-utils';
 import Card from '../Card.vue';
 
+jest.mock('@justeat/f-services');
```
SachaZvetelman, line 5: Not sure if it works, but if it does, can we move the `import` on top of this `jest.mock` line so we keep all the imports together?

packages/f-card/src/components/tests/Card.test.js
```diff
@@ -162,4 +174,96 @@ describe('Card', () => {
             });
         });
     });
+
+    describe('computed', () => {
+        let mockedLocale;
+        let propsData;
+        let mockedTennats;
```
SachaZvetelman: Typo: `mockedTenants`.

packages/f-card/src/components/tests/Card.test.js
```diff
@@ -162,4 +174,96 @@ describe('Card', () => {
...
+        });
+
+        afterEach(() => {
+            jest.resetAllMocks();
+        });
+
+        describe('cardLocale', () => {
+            it('should call the `getLocale` method from `sharedServices` and return correct locale', () => {
+                // Arrange
+                const mockedI18n = () => 'en-AU'
```
SachaZvetelman: Probably worth extracting `en-AU` to a `const` since we're using it everywhere in the file.

packages/f-card/src/components/tests/Card.test.js
```diff
@@ -162,4 +174,96 @@ describe('Card', () => {
             });
         });
     });
+
+    describe('computed', () => {
+        let mockedLocale;
```
siggerzz: Are we able to use `const` for some of these?

SachaZvetelman: Personal preference, but I'd actually only prefix with `mocked` those that are actually mocks. In this case, unless I'm wrong, `mockedLocale` is just a hardcoded locale you're using (`en-AU`).

SachaZvetelman (commented):
Good first PR, @FayeCarter, nice job 👏

packages/f-card/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-card",
   "description": "Fozzie Card Component – Used for providing wrapper card styling to an element (or group of elements)",
-  "version": "0.4.0",
+  "version": "0.4.1",
```
siggerzz: I _think_ we only do patches (incrementing 3rd digit) for bug fixes?

I usually do a minor release for tests, but perhaps @ashleynolan can confirm 👍

packages/f-card/src/components/tests/Card.test.js
```diff
@@ -162,4 +174,96 @@ describe('Card', () => {
...
+        let propsData;
+        let mockedTennats;
+        let mockedCardLocale;
+
+        beforeEach(() => {
+            mockedLocale = 'en-AU';   
+            propsData = {
+                locale: mockedLocale
+            };
+            mockedTennats = {
```
siggerzz: Change to `mockedTenants`

siggerzz (changes requested):
Great job! 🎉 

Just a few minor comments, but looks fine otherwise!

packages/f-card/package.json
DamianMullins: Patch versions are for backwards compatible bug fixes. This one should be a minor version bump as you say @siggerzz.

DamianMullins: https://semver.org

DamianMullins: Saying that, if the changes here are only for the tests then maybe it should be a minor version bump as nothing will have changed for the end-user.

packages/f-card/src/components/tests/Card.test.js
FayeCarter: Hi, I've moved all of the variables in the beforeEach to consts. Is that what you meant?

siggerzz (approved):

BanksySan (approved):

PR #324, opened 2020-09-22
f-content-cards@2.1.0: Add display logic

asherstoppard:
Adds time based checks, brand checks and type checks against provided filters when displaying content cards.

## UI Review Checks

- **N/A** This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist)


packages/f-content-cards/package.json
```diff
@@ -38,6 +38,7 @@
   "dependencies": {
     "@justeat/f-services": "1.0.1",
     "core-js": "3.6.5",
+    "date-fns": "2.16.1",
```
minrwhite: Pretty sure this should be in devDependencies, as this dependency should be compiled into the resulting artifact on build.

packages/f-content-cards/src/services/_tests/contentCard.service.test.js
```diff
@@ -383,5 +383,29 @@ describe('`contentCardService`', () => {
...
+            ];
+            const options = {
+                enabledCardTypes: ['Home_Promotion_Card_1', 'Home_Promotion_Card_2']
+            };
+            const { cards: output } = new ContentCardService({ cards }, options)
+                .applyCardLimit(1)
+                .output();
+
+            expect(output).toHaveLength(1);
+            expect(output[0].type).toBe('Home_Promotion_Card_2');
```
minrwhite: This expectation is failing in the tests.

minrwhite (commented):
A few comments - plus did we have the requirement that we were ordering preference of cards based on the given `enabledCardTypes` array?

packages/f-content-cards/src/services/utils/isCardCurrentlyActive.js
```diff
@@ -0,0 +1,32 @@
+import parse from 'date-fns/parse';
```
DamianMullins: Is there any benefit of importing these functions separately rather than via named imports? Do named exports not tree-shake?

packages/f-content-cards/package.json
DamianMullins: Agree. Also, bundling this into the package will [add ~18kb (gzipped)](https://bundlephobia.com/result?p=date-fns@2.16.1) to final size, is that acceptable?

asherstoppard: Makes sense, I'll get it switched over :+1:

packages/f-content-cards/src/services/utils/isCardCurrentlyActive.js
asherstoppard: No you're correct, named imports should tree-shake. I'll run a test just to be sure

minrwhite:
Happy that the specific points raised were addressed - but please could you answer:

> did we have the requirement that we were ordering preference of cards based on the given `enabledCardTypes` array?

If so is that executed in this PR, or is that for a follow-up change?

asherstoppard:
@benwhite-deltas We did agree that but this PR only introduces that change if the content card limit is equal to 1 to give some breathing room in terms of looking into performance. I'll get another ticket raised in the BAU backlog to introduce a better pattern :+1:

minrwhite (approved):

kevinrodrigues (approved):

PR #325, opened 2020-09-22
f-card@v0.6.0 - Apply appropriate margins when isPageContentWrapper

tommcclean:
- A small non functional UI change which adds margin to the top and bottom of the F-Card when 'isPageContentWrapper' is true.
- This enables the component to not sit flush with the F-Header on the new Registration Page
---

## UI Review Checks

F-Card change
image

Which enables
image

Which enables
image

## Browsers Tested

- [x] Chrome (latest)
- [x] Internet Explorer 11
- [x] Chrome - Emulated Mobile and Tablet

### DIFF
--- a/packages/f-card/src/components/Card.vue
+++ b/packages/f-card/src/components/Card.vue
@@ -107,6 +107,7 @@ $card--pageContentWrapper-width           : 460px;
         transition: 250ms padding ease-in-out;
         padding-left: 6%;
         padding-right: 6%;
+        margin: spacing() * 5 0;
 
         @include media('>=narrow') {
             padding-left: 10%;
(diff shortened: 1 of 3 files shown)


packages/f-card/src/components/Card.vue
```diff
@@ -107,6 +107,7 @@ $card--pageContentWrapper-width           : 460px;
         transition: 250ms padding ease-in-out;
         padding-left: 6%;
         padding-right: 6%;
+        margin: spacing() * 5 0;
```
minrwhite: I think 
```suggestion
        margin: spacing(x5) 0;
```
would be more idiomatic?

tommcclean: Will give it a go, just copied the syntax in the rest of the file :D

minrwhite (approved):

laurenrodney (approved):

PR #327, opened 2020-09-23
f-registration@0.26.2 - Revert syntax in registration service back to ES6

siggerzz:
Reverting changes in v0.26.1

### DIFF
--- a/packages/f-registration/src/services/RegistrationServiceApi.js
+++ b/packages/f-registration/src/services/RegistrationServiceApi.js
@@ -1,15 +1,16 @@
-const axios = require('axios');
+import axios from 'axios';
 
-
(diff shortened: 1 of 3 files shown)


tommcclean (approved):

packages/f-registration/src/services/RegistrationServiceApi.js
```diff
@@ -1,15 +1,16 @@
-const axios = require('axios');
+import axios from 'axios';
 
-
-exports.createAccount = async (url, tenant, data) => {
-    const config = {
-        method: 'post',
```
kevinrodrigues, line 6: Not your change but do we need to set a post if we're using the axios post aliases here?

ashleynolan (approved):

PR #312, opened 2020-09-15
f-metadata@3.0.0-beta.1 〰 Begins decoupling of f-content-cards

minrwhite:
### Changed
- Begins the decoupling of f-metadata from f-content-cards, allowing callbacks for multiple consuming components to be registered
- Ports content cards service modue from f-content-cards to perform the card data model manipulation before cards are reported back to consuming components

---

## UI Review Checks

- [x] README and/or UI Documentation has been updated
- [x] Unit tests have been created & updated
- [ ] ~~This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist)~~ (Not relevant as no user-facing UI updated)

## Browsers Tested

- [ ] Chrome (latest)
- [ ] Internet Explorer 11
- [ ] Mobile (Please list device/browser – Ideally one iPhone model and one Android)

### List any other browsers that this PR has been tested in:

- [View the Browser Support Checklist](http://fozzie.just-eat.com/documentation/general/browser-support)

PR #293, opened 2020-08-26
fozzie-components@v2.0.0 – JustEatBasis typography updates

ashleynolan:
PR for v2.0.0 of fozzie components – updating fonts to new JustEatBasis.

This PR should only be merged once we have a release date confirmed for the font changes – in the meantime though, the components changes will be released as beta versions so that the new versions of the header/footer etc can be released and pulled in to the vertical codebases. These will then move to full non-beta releases once we roll this out and don't need to support the old versions.

---

## UI Review Checks

- [x] README and/or UI Documentation has been updated
- [x] This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist)

## Browsers Tested

- [x] Chrome (latest)
- [x] Internet Explorer 11
- [x] Mobile (Please list device/browser – Ideally one iPhone model and one Android)

PR #329, opened 2020-09-23
f-content-cards@v2.1.0-beta.1

minrwhite:
The matching piece to complete the work from PR #312

Merging to a separate branch than master so as to keep master clean until we're certain this is working in the way we want to, while still being able to publish for consumption by other parties.

v2.1.0-beta.1
*October 5, 2020*

### Added
- `@custom-cards-callback` event dispatched on mount to expose a callback for consuming code to inject
  custom content cards

### Removed
- Card filtering/transforming/sorting logic moved to `f-metadata`; only the absolute limit number remains

### Changed
- Bumped dependency on f-metadata to version `3.0.0-beta.1`

---

## UI Review Checks


ashleynolan:
I'm just rebasing [this PR](https://github.com/justeat/fozzie-components/pull/293) so that we can get all the new components out of beta. If you give me 30 mins, should be able to get that reviewed so that you can just release this as a standard release after that?

minrwhite:
@ashleynolan I'd still like to release this as a beta because of the changes it involves, but happy to do that on top of the typography changes once that's merged. Will mark this as do not merge, but still fish for comments/reviews for now.

asherstoppard (approved):

kevinrodrigues (approved):

packages/f-content-cards/CHANGELOG.md
```diff
@@ -4,6 +4,21 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v2.1.0-beta.1
```
ashleynolan, line 7: Might be worth noting why this is a beta release – is it the link to `f-metadata` that's in beta?

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -116,11 +118,16 @@ export default {
             copy: { ...localeConfig },
             titleCard: {},
             hasLoaded: false,
-            loadingCard
+            loadingCard,
+            brazeDispatcher: null
```
ashleynolan, line 122: Only thing I'd be a bit wary of is naming these with respect to Braze – it makes sense now, but if we were ever to move away from Braze to another service in future, it would mean renaming all of these variables (just look at `appboy` 😆).

Not a crucial thing, but if there's a more generic name that makes sense for these I'd rename them to that instead.

packages/f-content-cards/CHANGELOG.md
minrwhite: Essentially yes - the semi decoupling between f-metadata and f-content-cards - hence why they're both `-beta` suffixed, for now. I'll add a message in.

packages/f-content-cards/src/components/ContentCards.vue
minrwhite: That's fair - I could name it with respect to `metadata` as that's the module this is intended to adapt - the whole intent of the decoupling piece is so that f-content-cards can consume cards from other sources - initially just custom but maybe adding more marketing sources down the road. Will change references from `braze` to `metadata` in an upcoming PR if you're happy with that?

ashleynolan: Yeh, makes sense. Happy for you to call it whatever you think makes the most sense 👍

PR #334, opened 2020-09-25
f-form-field@0.7.0: Add `inlineNarrow` label-style

FayeCarter:
v0.7.0
*September 29, 2020*

### Added
- `inlineNarrow` label style that changes label positioning based on window size.
- `form-field` tests for `inlineNarrow` style.

### Changed
- `babel.config.js` to match other packages as it was causing `ReferenceError: regeneratorRuntime is not defined`.

## UI Review Checks


packages/f-form-field/src/components/tests/FormField.test.js
```diff
@@ -149,17 +148,17 @@ describe('FormField', () => {
...
-                it('should append the label inline with input when window size is mobile', async () => {
+                it('should append the label inline with input when window size is mobile', () => {
                     // Arrange
                     const propsData = {
                         labelStyle: 'inlineNarrow',
                         labelText: 'Test Label'
                     };
 
-                    resizeWindow(750);
+                    resizeWindow(767);
```
SachaZvetelman: Is it worth extracting this `767` to indicate exactly what it is?

FayeCarter: 👍

simonsandersje (approved):

packages/f-form-field/src/components/FormField.vue
```diff
@@ -36,81 +40,128 @@
...
         uniqueId () {
             return `formField-${(this.$attrs.name ? this.$attrs.name : this._uid)}`;
         },
+
         testId () {
             return this.dataTestId || this.$attrs.name || false;
+        },
+
+        isInline () {
+            if (typeof (window) !== 'undefined' && this.windowWidth < MOBILE_WIDTH && this.labelStyle === 'inlineNarrow') {
```
tommcclean: This `typeof(window)...` is not needed if you are using `this.windowWidth` as window is not called in this computed

packages/f-form-field/src/components/FormField.vue
```diff
@@ -36,81 +40,128 @@
...
+            } else if (this.labelStyle === 'inline') {
+                return true;
+            }
+
+            return false;
         }
     },
+
+    mounted () {
+        window.addEventListener('resize', this.updateWidth);
```
kevinrodrigues: I would also destroy the event listener in a `destroyed` lifecycle hook.

packages/f-form-field/src/components/FormField.vue
```diff
@@ -36,81 +40,128 @@
...
+        this.updateWidth();
+    },
+
     methods: {
         updateValue (event) {
             this.$emit('input', event.target.value);
+        },
+
+        updateWidth () {
```
kevinrodrigues, line 163: This one should be `debounced` to avoid multiple calls when the resize event is triggered. Something like below.

```javascript
updateWidth: debounce(function resize () {
   this.windowWidth = window.innerWidth;
}, 100),

File: packages/f-form-field/src/components/FormField.vue [hunk 5]
```diff
@@ -36,81 +40,128 @@
...
+            this.windowWidth = window.innerWidth;
```
kevinrodrigues: Do you need to verify that window exists?

packages/f-form-field/src/components/tests/FormField.test.js
```diff
@@ -107,5 +106,73 @@ describe('FormField', () => {
...
+                // Act
+                const wrapper = shallowMount(FormField, { propsData });
+                const formLabel = wrapper.find('form-label-stub');
+
+                // Assert
+                expect(formLabel.attributes('labelstyle')).toBe(definedType);
+            });
+
+            describe('when set to `inlineNarrow`', () => {
+                const eventName = 'resize';
```
kevinrodrigues: I think this line down to line 134 can all go in a `beforeEach`?

packages/f-form-field/src/components/FormField.vue
```diff
@@ -36,81 +40,128 @@
...
+
         dataTestId: {
             type: String,
             default: ''
         }
     },
+
+    data () {
+        return {
+            windowWidth: ''
```
kevinrodrigues: Might be best left as `null` instead of a empty string. The value would normally be undefined or numeric.

packages/f-form-field/src/components/FormField.vue
SachaZvetelman: We could simplify this to:
```
isInline () {
    return (this.windowWidth < MOBILE_WIDTH && this.labelStyle === 'inlineNarrow') ||
        this.labelStyle === 'inline';
}
```

packages/f-form-field/src/components/tests/FormField.test.js
```diff
@@ -107,5 +106,73 @@ describe('FormField', () => {
                 expect(formInput.attributes('data-test-id')).toBe(undefined);
             });
         });
+
+        describe('labelStyle ::', () => {
+            it.each(VALID_LABEL_STYLES)('should set the type of form label element as expected', definedType => {
```
SachaZvetelman: You can include the name of each label in the test name so you don't get the 3 tests with the same name. If you do `%s` within the test name, it will replace it with each element of `VALID_LABEL_STYLES` 😄

packages/f-form-field/src/components/FormField.vue
FayeCarter: 👍

packages/f-form-field/src/components/FormField.vue
FayeCarter: Thanks, I've changed this

packages/f-form-field/src/components/FormField.vue
FayeCarter: 👍

packages/f-form-field/src/components/FormField.vue
FayeCarter: I've added this in

packages/f-form-field/src/components/tests/FormField.test.js
FayeCarter: Oooh, that is cool 👍

packages/f-form-field/src/components/tests/FormField.test.js
```diff
@@ -1,7 +1,7 @@
 import { shallowMount } from '@vue/test-utils';
 import FormField from '../FormField.vue';
-import { DEFAULT_INPUT_TYPE, VALID_INPUT_TYPES } from '../../constants';
-
+import { DEFAULT_INPUT_TYPE, VALID_INPUT_TYPES, VALID_LABEL_STYLES } from '../../constants';
+import _ from 'lodash';
```
SachaZvetelman: I might be blind, but where are we using `lodash` in this file? :D

FayeCarter: This is where I was trying to bring in debounce so `updateWidth` isn't called as frequently. It didn't work, I was meant to take it out🤦‍♀️

tommcclean: I use a debounce in my personal website that doesn't use Lodash (Cos lodash constantly throws up security vulns): https://github.com/tommcclean/tommcclean.me/blob/master/components/controllers/slideWrapper.vue#L22

FayeCarter: Stolen, thanks!

packages/f-form-field/src/components/tests/FormField.test.js
FayeCarter: 👍

kevinrodrigues (approved):

packages/f-form-field/src/components/tests/FormField.test.js
```diff
@@ -107,5 +106,80 @@ describe('FormField', () => {
...
+                const formLabel = wrapper.find('form-label-stub');
+
+                // Assert
+                expect(formLabel.attributes('labelstyle')).toBe(definedType);
+            });
+
+            describe('when set to `inlineNarrow`', () => {
+                let eventName;
+                let resizeWindow;
+                let MOBILE;
```
SachaZvetelman, line 128: This can be a `const` that it set here. No need to include this in the `beforeEach`. Same with most of these variables.

packages/f-form-field/src/services/debounce.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function debounce (fn, delay) {
+    let timeoutID = null;
```
SachaZvetelman, line 2: Is there any reason to declare this variable here?
Nit: should be `timeoutId`, I think.

packages/f-form-field/src/components/FormField.vue
```diff
@@ -35,83 +39,132 @@
...
 import FormLabel from './FormLabel.vue';
+import Debounce from '../services/debounce';
 import tenantConfigs from '../tenants';
-import { VALID_INPUT_TYPES, DEFAULT_INPUT_TYPE, VALID_LABEL_STYLES } from '../constants';
+import {
+    VALID_INPUT_TYPES,
+    DEFAULT_INPUT_TYPE,
+    VALID_LABEL_STYLES,
+    MOBILE_WIDTH
+} from '../constants';
```
BanksySan, line 49: Is this styling change inline with the conventions?

packages/f-form-field/src/components/FormField.vue
```diff
@@ -35,83 +39,132 @@
...
         locale: {
             type: String,
             default: ''
         },
+
         labelText: {
             type: String,
             default: ''
         },
+
```
BanksySan, line 70: Is the addition of new lines the convention?

PR #335, opened 2020-09-28
Minor - Don't use DIST for story file

tommcclean:
Storybook falls over because the vue-icons package uses Dist to import its components. This means it won't work for anybody until they CD into Vue icons and build it; which is undesirable.

This PR corrects the story to load the components from the src directory.

Not versioned because I don't intend to publish vue-icons.

### DIFF
--- a/packages/f-vue-icons/stories/Icons.stories.js
+++ b/packages/f-vue-icons/stories/Icons.stories.js
@@ -1,6 +1,6 @@
 import { withA11y } from '@storybook/addon-a11y';
 
-import * as IconsComponents from '../dist/f-vue-icons.es';
+import * as IconsComponents from '../src/index';
 import './icons.css';
 
 const icons = {};


DamianMullins (approved):

kevinrodrigues (approved):

kevinrodrigues:
There's a entrypoint limit warning in the build is this something that has always appeared?

tommcclean:
> There's a entrypoint limit warning in the build is this something that has always appeared?

Do you mean the Percy stuff? I just messaged Ben S about that as I see it failing on all the newest PR's - we ran out of Percy credit I think.

*Edit: I see it too now, yeah seems to be happening to all the latest PR's too

kevinrodrigues:
> > There's a entrypoint limit warning in the build is this something that has always appeared?
> 
> Do you mean the Percy stuff? I just messaged Ben S about that as I see it failing on all the newest PR's - we ran out of Percy credit I think.
> 
> *Edit: I see it too now, yeah seems to be happening to all the latest PR's too

Yeah looks percy related.

kevinrodrigues:
> > There's a entrypoint limit warning in the build is this something that has always appeared?
> 
> Do you mean the Percy stuff? I just messaged Ben S about that as I see it failing on all the newest PR's - we ran out of Percy credit I think.
> 
> *Edit: I see it too now, yeah seems to be happening to all the latest PR's too

But also there's two warnings after the icons are compiled. So not sure if it's related or just something that has always happened. Worth checking if the change has increased the bundle at all incase.

Screen Shot 2020-09-28 at 11 25 42 AM

tommcclean:
> > > There's a entrypoint limit warning in the build is this something that has always appeared?
> > 
> > 
> > Do you mean the Percy stuff? I just messaged Ben S about that as I see it failing on all the newest PR's - we ran out of Percy credit I think.
> > *Edit: I see it too now, yeah seems to be happening to all the latest PR's too
> 
> But also there's two warnings after the icons are compiled. So not sure if it's related or just something that has always happened. Worth checking if the change has increased the bundle at all incase.
> 
> Screen Shot 2020-09-28 at 11 25 42 AM

Yeah that warning pops up in other PR's in the list also, I think theres something underlying we need to fix; not introduced by this though. I would "guess" it is something to do with Vue CLI. the story file doesn't contribute to the build as it isn't output.

PR #338, opened 2020-09-28
f-header@4.1.0 - Added property to show/hide the Help link and improved hamburger icon behaviour.

SachaZvetelman:
v4.1.0
*September 28, 2020*

### Added
- Property to show/hide the Help link.

### Changed
- Hamburger icon. Hide it if there are no visible links to be shown.
- Storybook property from free text to boolean for "Show login/user info".

## UI Review Checks

**Without links:**
image

**With links:**
image

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/f-header/src/components/Navigation.vue
+++ b/packages/f-header/src/components/Navigation.vue
@@ -177,9 +177,10 @@
                     </a>
                 </li>
 
-                <li class="c-nav-list-item c-nav-list-item--support">
+                <li
+                    v-if="showHelpLink"
+                    class="c-nav-list-item c-nav-list-item--support">
                     <a
-                        v-if="showHelpLink"
                         :href="help.url"
                         :data-trak='`{
                                 "trakEvent": "click",
--- a/packages/f-header/src/components/Navigation.vue
+++ b/packages/f-header/src/components/Navigation.vue
@@ -379,8 +379,7 @@ export default {
             return this.showOffersLink ||
                 this.showHelpLink ||
                 this.showDeliveryEnquiry ||
-                this.showLoginInfo ||
-                this.userInfo;
+                this.showLoginInfo;
         }
     },
 
--- a/packages/f-header/src/components/Navigation.vue
+++ b/packages/f-header/src/components/Navigation.vue
@@ -1,5 +1,8 @@
 <template>
-    <nav class="c-nav c-nav--global">
+    <nav
+        v-if="hasNavigationLinks"
+        class="c-nav c-nav--global"
+        data-js-test="nav-container">
         <button
             :class="['c-nav-trigger c-nav-toggle is-hidden--noJS', navToggleThemeClass, {
                 'is-open': navIsOpen
@@ -370,6 +373,14 @@ export default {
          */
         getAnalyticsBlob () {
             return window.localStorage.getItem('je-analytics') || false;
+        },
+
+        hasNavigationLinks () {
+            return this.showOffersLink ||
+                this.showHelpLink ||
+                this.showDeliveryEnquiry ||
+                this.showLoginInfo ||
+                this.userInfo;
         }
     },
 
--- a/packages/f-header/src/components/Navigation.vue
+++ b/packages/f-header/src/components/Navigation.vue
@@ -176,6 +176,7 @@
 
                 <li class="c-nav-list-item c-nav-list-item--support">
                     <a
+                        v-if="showHelpLink"
                         :href="help.url"
                         :data-trak='`{
                                 "trakEvent": "click",
@@ -273,6 +274,11 @@ export default {
         showOffersLink: {
             type: Boolean,
             default: false
+        },
+
+        showHelpLink: {
+            type: Boolean,
+            default: true
         },
 
         showLoginInfo: {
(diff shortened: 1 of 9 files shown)


packages/f-header/src/components/Navigation.vue
```diff
@@ -176,6 +179,7 @@
 
                 <li class="c-nav-list-item c-nav-list-item--support">
                     <a
+                        v-if="showHelpLink"
```
kevinrodrigues: Looks like if this is `false` will you be left with a empty `li` element?

packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -188,6 +190,45 @@ describe('Navigation', () => {
...
+            propsData,
+            computed: {
+                hasNavigationLinks () {
+                    return true;
+                }
+            }
+        });
+
+        // Assert
+        expect(wrapper.find('[data-js-test="nav-container"]').exists()).toBe(true);
```
BanksySan, line 210: Do we need to check that there's exactly one instance, rather than at least one?

packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -571,4 +665,90 @@ describe('Navigation', () => {
             expect(wrapper.find('[data-js-test="user-info-icon"]').classes()).toContain('is-hidden');
         });
     });
+
+
+    describe('hasNavigationLinks', () => {
+        it('should return true if `showOffersLink` is true', async () => {
```
kevinrodrigues: For these few tests you could just combine it like how you did the last test? i.e set all the `propsData` that are required to `true` and pass that into one single `IT` block instead of 4 separate ones? Happy as it is now tho just a thought.

BanksySan (changes requested):
In `f-registration` we have a layer that hides the logic of getting a tested property from the test method.
e.g. `packages/f-registration/test-utils/component-objects/f-registration.component.js`
Do we want these to be in a similar pattern?

packages/f-header/src/components/tests/Navigation.test.js
SachaZvetelman: Your suggestion would be ideal for an `AND` scenario, but in this case, we have multiple `ORs`. If any of those props is `true`, `hasNavigationLinks` should return `true`. Having one test with the four props set to `true` wouldn't be a very meaningful test, if it makes sense? But let me know if I misunderstood your suggestion!

packages/f-header/src/components/Navigation.vue
SachaZvetelman: Good catch! Thanks.

packages/f-header/src/components/tests/Navigation.test.js
SachaZvetelman: I personally would leave it as it is since it's consistent with the rest of the tests in `f-header`, and counting we have exactly one element won't add a lot of value. In other cases, it would. I'm happy to change this is more people agree with you, for sure.

SachaZvetelman:
@BanksySan, we normally use the POM (Page Object Model) pattern for system/feature/e2e tests rather than unit tests. I think it would be an overkill in this scenario, since it's a unit test, and it's consistent with what we're already doing.

packages/f-header/src/components/tests/Navigation.test.js
kevinrodrigues: It was mostly an observation with the last test against these individual ones. I was thinking you can set them all to false (like your last test)  and then a second test with one as true? Because the computed prop only really cares about the OR statement part of the computed property returning true if one individual prop is truthy and will therefore show the whole navigation if truthy, so we're not too concerned which one passes here. Hence the `hasNavigationLinks` always being asserted in each test if that makes sense? (Just a thought so happy to approve as is)

SachaZvetelman: Ah, I see what you mean. However, if we split the tests (as they are now), we get the benefit of one of them failing if someone removes one of the conditions from the computed property (which I foresee happening, haha). Does it make sense? But I get your point.

xander-marjoram: I think you could pass in the name of each property and do something like:
```javascript
it.each([
    'showOffersLink',
    'showHelpLink',
    'showDeliveryEnquiry',
    'showLoginLink'
])('should return true if `%s` is true', async navLink => {
    // Arrange
    const propsData = {
        ...defaultPropsData,
        [navLink]: true
    };
   
    // Rest of test continues...
});

xander-marjoram: (Also, do these tests need to be `async`?)

SachaZvetelman: @xander-marjoram, yup, I originally did that! But then I remembered some people not being huge fans of that syntax here. I'm more than happy to do it that way if people are ok! Maybe that was a while ago, hehe 👍

BanksySan (approved):

packages/f-header/src/components/tests/Navigation.test.js
SachaZvetelman: Nope, only one of them needed to be `async`. Good catch. I copy-pasted, hehe.

xander-marjoram: @SachaZvetelman Personally I think it's good to use `.each` as long as it doesn't make it hard to read the test, and obviously the more it reduces duplication the better

xander-marjoram (approved):
Looks good to me. I do think we need to address the number of props at some point, it's beginning to get a bit bloated.

PR #339, opened 2020-09-28
f-vue-icons@1.3.0

luke-andrew-parry:
This PR updates the f-icons package and includes the bag celebrate icon.

image


packages/f-vue-icons/package.json
```diff
@@ -48,10 +48,11 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
+    "@justeat/f-icons": "2.2.0",
     "babel-helper-vue-jsx-merge-props": "2.0.3"
   },
   "devDependencies": {
-    "@justeat/f-icons": "2.1.0",
+    "@justeat/f-icons": "2.2.0",
```
adamashton, line 55: Is `@justeat/f-icons` needed in devDependencies any more?

DamianMullins: Yes, it's where this package gets the icons to convert into Vue components.

DamianMullins (approved):

packages/f-vue-icons/package.json
DamianMullins: I'll make a note to update the readme to include a bit more detail on how this package is architected and the process to add and updated icons.

adamashton: I meant if that it's specified in dependencies is it needed in devDependencies? I didn't think it was but I may be wrong.

kevinrodrigues (approved):

packages/f-vue-icons/package.json
DamianMullins: Ah good catch, no it shouldn't be in `dependencies` as well, only in `devDependencies`.

packages/f-vue-icons/package.json
```diff
@@ -48,10 +48,11 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
+    "@justeat/f-icons": "2.2.0",
```
DamianMullins, line 51: @luke-andrew-parry Was there any reason why this was added as a `dependency`? It shouldn't be required and means that consumers of this package will download `f-icons` unnecessarily. 

If it was in error, are you able to create a PR to revert this line?

PR #342, opened 2020-09-29
f-footer@v4.1.0 - Added F-Footer Webdriver Tests

LTurns:
## Added 
- data-test IDs to Footer Component 
- Test for Footer Icons 
- F-Footer Component-Object 

### Changed
- Npm scripts

### DIFF
--- a/packages/f-footer/src/components/IconList.vue
+++ b/packages/f-footer/src/components/IconList.vue
@@ -26,7 +26,8 @@
                         "category": "engagement",
                         "action": "footer",
                         "label": "${icon.gtm}"
-                    }`'>
+                    }`'
+                    data-test-id="footerIcon">
                     <component
                         :is="iconChoice"
                         v-bind="icon"
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "@vue/test-utils": "1.0.3",
     "@wdio/cli": "6.4.0",
     "@wdio/local-runner": "6.4.0",
-    "@wdio/mocha-framework": "6.4.0",
+    "@wdio/mocha-framework": "^6.4.0",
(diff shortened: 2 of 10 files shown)


siggerzz:
Build is failing due to an unrelated issue with the visual regression tests (that should have been removed by me in an earlier PR! 😕 )

siggerzz (approved):

xander-marjoram (approved):

packages/f-footer/src/components/IconList.vue
```diff
@@ -26,7 +26,8 @@
                         "category": "engagement",
                         "action": "footer",
                         "label": "${icon.gtm}"
-                    }`'>
+                    }`'
+                    data-test-id="footerIcon">
```
minrwhite, line 30: You should have access to the [`name`](https://github.com/justeat/fozzie-components/blob/master/packages/f-footer/src/tenants/en-GB.js#L216) key of the socialicons/appstore icons array here, I'd prefer if this was used here similar to:
```suggestion
                    :data-test-id="`footerIcon-${icon.name}`">
```
So that when we come to refer to the icons in `test-utils/component-objects/f-footer.component.js`, we're not reliant on any particular order of the icons.

minrwhite (commented):
One suggestion to prevent reliance on order of discovered elements by the testing framework - happy if you want to leave it as it is for now though.

packages/f-footer/src/components/IconList.vue
siggerzz: Ahhh yes, great shout! I'll pick this up with Lizzie next week after the hackathon. For now though, I think we're good to merge :)

PR #343, opened 2020-09-29
f-form-field@0.7.1: Fix debounce bug

FayeCarter:
v0.7.1
*September 29, 2020*

### Changed
- `debounce.js` lint issue and timeout variable name.

### DIFF
--- a/packages/f-form-field/src/components/tests/FormField.test.js
+++ b/packages/f-form-field/src/components/tests/FormField.test.js
@@ -123,10 +123,10 @@ describe('FormField', () => {
             });
 
             describe('when set to `inlineNarrow`', () => {
+                const MOBILE = 767;
+                const DESKTOP = 768;
                 let eventName;
                 let resizeWindow;
-                let MOBILE;
-                let DESKTOP;
 
                 beforeEach(() => {
                     eventName = 'resize';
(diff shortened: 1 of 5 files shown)


packages/f-form-field/src/components/tests/FormField.test.js
```diff
@@ -123,10 +123,10 @@ describe('FormField', () => {
             });
 
             describe('when set to `inlineNarrow`', () => {
+                const MOBILE = 767;
+                const DESKTOP = 768;
                 let eventName;
```
SachaZvetelman: I think you can `const` this one as well.

packages/f-form-field/src/services/debounce.js
```diff
@@ -1,12 +1,11 @@
-module.exports = function debounce (fn, delay) {
-    let timeoutID = null;
+export default function debounce (fn, delay, ...args) {
+    let timeoutId = null;
```
SachaZvetelman, line 2: Do we need the `timeoutId` declaration here? 🤔

packages/f-form-field/src/services/debounce.js
```diff
@@ -1,12 +1,11 @@
-module.exports = function debounce (fn, delay) {
```
kevinrodrigues, line 1: Was going to ask this in the previous PR but should this go in `f-services` or somewhere it can be shared across components? I can see this being duplicated across components in future. `f-services` looks to already be used in this component too..

FayeCarter: I'll do this as a separate PR

PR #341, opened 2020-09-29
fozzie-components@2.2.0 - Set up bundlewatch

xander-marjoram:
### Added
- Bundlewatch config for monitoring bundle file sizes.

---

## UI Review Checks

- [ ] README and/or UI Documentation has been [created|updated]

PR #344, opened 2020-09-29
fozzie-components@2.2.0 - Use BuildSize for bundle size PR checks

xander-marjoram:
### Added
- Build and store artifacts for bundle size comparisons with BuildSize.

---

## UI Review Checks


xander-marjoram:
It looks like the artifacts aren't being found by BuildSize. Also, the last commit on the project was 9 months ago so I don't think it's actively maintained, I suggest we use BundleWatch as per #341

PR #345, opened 2020-09-30
f-registration@0.31.0 Adding accessibility tests

adamashton:
### Added
- Accessibility tests that run as part of the pipeline for f-registration


packages/f-registration/package.json
```diff
@@ -58,6 +58,9 @@
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.4.6",
     "@vue/test-utils": "1.0.3",
+    "assert": "2.0.0",
```
siggerzz: We're currently using the `expect` syntax for assertions, so you shouldn't need this dep.

In terms of `axe-core` / `axe-reports`, I'd personally add the deps to the root package.json, that way all packages have access to Axe :+1:

packages/f-registration/test-utils/helpers/axe-helper.js
```diff
@@ -0,0 +1,30 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
siggerzz: I'd also move this `test/test-utils` to the root level so all packages can have access to it, should Axe get added to those packages in the future

siggerzz (dismissed):

siggerzz (dismissed):

packages/f-registration/test/specs/component/axe-accessibility.spec.js
```diff
@@ -0,0 +1,40 @@
...
+        // Arrange
+        browser.url('http://localhost:8081');
+    });
+
+    it('a11y - should test f-registration component WCAG compliance', () => {
+        // Act
+        const axeResults = getAccessibilityTestResults();
+
+        // Assert
+        if (axeResults.violations.length > 0) {
```
kevinrodrigues, line 15: I'm not sure i follow this change, is this here so that we handle violation fails within the test?

test/utils/axe-helper.js
```diff
@@ -0,0 +1,63 @@
...
+                });
+            }
+          } catch(err) {
+            console.error(err)
+          }
+
+        AxeReports.processResults(results, 'csv', ciFilePath, false);
+    }
+
+        AxeReports.processResults(results, 'csv', localFilePath, false);
```
adammorr, line 65: Does this mean we end up with two csvs outputted if running in CI? Should there be an else here?

packages/f-registration/test/specs/component/axe-accessibility.spec.js
adamashton: If there are accessibility violations we process these results into a CSV which is available as a build artifact.

kevinrodrigues: I mean why do we need to have this in this particular test?

kevinrodrigues: You probably need a separate test altogether that handles the CSV behaviour i.e the `processResults` method itself.

test/utils/axe-helper.js
```diff
@@ -0,0 +1,63 @@
+const { source } = require('axe-core');
+const AxeReports = require('axe-reports');
+const { exec } = require("child_process");
+
+exports.getAccessibilityTestResults =  () => {
```
ashleynolan: Can you add JSDoc comments to these functions?

test/utils/axe-helper.js
adamashton: Good spot, fixed that now.

test/utils/axe-helper.js
adamashton: Yup, added

packages/f-registration/test/specs/component/axe-accessibility.spec.js
adamashton: There are only two accessibility tests. One is run on page load, and another is run when the error messages are displayed.  I'm not sure I follow your question.

kevinrodrigues: I might just be missing something super obvious so apologies if that's the case, but from looking at the test tho i dont see the need to invoke `processResults` (line 16) because your invoking an assertion that you expect to be `0`/`false` anyway. If there are any instances of `axeResults.violations.length` being greater than 0 then that should be tested & mocked separately imo. If the `axeResults.violations.length` are ever going to be greater than `0` this test fails (unless calling processResults sets it to zero again?)

adamashton: We want the test and build to fail if there are accessibility issues. The if statement just saves us creating a build artifact unnecessarily if there are no violations.

kevinrodrigues: Ah i see makes sense now. I think it's worth adding a note about that then somewhere.

kevinrodrigues (dismissed):

packages/f-registration/test/specs/component/axe-accessibility.spec.js
adamashton: Yeh, will do as it's not obvious to someone who hasn't worked on it!

kevinrodrigues (approved):

adammorr (approved):

dandel10n (approved):

PR #353, opened 2020-10-05
generator-component@1.1.0 - Update component generator template scripts

siggerzz:


ashleynolan (dismissed):

SachaZvetelman:
@siggerzz, rather than asking the user to delete the unused files, why don't we ask the user if they need these files when presenting the wizard? It shouldn't be too difficult to do. I'm happy to pair on this if you need.

SachaZvetelman (dismissed):

SachaZvetelman:
@siggerzz, just to confirm, was hardcoding `8080` intentional?

SachaZvetelman (dismissed):

DamianMullins (dismissed):

test/infrastructure/healthcheck.js
```diff
@@ -1,7 +1,7 @@
 
 const axios = require('axios');
 
-const address = `http://localhost:${process.argv[2]}`;
+const address = `http://localhost:8080`;
```
kevinrodrigues: You can just use a standard string here instead of the template now.

DamianMullins (approved):

kevinrodrigues (approved):

PR #354, opened 2020-10-06
f-checkout@0.1.0 - First commit with the base component generated by generator-component.

SachaZvetelman:
v0.1.0
*October 6, 2020*

### Added
- Component structure and basic configuration (created using `generator-component`).

### DIFF
--- a/packages/f-checkout/babel.config.js
+++ b/packages/f-checkout/babel.config.js
@@ -1,0 +1,24 @@
+module.exports = api => {
+    // Use `isTest` to determine what presets and plugins to use with jest
+    const isTest = api.env('test');
+    const presets = [];
+    const plugins = [
+        '@babel/plugin-proposal-optional-chaining' // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
+    ];
+    const builtIns = (api.env('development') ? 'entry' : false);
+
+    if (!isTest) {
+        api.cache(true); // Caches the computed babel config function – https://babeljs.io/docs/en/config-files#apicache
+        presets.push(['@vue/app', { useBuiltIns: builtIns }]);
+    }
+
+    // Alias for @babel/preset-env
+    // Hooks into browserslist to provide smart Babel transforms
+    // https://babeljs.io/docs/en/babel-preset-env
+    presets.push('@babel/env');
+
+    return {
+        presets,
+        plugins
+    };
+};
--- a/packages/f-checkout/babel.config.js
+++ b/packages/f-checkout/babel.config.js
@@ -10,12 +10,14 @@ module.exports = api => {
     if (!isTest) {
         api.cache(true); // Caches the computed babel config function – https://babeljs.io/docs/en/config-files#apicache
         presets.push(['@vue/app', { useBuiltIns: builtIns }]);
+        // Alias for @babel/preset-env
+        // Hooks into browserslist to provide smart Babel transforms
+        // https://babeljs.io/docs/en/babel-preset-env
+        presets.push('@babel/env');
+    } else {
+        // use current node version for transpiling test files
+        presets.push(['@babel/env', { targets: { node: 'current' } }]);
     }
-
-    // Alias for @babel/preset-env
-    // Hooks into browserslist to provide smart Babel transforms
-    // https://babeljs.io/docs/en/babel-preset-env
-    presets.push('@babel/env');
 
     return {
         presets,
--- a/packages/f-checkout/src/assets/scss/common.scss
+++ b/packages/f-checkout/src/assets/scss/common.scss
@@ -1,7 +1,7 @@
 // Set $theme variable so that fozzie doesn't complain when vars are imported below
 // n.b. This variables isn't actually used (unless there are theme specific vars from fozzie)
 // Instead, the theme is handled via a data-attribute on the component (via the `@mixin theme()` below)
-$theme: 'je' !default;
+$theme: 'jet' !default;
 
 @import '~@justeat/f-utils/src/scss/functions/index';
 @import '~@justeat/f-utils/src/scss/mixins/index';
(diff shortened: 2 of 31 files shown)


DamianMullins (dismissed):

kevinrodrigues (dismissed):

packages/f-checkout/babel.config.js
```diff
@@ -0,0 +1,24 @@
+module.exports = api => {
+    // Use `isTest` to determine what presets and plugins to use with jest
+    const isTest = api.env('test');
+    const presets = [];
+    const plugins = [
+        '@babel/plugin-proposal-optional-chaining' // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
+    ];
+    const builtIns = (api.env('development') ? 'entry' : false);
+
+    if (!isTest) {
```
ashleynolan, line 10: Think this is something that'll need updating in the generator – we've switched the other modules to transpile the tests using Node, rather than using broswerlist transforms (as it causes issues with things like async/await in the tests).

Example of how this should look now is > https://github.com/justeat/fozzie-components/blob/master/packages/f-registration/babel.config.js#L11-L21

packages/f-checkout/src/demo/Index.vue
```diff
@@ -0,0 +1,29 @@
...
+        <meta
+            name="viewport"
+            content="width=device-width, initial-scale=1">
+        <vue-checkout
+            locale="en-GB" />
+    </div>
+</template>
+
+<script>
+import VueCheckout from '@/components/Checkout.vue';
```
kevinrodrigues, line 15: Just a minor question, should this be called `VueCheckout`? Would just `CheckOut` be fine here?

packages/f-checkout/babel.config.js
SachaZvetelman: @siggerzz, would you mind including this change in the generator as part of your other changes? 🙏 
If it adds too much complexity to your ticket, I'm happy to create a Jira for this.
In the meantime, I'll manually update this here. Thanks, @ashleynolan 👍

packages/f-checkout/src/assets/scss/common.scss
```diff
@@ -0,0 +1,22 @@
+// Set $theme variable so that fozzie doesn't complain when vars are imported below
+// n.b. This variables isn't actually used (unless there are theme specific vars from fozzie)
+// Instead, the theme is handled via a data-attribute on the component (via the `@mixin theme()` below)
+$theme: 'je' !default;
```
ashleynolan: Not an issue for this PR, but we need to update all these theme references across all components to `jet` now rather than `je`. Will just need a ticket added to our backlog. Happy to add this though if you don't have the chance (just let me know)

packages/f-checkout/src/demo/Index.vue
```diff
@@ -0,0 +1,29 @@
...
+
+export default {
+    components: { VueCheckout }
+};
+</script>
+
+<style>
+@import url('https://fonts.googleapis.com/css?family=Ubuntu:300,500&display=swap');
```
ashleynolan: Don't need these anymore – they're for the old fonts that aren't used.

ashleynolan (commented):
Looks good – couple of small things spotted that will need updating, but only minor things.

packages/f-checkout/babel.config.js
SachaZvetelman: Actually, nevermind, @siggerzz. It seems we have a bunch of updates we need to do to the `generator-component`. I'll create a ticket.

packages/f-checkout/src/demo/Index.vue
SachaZvetelman: This is the naming we've been following, I believe (e.g. `f-footer`, `f-header`), but I'm happy to change this and the generator. Thoughts?

packages/f-checkout/src/assets/scss/common.scss
SachaZvetelman: Just so that I fully understand, do we already have a `jet` theme and we need to update all the references to be `jet` instead of `je` or do we need to create that new theme?

packages/f-checkout/src/demo/Index.vue
ashleynolan: I'm happy with it being VueCheckout for the demo component – when other people import the component they can call it whatever they want to anyway, so it doesn't make much difference what we call it here.

`Header` and `Footer` need an extra word (thus having `VueHeader`) as Vue recommends using two-word component names so that they never clash with standard HTML elements. So we'd need to have some sort of prefix for those components anyway.

packages/f-checkout/src/assets/scss/common.scss
ashleynolan: Yeh, so the theme variable is set [in fozzie too here](https://github.com/justeat/fozzie/blob/master/src/scss/settings/_theme.scss). It was updated to `jet` as part of the project orange/rebrand work recently.

SachaZvetelman: @ashleynolan, cool. I'm updating this PR and I'm including this change in the ticket I'm creating for updating `generator-component`. Would you mind creating a ticket for updating this everywhere else, please?

packages/f-checkout/src/tenants/index.js
```diff
@@ -0,0 +1,21 @@
+import uk from './en-GB';
+import au from './en-AU';
+import nz from './en-NZ';
+import dk from './da-DK';
+import es from './es-ES';
+import ie from './en-IE';
+import it from './it-IT';
```
ashleynolan: QQ – As Italy is in the process of being migrated, should we include IT translations in our components now?

SachaZvetelman: Good point, we can remove this now, I think.

packages/f-checkout/src/assets/scss/common.scss
ashleynolan: Created [ticket here](https://skipthedishes.atlassian.net/browse/WCB-1216)

ashleynolan (approved):

kevinrodrigues (approved):

packages/f-checkout/src/demo/Index.vue
kevinrodrigues: That makes sense, didn't realise it was generated too. 👍

DamianMullins (approved):

PR #356, opened 2020-10-09
f-checkout@0.2.0 - Replicate ConsumerWeb layout

FayeCarter:
F-checkout component to replicate what we have in ConsumerWeb UK.

To be done in a separate PR:
- Extract parts of the view that are tenant-specific to tenant config files (e.g. placeholders, labels, etc).

## UI Review Checks
**Full Width**
|ConsumerWeb             |  f-checkout|
|:-------------------------:|:-------------------------:|
|image | image|

**Mid Width**
|ConsumerWeb             |  f-checkout|
|:-------------------------:|:-------------------------:|
| image | image|

**Narrow Width**
|ConsumerWeb             |  f-checkout|
|:-------------------------:|:-------------------------:|
| image | image |


packages/f-checkout/src/components/AllergyNote.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+    <a :class="$style['c-allergyNote']">
+        {{ text }}
+    </a>
+</template>
+
+<script>
+export default {
+    name: 'SelectorComponent',
```
DamianMullins: You shouldn't need to set the name - Vue CLI will handle setting that to be the same as the filename.

packages/f-checkout/src/components/AllergyNote.vue
```diff
@@ -0,0 +1,35 @@
...
+
+    props: {
+        text: {
+            type: String,
+            default: 'If you or someone you’re ordering for has a food allergy or intolerance, click here'
```
DamianMullins: I think I might have asked this recently but is it OK to hard-coded English here?

packages/f-checkout/src/components/AllergyNote.vue
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+.c-allergyNote {
+    display: block;
+    margin: auto;
+    padding: 15px;
```
DamianMullins: Should we be using `spacing(x2) here? Or is this value separate from our spacing rules?

packages/f-checkout/src/components/Button.vue
```diff
@@ -0,0 +1,184 @@
+<template>
+    <button
+        type='submit'
+        :class="[
+            $style['o-btn'],
+            $style[buttonStyleClass],
```
DamianMullins: You have two styles of conditional class assignment here and the line below. I'd suggest picking one, if possible, and going with that?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -3,47 +3,151 @@
...
+            :class="$style['c-card-padding']">
+            <form action="post">
+                <form-field
+                    v-model="mobileNumber"
+                    name="mobile-number"
+                    data-test-id="input-mobile-number"
+                    label-text="Mobile number"
+                    input-type="text"
+                    label-style="inline" />
+                <div :class="$style['l-address-group']">
```
DamianMullins: It might be good to add some space between the elements in this component as it makes it a little easier to scan if there's a bit of breathing space.

packages/f-checkout/src/components/AllergyNote.vue
FayeCarter: I think this will change in the next couple of PRs. At the moment I'm just hardcoding so I can get a basic layout in place.

packages/f-checkout/src/components/AllergyNote.vue
FayeCarter: 👍  Thanks!

packages/f-checkout/src/components/AllergyNote.vue
FayeCarter: I hadn't come across spacing rules yet, but I will have a look, thanks

packages/f-checkout/src/components/Button.vue
FayeCarter: 👍

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Will do

packages/f-checkout/src/components/Button.vue
```diff
@@ -0,0 +1,184 @@
...
+            type: String,
+            default: ''
+        },
+        isFullWidth: {
+            type: Boolean,
+            default: false
+        }
+    },
+    computed: {
+        buttonStyleClass () {
```
tommcclean: Optional: If you eliminated this computed by setting a default value on the prop, you could make this component functional. (Note: If functional you need to use `props.isFullWidth` in the template)

packages/f-checkout/src/components/UserNote.vue
```diff
@@ -0,0 +1,71 @@
+<template>
+    <div :class="$style['c-userNote']">
+        <h4>
+            Leave a note
+        </h4>
+        <p>
+            {{ noteText }}
+        </p>
+        <textarea
+            id="c-fareaorm-text-"
```
tommcclean: Is this ID used for anything, should be it be a prop value or use `_uid` to be unique

packages/f-checkout/src/components/AllergyNote.vue
kevinrodrigues: I think I asked this same question a while back on a PR before (sometime earlier in the year) and the response was that we should use the `name`. I agree tho we shouldn't need it :)

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -2,9 +2,16 @@ import { shallowMount } from '@vue/test-utils';
...
 describe('Checkout', () => {
+    const propsData = {};
+
     it('should be defined', () => {
-        const propsData = {};
         const wrapper = shallowMount(VueCheckout, { propsData });
         expect(wrapper.exists()).toBe(true);
     });
+
+    it('has a button', () => {
```
kevinrodrigues: `should have a button` :)

packages/f-checkout/src/components/AllergyNote.vue
```diff
@@ -0,0 +1,35 @@
...
+    padding-bottom: 0;
+    text-align: center;
+    color: $color-link-default;
+    text-decoration: underline;
+    font-weight: $font-weight-bold;
+
+    &:hover {
```
kevinrodrigues: &:focus should be added here too.

packages/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: 🤦‍♀️ Thanks

packages/f-checkout/src/components/UserNote.vue
FayeCarter: It's used for nothing at the moment, I honestly don't remember writing that in as an ID. I'll change that, thanks

kevinrodrigues:
@FayeCarter Will there be some more style changes as the new version looks a little different to CW (padding/ icon/font sizes etc)? It's probably worth running the styles by a UI designer. Someone like Bruno Molena or Tom Oxlade could help here.

FayeCarter:
@kevinrodrigues, yes, still working on style changes. I just wanted to get a PR out early so I can start working on feedback

packages/f-checkout/package.json
```diff
@@ -41,6 +43,8 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
+    "@justeat/f-card": "^0.6.0",
```
ashleynolan: Can you remove the carat's `^` from these packages to lock the versions down

packages/f-checkout/src/components/Button.vue
```diff
@@ -0,0 +1,87 @@
...
+<script>
+export default {
+    name: 'ButtonComponent'
+};
+</script>
+
+<style lang="scss" module>
+
+$btn-default-weight                 : 500;
+$btn-default-height                 : 2.6;
```
ashleynolan: Doesn't look like some of these variables are used below? (like `$btn-default-height`)

Probably worth getting rid of any aren't needed and adding them in as they're requred.

FayeCarter: Yes, will do

packages/f-checkout/package.json
FayeCarter: Noooo, I missed it again! Yes I'll change this

packages/f-checkout/src/components/Button.vue
```diff
@@ -0,0 +1,78 @@
...
+$btn-default-hozPadding             : 2em;
+$btn-default-vertPadding            : 12px;
+
+$btn-primary-bgColor                : $orange;
+$btn-primary-bgColor--hover         : $orange--dark;
+$btn-primary-textColor              : $white;
+
+$btn-disabled-bgColor               : $grey--light;
+
+.o-btn {
```
SachaZvetelman: Don't we have this class defined in `fozzie` already? https://github.com/justeat/fozzie/blob/2b87c3a0e39fd5aacca029cf4780a1a25d21ba58/src/scss/objects/_buttons.scss#L78

We shouldn't be copy-pasting, but re-using the existing styles.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -3,47 +3,173 @@
...
+        padding: 25px 80px 50px 80px;
+    }
 
+    h1 {
+        font-size: 20px;
+        margin-bottom: spacing(x2);
+        color: $grey--darkest;
+    }
+
+    .c-formField {
```
SachaZvetelman, line 172: Same here. Isn't this defined by the `f-form-field` component?

SachaZvetelman (commented):
Looking good!
I think many of the styles should be already coming from `fozzie`. I recommend taking another look at that so we don't end up with a lot of duplicate styles.

packages/f-checkout/src/components/AllergyNote.vue
```diff
@@ -0,0 +1,35 @@
...
+.c-allergyNote {
+    display: block;
+    margin: auto;
+    margin-top: spacing(x2) ;
+    padding: 0 spacing(x3);
+    text-align: center;
+    color: $color-link-default;
+    text-decoration: underline;
+    font-weight: $font-weight-bold;
+    font-size: 16px;
```
xander-marjoram: Is there a font-size variable you can use? (Happy for you to look at this in the next PR)

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -3,47 +3,176 @@
...
+                    label-text="Mobile number"
+                    input-type="text"
+                    label-style="inline" />
+
+                <div :class="$style['l-address-group']">
+                    <form-field
+                        v-model="address.lineOne"
+                        :class="$style['c-formField']"
+                        name="address-line-one"
+                        data-test-id="input-address-line-one"
```
xander-marjoram: Just curious if you had an issue with `"input-address-line-1"`? It might make it easier to find the field in an automated way in future

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -3,47 +3,176 @@
...
+                </button>
+            </form>
+        </card>
     </div>
 </template>
 
 <script>
 import { globalisationServices } from '@justeat/f-services';
+import Card from '@justeat/f-card';
+import '@justeat/f-card/dist/f-card.css';
```
xander-marjoram, line 85: I think you can actually move these CSS imports down into the `style` section with `@import`

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -3,47 +3,176 @@
...
+import FormField from '@justeat/f-form-field';
+import '@justeat/f-form-field/dist/f-form-field.css';
+import FormSelector from './Selector.vue';
+import UserNote from './UserNote.vue';
+import AllergyNote from './AllergyNote.vue';
```
xander-marjoram: And as a **nit**, please can you alphabetise these here and in the components block below?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -3,47 +3,176 @@
...
+        height: 50px;
+    }
 
+
+    .l-address-group {
+        margin: spacing(x2) 0 spacing(x4) 0;
+        @include font-size(body-s);
+
+        div {
+            margin-bottom: -17px;
```
xander-marjoram, line 187: Would `-spacing(x2)` be good enough?

packages/f-checkout/src/components/UserNote.vue
```diff
@@ -0,0 +1,73 @@
...
+<script>
+export default {
+    props: {
+        noteText: {
+            type: String,
+            default: "Leave a note for the restaurant with anything they need to know (e.g. the doorbell doesn't work). Do not include details about any allergies here."
+        },
+        notePlaceholder: {
+            type: String,
+            default: "e.g. the doorbell doesn't work. Do not include details about any allergies here. \n\nWe're working with resturants to cut waste. Please don't ask for plastic cutlery."
```
xander-marjoram, line 27: I know you'll replace/extract this text, but there's a typo: "resturants"

packages/f-checkout/src/components/UserNote.vue
```diff
@@ -0,0 +1,73 @@
...
+        font-weight: 600;
+        margin-top: spacing(x3);
+        color: $font-colour;
+        font-size: 16px;
+    }
+
+    div {
+        margin-top: 5px;
+        font-size: 14px;
+        // color: $grey--dark;
```
xander-marjoram: Can this be removed?

packages/f-checkout/package.json
```diff
@@ -41,6 +43,8 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
+    "@justeat/f-card": "0.6.0",
+    "@justeat/f-form-field": "0.7.1",
     "@justeat/f-services": "1.0.0",
     "axios": "0.20.0"
```
xander-marjoram: Out of interest, why does this need to depend on axios? Is it possible to use axiosServices from f-services instead?

packages/f-checkout/src/components/AllergyNote.vue
```diff
@@ -0,0 +1,35 @@
...
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+.c-allergyNote {
+    display: block;
+    margin: auto;
+    margin-top: spacing(x2) ;
```
xander-marjoram: Errant space at the end 😄

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Thanks, I've updated this

packages/f-checkout/src/components/Checkout.vue
FayeCarter: You can! Thanks

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Annoyingly none of the spacing seem to work for it, but I'll keep experimenting

packages/f-checkout/src/components/UserNote.vue
FayeCarter: 🤦‍♀️

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -3,47 +3,188 @@
...
+                    v-model="mobileNumber"
+                    name="mobile-number"
+                    data-test-id="input-mobile-number"
+                    label-text="Mobile number"
+                    input-type="text"
+                    label-style="inline" />
+
+                <div :class="$style['l-addressGroup']">
+                    <form-field
+                        v-model="address.lineOne"
```
xander-marjoram: Please could you update the variable names to `line1`, etc, for consistency?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -3,47 +3,188 @@
...
     data () {
         const locale = globalisationServices.getLocale(tenantConfigs, this.locale, this.$i18n);
         const localeConfig = tenantConfigs[locale];
         const theme = globalisationServices.getTheme(locale);
 
         return {
             copy: { ...localeConfig },
-            theme
+            theme,
+            firstName: 'faye',
```
xander-marjoram: Probably want a more sensible default 😄

FayeCarter: 😂 I'll change it to Xander. Thanks

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Turns out it breaks the build. But I'll keep looking if there's a way

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -3,47 +3,187 @@
...
+        FormSelector,
+        UserNote
+    },
+
     props: {
         locale: {
             type: String,
             default: ''
+        },
+        buttonText: {
```
SachaZvetelman: Not sure `buttonText` and `allergyText` should be `props`. 
They look more like `data` to me. And in the future, they will be imported from a tenant config file.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -3,47 +3,187 @@
...
+        }
     }
 };
 </script>
 
 <style lang="scss" module>
+$font-family                              : $font-family-base;
+$font-colour                              : $color-text;
+$font-weight                              : $font-weight-base;
+$font-weight-bold                         : $font-weight-bold;
```
SachaZvetelman: Do we need these?

packages/f-checkout/src/components/Selector.vue
```diff
@@ -0,0 +1,69 @@
...
+
+    .o-selector-input {
+        height: 53px;
+        width: 100%;
+        padding: 0.5rem;
+        border: $form-input-borderWidth solid $form-input-borderColour;
+        border-radius: $form-input-borderRadius;
+        cursor: pointer;
+    }
+}
```
SachaZvetelman: Aren't these available in a `fozzie` repo already that we're importing? Similar to `.o-btn` for example.

packages/f-checkout/package.json
FayeCarter: I would imagine that this is possible. I know that this has been setup in the same was as `f-registration`. Is it something that should be changed in this PR? Or in a separate one?

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Probably not. 😄

packages/f-checkout/src/components/Checkout.vue
ashleynolan: Don't think you need to here – so as this component gets bundled up, it creates as output a JS module file and a css file (which this will be bundled into).

So it's only in Magikarp that we need to make sure any dependency CSS files use `@import` in the SCSS blocks (so that they render without JS).

I imagine it breaks the build because WebPack is rubbish 😆

packages/f-checkout/package.json
xander-marjoram: Happy for this to be addressed in another PR

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -3,47 +3,216 @@
...
+            margin-top: 0;
+        }
+    }
+
+    h1 {
+        @include font-size(heading-s);
+        margin-bottom: spacing(x2);
+    }
+
+    .c-formField {
```
FayeCarter, line 172: This will be removed with updates to `f-form-field`

packages/f-checkout/package.json
```diff
@@ -32,7 +32,8 @@
     "demo": "vue serve --port 8080 ./src/demo/Index.vue",
     "lint": "vue-cli-service lint",
     "lint:fix": "yarn lint --fix",
-    "test": "vue-cli-service test:unit --json --outputFile=./src/components/tests/.jest-test-results.json",
+    "test": "vue-cli-service test:unit",
```
minrwhite: I think this is removing code that @siggerzz added for showing tests in storybook - possibly as a result of a mis-merge of the fragment?

simonsandersje (dismissed):

packages/f-checkout/stories/Checkout.stories.js
```diff
@@ -1,20 +1,15 @@
...
+        '<vue-checkout />'
 });
 
-Checkout.parameters = {
-  jest: ['Checkout.test.js'],
-};
-
-Checkout.storyName = 'f-checkout';
\ No newline at end of file
+CheckoutComponent.storyName = 'f-checkout';
```
minrwhite: Again, I think this is removing some code for the `withTests` decorator.

packages/f-checkout/package.json
FayeCarter: 🤦‍♀️ Sorry, I'll put it back in.

packages/f-checkout/stories/Checkout.stories.js
FayeCarter: 👍 Thanks Ben

packages/f-checkout/package.json
FayeCarter: @benwhite-deltas and @siggerzz I think I've put everything back together again. Can you see If missed anything please?

minrwhite: The fixed file looks good to me @FayeCarter - one last thing is that `packages/f-checkout/src/components/tests/.jest-test-results.json` should be excluded by [this line in `.gitignore`](https://github.com/justeat/fozzie-components/blob/874660e2bf650c7bbc5da4c920664c6cb25c847f/.gitignore#L32) - please remove this file and then we should be good.

FayeCarter: Removed. Thank you 😃

minrwhite (approved):

siggerzz (approved):

xander-marjoram (approved):

PR #357, opened 2020-10-09
f-error-message@v0.1.0

ross-gratton:
Created generic error message component `f-error-message`. This PR is the creation and functionality of the component, the tests are still to be added.

Has been configured for storybook
Screenshot 2020-10-09 at 11 25 24


ashleynolan:
@ross-gratton Just a quick note – PR's/commits should refer to the package and package number, rather than the JIRA number (as that's internal facing). So this will be f-error-message@v0.1.0

packages/f-error-message/babel.config.js
```diff
@@ -0,0 +1,24 @@
+module.exports = api => {
+    // Use `isTest` to determine what presets and plugins to use with jest
+    const isTest = api.env('test');
+    const presets = [];
+    const plugins = [
+        '@babel/plugin-proposal-optional-chaining' // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
+    ];
+    const builtIns = (api.env('development') ? 'entry' : false);
+
+    if (!isTest) {
```
ashleynolan, line 10: Think this is something that'll need updating in the generator – we've switched the other modules to transpile the tests using Node, rather than using broswerlist transforms (as it causes issues with things like async/await in the tests).

Example of how this should look now is > https://github.com/justeat/fozzie-components/blob/master/packages/f-registration/babel.config.js#L11-L21

packages/f-error-message/src/demo/Index.vue
```diff
@@ -0,0 +1,29 @@
...
+<script>
+import ErrorMessage from '@/components/ErrorMessage.vue';
+
+export default {
+    components: { ErrorMessage }
+};
+</script>
+
+<style>
+@import url('https://fonts.googleapis.com/css?family=Ubuntu:300,500&display=swap');
```
ashleynolan: Can delete these 2 lines – they're from the old fonts that got recently replaced

packages/f-error-message/src/components/ErrorMessage.vue
```diff
@@ -0,0 +1,49 @@
...
+        hasSlotData () {
+            return !!this.$slots.default;
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+
+.c-error-message {
```
ashleynolan: Should be `c-errorMessage` to match the scss naming convention (and then `c-errorMessage-icon` below)

packages/f-error-message/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+# Changelog
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+
+v0.1.0
+------------------------------
+*January 1, 2020*
```
simonsandersje: Time travelling again Ross?

Set the date to today's date and under 'Added' put something like "Initial commit"

packages/f-error-message/src/components/ErrorMessage.vue
```diff
@@ -0,0 +1,52 @@
...
+    }
+};
+</script>
+
+<style lang="scss" module>
+
+.c-errorMessage {
+    display: flex;
+    align-items: center;
+    color: $red;
```
simonsandersje: IMHO might be better to use the generic colour def: $color-text--danger

ross-gratton: @ashleynolan are you happy with me making this change, the use of `$red` was extracted directly from the `f-registration` component.

ashleynolan: Yeh, makes sense 👍

ashleynolan (dismissed):

ashleynolan:
Looks like you need to include `f-vue-icons` as a dependency – the build seems to be failing on that atm

siggerzz (dismissed):

packages/f-error-message/src/components/tests/ErrorMessage.test.js
```diff
@@ -0,0 +1,37 @@
+import { shallowMount } from '@vue/test-utils';
+import ErrorMessage from '../ErrorMessage.vue';
+
+const propsData = {
+    slots: {
+        default: '<div class="mock-message">Here is a message</div>'
+    }
+};
+
+describe('ErrorMessage', () => {
```
kevinrodrigues, line 10: Could we tidy up the spacing here please (Looks like there's new lines before most of the `const` & spaces after the `it`)? Also add Arrange, Act & Assert too?

packages/f-error-message/src/components/tests/ErrorMessage.test.js
```diff
@@ -0,0 +1,37 @@
...
+
+    it('should be defined', () => {
+        const wrapper = shallowMount(ErrorMessage, {});
+        expect(wrapper.exists()).toBe(true);
+    });
+
+    it ('should render if slot content is provided', () => {
```
kevinrodrigues: This test and the very last test looks similar. You could just test that the inner content is rendered which would automatically mean that the element exists?

packages/f-error-message/src/components/tests/ErrorMessage.test.js
```diff
@@ -0,0 +1,37 @@
...
+        const wrapper = shallowMount(ErrorMessage, {});
+        const content = wrapper.find('[data-test-id="content"]');
+        expect(content.exists()).toBe(false);
+    });
+
+    it ('should correctly render the content passed in a slot', () => {
+
+        const wrapper = shallowMount(ErrorMessage, propsData);
+        const content = wrapper.find('[data-test-id="content"]');
+        expect(content.element.innerHTML).toBe(propsData.slots.default);
```
kevinrodrigues, line 42: Would this work instead?

`expect(content.find('.mock-message').exists()).toBeTruthy()`

packages/f-error-message/src/components/tests/ErrorMessage.test.js
ross-gratton: Do you mean remove the "should be defined" test as all the other tests will fail anyway should the wrapper not exist?

packages/f-error-message/src/components/tests/ErrorMessage.test.js
ross-gratton: It would, however it would tightly couple that test to the structure of the default message. The current method allows the default message to take any form and not fail in the future. We're only concerned with the message matching, not the set structure of it. Should a future test need to change the default, this test wont need any attention to correct.

kevinrodrigues: But this is checking the structure unless i've read it wrong? You're checking that the innerHTML is equal to the default mocked html? The example i provided is effectively the same thing here.

packages/f-error-message/src/components/tests/ErrorMessage.test.js
kevinrodrigues: Here i mean that you probably don't need this because `propsData` is already being set in the last test so it's effectively testing this scenario anyway.

packages/f-error-message/src/components/ErrorMessage.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+    <p
```
kevinrodrigues, line 2: Will we add any accessibility roles to this component to warn users about errors?

packages/f-error-message/src/components/tests/ErrorMessage.test.js
kevinrodrigues: I think i know what you mean now. Just not super clear from the test imo.

ross-gratton: Yeah the structure isn't really a concern, it's just testing that whatever gets rendered matches what gets sent through the slot. So essentially validating the component is working as expected.

packages/f-error-message/src/components/ErrorMessage.vue
ross-gratton: There is a separate ticket in f-registration that deals with accessibility in errors. I think we should get both this work and that work through and then raise a ticket to integrate this ticket into f-registration and make any accessibility adjustments.

simonsandersje (dismissed):

packages/f-error-message/src/components/ErrorMessage.vue
kevinrodrigues: So will the accessibility changes eventually be made in here? For me it would be just adding `assertive` to a aria live attribute. Just keen to know because if this component is reused in another component other than the registration page we might end up losing some accessibility. Not a blocker for merging if you have a ticket already

ross-gratton: I'll check that we have a ticket created to handle this. But, yes there will be some more work in this component, it'll mostly be around adding the correct aria attributes to ensure that the form fields are correctly associated with the error messages.
AFAIK if we put the aria live assertive directly on the error they'll actually just end up cancelling each other out and you'll only hear the last error. This has all been addressed in getting the f-registration errors to be accessible. 
This is why my feeling is that the accessibility work that is required for this component should be done while integrating with an already accessible consumer to ensure that we have to desired outcome.

packages/f-error-message/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+# Changelog
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+
+v0.1.0
+------------------------------
+*October 6, 2020*
```
ashleynolan, line 17: Probably needs a bump now 😆

packages/f-error-message/package.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "keywords": [
+    "fozzie"
+  ],
+  "scripts": {
+    "prepublishOnly": "yarn lint && yarn test && yarn build",
+    "build": "vue-cli-service build --target lib --name f-error-message ./src/index.js",
+    "demo": "vue serve --open ./src/demo/Index.vue",
+    "lint": "vue-cli-service lint",
+    "lint:fix": "yarn lint --fix",
```
ashleynolan, line 32: Could you add `"lint:style": "vue-cli-service lint:style",` below the `lint:fix` line

packages/f-error-message/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "extends @justeat/browserslist-config-fozzie"
+  ],
+  "dependencies": {
+    "@justeat/f-vue-icons": "1.0.0"
+  },
+  "peerDependencies": {
+    "@justeat/browserslist-config-fozzie": ">=1.1.1"
+  },
+  "devDependencies": {
+    "@vue/cli-plugin-babel": "4.4.6",
```
ashleynolan, line 47: And `"@samhammer/vue-cli-plugin-stylelint": "2.0.1",` at the top of the `devDependencies` list

packages/f-error-message/vue.config.js
```diff
@@ -0,0 +1,17 @@
...
+            .rule('scss-importer')
+            .test(/\.scss$/)
+            .use('importer')
+            .loader('sass-loader')
+            .options({
+                importer: magicImporter(),
+                // eslint-disable-next-line quotes
+                data: `@import "@/assets/scss/common.scss";`
+            });
+    }
```
ashleynolan, line 19: …and 

```js
    },
    pluginOptions: {
        lintStyleOnBuild: true
    }
```

packages/f-error-message/src/components/ErrorMessage.vue
```diff
@@ -0,0 +1,52 @@
...
+<script>
+import { WarningIcon } from '@justeat/f-vue-icons';
+
+export default {
+    name: 'ErrorMessage',
+    components: {
+        WarningIcon
+    },
+    props: {
+        dataTestId: {
```
ashleynolan: Is this used anywhere? Can't see it referenced in the template (can see a `dataTestTitle` but not sure if that's meant to be linked?)

ross-gratton: Good catch, you're absolutely right, that prop should be dataTestTitle. Have fixed in new commit.

packages/f-error-message/CHANGELOG.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Added
+- Initial commit
+
+v0.2.0
```
ashleynolan, line 7: Sorry I keep noticing things separately 🐼  Later versions should appear at the top of the file, not below (so the latest version is always at the top of the `CHANGELOG`

Pretty sure that's it then 😆

ross-gratton: No probs, fixing now.

ashleynolan (dismissed):

kevinrodrigues (dismissed):

ashleynolan:
Looks like you're getting a build error – may be because you need to merge/rebase/remerge latest master back into this branch

siggerzz (approved):

ashleynolan (approved):

PR #358, opened 2020-10-09
f-registration@0.34.0

luke-andrew-parry:
This PR updates the custom events that are emitted satisfy hooks required for pushing tracking into GA.

### CreateAccountStart

This event fires then the form is first clicked or tabbed into.

### CreateAccountInlineError

This event first when a form field loses focus and is invalid.

### CreateAccountFailure 

This event has been updated to pass in the payload a list of valid & invalid fields so that these can be passed into GA.


cmyhill-justeat (dismissed):

packages/f-registration/package.json
```diff
@@ -44,7 +44,7 @@
   ],
   "dependencies": {
     "@justeat/f-card": "0.6.0",
-    "@justeat/f-form-field": "0.7.1",
+    "@justeat/f-form-field": "^0.7.1",
```
adamashton: I think this should be an exact version.

luke-andrew-parry: Yes, it should. I'll revert this.

adamashton (approved):

cmyhill-justeat (approved):

PR #360, opened 2020-10-12
f-registration@0.34.0

luke-andrew-parry:
This PR updates the custom events that are emitted satisfy hooks required for pushing tracking into GA.

### CreateAccountStart

This event fires then the form is first clicked or tabbed into.

### CreateAccountInlineError

This event first when a form field loses focus and is invalid.

### CreateAccountFailure 

This event has been updated to pass in the payload a list of valid & invalid fields so that these can be passed into GA.


kevinrodrigues (approved):

adamashton (approved):

simonsandersje (approved):

packages/f-registration/src/components/Registration.vue
```diff
@@ -216,6 +219,25 @@ import EventNames from '../event-names';
  */
 const meetsCharacterValidationRules = value => /^[\u0060\u00C0-\u00F6\u00F8-\u017Fa-zA-Z-' ]*$/.test(value);
 
+const formValidationState = $v => {
```
ashleynolan, line 222: Could this be a `method` defined with the other methods below? (just wondering if there was a reason for it being separate)

(Sorry for late review – just interested to know!)

PR #362, opened 2020-10-13
generator-component@v1.4.0 - added accessibility test template

LTurns:
In the generator component: 

### Added
- Axe Accessibility Spec template in its own Accessibility folder

### Changed
- Refactored test by moving conditional statement from axe-accessibility.spec to axe-helper


packages/generator-component/generators/app/index.js
```diff
@@ -48,8 +48,8 @@ module.exports = class extends Generator {
                                 .replace(/(f-skeleton)/g, `f-${nameTransformations.class}`)
                                 .replace(/__/g, ''); // We don't want to have file names such as .test.js or .stories.js, otherwise Jest or Storybook will pick them up from the templates folder.
         }));
-
-        let ignoreTestPattern = this.answers.needsUITests ? [] : ["**/*/test", '**/*/test-utils/component-objects']
+
```
siggerzz: Nit: This white space can be removed

packages/generator-component/generators/app/templates/test/specs/accessibility/axe-accessibility.spec.js
```diff
@@ -0,0 +1,21 @@
+import { getAccessibilityTestResults, processResults } from '../../../../../test/utils/axe-helper';
+
+describe('Accessibility tests', () => {
+    beforeEach(() => {
+        // Arrange
+        browser.url('http://localhost:8081');
+    });
+
+    it('a11y - should test f-registration component WCAG compliance', () => {
```
siggerzz: This string still mentions f-registration

siggerzz (changes requested):
Minor comments, but looks good otherwise!

packages/generator-component/generators/app/templates/test/specs/accessibility/axe-accessibility.spec.js
LTurns: ah well spotted, will change now!

simonsandersje (dismissed):
LGTM - apart from Ben's comment about the f-reg mention

siggerzz (dismissed):
🎉

packages/generator-component/generators/app/templates/test/specs/accessibility/axe-accessibility.spec.js
```diff
@@ -0,0 +1,21 @@
...
+        // Arrange
+        browser.url('http://localhost:8081');
+    });
+
+    it('a11y - should test f-<%= name.class %> component WCAG compliance', () => {
+        // Act
+        const axeResults = getAccessibilityTestResults();
+
+        // Assert
+        if (axeResults.violations.length > 0) {
```
SachaZvetelman: I personally think having logic (e.g. conditionals, loops) within tests is not a good idea. Is this something we must have or can we find a way to avoid this? cc @LTurns and @siggerzz.

siggerzz: Yeah, this is mandatory. A11y tests should only fail if there are any Axe violations. All tests look identical, the only difference is the component that gets loaded in the page.

SachaZvetelman: Well, we could potentially amend `processResults` so it checks if there are violations inside it, instead of requiring putting this conditional in every test we want to write in the future for accessibility. Unless I'm wrong here.
I'm normally quite against logic in tests, because...Who's testing the logic in those tests is correct?
I'm happy to get more opinions here, and if most of the people are ok with having conditionals in tests, then I'll shut up 😬, but in general, it's not a practice I'm very happy with.

SachaZvetelman: Sorry, one more question. Is this the one and only test we'll have for accessibility in our components or are we going to be writing more tests in the future? If it's this one and nothing else, not a big deal and ignore my comment above.

DamianMullins: I'm with @SachaZvetelman on this, conditions in tests should be avoided as it complicates the tests, makes them harder to follow, and could lead to failures which have nothing to do with the SUT. 

As @SachaZvetelman mentioned, why not wrap up the `processResults` call in the `getAccessibilityTestResults` function? I', wondering if you could check the violations and do the additional work before returning?

siggerzz: @SachaZvetelman Yeah, this _should_ be the only accessibility test for most components. Happy to address this though :)

LTurns: Yeah I agree there's probably a simple way to do this, will take a look at the logic now! :)

ashleynolan:
@LTurns Just an FYI – I've also raised a PR for the generator just now with some small updates. Mine are quite minor changes and saw you were addressing comments, so put mine as `v1.2.0` too incase it gets approved first, but happy to change if this is ready for review/merge (and I'll merge your changes into mine then).

LTurns:
> @LTurns Just an FYI – I've also raised a PR for the generator just now with some small updates. Mine are quite minor changes and saw you were addressing comments, so put mine as `v1.2.0` too incase it gets approved first, but happy to change if this is ready for review/merge (and I'll merge your changes into mine then).

No problem @ashleynolan, I'm still testing my changes locally so probably unlikely to merge before you. Will rebase before I push up again. 👍

package.json
```diff
@@ -49,7 +49,7 @@
     "babel-jest": "26.1.0",
     "babel-loader": "8.1.0",
     "bundlewatch": "0.3.1",
-    "chromedriver": "84.0.1",
+    "chromedriver": "85.0.1",
```
siggerzz: This can be reverted

packages/f-form-field/test/specs/README.md
```diff
@@ -0,0 +1,3 @@
+# Usage
```
siggerzz: This can be removed

packages/generator-component/CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v1.4.0
+------------------------------
+*October 14, 2020*
```
siggerzz: Date needs updating

test/utils/axe-helper.js
```diff
@@ -66,4 +75,4 @@ exports.processResults = (results, componentName) => {
     }
 
     console.error(`Expected no accessibility violations. Found: ${results.violations.length}`);
-};
+};
```
siggerzz: Add a new line at the end of this one

siggerzz (changes requested):
Few minor changes!

LTurns:
> Few minor changes!

Just changed @siggerzz thank you!

siggerzz (dismissed):

siggerzz (approved):

simonsandersje (approved):

PR #363, opened 2020-10-13
f-alert@0.1.0 - Initial version with support for `danger`, `warning`, `success` and `info` dismissable/non-dismissable alerts

SachaZvetelman:
v0.1.0
*October 20, 2020*

### Added
- Initial version with support for `danger`, `warning`, `success` and `info` dismissable/non-dismissable alerts.

---

## UI Review Checks

**Desktop success dismissable alert:**
image

**Desktop danger non-dismissable alert:**
image

**Mobile warning non-dismissable alert:**
image

**Mobile info dismissable alert:**
image

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/f-alert/src/components/tests/Alert.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+    describe('methods', () => {
+        describe('dismiss', () => {
+            it('should set `isDismissed` to `true`', () => {
+                // Arrange
+                const propsData = {};
+                const wrapper = shallowMount(VueAlert, { propsData });
+
+                // Act
+                wrapper.vm.dismiss();
```
FayeCarter, line 145: 100% out of curiosity. Would you ever call this method by triggering a click of the dismiss button?

SachaZvetelman: Absolutely! The existing test basically verifies the `dismiss()` method does what it's supposed to do.
There will be another test to verify when the close button is clicked, the `dismiss()` method is called. Does it make sense? 😄

FayeCarter: 👍 good to know, thanks!

packages/f-alert/src/components/Alert.vue
```diff
@@ -0,0 +1,168 @@
+<template>
+    <div
```
kevinrodrigues, line 2: Will there be any accessibility `roles` added to this component in the next few prs? We should have something that alerts the user to potential alerts.

packages/f-alert/src/assets/scss/common.scss
```diff
@@ -0,0 +1,22 @@
...
+@import '~@justeat/fozzie/src/scss/settings/variables';
+@import '~@justeat/fozzie/src/scss/base/reset';
+@import '~@justeat/fozzie-colour-palette/src/scss/';
+
+/**
+ * Component Variables
+ */
+// $alert-background-color: $brand--orange;
+
+@mixin theme($type) {
```
ashleynolan: You can delete this theme mixin (it's not needed anymore – I've got a PR incoming for the generator with some updates!)

packages/f-alert/src/components/Alert.vue
```diff
@@ -0,0 +1,168 @@
...
+        v-if="!isDismissed"
+        :class="[$style['c-alert'],
+                 $style[`c-alert--${type}`]]"
+        data-test-id='alert-component'>
+        <div
```
ashleynolan, line 8: Could just call this `c-alert-header` (or if you want to use container could call it `c-alert-headingContainer`)

packages/f-alert/src/components/Alert.vue
```diff
@@ -0,0 +1,168 @@
...
+            :class="$style['c-alert-heading-container']">
+            <component
+                :is="icon"
+                :class="[$style['c-icon--alert'],
+                         $style['c-alert-icon'],
+                         $style[`c-alert-icon--${type}`]]" />
+            <!-- TODO: make this dynamic. See https://skipthedishes.atlassian.net/browse/WCB-1219 -->
+            <h2
+                :class="$style['c-alert-heading']">
+                <slot name="heading" />
```
ashleynolan: Will this always be just text? If so, I'd probably make it a prop, so that only strings can be passed in.

packages/f-alert/src/components/Alert.vue
```diff
@@ -0,0 +1,168 @@
...
+    .c-alert-icon {
+        height: 20px;
+        margin: 0 spacing(x1.5) 0 spacing();
+    }
+
+        .c-alert-icon--warning {
+            margin-right: 10px;
+        }
+
+    button.c-alert-dismiss { // TODO: Needed more specificity here.
```
ashleynolan, line 161: Did you need the added specificity to override the `o-btn` styles? Or some styles from elsewhere?

packages/f-alert/stories/Alert.stories.js
```diff
@@ -0,0 +1,32 @@
+import { storiesOf } from '@storybook/vue';
+import VueAlert from '../src/components/Alert.vue';
+import VueCard from '../../f-card/src/components/Card.vue';
+
+
+storiesOf('Components', module)
```
ashleynolan: Another annoying one that I noticed yesterday when starting the button component – this is using Storybook syntax. `storiesOf` uses the legacy API.

If you look at the majority of the other components stories, they should use the `export` syntax (which is very similar, just need a little reworking).

I'm updating this in the generator too (right now) 😄

packages/f-alert/src/components/Alert.vue
SachaZvetelman: I cannot think of a scenario in which someone will want HTML in the title, so yup, sounds good, will change. Good point.

packages/f-alert/src/tenants/index.js
```diff
@@ -0,0 +1,19 @@
+import uk from './en-GB';
+import au from './en-AU';
+import nz from './en-NZ';
+import dk from './da-DK';
```
tommcclean, line 4: Do we need EU countries?

SachaZvetelman: For the time being, yes, we do 🙂

packages/f-alert/src/components/Alert.vue
SachaZvetelman: It's actually `.o-btn--icon` the one screwing things up. What do you recommend? :)

ashleynolan: I'd probaby leave it like it is for now. What would be nice to do would be to extend the `f-button` component that I'm adding to handle icons like this too and then you could pull that component in for this potentially (and that would fix this issue).

packages/f-alert/src/components/Alert.vue
DamianMullins: You can support both scenarios like this to allow for flexibility.

```vue
<slot name="heading">
  {{ heading }}
</slot>
```

SachaZvetelman: @DamianMullins, that's a nice one. I think for now I'll keep it as a `prop`, but it's good to know about that alternative.

packages/f-alert/src/components/Alert.vue
```diff
@@ -0,0 +1,176 @@
...
+                    :class="[$style['c-icon--cross'], $style['c-alert-dismiss-icon']]"
+                />
+                <span class="is-visuallyHidden">
+                    {{ copy.dismissAlertText }}
+                </span>
+            </button>
+        </div>
+        <div
+            :class="$style['c-alert-content']">
+            <slot name="content" />
```
ashleynolan: As this is the only slot now, you can just use `<slot />` here. Then anything you put inside the component tags when it's called will just get output in this slot.

packages/f-alert/src/components/Alert.vue
```diff
@@ -0,0 +1,176 @@
...
+                {{ heading }}
+            </h2>
+            <button
+                v-if="isDismissable"
+                type="button"
+                :class="[$style['c-alert-dismiss'], 'o-btn o-btn--icon']"
+                data-test-id="alert-dismiss"
+                @click="dismiss">
+                <cross-icon
+                    :class="[$style['c-icon--cross'], $style['c-alert-dismiss-icon']]"
```
ashleynolan: Doesn't look like you're using `c-icon--cross` anywhere atm?

packages/f-alert/stories/Alert.stories.js
```diff
@@ -0,0 +1,36 @@
...
+        locale="en-GB"
+        is-rounded
+        has-outline
+        is-page-content-wrapper>
+        <vue-alert
+            locale="en-GB"
+            type="success"
+            :isDismissable="true"
+            heading="Title of the alert">
+            <template v-slot:content>You can put any HTML here, buddy.</template>
```
ashleynolan: Don't need the template tag here if you change the component to just use the default slot tag

packages/f-alert/src/demo/Index.vue
```diff
@@ -0,0 +1,31 @@
...
+    <div>
+        <meta
+            name="viewport"
+            content="width=device-width, initial-scale=1">
+        <vue-alert
+            locale="en-GB"
+            type="success"
+            :isDismissable="true"
+            heading="Title of the alert">
+            <template v-slot:content>You can put any HTML here!</template>
```
ashleynolan: Ditto to the comment below

packages/f-alert/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+# Changelog
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+
+v0.1.0
+------------------------------
+*October 13, 2020*
```
FayeCarter: This is the only thing I can see to change. Sorry

FayeCarter: Never mind

FayeCarter (approved):

DamianMullins (approved):

packages/f-alert/src/components/Alert.vue
```diff
@@ -0,0 +1,175 @@
...
+        data-test-id="alert-component"
+        role="alert">
+        <div
+            :class="$style['c-alert-headingContainer']">
+            <component
+                :is="icon"
+                :class="[$style['c-alert-icon'],
+                         $style[`c-alert-icon--${type}`]]"
+                data-test-id="alert-icon" />
+            <!-- TODO: make this dynamic. See https://skipthedishes.atlassian.net/browse/WCB-1219 -->
```
xander-marjoram, line 15: I don't think infosec will like this

PR #365, opened 2020-10-13
f-registration@v0.37.0

luke-andrew-parry:
This PR makes a number of styling tweaks to align closer to the Pie designs. There will be further changes to follow when amendments have been made to the `f-form-field` component.

## UI Review Checks
iPhone (iOS14 Safari)

image

Google Pixel 4 (Android 11 Chrome)

image

## Browsers Tested

- [x] Chrome (latest)
- [x] Mobile


packages/f-registration/src/components/Registration.vue
```diff
@@ -472,18 +471,12 @@ export default {
     margin-top: spacing(x2);
 }
 
-.c-loginLink {
+.c-loginLink, .c-legal-hyperlinks {
```
ashleynolan: Smal NIT: Can you put these selectors on separate lines?  Just makes it a bit clearer

packages/f-registration/src/demo/Index.vue
```diff
@@ -10,8 +10,7 @@
         <registration
             locale="en-GB"
             :login-settings="{
-                preLinkText: 'Already have an account?',
```
ashleynolan: Is it worth getting rid of the reference to this in the template (.vue file)

packages/f-registration/src/components/Button.vue
```diff
@@ -41,9 +41,9 @@ export default {
 
 $btn-default-bgColor                : $grey--lighter;
 $btn-default-text-colour            : $grey--dark;
-$btn-default-weight                 : 500;
+$btn-default-weight                 : bold;
```
ashleynolan: Can you change this to `$font-weight-bold`? That links to a variable that [handles this in fozzie](https://github.com/justeat/fozzie/blob/master/src/scss/settings/_variables.scss#L57)

packages/f-registration/src/components/Button.vue
```diff
@@ -41,9 +41,9 @@ export default {
...
 $btn-default-height                 : 2.6;
-$btn-default-font-size              : body-l;
+$btn-default-font-size              : 20px;
```
ashleynolan: Should use `heading-s` here (so it links into the font-map keys in fozzie). I'm pretty sure you could then just remove the line-height declation on line 86, as the mixin below should output `line-height: 1.5` in the styles when it's called.

packages/f-registration/src/components/Registration.vue
```diff
@@ -5,17 +5,16 @@
             :data-theme-registration="theme"
             :card-heading="title"
             is-rounded
-            has-outline
             is-page-content-wrapper
             card-heading-position="center"
             data-test-id="registration-component"
             :class="$style['c-card-padding']">
-            <p
+            <div
```
ashleynolan: It's more semantic to keep this text in a paragraph tag (as links with no block tag `p` don't always read correctly to screenreaders).

is this change style related?

packages/f-registration/src/components/Registration.vue
```diff
@@ -164,14 +163,15 @@
...
+                    class="submit-button-padding"
                     data-test-id="create-account-submit-button"
                     button-style="primary"
                     is-full-width
                     :disabled="shouldDisableCreateAccountButton">
                     {{ buttonText }}
                 </form-button>
             </form>
-            <p :class="$style['c-legal-hyperlinks']">
+            <div :class="$style['c-legal-hyperlinks']">
```
ashleynolan: Same here as above – these should stay as `p` tags ideally. If it's style related, we probably need to look at the styles and change those a little.

packages/f-registration/src/components/Registration.vue
```diff
@@ -472,31 +472,37 @@ export default {
...
-    }
-}
-
+.c-loginLink,
 .c-legal-hyperlinks {
+    text-align: center;
     a {
         color: $blue;
         text-decoration: none;
+        font-weight: bold;
```
ashleynolan: Should use `$font-weight-bold` here instead of `bold`

packages/f-registration/src/components/Registration.vue
```diff
@@ -472,31 +472,37 @@ export default {
...
     }
 }
 
 .c-card-padding {
-    padding-top: 30px;
+    padding-top: 55px;
```
ashleynolan: Probably worth checking with Tom, but is this padding specific to the registration component?

packages/f-registration/src/components/Registration.vue
```diff
@@ -472,31 +472,37 @@ export default {
...
+    padding-bottom: 50px;
+
+    @include media('<mid') {
+        padding-bottom: 32px;
```
ashleynolan: Can use `spacing(x4)` here instead of `32px`

packages/f-registration/src/components/Registration.vue
```diff
@@ -472,31 +472,37 @@ export default {
...
         color: $blue;
         text-decoration: none;
+        font-weight: bold;
     }
 }
 
 .c-card-padding {
-    padding-top: 30px;
+    padding-top: 55px;
+    padding-bottom: 50px;
```
ashleynolan: We tend to use padding as a multiple of 8px throughout the site (defined by design). So this would likely either be `spacing(x6)` or  `spacing(x7)` (for 48px or 56px) depending on which fits the component best.

packages/f-registration/src/components/Registration.vue
```diff
@@ -472,31 +472,37 @@ export default {
...
+        padding-bottom: 32px;
+    }
 }
 
 .bag-icon {
     width: 97px;
     height: 78px;
+    top: 80px;
+    left: 20px;
+    position: relative;
```
ashleynolan: I'd be tempted to make this icon `position: absolute` and pass the icon inside the card component (as `absolute` is a bit less flakey to change than relatively positioned elements).

packages/f-registration/src/components/Registration.vue
```diff
@@ -472,31 +472,37 @@ export default {
...
+    @include media('<mid') {
```
ashleynolan, line 511: Can you put this media query below the margin and display lines below (just our convention in the styles)

packages/f-registration/src/components/Registration.vue
luke-andrew-parry: Yeah, it's style-related as it seems to add a bunch of padding to `p` tags. I'll revert and we can address the styles assigned to these.

packages/f-registration/src/components/Registration.vue
luke-andrew-parry: Yeah, it's style-related as it seems to add a bunch of padding to `p` tags.

packages/f-registration/src/components/Registration.vue
luke-andrew-parry: I added this in to give some extra space between the bag icon and the title.

packages/f-registration/src/components/Registration.vue
luke-andrew-parry: Sure,  I'll have a look at this.

luke-andrew-parry: @ashleynolan I've had a look at changing this to be absolute and it's turning out to be quite fiddly and from what I can see requiring a number of different top/left values base on screen size. 

Did you have a specific solution in mind?

packages/f-registration/src/components/Registration.vue
luke-andrew-parry: Yeah, it's style-related as it seems to add a bunch of padding to `p` tags. I'll revert and we can address the styles assigned to these.

packages/f-registration/src/tenants/en-GB.js
```diff
@@ -26,7 +26,7 @@ export default {
     },
 
     labels: {
-        createAccountTitle: 'Create an account',
+        createAccountTitle: 'Create account',
```
adamashton, line 29: Is this coming from the zeplin designs? The spreadsheet I used might have been wrong again.

luke-andrew-parry: Yep. I'd get access to the zeplin designs and use that as the source of truth.

adamashton (dismissed):

packages/f-registration/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.36.0
+------------------------------
+*October 16, 2020*
```
ashleynolan: Super NIT – can you bump the date and put a line break between the date and the first title on line 9?

packages/f-registration/src/demo/Index.vue
```diff
@@ -29,6 +29,7 @@ export default {
 
 body {
     margin: 0;
+    background: #f9fafb;
```
ashleynolan: Can this styling not be applied to the registration component? You could potentially give the outside div in `Registration.vue` a classname `c-registration` and then put the background colour on that I think.

packages/f-registration/src/demo/Index.vue
```diff
@@ -29,6 +29,7 @@ export default {
...
 }
 
 .registration-form {
```
ashleynolan: Similarly for this – you could add a class `c-registration-form` on the form element in `Registration.vue` and put this spacing on that class.

packages/f-registration/src/demo/Index.vue
luke-andrew-parry: These styles aren't related to the registration component. The background colour is the greyish colour for the containing page. I've moved the other style into the component though.

simonsandersje (dismissed):

AlixRichards (dismissed):

tommcclean (dismissed):

packages/f-registration/src/components/Registration.vue
```diff
@@ -1,15 +1,15 @@
 <template>
-    <div>
-        <bag-celebrate-icon :class="$style['bag-icon']" />
+    <div
+        :class="$style['c-registration-form']">
```
ashleynolan: Could this just be `c-registration`? (the extra dash in our naming scheme implies there's a parent class of `c-registration` above `c-registration-form`)

packages/f-registration/src/components/Registration.vue
```diff
@@ -165,6 +165,7 @@
                 </form-field>
 
                 <form-button
+                    class="submit-button-padding"
```
ashleynolan: I'd go with `c-registration-btn` for this (here's our [naming scheme for classes](https://fozzie.just-eat.com/documentation/css/css-naming) too – dashes indicate element relationship and we use camelCase for joined words)

ashleynolan (commented):
Couple of super small classname suggestions – then LGTM

packages/f-registration/src/components/Registration.vue
luke-andrew-parry: I actually got rid of this style and forgot to remove the `class` attribute.

ashleynolan (changes requested):
Couple of super small classname suggestions – then LGTM

ashleynolan (approved):

AlixRichards (approved):

PR #371, opened 2020-10-15
f-registration@0.35.0 - Move text into translations file

adamashton:
v0.35.0
*October 16, 2020*
### Added
- Moved text into translations file 
- Support for locale en-GB

### Changed
- Prop for showing login link changed to boolean called `showLoginLink`

## UI Review Checks

image

## Browsers Tested

- [X] Chrome (latest)

### DIFF
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -13,7 +13,8 @@
             <p
                 v-if="shouldShowLoginLink"
                 :class="$style['c-loginLink']"
-                data-test-id="create-account-login-link">
+                data-test-id="create-account-login-link"
+                @click="visitLoginPage">
                 <a :href="copy.navLinks.login.url">{{ copy.navLinks.login.text }}</a>
             </p>
             <form
@@ -371,6 +372,10 @@ export default {
                 this.$emit(EventNames.CreateAccountStart);
                 this.formStarted = true;
             }
+        },
+
+        visitLoginPage () {
+            this.$emit(EventNames.VisitLoginPage);
         },
 
         formFieldBlur (field) {
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -11,7 +11,7 @@
             data-test-id="registration-component"
             :class="$style['c-card-padding']">
             <p
-                v-if="shouldShowLoginLink"
+                v-if="showLoginLink"
                 :class="$style['c-loginLink']"
                 data-test-id="create-account-login-link"
                 @click="visitLoginPage">
@@ -263,7 +263,7 @@ export default {
         },
         showLoginLink: {
             type: Boolean,
-            default: false
+            default: true
         }
     },
 
@@ -326,9 +326,6 @@ export default {
         },
         shouldShowPasswordMinLengthError () {
             return !this.$v.password.minLength && this.$v.password.$dirty;
-        },
-        shouldShowLoginLink () {
-            return this.showLoginLink;
         },
         tenant () {
             return {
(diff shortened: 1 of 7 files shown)


packages/f-registration/src/tenants/en-GB.js
```diff
@@ -18,6 +18,44 @@ export default {
             text: 'Cookie Policy',
             url: '/info/cookies-policy',
             suffix: '.'
+        },
+        alreadyHaveAnAccount: {
+            text: 'Already on Just Eat?'
```
luke-andrew-parry: Do you think we should have the `login` link here instead of passing it into the component? I can't see it changing anytime soon.

packages/f-registration/src/tenants/en-GB.js
```diff
@@ -18,6 +18,44 @@ export default {
             text: 'Cookie Policy',
             url: '/info/cookies-policy',
             suffix: '.'
+        },
+        alreadyHaveAnAccount: {
```
luke-andrew-parry: Also, maybe rename the object to `login`?

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -63,24 +57,7 @@ describe('Registration', () => {
...
-                }
-            });
-
-            const loginLink = wrapper.find("[data-test-id='create-account-login-link']");
-
-            expect(loginLink.exists()).toBe(false);
-        });
-
-        it('shoud not show the login link if loginSettings prop not set', () => {
+        it('shoud not show the login link if login url prop not set', () => {
```
ashleynolan: Small typo – `should`

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -47,14 +43,12 @@ describe('Registration', () => {
             expect(loginLink.exists()).toBe(true);
         });
 
-        it('should not show the login link if loginSettings prop set but linkText not set.', () => {
+        it('should not show the login link if login url not set.', () => {
```
ashleynolan: Would change this to be `should not show the login link if login url is an empty string` (as it is technically set).

packages/f-registration/src/components/Registration.vue
```diff
@@ -258,19 +258,19 @@ export default {
         },
         title: {
```
ashleynolan: Do you need these props now (`title` and `buttonText`)? Would imagine they are only really needed if you'd like to override the translation files. Happy for them to stay there if you think that flexibility could be useful though.

packages/f-registration/src/tenants/en-GB.js
adamashton: Sure 👍

packages/f-registration/src/components/Registration.vue
adamashton: Happy to remove them actually

cmyhill-justeat (dismissed):

smarkey (dismissed):

packages/f-registration/src/components/Registration.vue
```diff
@@ -256,21 +257,13 @@ export default {
...
         createAccountUrl: {
             type: String,
             required: true
         },
-        loginSettings: {
-            type: Object,
-            default: () => {}
+        showLoginLink: {
+            type: Boolean,
+            default: false
```
luke-andrew-parry: Maybe default this to `true`? I think it's likely going to be the _exception_ that we don't want to show the login link.

packages/f-registration/src/components/Registration.vue
```diff
@@ -335,7 +328,7 @@ export default {
             return !this.$v.password.minLength && this.$v.password.$dirty;
         },
         shouldShowLoginLink () {
```
kevinrodrigues: Do you need to now have this computed prop now that we're only checking the `showLoginLink` prop?

adamashton: Yeh, good point

luke-andrew-parry (approved):

cmyhill-justeat (approved):

PR #372, opened 2020-10-16
f-checkout@0.3.0 – Selector to display label above time when time selected

FayeCarter:
At the moment when a time is selected it is displayed directly on top of the label text. This fix is to move the label above the selected time

*Selector is currently a placeholder to to build the layout for `f-checkout`. All values are currently hardcoded. A new fozzie component will be built to replace it.*

v0.3.0
*October 21, 2020*

### Changed
- Selector to hide label when time selected.

### Added
- Selector unit tests

## UI Review Checks
| Empty |Time Selected |
|:-------------------------:|:-------------------------:|
| image | image|

### DIFF
--- a/packages/f-checkout/CHANGELOG.md
+++ b/packages/f-checkout/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.3.0
 ------------------------------
-*October 20, 2020*
+*October 21, 2020*
 
 ### Changed
 - Selector to hide label when time selected.
--- a/packages/f-checkout/CHANGELOG.md
+++ b/packages/f-checkout/CHANGELOG.md
@@ -3,6 +3,7 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
 v0.3.0
 ------------------------------
 *October 20, 2020*
--- a/packages/f-checkout/CHANGELOG.md
+++ b/packages/f-checkout/CHANGELOG.md
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.2.1
+------------------------------
+*October 16, 2020*
+
+### Changed
+- Selector to hide label when time selected.
+
+
 v0.2.0
 ------------------------------
 *October 15, 2020*
--- a/packages/f-checkout/CHANGELOG.md
+++ b/packages/f-checkout/CHANGELOG.md
@@ -5,11 +5,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.2.1
 ------------------------------
-*October 16, 2020*
+*October 19, 2020*
 
 ### Changed
 - Selector to hide label when time selected.
 
+### Added
+- Selector unit tests
 
 v0.2.0
 ------------------------------
--- a/packages/f-checkout/CHANGELOG.md
+++ b/packages/f-checkout/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.3.0
 ------------------------------
-*October 19, 2020*
+*October 20, 2020*
 
 ### Changed
 - Selector to hide label when time selected.
--- a/packages/f-checkout/CHANGELOG.md
+++ b/packages/f-checkout/CHANGELOG.md
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-v0.2.1
+v0.3.0
 ------------------------------
 *October 19, 2020*
 
--- a/packages/f-checkout/CHANGELOG.md
+++ b/packages/f-checkout/CHANGELOG.md
@@ -13,6 +13,7 @@ v0.3.0
 ### Added
 - Selector unit tests
 
+
 v0.2.0
 ------------------------------
 *October 15, 2020*
--- a/packages/f-checkout/package.json
+++ b/packages/f-checkout/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-checkout",
   "description": "Fozzie Checkout – Fozzie Checkout Component",
-  "version": "0.2.1",
+  "version": "0.3.0",
   "main": "dist/f-checkout.umd.min.js",
   "files": [
     "dist",
--- a/packages/f-checkout/package.json
+++ b/packages/f-checkout/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-checkout",
   "description": "Fozzie Checkout – Fozzie Checkout Component",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "main": "dist/f-checkout.umd.min.js",
   "files": [
     "dist",
--- a/packages/f-checkout/src/components/Selector.vue
+++ b/packages/f-checkout/src/components/Selector.vue
@@ -1,19 +1,19 @@
 <template>
     <div
+        data-test-id="form-select"
         :class="[
-            'o-form-select',
-            { 'o-form-select--float': selectedTime }
-        ]"
-        data-test-id="form-select">
+            $style['o-form-select'],
+            (!selectedTime ? '' : $style['o-form-select--float'])
+        ]">
         <label
             for="delivery-time"
-            class="o-form-select-label">
+            :class="$style['o-form-select-label']">
             Delivery time
         </label>
         <select
             id="delivery-time"
             v-model="selectedTime"
-            class="o-form-select-input">
+            :class="$style['o-form-select-input']">
             <option
                 v-for="(time, index) in deliveryTimes"
                 :key="index"
@@ -41,7 +41,7 @@ export default {
 };
 </script>
 
-<style lang="scss">
+<style lang="scss" module>
 $form-label-colour                        : $grey--dark;
 $form-input-colour                        : $color-text;
 $form-input-bg                            : $white;
--- a/packages/f-checkout/src/components/Selector.vue
+++ b/packages/f-checkout/src/components/Selector.vue
@@ -76,7 +76,7 @@ $form-input-borderColour--focus           : $grey--dark;
     .o-form-select-input {
         height: 100%;
         width: 100%;
-        padding: 3px;
+        padding: spacing(x0.5);
         border: none;
         cursor: pointer;
         color: $color-text;
@@ -90,8 +90,8 @@ $form-input-borderColour--focus           : $grey--dark;
  */
 .o-form-select--float {
     .o-form-select-label {
-        @include font-size(body-s, false);
-        line-height: 1.29;
+        @include font-size(body-s);
+        // line-height: 1.29;
         top: 15px;
     }
 
--- a/packages/f-checkout/src/components/Selector.vue
+++ b/packages/f-checkout/src/components/Selector.vue
@@ -76,7 +76,7 @@ $form-input-borderColour--focus           : $grey--dark;
     .o-selector-input {
         height: 100%;
         width: 100%;
-        padding: 2.8px;
+        padding: 3px;
         border: none;
         cursor: pointer;
         color: $color-text;
--- a/packages/f-checkout/src/components/Selector.vue
+++ b/packages/f-checkout/src/components/Selector.vue
@@ -1,19 +1,19 @@
 <template>
     <div
         :class="[
-            $style['o-selector'],
-            (!selectedTime ? '' : 'o-selector--float')
+            'o-selector',
+            { 'o-selector--float': selectedTime }
         ]"
         data-test-id="selector">
         <label
             for="delivery-time"
-            :class="$style['o-selector-label']">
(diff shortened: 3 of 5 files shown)


packages/f-checkout/src/components/Selector.vue
```diff
@@ -1,20 +1,25 @@
...
-                v-for="(time, index) in deliveryTimes"
-                :key="index"
-                :value="time">
-                {{ time }}
-            </option>
-        </select>
+    <div>
+        <div :class="$style['o-selector']">
+            <label
+                v-if="!selectedTime"
```
kevinrodrigues: Maybe a visibility hidden css style would be better here (should be a helper available in fozzie for this), I think keeping the label that's associated with the select is important accessibility wise (screen readers). You can still make use of the `selectedTime` value but just to toggle the visibility via css instead.

FayeCarter: OK, cool, I'll take a look. Thank you!

simonsandersje (dismissed):

packages/f-checkout/src/components/Selector.vue
FayeCarter: @kevinrodrigues I've added a class. Is this the one you were thinking of?
https://github.com/justeat/fozzie/blob/master/src/scss/trumps/_utilities.scss#L58

packages/f-checkout/src/components/Selector.vue
```diff
@@ -1,20 +1,27 @@
...
-                {{ time }}
-            </option>
-        </select>
+    <div>
+        <div :class="$style['o-selector']">
+            <label
+                for="delivery-time"
+                :class="[
+                    $style['o-selector-label'],
+                    (!selectedTime ? '' : 'is-visuallyHidden')
```
kevinrodrigues: You should be able to just do: `!!selectedTime: 'is-visuallyHidden'`

packages/f-checkout/src/components/tests/Selector.test.js
```diff
@@ -0,0 +1,27 @@
...
+    const propsData = {};
+
+    it('should be defined', () => {
+        const wrapper = shallowMount(Selector, { propsData });
+        expect(wrapper.exists()).toBe(true);
+    });
+
+    describe('data ::', () => {
+        describe('selectedTime ::', () => {
+            it('should add class to visually hide label when not null', async () => {
```
kevinrodrigues: You could also have a test for when the label should display? i.e when `selectedTime: null`?

kevinrodrigues:
More out of interest and maybe something I should have asked first. How comes we want to hide the label when a time is selected? Would it not just be showing as it currently does on live? Seems to stay regardless of the time selection.

packages/f-checkout/src/components/Selector.vue
FayeCarter: I was trying that originally and it kept breaking, but I'll have a look

packages/f-checkout/src/components/tests/Selector.test.js
FayeCarter: Good point

FayeCarter:
That's a great point. I went for hiding it so that it visually matches how the form fields are currently working. But I'm not sure if that was the right decision now.
Rebuilding now 😄

packages/f-checkout/src/components/Selector.vue
ashleynolan: Would it not be `!!selectedTime` if it's meant to only show when true?

packages/f-checkout/src/components/Selector.vue
```diff
@@ -39,32 +51,47 @@ $form-input-borderColour                  : $grey--light;
...
+    border-radius: $form-input-borderRadius;
 
     .o-selector-label {
         display: block;
-        color: $form-label-colour;
         position: absolute;
         top: 50%;
-        left: 0.5rem;
         transform: translateY(-50%);
+        padding: 0.5rem;
```
xander-marjoram: Should we be using a pixel value for padding instead?

ross-gratton (commented):
General question: Is this component reusable?
It looks like theres a number of hard coded references to selected time in IDs and variables. This leads me to think that it's not a reusable component.
If that's the case the name of the name of the component itself is probably too generic.

packages/f-checkout/src/components/Selector.vue
FayeCarter: Can do

FayeCarter:
@ross-gratton At the moment it's not reusable, but there's a ticket in place for it to move it to a reusable component and new fozzie-component. 
At the moment this is just a placeholder until the new one gets built so everything is hardcoded. Happy to change names if you want though?

xander-marjoram (dismissed):

packages/f-checkout/src/components/Selector.vue
ashleynolan: I'm going to be super annoying – I'd probably try and make this an `em` if you can to make it relative to the font-size of the element. Will make it slightly more flexible if the size of the element were to change in future.

packages/f-checkout/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-checkout",
   "description": "Fozzie Checkout – Fozzie Checkout Component",
-  "version": "0.2.0",
+  "version": "0.2.1",
```
ashleynolan: Should be a minor version bump as patches should only be for really tiny changes (or small bug fixes)

FayeCarter: Ahh ok

packages/f-checkout/src/components/Selector.vue
```diff
@@ -1,13 +1,19 @@
 <template>
-    <div :class="$style['o-selector']">
+    <div
+        :class="[
+            'o-selector',
```
ashleynolan: Don't need to do it as part of this PR (unless you really want to) but I'd update this to be `o-select`, mainly as objects (`o-`) should link to HTML tags (so this one would be the select tag).

Could also be `o-form-select` if you wanted to make it super explicit that it's a form select (might be worth looking at other component libraries to see what they call these sort of components).

FayeCarter: Happy to change this now

packages/f-checkout/src/components/Selector.vue
```diff
@@ -39,32 +51,47 @@ $form-input-borderColour                  : $grey--light;
...
-        border: $form-input-borderWidth solid $form-input-borderColour;
-        border-radius: $form-input-borderRadius;
+        padding: 3px;
+        border: none;
         cursor: pointer;
+        color: $color-text;
+    }
+}
+
+.o-selector--float {
```
ashleynolan: Always a good idea to add a small comment above modifier to explain what they do. Just saves someone having to try and work out what the styling is doing to the base component.

DamianMullins:
@FayeCarter On the re-usable component question from @ross-gratton, I'm wondering how checkout is going to be used outside of the one place on the consumer-facing website?

tommcclean (dismissed):

packages/f-checkout/CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
...
+v0.3.0
+------------------------------
+*October 20, 2020*
+
+### Changed
+- Selector to hide label when time selected.
+
+### Added
+- Selector unit tests
+
```
xander-marjoram, line 16: The nittiest of nits: can you make sure you have two empty lines between entries please

FayeCarter: 😄 I'll add this in

packages/f-checkout/src/components/tests/Selector.test.js
```diff
@@ -0,0 +1,36 @@
...
+                const selector = wrapper.find("[data-test-id='form-select']");
+
+                wrapper.setData({ selectedTime: 'testTime' });
+                await wrapper.vm.$nextTick();
+
+                // Assert
+                expect(selector.classes()).toContain('o-form-select--float');
+            });
+
+            it('should remove class to display label centrally when  null', () => {
```
SachaZvetelman: Nit: extra whitespace before `null`.

packages/f-checkout/src/components/Selector.vue
```diff
@@ -38,33 +50,53 @@ $form-input-borderWidth                   : 1px;
...
+
+/**
+ * Modifier – .o-form-select--float
+ *
+ * Moves the select label to sit above chosen option
+ */
+.o-form-select--float {
+    .o-form-select-label {
+        @include font-size(body-s, false);
+        line-height: 1.29;
```
SachaZvetelman: This seems a rather specific `line-height`. Is it just random or it comes from an existing style we're duplicating?

packages/f-checkout/src/components/Selector.vue
```diff
@@ -38,33 +50,53 @@ $form-input-borderWidth                   : 1px;
...
 
-    .o-selector-input {
-        height: 53px;
+    .o-form-select-input {
+        height: 100%;
         width: 100%;
-        padding: 0.5rem;
-        border: $form-input-borderWidth solid $form-input-borderColour;
-        border-radius: $form-input-borderRadius;
+        padding: 3px;
```
SachaZvetelman: Could we use `spacing()` here?

SachaZvetelman:
@DamianMullins, I think the answer to your question at this stage is "we don't know". But we're building components that may be reusable in different parts in this monorepo. `f-registration` is another case. 
Probably @ashleynolan will be able to give you a better answer, though.

DamianMullins:
@SachaZvetelman OK, I thought the general process was to move things into Fozzie once we identify that they are being used in more than one place is all. So, for example, f-mega-modal was first developed and proved to work well in MenuWeb, following that a few engineers expressed interest in using it outside of menu so I (eventually!) moved it into Fozzie.

I'm not saying this approach is wrong or needs to change by the way, I'm just curious about how the decision was made so I can keep up to date with what is happening in the web chapter 😸

packages/f-checkout/src/components/Selector.vue
FayeCarter: Hadn't realised that there was a `spacing(x0.5)`. I'll add it in.

packages/f-checkout/src/components/Selector.vue
FayeCarter: This was to match a zeplin design. But putting it back to the standard appears to not change it at all. So I will remove it 😂

tommcclean:
We want to make sure components are written in the mono repo so that we can keep magikarp thin, lean and mean. And make sure that we don't have so much code going into Magi, that we get conflict hell. Having components in the mono repo clearly gives ownership to those who own it; whereas munged into Magi, its hard to know who owns what

DamianMullins:
@tommcclean, In that case, I'm not sure moving those components into the public domain is the best idea? Also, aren't you just moving the problem out of Magi and into this repo instead? How do you know who owns what?

There's already a lot going on in this repo, it's hard enough to keep up as it is - multiple PRs touching the same project with conflicting versions, sweeping changes affecting all components, new components being added etc. 

It feels like the repo is losing focus a bit but that might just be me, maybe I'm a little out of the loop or something 😄

tommcclean:
Yeah some of the same concerns with me around it growing; I think thats the point of a mono repo though, probably just the organisation of stuff needs breaking out? 

When you look at other mono repo's in JE (Like the IOS team's, for instance) they organise the stuff differently, but are much larger overall than ours and have more people working in them. (2K pr's opposed to our 300 too)

Cos the components are self contained, you shouldn't end up with conflicts when a team wants to update their component as they only change the component directory.

The alternative would be building a monolith in Magikarp; which we already know we don't want to do. Or having mono repo's for each team, but that seems against the point; either way not my decision just chiming in :D

packages/f-checkout/src/components/tests/Selector.test.js
```diff
@@ -0,0 +1,36 @@
...
+    });
+
+    describe('data ::', () => {
+        describe('selectedTime ::', () => {
+            it('should add class to display label above option when not null', async () => {
+                // Arrange & Act
+                const wrapper = shallowMount(Selector, { propsData });
+                const selector = wrapper.find("[data-test-id='form-select']");
+
+                wrapper.setData({ selectedTime: 'testTime' });
```
SachaZvetelman: Nit: isn't this the `// Act` part actually?

SachaZvetelman (dismissed):

SachaZvetelman:
Ah, I've approved it, but actually, I think the version is not right. If it's a fix, and we're not adding new functionality, we should change the third number of the version instead.

packages/f-checkout/src/components/Selector.vue
```diff
@@ -25,11 +31,17 @@ export default {
...
+    data () {
+        return {
+            selectedTime: null
+        };
     }
 };
 </script>
 
-<style lang="scss" module>
+<style lang="scss">
```
jamieomaguire: do we scope inline template styles?

ashleynolan: We do usually – what's the reason for not using CSS modules for this? (if you were seeing any issues, let me know and can help resolve them)

ashleynolan (commented):
Looks good – just one small comment re the `modules` attribute

jamieomaguire (dismissed):

packages/f-checkout/src/components/Selector.vue
FayeCarter: It was breaking the tests. Which I'm now realising was probably not a good reason to change it.
I have a test that checks if a modifier class is applied. The class works correctly in the browser, but when testing it wouldn't bind any classes.
I'm doing another PR on this component at the moment so I can work out how to get that working properly in the next PR if that's ok, or have a look in this one.

FayeCarter: Hey @ashleynolan and @jamieomaguire! modules are now back in place.

ashleynolan (approved):

SachaZvetelman (approved):

PR #374, opened 2020-10-19
f-metadata@3.0.0-beta.3 - Add Grouped output option via alternative callback / API to f-metadata

matthewhardern:
Request was made to allow the braze dispatcher library to support grouped by title cards, I have modified the code and added a new attribute `groups`.  I have added groups variable to `ContentCards` service, which `output` now uses. 

The new output is only applied to the new callback `handleContentCardsGrouped`. If this format is wanted when consuming the f-metadata API you must add this to the callback functions list. 

Added Test to cover the grouping output for `ContentCards` service and updated the `BrazeDispatcher` tests to account for the grouped output inside the `contentCardsHandler` function. 

I have changed the `arrangeCardsByTitles` to include `startsWith` lodash function (startsWith is es15 only so if our babel transpile allows es15 to es5 then I can sub in the es15 version). It is only imported in the f-metadata component, individually loaded. I included this in the as a result of some card json supplied to me, which showed `Terms_And_Conditions_Card_2` being used as type rather than the previous checked value `Terms_And_Conditions_Card`. If this is not correct let me know and I will remove it / put it back to just looking for `Terms_And_Conditions_Card` and `Header_Card`. 

---

## UI Review Checks

### DIFF
--- a/packages/f-metadata/src/BrazeDispatcher.js
+++ b/packages/f-metadata/src/BrazeDispatcher.js
@@ -43,19 +43,30 @@ function interceptInAppMessagesHandler (message) {
 function contentCardsHandler (postCardsAppboy) {
     this.refreshRequested = false;
 
-    const {
-        cards,
-        rawCards
-    } = new ContentCards(postCardsAppboy, { enabledCardTypes: this.dispatcherOptions.enabledCardTypes })
+    // don't output straight away as we need to create grouped output as well as the normal output
+    const preCallbackSelection = new ContentCards(postCardsAppboy, { enabledCardTypes: this.dispatcherOptions.enabledCardTypes })
         .removeDuplicateContentCards()
         .filterCards()
-        .getTitleCard()
-        .arrangeCardsByTitles()
-        .output();
+        .getTitleCard();
 
-    this.cardsCallbacks.forEach(callback => callback(cards));
-
-    this.rawCards = rawCards;
+    if (this.cardsCallbacks.length > 0) {
+        // cards and raw cards
+        const { cards, rawCards } = preCallbackSelection.output();
+        // cards in ungrouped format returned via callback
+        this.cardsCallbacks.forEach(callback => callback(cards));
+        // set raw cards
+        this.rawCards = rawCards;
+    }
+    if (this.groupedCardsCallback.length > 0) {
+        // grouped cards
+        const { groups, rawCards } = preCallbackSelection
+            .arrangeCardsByTitles()
+            .outputGroups();
+        // cards in grouped format returned via callback
+        this.groupedCardsCallback.forEach(callback => callback(groups));
+        // set raw cards only if above is not set
+        if (this.rawCards.length === 0) this.rawCards = rawCards;
+    }
 
     this.rawCardIndex = Object.fromEntries(this.rawCards.map((rawCard, idx) => [rawCard.id, idx]));
 }
@@ -95,6 +106,7 @@ class BrazeDispatcher {
         this.dispatcherOptions = null;
 
         this.cardsCallbacks = [];
+        this.groupedCardsCallback = [];
         this.inAppMessageClickEventCallbacks = [];
         this.inAppMessagesCallbacks = [];
 
@@ -147,6 +159,7 @@ class BrazeDispatcher {
         if (callbacks.interceptInAppMessageClickEvents) this.inAppMessageClickEventCallbacks.push(callbacks.interceptInAppMessageClickEvents);
         if (callbacks.interceptInAppMessages) this.inAppMessagesCallbacks.push(callbacks.interceptInAppMessages);
         if (callbacks.handleContentCards) this.cardsCallbacks.push(callbacks.handleContentCards);
+        if (callbacks.handleContentCardsGrouped) this.groupedCardsCallback.push(callbacks.handleContentCardsGrouped);
 
         // Note that appBoyPromise will not be set if this is the first time running this method -
         // this is a guard against initialise() being called more than once, and attempting to
--- a/packages/f-metadata/src/BrazeDispatcher.js
+++ b/packages/f-metadata/src/BrazeDispatcher.js
@@ -36,6 +36,25 @@ function interceptInAppMessagesHandler (message) {
 }
 
 /**
+ * Takes Callbacks and fires each one with appropriate cards.
+ * @param callbacks
+ * @param contentCardsSelection
+ * @param group
+ * @returns {*[]|*}
+ */
+function cardCallbackHandler (callbacks, contentCardsSelection, group = false) {
+    if (callbacks.length > 0) {
+        const { cards, rawCards } = group ? contentCardsSelection
+            .arrangeCardsByTitles()
+            .outputGroups() :
+            contentCardsSelection.output();
+        callbacks.forEach(callback => callback(cards));
+        return rawCards;
+    }
+    return [];
+}
+
+/**
  * Internal handler for content cards that dispatches to component-registered callbacks
  * @param postCardsAppboy
  * @this BrazeDispatcher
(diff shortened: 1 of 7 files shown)


packages/f-metadata/src/BrazeDispatcher.js
```diff
@@ -43,19 +43,30 @@ function interceptInAppMessagesHandler (message) {
...
     this.refreshRequested = false;
 
-    const {
-        cards,
-        rawCards
-    } = new ContentCards(postCardsAppboy, { enabledCardTypes: this.dispatcherOptions.enabledCardTypes })
+    // don't output straight away as we need to create grouped output as well as the normal output
+    const preCallbackSelection = new ContentCards(postCardsAppboy, { enabledCardTypes: this.dispatcherOptions.enabledCardTypes })
         .removeDuplicateContentCards()
         .filterCards()
```
matthewhardern, line 52: This function in contentCards service expects a brands parameter, should it be missing here?

packages/f-metadata/src/BrazeDispatcher.js
```diff
@@ -43,19 +43,30 @@ function interceptInAppMessagesHandler (message) {
...
+    if (this.groupedCardsCallback.length > 0) {
+        // grouped cards
+        const { groups, rawCards } = preCallbackSelection
+            .arrangeCardsByTitles()
+            .outputGroups();
+        // cards in grouped format returned via callback
+        this.groupedCardsCallback.forEach(callback => callback(groups));
+        // set raw cards only if above is not set
+        if (this.rawCards.length === 0) this.rawCards = rawCards;
+    }
```
SachaZvetelman: It would be good to remove these conditionals from this function, in my humble opinion (maybe extract this to a separate function). At the same time, there's some repeated functionality between the two blocks, so maybe we could extract this into a common function?
I'd remove the comment as well 😄

packages/f-metadata/src/services/_tests/contentCard.service.test.js
```diff
@@ -302,4 +302,78 @@ describe('`contentCardService`', () => {
             });
         });
     });
+
+    describe('`arrangeCardsByTitles` method', () => {
+        it('should group content cards by type where type starts with either `Header_Card` or `Terms_And_Conditions_Card`', () => {
+            // Arrange
+            const cards = [
```
SachaZvetelman: Could we please extract this out of the test so the test ends up being smaller, concise and more readable?

packages/f-metadata/src/services/_tests/contentCard.service.test.js
```diff
@@ -302,4 +302,78 @@ describe('`contentCardService`', () => {
...
+                }
+            ];
+            const service = new ContentCardService({ cards });
+
+            // Act
+            const {
+                groups: contentCards
+            } = service.arrangeCardsByTitles().outputGroups();
+
+            expect(contentCards).toHaveLength(4);
```
SachaZvetelman, line 376: Nit: missing `// Assert`

packages/f-metadata/src/tests/BrazeDispatcher.test.js
```diff
@@ -257,9 +275,13 @@ describe('BrazeDispatcher operation', () => {
...
                     rawCards
                 });
+                ContentCards.prototype.outputGroups.mockReturnValue({
+                    groups: groupedCards,
+                    rawCards
+                });
             });
 
-            it('should instantiate ContentCards and call required methods in order', () => {
+            it('should instantiate ContentCards and call required methods in order up to output method', () => {
```
SachaZvetelman: I think we can be more specific with the test name here. Something like: `should instantiate ContentCards and call `filterCards` and `getTitleCard` or similar?

packages/f-metadata/src/tests/BrazeDispatcher.test.js
```diff
@@ -269,8 +291,24 @@ describe('BrazeDispatcher operation', () => {
...
-                    .toHaveBeenCalledBefore(contentCardsInstance.arrangeCardsByTitles);
+            });
+
+            it('should call the arrangeCardsByTitles method before outputGroups method is called', () => {
+                // Arrange & Act
+                contentCardsHandler();
+
+                const [contentCardsInstance] = ContentCards.mock.instances;
+
+                expect(contentCardsInstance.arrangeCardsByTitles)
```
SachaZvetelman, line 291: Same nit: missing `// Assert`

packages/f-metadata/src/BrazeDispatcher.js
matthewhardern: @SachaZvetelman Agree will do my best to make as much shared functionality as possible.

packages/f-metadata/src/services/_tests/contentCard.service.test.js
matthewhardern: yes can extract

packages/f-metadata/src/BrazeDispatcher.js
minrwhite: Good call - that should be reinstated at some point; happy for it to remain out while we're doing `beta` releases of this new version of the component, but worth raising a ticket on the backlog to look at this.

packages/f-metadata/src/BrazeDispatcher.js
```diff
@@ -43,19 +62,15 @@ function interceptInAppMessagesHandler (message) {
...
-        .getTitleCard()
-        .arrangeCardsByTitles()
-        .output();
-
-    this.cardsCallbacks.forEach(callback => callback(cards));
+        .getTitleCard();
 
-    this.rawCards = rawCards;
+    this.rawCards = cardCallbackHandler(this.cardsCallbacks, contentCardsSelection);
+    const rawCards = cardCallbackHandler(this.groupedCardsCallback, contentCardsSelection, true);
```
matthewhardern: @SachaZvetelman added new function to remove conditionals and with optional group parameter. Reduces duplication as much as I can. If all good I will make test for new function.

matthewhardern: resolving as outdated

packages/f-metadata/src/tests/BrazeDispatcher.test.js
matthewhardern: added asserts where needed

packages/f-metadata/src/tests/BrazeDispatcher.test.js
matthewhardern: updated to now include `[removeDuplicateContentCards, filterCards, getTitleCard, output]`

packages/f-metadata/src/services/_tests/contentCard.service.test.js
matthewhardern: I have moved them above the test

packages/f-metadata/src/BrazeDispatcher.js
matthewhardern: Having changed the approach this is now outdated, @SachaZvetelman I have changed this based on our chat to now be an attribute on output  return value.

packages/f-metadata/src/BrazeDispatcher.js
matthewhardern: I have created an issue for this in the project

packages/f-metadata/src/tests/BrazeDispatcher.test.js
```diff
@@ -273,15 +273,12 @@ describe('BrazeDispatcher operation', () => {
...
-                    rawCards
-                });
-                ContentCards.prototype.outputGroups.mockReturnValue({
                     groups: groupedCards,
                     rawCards
                 });
             });
 
-            it('should instantiate ContentCards and call required methods in order up to output method', () => {
+            it('should instantiate ContentCards and call required methods in order of [removeDuplicateContentCards, filterCards, getTitleCard, output]', () => {
```
minrwhite, line 281: _Great_ change! 💯

minrwhite (dismissed):

packages/f-metadata/src/tests/BrazeDispatcher.test.js
matthewhardern: @SachaZvetelman made a good point so I changed it

packages/f-metadata/src/services/_tests/contentCard.service.test.js
```diff
@@ -302,4 +302,78 @@ describe('`contentCardService`', () => {
...
+            },
+            {
+                title: 'Full terms and conditions here.',
+                extras: {
+                    order: '100',
+                    custom_card_type: 'Terms_And_Conditions_Card_2' // eslint-disable-line camelcase
+                }
+            }
+        ];
+        it('should group content cards by type where type starts with either `Header_Card` or `Terms_And_Conditions_Card`', () => {
```
SachaZvetelman, line 366: Ultra nit: could you please add an empty line before the `it`?

packages/f-metadata/src/services/_tests/contentCard.service.test.js
```diff
@@ -302,4 +302,78 @@ describe('`contentCardService`', () => {
...
+            // Arrange
+            const service = new ContentCardService({ cards });
+            // Act
```
SachaZvetelman, line 370: Same here. It helps in terms of readability.

SachaZvetelman (dismissed):

packages/f-metadata/src/services/_tests/contentCard.service.test.js
matthewhardern: sorted

SachaZvetelman (approved):

minrwhite (approved):

PR #375, opened 2020-10-20
f-alert@0.1.1 - Removed unwanted comment.

SachaZvetelman:
v0.1.1
*October 20, 2020*

### Changed
- Removing unwanted comment from the code.


FayeCarter (approved):

packages/f-alert/src/components/Alert.vue
```diff
@@ -12,7 +12,7 @@
                 :class="[$style['c-alert-icon'],
                          $style[`c-alert-icon--${type}`]]"
                 data-test-id="alert-icon" />
-            <!-- TODO: make this dynamic. See https://skipthedishes.atlassian.net/browse/WCB-1219 -->
+            <!-- TODO: make this dynamic. Find related ticket ;) -->
```
jamieomaguire, line 15: jokes aside, are people going to need to reference this ticket in the future? :P

jamieomaguire (approved):

packages/f-alert/src/components/Alert.vue
SachaZvetelman: We'll have to find it probably, yes. But we'll be working on it soon, so hopefully, we won't forget what ticket we are talking about 🤣

DamianMullins: If you already have a ticket for this do you need a comment at all? You could link to the relevant places in Jira? Just thinking about any future comments being added.

PR #376, opened 2020-10-20
f-checkout@0.5.0 - Add f-checkout component tests

siggerzz:
### Added
- New functionality to component object
- New component tests


packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -1,4 +1,68 @@
...
+
+const userNoteInput = () => $('[data-test-id="user-note"] textarea');
+
+const allergenLink = () => $('[data-test-id="allergy-button"]');
+
+const deliveryTimeDropdown = () => $('[data-test-id="delivery-time"]');
+
+const goToPaymentButton = () => $('[data-test-id="confirm-payment-submit-button"]');
+
 exports.waitForCheckoutComponent = () => checkoutComponent().waitForExist();
```
LTurns, line 18: more a q than a comment - do we need a test that the whole component is displayed if by default it must be when we test the other elements inside of it? Noticed we didn't do that for f-footer but happy to add in!

siggerzz: `waitForExist()` only checks for presence in the DOM, not that it's necessarily displayed, so I'd still have tests for visibility :)

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -4,10 +4,35 @@ describe('f-checkout component tests', () => {
...
         // Arrange
         browser.url('http://localhost:8080');
+        CheckoutComponent.waitForCheckoutComponent();
     })
 
     it('should display the f-checkout component', () => {
         // Assert
         expect(CheckoutComponent.isCheckoutComponentDisplayed()).toBe(true);
     });
+
```
LTurns, line 14: missing a few semi-colons if I'm nit-picking, but looks like we're waiting for an assert anyway

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
LTurns: ah okay awesome, makes sense!

LTurns (dismissed):

simonsandersje (dismissed):

packages/f-checkout/src/components/Selector.vue
```diff
@@ -67,7 +65,6 @@ $form-input-borderColour--focus           : $grey--dark;
     background-color: $form-input-bg;
     border: $form-input-borderWidth solid $form-input-borderColour;
     border-radius: $form-input-borderRadius;
-
```
ashleynolan: Good idea to keep these line breaks – makes the styles more readable (not sure if this is being auto-formatted by your code editor?)

siggerzz: Yeah, this was the result of a bad merge conflict. Will sort this out 👍

simonsandersje (approved):

LTurns (approved):

PR #378, opened 2020-10-20
f-registration@v0.36.0

luke-andrew-parry:
This PR adds handling for `403` responses when an account is created. This can happen when the authentication attempt on the newly created account gets either blocked by ravelin or a Recaptcha is served. In these instances, we need to navigate the user to the login page to manually log in.

### DIFF
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -422,6 +422,7 @@ export default {
                         this.shouldShowEmailAlreadyExistsError = true;
                     } else if (thrownErrors.some(thrownError => thrownError.errorCode === '403')) {
                         this.$emit(EventNames.LoginBlocked);
+
                         shouldEmitCreateAccountFailure = false;
                     } else {
                         this.genericErrorMessage = thrownErrors[0].description || 'Something went wrong, please try again later';
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -415,19 +415,23 @@ export default {
                 if (error && error.response && error.response.data && error.response.data.errors) {
                     thrownErrors = error.response.data.errors;
                 }
-
-                this.$emit(EventNames.CreateAccountFailure, thrownErrors);
+                let shouldEmitCreateAccountFailure = true;
 
                 if (Array.isArray(thrownErrors)) {
                     if (thrownErrors.some(thrownError => thrownError.errorCode === '409')) {
                         this.shouldShowEmailAlreadyExistsError = true;
                     } else if (thrownErrors.some(thrownError => thrownError.errorCode === '403')) {
-                        window.location.href = this.copy.navLinks.login.url;
+                        this.$emit(EventNames.LoginBlocked);
+                        shouldEmitCreateAccountFailure = false;
                     } else {
                         this.genericErrorMessage = thrownErrors[0].description || 'Something went wrong, please try again later';
                     }
                 } else {
                     this.genericErrorMessage = error;
+                }
+
+                if (shouldEmitCreateAccountFailure) {
+                    this.$emit(EventNames.CreateAccountFailure, thrownErrors);
                 }
             } finally {
                 this.shouldDisableCreateAccountButton = false;
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -416,16 +416,19 @@ export default {
                     thrownErrors = error.response.data.errors;
                 }
 
+                this.$emit(EventNames.CreateAccountFailure, thrownErrors);
+
                 if (Array.isArray(thrownErrors)) {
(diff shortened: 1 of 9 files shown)


adamashton (dismissed):

packages/f-registration/src/components/Registration.vue
```diff
@@ -416,16 +416,19 @@ export default {
                     thrownErrors = error.response.data.errors;
                 }
 
+                this.$emit(EventNames.CreateAccountFailure, thrownErrors);
+
                 if (Array.isArray(thrownErrors)) {
                     if (thrownErrors.some(thrownError => thrownError.errorCode === '409')) {
                         this.shouldShowEmailAlreadyExistsError = true;
+                    } else if (thrownErrors.some(thrownError => thrownError.errorCode === '403')) {
+                        window.location.href = this.copy.navLinks.login.url;
```
tommcclean: Just a question: Do we want to do the navigation here or in the parent component, I ask because I think when an account is registered successfully; the website routes the user back to the homepage (not the component), does this mean the user doesn't seem a message saying why? (is that desirable)

luke-andrew-parry: > Just a question: Do we want to do the navigation here or in the parent component, I ask because I think when an account is registered successfully; the website routes the user back to the homepage (not the component),

Good point. I can change this to emit an event instead and hook into it on the parent component to redirect.

> does this mean the user doesn't seem a message saying why? (is that desirable)

This is how it behaves currently. It's quite an edge case but I agree it's not a very nice experience for the user.

smarkey (dismissed):
LGTM from test perspective

adamashton (dismissed):

packages/f-registration/src/components/Registration.vue
```diff
@@ -415,17 +415,24 @@ export default {
                 if (error && error.response && error.response.data && error.response.data.errors) {
                     thrownErrors = error.response.data.errors;
                 }
+                let shouldEmitCreateAccountFailure = true;
 
                 if (Array.isArray(thrownErrors)) {
                     if (thrownErrors.some(thrownError => thrownError.errorCode === '409')) {
                         this.shouldShowEmailAlreadyExistsError = true;
+                    } else if (thrownErrors.some(thrownError => thrownError.errorCode === '403')) {
```
cmyhill-justeat, line 428: Would extending this to a switch statement be more elegant?

tommcclean (dismissed):

packages/f-registration/src/components/tests/Registration.test.js
```diff
@@ -133,6 +133,26 @@ describe('Registration', () => {
                 }
             });
 
+            it('should emit login blocked event when service responds with a 403', async () => {
+                // Arrange
+                const err = { response: { data: { faultId: '123', traceId: '123', errors: [{ description: 'Not authorized.', errorCode: '403' }] } } };
```
xander-marjoram: Nit: Isn't 403 Forbidden and 401 is Unauthorized?

xander-marjoram (commented):
Code looks good, just one small comment

packages/f-registration/src/components/tests/Registration.test.js
luke-andrew-parry: Yes, that should be forbidden.

packages/f-registration/src/components/Registration.vue
luke-andrew-parry: I tried it but to be honest with the amount of indenting I think it looks more verbose and noisy 😅 

```                    
if (thrownErrors.some(thrownError => thrownError.errorCode === '409')) {
    this.shouldShowEmailAlreadyExistsError = true;
} else if (thrownErrors.some(thrownError => thrownError.errorCode === '403')) {                        
    this.$emit(EventNames.LoginBlocked)
    shouldEmitCreateAccountFailure = false;
} else {
    this.genericErrorMessage = thrownErrors[0].description || 'Something went wrong, please try again later';
}
```
vs
```
thrownErrors.forEach(thrownError => {
    switch (thrownError.error) {
        case '409':
            this.shouldShowEmailAlreadyExistsError = true;
            break;
        case '403':
            this.$emit(EventNames.LoginBlocked);
            shouldEmitCreateAccountFailure = false;
            break;
        default:
            this.genericErrorMessage = thrownErrors[0].description || 'Something went wrong, please try again later';
    }
});
```

cmyhill-justeat: Yeah not so keen on iterating the entire collection would rather reduce it first.  Yeah looks like switch it not the best solution here

cmyhill-justeat (dismissed):

luke-andrew-parry:
@cmyhill-justeat @tommcclean @smarkey @adamashton Would you mind putting your reviews back on, please? Had to push up a small test change.

cmyhill-justeat (dismissed):

adamashton (dismissed):

xander-marjoram (approved):

smarkey (approved):

PR #379, opened 2020-10-21
f-form-field@v0.8.0 - Added component tests

LTurns:
Component tests for small component, f-form-field: 

### Added 

- data-test IDs to FormField Component 
- Test for FormField 
- FormField Component-Object


tommcclean (dismissed):

packages/f-form-field/CHANGELOG.md
```diff
@@ -3,6 +3,15 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.8.1
```
xander-marjoram: Should this be v0.8.0?

LTurns: Ah yes, thank you!

packages/f-form-field/src/components/FormField.vue
```diff
@@ -1,7 +1,8 @@
 <template>
     <div
         :data-theme-formfield="theme"
-        :class="$style['c-formField']">
+        :class="$style['c-formField']"
+        data-test-id="form-field-component">
```
xander-marjoram, line 5: I'm wondering if `-component` might be unnecessary. Would "form-field" clash with an existing data-test-id?

packages/f-form-field/CHANGELOG.md
xander-marjoram: Don't forget the package.json 😸

packages/f-form-field/src/components/FormField.vue
LTurns: Ooh it doesn't clash so I guess it's a case of preference - 'component' is added in f-footer, maybe because this id is about encompassing the whole component? But happy to remove if that's still clear - what d'you think?

packages/f-form-field/src/components/FormField.vue
```diff
@@ -28,7 +29,8 @@
                 :label-style="normalisedLabelStyle"
                 :for="uniqueId"
                 :is-inline="isInline"
-                data-js-test="inlineLabel">
+                data-js-test="inlineLabel"
+                data-test-id="testLabel">
```
SachaZvetelman, line 34: Do we need two test attributes? :)

packages/f-form-field/test-utils/component-objects/f-form-field.component.js
```diff
@@ -0,0 +1,21 @@
...
+const testLabel = () => $('[data-test-id="testLabel"]');
+const fieldInput = () => $('input');
+
+/**
+ * @description
+ * Inputs user's first name into the form-field component.
+ *
+ * @param {Object} userInput
+ * @param {String} userInput.firstName The user's first name
+ */
```
SachaZvetelman: Not sure what this description belongs to. Maybe it's just me, though.

packages/f-form-field/test-utils/component-objects/f-form-field.component.js
```diff
@@ -0,0 +1,21 @@
...
+
+exports.waitForFormField = () => formFieldComponent().waitForExist();
+
+exports.displayUserInput = (userInput) => {
+    fieldInput().setValue(userInput.firstName);
+    fieldInput().getValue()
```
SachaZvetelman: Nit: missing `;`

packages/f-form-field/test-utils/component-objects/f-form-field.component.js
```diff
@@ -0,0 +1,21 @@
+const formFieldComponent = () => $('[data-test-id="form-field-component"]');
+const testLabel = () => $('[data-test-id="testLabel"]');
+const fieldInput = () => $('input');
```
SachaZvetelman: Is this selector ok as it is or should we be more specific? cc @siggerzz

packages/f-form-field/src/components/FormField.vue
LTurns: Yeah I'm not sure - I think it might be better having two different ids in case things change on the unit test side of things, but I've just thought that if we end up having one shared page object for both unit and component tests we might be able to avoid this? @siggerzz

packages/f-form-field/test-utils/component-objects/f-form-field.component.js
LTurns: copied this structure from f-registration but happy to remove, agree it might seem a bit random

packages/f-form-field/src/components/FormField.vue
siggerzz: @xander-marjoram This is autogenerated by the yeoman generator. I'd prefer to leave as in for now as it's consistent with what's already in the repo.

We can always discuss a better naming convention if the future though if you like 👍

packages/f-form-field/src/components/FormField.vue
siggerzz: > Do we need two test attributes? :)
No. 😄 

However, we've kept them separate as we've had problems in past repos where the UI has changed and `data-test-id's` has been accidentally been removed.

As Lizzie mentioned, we have some ideas on how we can improve this and use the same selectors for both Unit and Component tests, however, I'd prefer to change this as part of a future ticket.

packages/f-form-field/test-utils/component-objects/f-form-field.component.js
siggerzz: @SachaZvetelman This is JSDoc syntax :) - https://jsdoc.app/tags-description.html

packages/f-form-field/test-utils/component-objects/f-form-field.component.js
siggerzz: Yep, as @SachaZvetelman mentioned, this should be more specific. The current selector would match with any `inputs` within the DOM, which could cause test flakiness

packages/f-form-field/test-utils/component-objects/f-form-field.component.js
DamianMullins: It doesn't appear to be attached to any functions though? Looks like it should be above the `displayUserInput` below?

packages/f-form-field/test-utils/component-objects/f-form-field.component.js
LTurns: thought that might be the case - @siggerzz the data-test-id attribute is already being used in jest for a prop. Is there another naming convention I could use or another attribute I could grab? WebdriverIO doesn't seem to like Vue attributes with a colon at the front

packages/f-form-field/test-utils/component-objects/f-form-field.component.js
LTurns: ah I see what you mean @DamianMullins, I'll move it to just before the function 👍

packages/f-form-field/test-utils/component-objects/f-form-field.component.js
```diff
@@ -0,0 +1,18 @@
+const formFieldComponent = () => $('[data-test-id="form-field-component"]');
+const testLabel = () => $('[data-test-id="testLabel"]');
+const input = () => $('[data-testid="testId"]');
```
LTurns: @siggerzz struggling to find a way of referencing the :data-test-id attribute - tried using Xpath and CSS selectors but it doesn't seem to like it. I could use this work around so we have two separate test ids? :)

simonsandersje (approved):

packages/f-form-field/src/components/FormField.vue
xander-marjoram: @siggerzz Do we have a ticket for this future work? I can see us never coming back to it otherwise

xander-marjoram (approved):
LGTM. Can we get a ticket raised for unifying the data-test-id vs data-js-test attributes? Would be better to pick one and stick with it.

LTurns:
> LGTM. Can we get a ticket raised for unifying the data-test-id vs data-js-test attributes? Would be better to pick one and stick with it.

@xander-marjoram thank you - yep good idea. I'll make a ticket for this!

packages/f-form-field/src/components/FormField.vue
ashleynolan: What @xander-marjoram said – would be good to get this resolved early, else this could become a fairly massive change once this pattern is set

PR #380, opened 2020-10-21
fozzie-components@v2.5.0 & f-button@v0.1.0 – fozzie package updates and initial f-button component setup

ashleynolan:
### Added
- First iteration of button component. Contains variants for primary, secondary, ghost, outline and btnLink.

---

### Added
- Upgrading `fozzie-colour-palette` and `fozzie` package versions.

---

## UI Review Checks

Screen Shot 2020-10-21 at 11 01 16

- [x] README and/or UI Documentation has been created
- [ ] Unit tests have been created (will follow in separate PR)

## Browsers Tested

- [x] Chrome (latest)

### DIFF
--- a/packages/f-button/src/components/Button.vue
+++ b/packages/f-button/src/components/Button.vue
@@ -1,0 +1,340 @@
+<template>
+    <button
+        :class="[
+            $style['o-btn'],
+            $style[`o-btn--${buttonType}`],
+            $style[`o-btn--size${buttonSizeClassname}`],
+            (fullWidth ? $style['o-btn--fullWidth'] : '')
+        ]"
+        data-test-id='button-component'>
+        <slot />
+    </button>
+</template>
+
+<script>
+
+export default {
+    name: 'VueButton',
+    components: {},
+    props: {
+        buttonType: {
+            type: String,
+            default: 'primary'
+        },
+        buttonSize: {
+            type: String,
+            default: 'medium'
+        },
+        fullWidth: {
+            type: Boolean,
+            default: false
+        }
+    },
+    computed: {
+        /**
+         * Converts the buttonSize prop into a normalised classname (that fit with our class naming scheme)
+         */
+        buttonSizeClassname () {
+            if (this.buttonSize === 'xsmall') {
+                return this.buttonSize.slice(0, 2).toUpperCase() + this.buttonSize.slice(2); // xsmall -> XSmall
+            }
+            return this.buttonSize.charAt(0).toUpperCase() + this.buttonSize.slice(1); // capitalize the first letter of the prop
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+
+$btn-default-bgColor            : $grey--lighter;
+$btn-default-bgColor--hover     : $grey--light;
+$btn-default-borderRadius       : $border-radius;
+$btn-default-font-family        : $font-family-base;
+$btn-default-font-size          : 'body-l';
+$btn-default-weight             : $font-weight-bold;
+$btn-default-padding            : 11px 1.5em 13px;
+$btn-default-outline-color      : $color-focus-outline;
+
+$btn-primary-bgColor            : $blue;
+$btn-primary-bgColor--hover     : $blue--dark;
+$btn-primary-bgColor--focus     : $blue--darkest;
+$btn-primary-textColor          : $white;
+
+$btn-secondary-bgColor          : $blue--offWhite;
+$btn-secondary-bgColor--hover   : $blue--offWhite--dark;
+$btn-secondary-bgColor--active  : $blue--offWhite--darkest;
+$btn-secondary-textColor        : $blue;
+
+$btn-outline-bgColor            : transparent;
+$btn-outline-bgColor--hover     : rgba($black, 0.08);
+$btn-outline-bgColor--active    : rgba($black, 0.12);
+$btn-outline-textColor          : $black;
+$btn-outline-border-color       : $grey--light;
+// TODO: checking with design if $btn-outline-border-color should change on hover/active
+
+$btn-ghost-bgColor              : transparent;
+$btn-ghost-bgColor--hover       : rgba($black, 0.08);
+$btn-ghost-bgColor--active      : rgba($black, 0.12);
+$btn-ghost-textColor            : $blue;
+
+$btn-disabled-bgColor           : $grey--light;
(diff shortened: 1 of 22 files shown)


kevinrodrigues (dismissed):

kevinrodrigues:
@ashleynolan LGTM there's a conflict tho.

ashleynolan:
@kevinrodrigues should be resolved now 👍

kevinrodrigues (dismissed):

packages/f-button/src/components/Button.vue
```diff
@@ -0,0 +1,340 @@
...
+
+export default {
+    name: 'VueButton',
+    components: {},
+    props: {
+        buttonType: {
+            type: String,
+            default: 'primary'
+        },
+        buttonSize: {
```
DamianMullins, line 24: I'm still not sold on using string values for props, but if that's the way we want to go then I think we should add a validator here to make sure valid values are being set?

packages/f-button/src/components/Button.vue
```diff
@@ -0,0 +1,340 @@
...
+$btn-sizeLarge-padding          : 13px 1.2em 15px;
+
+$btn-sizeSmall-padding          : 7px 1em 9px;
+
+$btn-sizeXSmall-padding         : 5px 0.5em 7px;
+$btn-sizeXSmall-lineHeight      : 1;
+
+
+.o-btn {
+    display: inline-block;                      /* [1] */
```
DamianMullins: Do you need the comments on these lines?

packages/f-button/src/components/Button.vue
```diff
@@ -0,0 +1,340 @@
...
+    font-family: $btn-default-font-family;      /* [3] */
+    @include font-size($btn-default-font-size); /* [3] */
+    cursor: pointer;                            /* [4] */
+    margin: 0;                                  /* [5] */
+    padding: $btn-default-padding;              /* [5, 6] */
+    overflow: visible;                          /* [7] */
+    text-align: center;
+    font-weight: $btn-default-weight;
+
+    // You may want to change this
```
DamianMullins: Who is this comment meant for and why would they want to change the value?

packages/f-button/src/components/Button.vue
```diff
@@ -0,0 +1,340 @@
...
+    name: 'VueButton',
+    components: {},
+    props: {
+        buttonType: {
+            type: String,
+            default: 'primary'
+        },
+        buttonSize: {
+            type: String,
+            default: 'medium'
```
jamieomaguire, line 26: if there's only a select few options that the size could be i.e. xsmall, medium etc (I don't actually know whether or not this is the case!), is it worth adding any prop validation here?

packages/f-button/src/components/Button.vue
```diff
@@ -0,0 +1,340 @@
...
+}
+
+/**
+ * ==========================================================================
+ * Btn Background Colour modifiers
+ * ==========================================================================
+ */
+
+/**
+ * Modifier – .o-btn--primary
```
DamianMullins, line 156: Is the class name necessary here when it's declared a few lines down?

packages/f-button/src/components/Button.vue
```diff
@@ -0,0 +1,340 @@
...
+.o-btn--sizeLarge {
+    @include font-size($btn-sizeLarge-font-size);
+    padding: $btn-sizeLarge-padding;
+
+    &.o-btn--primary {
+        background-color: $orange;
+    }
+}
+
+// .o-btn--sizeMedium {}
```
DamianMullins: Can we remove this?

jamieomaguire (dismissed):

packages/f-button/src/components/Button.vue
ashleynolan: Think this was a hangover from old button styles – I'll remove them

packages/f-button/src/components/Button.vue
ashleynolan: Again – copy/paste from `fozzie`. Don't think we need this anymore though 😆

packages/f-button/src/components/Button.vue
ashleynolan: Yep – I'm doing a separate PR for unit tests, so if it's ok, I'll add this to that PR?

packages/f-button/src/components/Button.vue
ashleynolan: I kinda like it, but I guess that's just personal preference. Happy to remove if you think it's not needed?

packages/f-button/src/components/Button.vue
ashleynolan: Will do

packages/f-button/src/components/Button.vue
DamianMullins: I'm happy to keep it if it's adding value, sure. Explicitly marking it as a modifier is nice.

packages/f-button/src/components/Button.vue
```diff
@@ -0,0 +1,340 @@
...
+            if (this.buttonSize === 'xsmall') {
+                return this.buttonSize.slice(0, 2).toUpperCase() + this.buttonSize.slice(2); // xsmall -> XSmall
+            }
+            return this.buttonSize.charAt(0).toUpperCase() + this.buttonSize.slice(1); // capitalize the first letter of the prop
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
```
DamianMullins, line 47: When I was testing pulling button components into another project I noticed that the generated scoped class styles couldn't easily be overridden as they were added after the styles where I was trying to override thanks to the cascade. Have you hit this issue before? Or would you have any recommendations on how to resolve that?

packages/f-button/src/components/Button.vue
jamieomaguire: ace!

jamieomaguire (approved):

DamianMullins (approved):

packages/f-button/src/components/Button.vue
ashleynolan: Probably worth catching up on hangout about this as would be good to get a bit more context. I'll drop you a msg 👍

DamianMullins: Yeah sounds good, I'll wait for this component to be released to npm and then set up a demo we can look at.

PR #381, opened 2020-10-21
f-checkout@0.4.0 - Replicate Collection layout

FayeCarter:
Allows `f-checkout` to be used for logged in Collection as well as Delivery.

*There will be more changes to follow with `f-form-field` and `f-button` updates/creation*

### Added
- Address component to be displayed if `checkoutMethod` is set to 'Delivery'.

### Removed
- Address related form fields from `f-checkout`.

### Changed
- Checkout Component to replicate Collection and Delivery in ConsumerWeb.
- Selector to uses `method` prop to change label text.

## UI Review Checks
Collection | Delivery
|---|---|
image | image


simonsandersje (dismissed):

packages/f-checkout/src/components/Address.vue
```diff
@@ -0,0 +1,88 @@
+<template>
+    <div data-test-id='address-component'>
+        <div :class="$style['l-addressGroup']">
```
ashleynolan: Not sure if there could be an accessibility benefit to defining this component as a fieldset? (as it is a group of related fields)

Doesn't need to to be resolved by this PR, but would be good to look into

packages/f-checkout/src/components/Address.vue
```diff
@@ -0,0 +1,88 @@
...
+            <form-field
+                v-model="address.line1"
+                :class="$style['c-formField']"
+                name="address-line-1"
+                data-test-id="input-address-line-1"
+                :label-text="labels.line1"
+                input-type="text"
```
ashleynolan: If you want to save yourself some code, `text` is the default type for the form-field component, so you shouldn't have to define the prop (and for the references below)

packages/f-checkout/src/components/Address.vue
```diff
@@ -0,0 +1,88 @@
+<template>
+    <div data-test-id='address-component'>
+        <div :class="$style['l-addressGroup']">
+            <form-field
+                v-model="address.line1"
+                :class="$style['c-formField']"
```
ashleynolan: More a question than a comment – does defining this classname here, does it effectively add two `c-formField` classes to the component? (as `c-formField` is also added to the `div` as part of the f-form-field component)

packages/f-checkout/src/components/Address.vue
```diff
@@ -0,0 +1,88 @@
...
+            }
+        };
+    }
+};
+</script>
+
+<style lang="scss" module>
+.c-formField {
+    input {
+        height: 50px;
```
ashleynolan: Any changes to the height of the inputs should happen as part of the `f-form-field` component (so that the height of `input` elements are consistent everywhere (I'm updating that to the PIE version today, so that might be one of the values that updates :) )

packages/f-checkout/src/components/Address.vue
```diff
@@ -0,0 +1,88 @@
...
+    }
+    label {
+        @include font-size(body-l);
```
ashleynolan: Same with the label – should be consistent across all form-fields

ashleynolan (changes requested):

packages/f-checkout/src/constants.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export const VALID_CHECKOUT_METHOD = ['Collection', 'Delivery'];
```
minrwhite: I'd say it's worth separating out the values from this array into individual `const`s that are also exported that you can then import for use in comparisons like the `v-if` in address-block. Additionally, you should probably re-export these in `index.js` as named exports so that consuming applications can use the values directly.

packages/f-checkout/stories/Checkout.stories.js
```diff
@@ -1,20 +1,27 @@
+import { withKnobs, select } from '@storybook/addon-knobs';
```
minrwhite: Knobs are deprecated as of storybook v6, please look into using controls if possible.

packages/f-checkout/src/components/Selector.vue
```diff
@@ -27,9 +27,14 @@
 <script>
 export default {
     props: {
-        deliveryTimes: {
+        times: {
             type: Array,
             default: () => ['', 'As soon as possible', 'Today in 5 minutes']
+        },
+
+        method: {
```
ashleynolan: `method` is a bit of a loaded word in JS – I'd probably call this `orderType` or `orderMethod`?

packages/f-checkout/src/components/Address.vue
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+.l-addressGroup {
+    margin: spacing(x2) 0 spacing(x4) 0;
+    @include font-size(body-s);
+
+    div {
```
minrwhite: Is this intended to apply to every `div` nested within an element with the `l-addressGroup` style applied? Even those nested within each other? It'd be better to specify a class here, if possible, unless there's a good reason for it.

packages/f-checkout/src/components/Address.vue
FayeCarter: Cool, I'll have a look

packages/f-checkout/src/components/Selector.vue
FayeCarter: Good point 😂

packages/f-checkout/src/constants.js
FayeCarter: Cool, I'll give this a go.

packages/f-checkout/src/components/Address.vue
FayeCarter: No good reason, just unnoticed 😂. Thanks

packages/f-checkout/src/components/Address.vue
FayeCarter: It does this 😬 
image
It's now removed.

packages/f-checkout/src/components/Selector.vue
```diff
@@ -8,7 +8,7 @@
         <label
             for="time-selection"
             :class="$style['o-form-select-label']">
-            {{ method }} time
+            {{ orderMethod }} time
```
minrwhite, line 11: Not an immediate concern, but at some point this should be made Localization (L10n) compatible, as even though this might not roll out to EU in lieu of them moving to takeaway's legacy platform, ANZ use the term 'pickup' rather than collection.

minrwhite (dismissed):

packages/f-checkout/src/components/Address.vue
FayeCarter: I've removed the height modification, it looks a little bit weird at the moment but I can update from the PIE design in another PR right?

ashleynolan: Yeh – this should update when the new f-form-field changes come in too (or if you need to change them, can work out how to change through a prop to that component)

packages/f-checkout/src/components/Address.vue
```diff
@@ -0,0 +1,77 @@
+<template>
+    <div data-test-id='address-component'>
+        <fieldset :class="$style['l-addressGroup']">
```
ashleynolan: I'd change the classname here to be `c-addressGroup`. We tend to only use `l-` for global layout helper classes (so those aren't used very often).

packages/f-checkout/src/components/Address.vue
```diff
@@ -0,0 +1,77 @@
...
+            <form-field
+                v-model="address.line1"
+                :class="$style['c-checkout-formField']"
```
ashleynolan: As you have the addressGroup class above now, you could use `c-addressGroup-formField` here to keep all the classname's in this SFC relative to this address component?

ashleynolan (commented):
Looks - good. Couple of super small class name suggestions, otherwise LGTM :shipit:

minrwhite (approved):

ashleynolan (approved):

PR #386, opened 2020-10-22
f-header@4.2.0 - Don't retrieve user details when the prop to show them is false

xander-marjoram:
Most of the lines are updating the yarn.lock 😬 

### Changed
- Don't fetch user details when showUserInfo prop is false.
- Use `axiosServices` from `f-services` instead of axios.
- Use latest `f-vue-icons`.

---

## UI Review Checks

## Browsers Tested
- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -1,78 +1,54 @@
 import { shallowMount } from '@vue/test-utils';
-import axios from 'axios';
-import loggedInUserMock from './__mocks__/api.account.details.json';
-import Navigation from '../Navigation.vue';
 
-const defaultPropsData = {
```
xander-marjoram, line 6: I moved these into `helpers/navigation.js`.

packages/f-header/src/components/tests/Navigation.test.js
```diff
@@ -531,70 +488,116 @@ describe('Navigation', () => {
...
             await wrapper.vm.fetchUserInfo();
 
-            expect(axios.get).toHaveBeenCalled();
+            // Assert
+            expect(mockGet).toHaveBeenCalled();
             expect(wrapper.vm.$data.userInfo).toBeDefined();
             expect(wrapper.vm.$data.userInfo).toEqual(asyncUserDetails);
         });
+
+        describe('(with spy)', () => {
```
xander-marjoram, line 494: Here is the new test coverage

tommcclean (dismissed):

packages/f-header/src/components/Navigation.vue
```diff
@@ -438,12 +440,14 @@ export default {
...
         // If userInfoProp wasn't passed we make a call for userInfo on mounted hook
         fetchUserInfo () {
-            const userDetailsPromise = axios.get(this.userInfoUrl, {
+            const client = axiosServices.createClient({
                 headers: {
                     credentials: 'same-origin'
                 }
             });
 
+            const userDetailsPromise = client.get(this.userInfoUrl);
```
DamianMullins: Is there any reason this needs to be split out? Is it possible to chain `createClient`, `get`, and `then` together?

simonsandersje (dismissed):

simonsandersje (approved):

packages/f-header/src/components/Navigation.vue
```diff
@@ -438,54 +440,55 @@ export default {
...
             });
 
-            userDetailsPromise.then(response => {
-                if (response.data.isAuthenticated) {
-                    this.userInfo = response.data;
-
-                    if (this.isOrderCountValid) {
-                        this.fetchOrderCountAndSave();
+            return client
+                .get(this.userInfoUrl)
```
matthewhardern, line 450: Have we considered moving this get request into an alternative service that is imported, would make this component far less bloated, also would make testing easier?

packages/f-header/src/components/Navigation.vue
```diff
@@ -438,54 +440,55 @@ export default {
...
-                }
-            })
-            .catch(err => {
-                if (this.errorLog) {
-                    this.errorLog('Unable to get order count from "fetchOrderCountAndSave', err);
-                }
-            });
-            return orderCountPromise;
+            return client
+                .get(this.orderCountUrl)
```
matthewhardern, line 478: Same as above comment

packages/f-header/src/components/Navigation.vue
xander-marjoram: I think we have plans to move the user requests out into another module at some point, perhaps f-services or a new package.
I can extract these out into a separate file now though to make that move a bit easier in future, and to hopefully tidy up the tests.

DamianMullins (approved):

packages/f-header/src/components/Navigation.vue
xander-marjoram: It's a bit of a strange one, because ultimately I don't think the navigation component should have anything to do with fetching the user data. I think we should leave this for now but I'll create a ticket for extracting all of this behaviour into another module, rather than leaving some props and computed variables hanging around inside the component and some of the functionality inside a JS module within f-header.

matthewhardern: I totally agree on that, so happy to resolve for now

PR #388, opened 2020-10-23
f-card@v0.8.0 - Added webdriver tests to small components

LTurns:
### Added 
- data-test IDs to Card Component 
- Test for Card
- Card Component-Object 

### Added
- data-test IDs to Alert Component 
- Basic tests for Alert 
- Alert Component-Object 

### Changed
- Small amendments to component test


simonsandersje (approved):

xander-marjoram (approved):

packages/f-card/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-card",
   "description": "Fozzie Card Component – Used for providing wrapper card styling to an element (or group of elements)",
-  "version": "0.7.0",
+  "version": "0.8.0",
```
ashleynolan, line 4: Small one – looks like you've bumped the package versions, but not the version in the CHANGELOGS (they're still at latest).

If you ever want to make a change without incrementing the individual package.json versions, use the `#globalconfig` tag in the PR title and it'll skip the CircleCI checks

packages/f-alert/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-alert",
   "description": "Fozzie Alert – Fozzie Alert Component",
-  "version": "0.1.1",
+  "version": "0.2.0",
```
ashleynolan, line 4: Same here – looks like you've bumped the package versions, but not the version in the CHANGELOGS (they're still at latest). Can you put in a change using `#trivial` tag in the PR title to move these version numbers? Otherwise can see this causing issues with version mismatches with the next release of these components.

packages/f-card/package.json
LTurns: Ah sorry that's my bad, got a little confused by latest release - will update the CHANGELOGS now! Good to know about #globalconfig

PR #395, opened 2020-10-27
f-checkout@v0.6.0 – Test added for visibility on displayed fields.

JoshuaNg2332:
---

This PR is to add tests for f-checkout. This test confirms the existence of fields on the page.

---

## Browsers Tested

- [X] Chrome (latest)


packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -17,6 +17,23 @@ describe('f-checkout component tests', () => {
         expect(CheckoutComponent.isAllergenLinkDisplayed()).toBe(true);
     }
 
+    it('should display all fields', () => {
+        // Act
+        var areFieldsDisplayed = function(){
```
matthewhardern: You can use `const` here as you never re-assign `areFieldsDisplayed`

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -17,6 +17,23 @@ describe('f-checkout component tests', () => {
...
+            var i;
+            var fields = CheckoutComponent.isCheckoutFormFieldsDisplayed();
```
matthewhardern: `isCheckoutFormFieldsDisplayed` does not currently exist in the imported utilities, and hence the reason for CI tests failing, did you forget to add it to git?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -17,6 +17,23 @@ describe('f-checkout component tests', () => {
...
+            for(i = 0; i < fields.values.length; i++){
```
matthewhardern: Having logic inside tests In my personal opinion is not best practice. This logic should be moved to a function that itself can be tested. This prevents the test failing because of this logic.

SachaZvetelman: Normally we try not to "create new functionality or logic" inside a test. For example, you are creating a function, having conditionals and loops. The problem with this approach is that nothing is testing your test if it makes sense. Happy to explain more on a Hangouts, if you want!

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -17,6 +17,23 @@ describe('f-checkout component tests', () => {
...
+                if(fields[i]==false){
+                    console.log(fields);
```
SachaZvetelman: Nit: remove from the test once it's ready 😄

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -17,6 +17,23 @@ describe('f-checkout component tests', () => {
         expect(CheckoutComponent.isAllergenLinkDisplayed()).toBe(true);
     }
 
+    it('should display all fields', () => {
+        // Act
+        var areFieldsDisplayed = function(){
+            var i;
+            var fields = CheckoutComponent.isCheckoutFormFieldsDisplayed();
+            for(i = 0; i < fields.values.length; i++){
+                if(fields[i]==false){
```
SachaZvetelman: I'm not entirely sure what we're checking here. Could you please clarify?
At the same time, remember the difference between `==` and `===`. 
In most scenarios, `if (!fields[i]) {` would suffice.

SachaZvetelman (commented):
A few comments. Feel free to ping me if you have any questions, of course.
Make sure you set up ESLint on your VS Code. There are many issues that will be automatically fixed for you when saving the file (e.g. spacing), and it will warn you of important things such as "you can use `const` here".

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -17,6 +17,23 @@ describe('f-checkout component tests', () => {
         expect(CheckoutComponent.isAllergenLinkDisplayed()).toBe(true);
     }
 
+    it('should display all fields', () => {
```
kevinrodrigues: I think you could switch the entire contents of this test out for a simpler test. You would only need to test that that the individual fields exist, similar idea to what you are already doing. The [util](https://github.com/justeat/fozzie-components/blob/12d5aab2756c5c38dc186d150e5a745e4a8824a1/packages/f-checkout/test-utils/component-objects/f-checkout.component.js) contains all the data test IDs for you to be able to either loop or individually test them without the `areFieldsDisplayed` setup.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
xander-marjoram: It looks like you're checking a list of boolean values to see if any of them are false, is that correct?

If so, you could use a function like [`filter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) to check this rather than using a for-loop to iterate through the array.

```javascript
const fields = CheckoutComponent.isCheckoutFormFieldsDisplayed();
const hiddenFields = fields.filter(f => !f); // This removes any "truthy" elements
expect(hiddenFields.length).toBe(0);
```

Also, nit: I would rename `isCheckoutFormFieldsDisplayed` to something like `areFormFieldsDisplayed`. You could even move the bit where you filter it down into the method itself, and just return true or false inside there. 😄

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
kevinrodrigues: Looks to me that this is exported ok?

matthewhardern: On master `f-checkout.component` in the test-utils directory does not contain `isCheckoutFormFieldsDisplayed` unless I am mistaken? Also Circle CI error is because isCheckoutFormFieldsDisplayed is not a function. So assumed it was undefined, happy to be corrected though

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
matthewhardern: The above is a good shout, not sure on opinions here, but you could do a `it.each` passing in the data test id's of all the fields.

kevinrodrigues: `it.each` is something we would usually use here 👍

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
kevinrodrigues: Looks ok to me on master, would need to pull it to see what was happening, it does seem to already be used in this file further up too tho.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
Festivejelly: @SachaZvetelman where would you put this logic then? It cant live in the page object model. All this is doing is asserting if fields are displayed. Its not changing the path the test would take, only the outcome of the assertion.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
Festivejelly: Aka test.each i guess? https://jestjs.io/docs/en/api

Tbh we were struggling to find examples on how to do this because it.each isnt an available function.

Also the test i'ds dont seem to be in a collection so how would one pass each one in?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
SachaZvetelman: @Festivejelly, my comment was about `console.log`, which shouldn't be there, since that is for dev purposes only. Or what part are you talking about exactly?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
matthewhardern: have a look at f-metadata `BrazeDispatcher.test.js` its got some good examples

xander-marjoram: @Festivejelly Yes `it` and `test` are interchangeable

Festivejelly: This is in Mocha. Your comments are relating to Jest afaik? We've added a new library to do a each on the it.

xander-marjoram: I think we're picturing something along these lines
```javascript
it.each([
    'field1',
    'field2'
])('should display %s field', testId => {
    // Test that element with data-test-id=testId is displayed
});
```

package.json
```diff
@@ -60,6 +60,7 @@
     "eslint-plugin-vue": "6.2.2",
     "include-media": "1.4.9",
     "lerna": "3.22.1",
+    "mocha-each": "^2.0.1",
```
siggerzz: You can remove the `^` from here 👍

wdio.conf.js
```diff
@@ -122,7 +122,7 @@ exports.config = {
     // Test reporter for stdout.
     // The only one supported by default is 'dot'
     // see also: https://webdriver.io/docs/dot-reporter.html
-    // reporters: ['dot'],
+    reporters: ['spec'],
```
siggerzz: This can be reverted

siggerzz (changes requested):
Few minor comments, but otherwise looks good to me 🎉

siggerzz (dismissed):

tommcclean (dismissed):

SachaZvetelman:
Only one question, and probably because I'm missing something, but why are we adding Mocha? :)

ashleynolan:
Can you remove the JIRA ID's from your commit titles?

ashleynolan (changes requested):
Just blocking this until the commit msgs are updated

JoshuaNg2332:
> Just blocking this until the commit msgs are updated

@ashleynolan  Updated commit messages now.

siggerzz:
@SachaZvetelman Mocha has always been part of the framework, we're just adding an additional mocha plugin. Mocha is the test framework that's use by WebDriverIO. (in the same way Jest is used for unit tests, or how we use Specflow in C#) 👍

SachaZvetelman:
That's what I was missing: didn't know we were using Mocha with WebDriverIO. Now I do 😬

siggerzz (dismissed):

packages/f-checkout/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.6.0
+------------------------------
+*October 28, 2020*
```
simonsandersje: Change to today's date before merging

ashleynolan (dismissed):

kevinrodrigues (dismissed):

siggerzz (dismissed):

simonsandersje (dismissed):

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -39,6 +39,24 @@ exports.submitCheckoutForm = (addressInfo) => {
...
+            return addressLine1Input().isDisplayed();
+        case 'addressline2':
+            return addressLine2Input().isDisplayed();
+        case 'addresscity':
+            return addressCityInput().isDisplayed();
+        case 'addresspostcode':
+            return addressPostcodeInput().isDisplayed();
+        default:
+            throw new Error(`Field ${fieldName} not recognized`);
+    }
```
SachaZvetelman: Sorry if I'm a bit late to the party, since you already have two approvals, but I feel this can be simplified.
For example, if you pass in the input as a parameter rather than the name, your function would pretty much be reduced to `return element.isDisplayed();`.

SachaZvetelman: On a similar note, the function name should probably be singular in the end. You're checking that one given field is displayed, not all the checkout form fields are displayed, if it makes sense.

JoshuaNg2332: I've had a chat with @siggerzz & @Festivejelly about this and we wanted to have a single test that checks these in one go rather than having individual functions that check visibility. If these were split into individual functions as part of the same test, one failure would stop the execution of the rest of the test.

SachaZvetelman: Hmmm, I think we're talking about different things here, hehe. Let's have a chat after my training, before we merge this, please.

SachaZvetelman (changes requested):
Just setting this "request changes" to avoid an accidental merge before we discuss those outstanding comments

siggerzz (dismissed):

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -15,6 +10,15 @@ const deliveryTimeDropdown = () => $('[data-test-id="delivery-time"]');
...
 
+exports.inputs = {
+    mobileNumber: () => $('[data-test-id="input-mobile-number"]'),
+    addressLine1: () => $('[data-test-id="input-address-line-1"]'),
+    addressLine2: () => $('[data-test-id="input-address-line-2"]'),
+    addressCity: () => $('[data-test-id="input-address-city"]'),
+    addressPostcode: () => $('[data-test-id="input-address-postcode"]')
+}
+
+// exports.mobileNumberInput = () => $('[data-test-id="input-mobile-number"]');
```
SachaZvetelman: Please remove commented code.

SachaZvetelman (dismissed):
Approved with a minor comment.

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
JoshuaNg2332: Removed. 😄

SachaZvetelman (dismissed):

siggerzz (dismissed):

siggerzz (dismissed):

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Latest (to be added to next release)
 ------------------------------
+*November 2, 2020*
+
+### Added
+- Added `mocha-each` to package.json.
+
+
+v2.9.0
```
ashleynolan, line 14: You can delete the latest line above by the looks of it and make this:

```
v2.9.0
*November 2, 2020*

### Added
- Added `mocha-each` to package.json.
- Added an entry to gitignore to ignore some local development tools for mocking http requests
```

ashleynolan (commented):
Super small comment then LGTM

CHANGELOG.md
JoshuaNg2332: @ashleynolan I've had a look at this and thought that it follows the existing convention? The version number lives below the actual update and then the one above reading `Latest (to be added to next release)` needs to be there?
@siggerzz Do correct me if I'm wrong here?

ashleynolan: So the version number should be above each update. We use the `Latest` title to indicate that the package has been updated but didn't require a version bump, so those `CHANGELOG` entries get merged into the next release (so the `latest` changes there currently, merge into your release)

ashleynolan: Tbh, I'm not sure why there's a `latest` entry in the root of the mono repo anyway – it never gets released as such, so the version should just get bumped every change. So bit confusing – if you update it to be as I showed though that should resolve the issue.

siggerzz (approved):

ashleynolan (approved):

ashleynolan:
Happy for this to be merged as long as you've checked the build is ok on CircleCI.

siggerzz:
All good on CircleCI! - https://app.circleci.com/pipelines/github/justeat/fozzie-components/2118/workflows/6d8dc34c-fe01-4a93-8f6a-a47f7425ce9c

PR #400, opened 2020-10-29
f-checkout@0.8.0 - Add server side and client side error handling

FayeCarter:
### Added
- Inline error handling to `Checkout` and `Address` components.
- Server side error handling.

## UI Review Checks
**Collection** | **Delivery** | **Server Error Messages**
|---|---|--|
image | image | image


packages/f-checkout/package.json
```diff
@@ -44,10 +44,13 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "axios": "0.20.0",
+    "@justeat/f-alert": "^0.1.1",
```
SachaZvetelman: Remove `^` so we pin the version 👍

packages/f-checkout/package.json
```diff
@@ -44,10 +44,13 @@
...
     "@justeat/f-card": "0.6.0",
     "@justeat/f-form-field": "0.7.1",
-    "@justeat/f-services": "1.0.0"
+    "@justeat/f-services": "1.0.0",
+    "axios": "0.20.0",
+    "flush-promises": "^1.0.2",
```
SachaZvetelman: And here.

packages/f-checkout/package.json
```diff
@@ -44,10 +44,13 @@
...
+    "vuelidate": "^0.7.6"
```
SachaZvetelman: And here.

@ashleynolan, @tommcclean, is this another candidate to add to `f-services`?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -115,7 +177,96 @@ export default {
 
         title () {
             return `${this.name}, confirm your details`;
+        },
+
+        shouldShowMobileNumberInvalidError () {
+            const mobileNumberInvalid = !this.$v.mobileNumber.required || !this.$v.mobileNumber.numeric || !this.$v.mobileNumber.minLength;
```
SachaZvetelman: Since it's a boolean, it should be `isMobileNumberInvalid`.
On a similar note, it might be easier to read if we invert the variable and call it `isMobileNumberValid`, but this is just a personal preference.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -98,13 +159,14 @@ export default {
...
-                lineOne: null,
-                lineTwo: null,
+                line1: null,
+                line2: null,
                 city: null,
                 postcode: null
             },
             buttonText: 'Go to payment',
-            delivery: CHECKOUT_METHOD_DELIVERY
+            delivery: CHECKOUT_METHOD_DELIVERY,
```
SachaZvetelman: Not sure what `delivery` is supposed to represent. Is it `delivery` or `checkout`? Maybe this should be a boolean? I might be missing some context here.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -115,7 +177,96 @@ export default {
...
+        shouldShowMobileNumberInvalidError () {
+            const mobileNumberInvalid = !this.$v.mobileNumber.required || !this.$v.mobileNumber.numeric || !this.$v.mobileNumber.minLength;
+            return mobileNumberInvalid && this.$v.mobileNumber.$dirty;
+        },
+
+        addressErrors () {
+            if (this.checkoutMethod === this.delivery) {
+                return {
+                    line1: {
+                        error: !this.$v.address.line1.required && this.$v.address.line1.$dirty,
```
SachaZvetelman: This pattern seems to be repeated in multiple places (required and dirty). Maybe we can extract this to a common function?

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +91,137 @@ describe('Checkout', () => {
...
     });
+
+    describe('with a working checkout service', () => {
+        describe('when checkoutMethod set to `Collection`', () => {
+            const propsData = { checkoutMethod: 'Collection' };
+
+            it('should emit success event when all fields are populated correctly', async () => {
+                // Arrange
+                const wrapper = mount(VueCheckout, { propsData });
+                // expect(wrapper.html()).toBe('1');
```
SachaZvetelman: Nit: remove comment.

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +91,137 @@ describe('Checkout', () => {
             });
         });
     });
+
+    describe('with a working checkout service', () => {
```
SachaZvetelman: Not sure I understand this description.

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +91,137 @@ describe('Checkout', () => {
...
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+                await wrapper.vm.$nextTick();
+
+                // Assert
+                expect(wrapper.vm.addressErrors).toBe(null);
+                expect(wrapper.emitted(EventNames.GoToPaymentFailure).length).toBe(1);
+                expect(wrapper.emitted(EventNames.GoToPaymentFailure)[0][0].invalidFields).toContain('mobileNumber');
```
SachaZvetelman: Shouldn't the verification here be about the address? I might be confused.

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +91,137 @@ describe('Checkout', () => {
...
+            });
+        });
+
+        describe('when checkoutMethod set to `Delivery`', () => {
+            const propsData = { checkoutMethod: 'Delivery' };
+
+            it('should emit success event when all fields are populated correctly', async () => {
+                // Arrange
+                const wrapper = mount(VueCheckout, { propsData });
+                // expect(wrapper.html()).toBe('1');
```
SachaZvetelman: Nit: remove.

packages/f-checkout/package.json
ashleynolan: Nope – if it's a dependency that's needed for a specific component, we should just include it in the `package.json` like this (or as a peerDependency).

If including it like this, then ideally we'd set-up webpack to chunk these dependencies so that they can be used (only really need to do once we know we'll have two occurrences of this package on the page at once, which I doubt would be the case for checkout).

Benefit of chunking it would also mean we could cache the chunked vuelidate JS for longer too (and across pages).

packages/f-checkout/package.json
```diff
@@ -44,10 +44,13 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "axios": "0.20.0",
+    "@justeat/f-alert": "^0.1.1",
     "@justeat/f-card": "0.6.0",
     "@justeat/f-form-field": "0.7.1",
-    "@justeat/f-services": "1.0.0"
+    "@justeat/f-services": "1.0.0",
+    "axios": "0.20.0",
```
ashleynolan, line 54: I don't think you need to include `axios` here, as it'll come down with `f-services` if you're using the axios methods.

packages/f-checkout/src/components/Address.vue
```diff
@@ -58,6 +108,13 @@ export default {
                 postcode: null
             }
         };
+    },
+
+    computed: {
+        localAddress: {
```
ashleynolan: Would be good to comment what `localAddress` is/represents (especially in comparison to `address`). Not knowing the context, I'd otherwise have to read through the whole component to try and work it out.

packages/f-checkout/src/components/Address.vue
```diff
@@ -74,4 +131,12 @@ export default {
         margin-bottom: -17px;
     }
 }
+
+.o-form-error {
+    display: flex;
+    align-items: center;
+    color: $red;
```
ashleynolan: Can use `$color-text--danger` here

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -115,7 +177,96 @@ export default {
 
         title () {
             return `${this.name}, confirm your details`;
+        },
+
+        shouldShowMobileNumberInvalidError () {
```
ashleynolan: Can you add JSDoc comments to any computed functions or `methods` below?

packages/f-checkout/src/components/Address.vue
ashleynolan: Are you using this computed to essentially `emit` to the `checkout` component? Not sure but I think it may be a bit tidier to use a watcher to do this if so (as would avoid needing two definitions of `address` in the component)

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -62,18 +91,50 @@ import AddressBlock from './Address.vue';
...
+        }
+    });
+
+    return {
+        validFields,
+        invalidFields
+    };
+};
+
+function isValidPostcode () {
```
ashleynolan: Is there a reason this wouldn't just be declared in the `methods` part of the component?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -62,18 +91,50 @@ import AddressBlock from './Address.vue';
 import FormSelector from './Selector.vue';
 import UserNote from './UserNote.vue';
 import tenantConfigs from '../tenants';
+import EventNames from '../event-names';
+
+const formValidationState = $v => {
```
ashleynolan: Same for this – could this be defined in the `methods`?

packages/f-checkout/src/components/Address.vue
FayeCarter: Can do, this class will be removed when `f-error-messages` is used as well. I should've added a note.

packages/f-checkout/src/components/Checkout.vue
FayeCarter: This was added in from review in a previous PR, but I'm not sure it's still needed 👍

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Sure, I'll give it a go 🙂

packages/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: Honestly, this was based off of f-registration so I'm not 100% sure. The aim was for it to be if there's not a server side error, but I'll update it with something a bit clearer

packages/f-checkout/src/components/Address.vue
FayeCarter: Yes that's what I'm trying to do. I tried a few approaches and this is the only way it worked for me yesterday.
I was hoping there was a better way. The double address definitely seems like overkill.
Thanks I'll have a look at using a watcher.

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Nope, I'll move it in.

packages/f-checkout/src/components/Checkout.vue
matthewhardern: Can I ask why we don't use $error on these, I not saying it should just more of a why are we not using $error ?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -115,7 +177,96 @@ export default {
...
         title () {
             return `${this.name}, confirm your details`;
+        },
+
+        shouldShowMobileNumberInvalidError () {
+            const mobileNumberInvalid = !this.$v.mobileNumber.required || !this.$v.mobileNumber.numeric || !this.$v.mobileNumber.minLength;
+            return mobileNumberInvalid && this.$v.mobileNumber.$dirty;
+        },
+
+        addressErrors () {
```
matthewhardern: Just a thought, have you considered using provide and inject so that instead of passing a giant list of address errors, you could do 
```
provide() {
    return {
      $v: this.$v
    }
}
``` 
on the parent. Then use 
```
inject: ['$v']
```
on the child to allow the child to have the access to the same instance of `$v`, allowing you to put the address errors in the address block making the parent less bloated.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -11,17 +18,31 @@
...
                     v-model="mobileNumber"
                     name="mobile-number"
                     data-test-id="input-mobile-number"
                     :label-text="copy.labels.mobileNumber"
-                    label-style="inline" />
+                    label-style="inline">
+                    // TODO: - Replace with f-error-message
+                    <template #error>
+                        <p
+                            v-if="shouldShowMobileNumberInvalidError"
```
matthewhardern: As an alternative way of doing this (if we wanted) we could loop over all the params under the `$v. mobileNumber.$params` with `v-for` accessing the index / key from the `v-for` to pull the error message from the messages object as long as the object key was the same.

```
<template v-for="(param, index) in $v. mobileNumber.$params" >
    {{ copy.validationMessages.mobileNumber[index] }}
</template>
```
Not a must but just a thought

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Ooh interesting that may save a whole load of issues. I haven't come across provide and inject. I'll have a look, thanks!

packages/f-checkout/package.json
FayeCarter: Axios is currently used in `CheckoutServiceApi`. I've left it in for now as I'm imagining it's going to be changed in later PRs. Is that OK?

packages/f-checkout/src/components/Checkout.vue
FayeCarter: I'll take a look

packages/f-checkout/src/components/Address.vue
```diff
@@ -2,62 +2,114 @@
...
+import ErrorMessage from '@justeat/f-error-message';
+import '@justeat/f-error-message/dist/f-error-message.css';
 import FormField from '@justeat/f-form-field';
 import '@justeat/f-form-field/dist/f-form-field.css';
 
 export default {
-    components: { FormField },
+    components: { FormField, ErrorMessage },
+
+    inject: ['shared'],
```
matthewhardern: recommend injecting the two `$v` and `copy` separately, makes it easier to read and understand. Nice work on getting it to run through provide and inject.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,24 +131,131 @@ export default {
...
             },
             buttonText: 'Go to payment',
-            delivery: CHECKOUT_METHOD_DELIVERY
+            genericErrorMessage: null
         };
     },
 
+    provide () {
+        const shared = {};
+        Object.defineProperties(shared, {
```
matthewhardern: again as above better if you just do them separately from a readability point of view.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,24 +131,131 @@ export default {
...
+                }
+            });
+
+            return {
+                validFields,
+                invalidFields
+            };
+        },
+
+        isValidPostcode () {
```
matthewhardern: Is this likely to be used in another component in the future?

packages/f-checkout/src/components/Checkout.vue
FayeCarter: I'm not sure this will work as it makes 3 error messages pop up at the same time. All 3 errors require the same error message at the moment.

packages/f-checkout/src/components/Address.vue
FayeCarter: Thanks, it was really good to learn about 😄. I'll separate them now

packages/f-checkout/src/components/Checkout.vue
FayeCarter: I can't see where it would be used at this point. Unless fozzie begins to extend into the user account components. I may be wrong though

kevinrodrigues: https://github.com/justeat/fozzie-components/pull/404/files#diff-e5a30357c1de2a2c914b4ad1279462e384f27637f6afed4d0ffc02eda10a71fdR17

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,24 +131,131 @@ export default {
...
+        },
+
+        isFormInvalid () {
+            this.$v.$touch();
+            return this.$v.$invalid;
         }
+    },
+
+    validations () {
+        if (this.checkoutMethod === 'Delivery') {
```
kevinrodrigues: You could move this to a computed prop and then reuse it here and also in the template above.

kevinrodrigues: It might also be better to write it like:

```javascript
const deliveryDetails = isDeliveryMethod === 'Delivery'
? {}
: {};

1. `{}` being the values you need to return.

2. `isDeliveryMethod` the computed for `this.checkoutMethod === 'Delivery'`

File: packages/f-checkout/src/components/tests/Checkout.test.js [hunk 23]
```diff
@@ -89,4 +91,88 @@ describe('Checkout', () => {
...
+                await wrapper.vm.$nextTick();
+
+                // Assert
+
+                expect(wrapper.vm.isMobileNumberInvalid).toBe(true);
+                expect(wrapper.emitted(EventNames.CheckoutFailure).length).toBe(1);
+                expect(wrapper.emitted(EventNames.CheckoutFailure)[0][0].invalidFields).toContain('mobileNumber');
+            });
+
+            it('should not show error message or emit failure event when the address input fields are not populated', async () => {
```
kevinrodrigues: Is this test correct? The expect below says that there should be an error unless i've read it wrong?

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +91,88 @@ describe('Checkout', () => {
...
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+
+                // Assert
+                expect(wrapper.emitted(EventNames.CheckoutSuccess).length).toBe(1);
+            });
+
+            it('should show error message and emit failure event when the mobile number field is not populated', async () => {
```
kevinrodrigues, line 185: Should there be a further test to check that the number meets the min & numeric requirement?

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Ahh I see. I'll put a note on it for now so it can be updated when it is in f-services?

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Right ok, that makes sense. I'll have a look, thanks!

packages/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: Ah yeah it's wrong. I'll update it

packages/f-checkout/src/components/Checkout.vue
FayeCarter: @kevinrodrigues just double checking, you mean like this right?
image

kevinrodrigues (dismissed):

packages/f-checkout/package.json
```diff
@@ -44,10 +44,14 @@
...
-    "axios": "0.20.0",
+    "@justeat/f-alert": "0.1.1",
     "@justeat/f-card": "0.6.0",
+    "@justeat/f-error-message": "0.2.0",
     "@justeat/f-form-field": "0.7.1",
-    "@justeat/f-services": "1.0.0"
+    "@justeat/f-services": "1.0.0",
+    "axios": "0.20.0",
+    "flush-promises": "1.0.2",
+    "vuelidate": "^0.7.6"
```
SachaZvetelman: This version should be pinned, as per my previous comment 🙂

packages/f-checkout/src/components/Address.vue
```diff
@@ -2,62 +2,114 @@
...
-        return {
-            address: {
-                line1: null,
-                line2: null,
-                city: null,
-                postcode: null
-            }
-        };
+    methods: {
+        checkValid (field) {
```
SachaZvetelman: I'd rename this to `isFieldValid` or similar, given it returns a `boolean`.

packages/f-checkout/src/components/Address.vue
```diff
@@ -2,62 +2,114 @@
...
+
+    computed: {
+        /*
+        * Validation methods return true if the validation conditions
+        * have not been met and the field has been `touched` by a user.
+        * The $dirty boolean changes to true when the user has focused/lost
+        * focus on the input field.
+        */
+
+        isAddressLine1Invalid () {
```
SachaZvetelman: I find the names a bit confusing, personally.
This computed property says `isInvalid`, but inside we call a function that says `checkValid`. One is negative ("invalid"), while the other one is positive ("valid"). I think we should do some renaming here. Happy to have a chat about this if you need!

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -11,24 +19,33 @@
...
                     v-model="mobileNumber"
                     name="mobile-number"
                     data-test-id="input-mobile-number"
                     :label-text="copy.labels.mobileNumber"
-                    label-style="inline" />
+                    label-style="inline">
+                    <template #error>
+                        <error-message
+                            v-if="isMobileNumberInvalid"
+                            data-test-id='error-mobile-number-empty'>
```
SachaZvetelman: Nit: `data-test-id` should be set with `"` (double quotes)

packages/f-checkout/src/components/Checkout.vue
SachaZvetelman: This has been marked as resolved, but I still see the same things I mentioned on my original comment.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,24 +131,159 @@ export default {
...
+            * have not been met and the field has been `touched` by a user.
+            * The $dirty boolean changes to true when the user has focused/lost
+            * focus on the input field.
+            */
+            const mobileNumberInvalid = !this.$v.mobileNumber.required || !this.$v.mobileNumber.numeric || !this.$v.mobileNumber.minLength;
+            return this.$v.mobileNumber.$dirty && mobileNumberInvalid;
+        },
+
+        isDeliveryMethod () {
+            return this.checkoutMethod === 'Delivery';
```
SachaZvetelman: We should probably extract this `"Delivery"` into a constant. I'd add it to our `constants` file.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,24 +131,159 @@ export default {
...
+
+        onFormSubmit () {
+            if (this.isFormInvalid()) {
+                const validationState = this.formValidationState(this.$v);
+                this.$emit(EventNames.CheckoutFailure, validationState);
+                return;
+            }
+
+            try {
+                this.$emit(EventNames.CheckoutSuccess);
```
SachaZvetelman: I'm not entirely familiar with the code you're writing, but under what circumstances would this actually fail and would cause the `catch` to be "triggered"? Emitting an event shouldn't ever fail. Please, correct me if I'm wrong here.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,24 +131,159 @@ export default {
...
+            } catch (error) {
+                let thrownErrors = error;
+                this.$emit(EventNames.CheckoutFailure, thrownErrors);
+
+                if (error && error.response && error.response.data && error.response.data.errors) {
```
SachaZvetelman: We're using the conditional chaining operator plugin in these projects, as far as I'm aware. That means you could do:
`if (error?.response?.data?.errors) {`

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,24 +131,159 @@ export default {
...
+                    thrownErrors = error.response.data.errors;
+                }
+                const shouldEmitCheckoutFailure = true;
+
+                if (Array.isArray(thrownErrors)) {
+                    this.genericErrorMessage = thrownErrors[0].description || 'Something went wrong, please try again later';
```
SachaZvetelman: If we get an array of errors, why do we should only the first one?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,24 +131,159 @@ export default {
...
+                } else {
+                    this.genericErrorMessage = error;
+                }
+
+                if (shouldEmitCheckoutFailure) {
+                    this.$emit(EventNames.CreateAccountFailure, thrownErrors);
```
SachaZvetelman: I'm struggling to understand here why we're throwing a `CreateAccountFailure` here. Could you please clarify?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,24 +131,159 @@ export default {
...
+        },
+
+        isFormInvalid () {
+            this.$v.$touch();
+            return this.$v.$invalid;
         }
+    },
+
+    validations () {
+        const deliveryDetails = this.isDeliveryMethod
```
SachaZvetelman: This shouldn't be `deliveryDetails` is it could be `collectionDetails`.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,24 +131,159 @@ export default {
...
+                    numeric,
+                    minLength: minLength(10)
+                }
+            }
+            : {
+                mobileNumber: {
+                    required,
+                    numeric,
+                    minLength: minLength(10)
+                }
```
SachaZvetelman: This bit is always needed, regardless of whether it's collection or delivery. Maybe you can add it always to the validation details and then only add the address ones if it's delivery.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -130,6 +297,11 @@ $line-height                              : 16px;
     color: $color-text;
     font-weight: $font-weight-base;
 
+    .c-checkout-alert {
+        width: 462px;
```
SachaZvetelman: Is this super-specific because it's the width of the checkout page?

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +91,116 @@ describe('Checkout', () => {
...
+                // Arrange
+                const wrapper = shallowMount(VueCheckout, { propsData });
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+                await wrapper.vm.$nextTick();
```
SachaZvetelman: I'm not so sure I understand why we're flushing promises and doing `$nextTick` in this case.
Besides that, `onFormSubmit` is not an `async` function, as far as I can tell, so I'm unsure why we're `await`ing here.

It might be me getting confused or missing some concepts here, but could you please clarify? (this applies to all the tests because I see we're flushing promises everywhere).

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +91,116 @@ describe('Checkout', () => {
...
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+
+                // Assert
+                expect(wrapper.vm.isMobileNumberInvalid).toBe(true);
+                expect(wrapper.emitted(EventNames.CheckoutFailure).length).toBe(1);
+                expect(wrapper.emitted(EventNames.CheckoutFailure)[0][0].invalidFields).toContain('mobileNumber');
+            });
+
+            it('should show error message and emit failure event when the mobile number field is populated with non numeric value', async () => {
```
SachaZvetelman, line 212: Is there any test in which we verify we're showing the right error message?

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +91,116 @@ describe('Checkout', () => {
...
+                const wrapper = mount(VueCheckout, { propsData });
+                wrapper.find('[data-test-id="input-mobile-number"]').setValue('07777777777');
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+
+                // Assert
+                expect(wrapper.vm.address.required).toBe(undefined);
+                expect(wrapper.emitted(EventNames.CheckoutSuccess).length).toBe(1);
```
SachaZvetelman, line 235: Might be a good idea to add an `expect` to check the `CheckoutFailure` event hasn't been emitted.

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +91,116 @@ describe('Checkout', () => {
...
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+                await flushPromises();
+
+                // Assert
+                expect(wrapper.emitted(EventNames.CheckoutSuccess).length).toBe(1);
+            });
+
+            it('should emit failure event when the address input fields are not populated', async () => {
```
SachaZvetelman: We should add more tests to check for valid postcodes.
In the future, these tests will be in `f-services`, when you move that RegEx for validations, and in here we'll just check basically that "the right validation is used".

packages/f-checkout/src/event-names.js
```diff
@@ -0,0 +1,15 @@
+const CheckoutSuccess = 'checkout-payment-success';
+const CheckoutFailure = 'checkout-payment-failure';
+const CheckoutStart = 'checkout-payment-start';
+const CheckoutInlineError = 'checkout-payment-inline-error';
+const VisitCheckoutPage = 'checkout-visit-payment-page';
```
SachaZvetelman: Should we rename to `CheckoutVisitPage` to keep it consistent?

SachaZvetelman (changes requested):

packages/f-checkout/package.json
FayeCarter: Noooooo I was sure these had all been removed 🤦‍♀️

packages/f-checkout/src/components/Checkout.vue
FayeCarter: It is, I'll move it to a variable

packages/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: I've now added an expect statement to each of the tests to get an error message. Not sure if it would be better to have them in separate tests though?

packages/f-checkout/src/components/Checkout.vue
FayeCarter: I may have borrowed some code and not updated it 😬

packages/f-checkout/src/components/Checkout.vue
FayeCarter: I've now added a method that makes this make slightly more sense. I hope

packages/f-checkout/src/components/Checkout.vue
FayeCarter: I can't get this to work for some reason. I'll keep exploring

packages/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: I've removed a lot of this now. I hadn't finished the full submit method

packages/f-checkout/src/components/Address.vue
```diff
@@ -2,62 +2,114 @@
...
 
 export default {
-    components: { FormField },
+    components: { FormField, ErrorMessage },
+
+    inject: ['$v', 'copy'],
 
     props: {
-        labels: {
+        value: {
```
SachaZvetelman: Not sure what `value` is. Let's try finding a name that describes what it is, please 🙂

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -2,7 +2,15 @@
...
         :data-theme="theme"
         :class="$style['c-checkout']"
-        data-test-id='checkout-component'>
+        data-test-id="checkout-component">
+        <alert
+            v-if="genericErrorMessage"
+            :locale="locale"
+            type="danger"
+            :class="$style['c-checkout-alert']"
+            heading="Error">
```
SachaZvetelman: `Error` should also come from the translation files.
Btw, we need to check if `Error` is the correct word to show here. We might have to ask the designers.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,39 +143,185 @@ export default {
...
+        const collectionDetails = {
+            mobileNumber: {
+                required,
+                numeric,
+                minLength: minLength(10)
+            }
+        };
+
+        if (this.isDeliveryMethod) {
+            collectionDetails.address = {
```
SachaZvetelman: Hmmm, `collectionDetails` doesn't make a lot of sense to me, sorry. But if it's `delivery`, then it's not `collection details` 😬

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +98,181 @@ describe('Checkout', () => {
             });
         });
     });
+
+    describe('when form submitted', () => {
+        function mountComponentAndAttachToDocument (propsData) {
+            const div = document.createElement('div');
+            document.body.appendChild(div);
+            return mount(VueCheckout, { propsData, attachTo: div });
+        }
```
SachaZvetelman: This is a bit of weird way to get and mount the component. I haven't seen it before. Is there any particular reason why we're doing it this way and not the way we do it in most of our `fozzie-components` tests?

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +98,181 @@ describe('Checkout', () => {
...
+                // Arrange
+                wrapper.find('[data-test-id="input-mobile-number"]').setValue('hs;-j`$e&1l');
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+                const mobileNumberEmptyMessage = wrapper.find('[data-test-id="error-mobile-number"]');
+
+                // Assert
+                expect(wrapper.vm.isMobileNumberValid).toBe(false);
+                expect(mobileNumberEmptyMessage.exists()).toBe(true);
```
SachaZvetelman: Probably you want to check what text you're showing, not just that it exists. `toMatchSnapshot` would do.

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +98,181 @@ describe('Checkout', () => {
...
+                wrapper = mountComponentAndAttachToDocument(propsData);
+            });
+
+            afterEach(() => {
+                wrapper.destroy();
+            });
+
+            it('should emit success event when all fields are populated correctly', async () => {
+                // Arrange
+                wrapper.find('[data-test-id="input-mobile-number"]').setValue('07777777777');
```
SachaZvetelman: Could we extract this information to a constant at the top of the file? We're reusing the address and the mobile number in some parts.

SachaZvetelman (changes requested):

packages/f-checkout/src/components/Address.vue
FayeCarter: @SachaZvetelman  - This seems to be the only way that it will work with a nested v-model. I've tried setting it to address (because it would make more sense that way), but that way it can't read the prop at all.
I'm hoping I'm wrong here though

packages/f-checkout/src/components/Checkout.vue
FayeCarter: It was delivery before. I think I changed it because of another review 🤦‍♀️. But agreed it makes more sense as delivery

packages/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: I couldn't get it to work how it is normally done. Just tried it again ... it works. It's changed now.

packages/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: I'll give it a go

packages/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: @SachaZvetelman Just wanted to check I've done this correctly now

packages/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: @SachaZvetelman Same with this, just wanted to check that I've written it as you expected

packages/f-checkout/babel.config.js
```diff
@@ -3,7 +3,10 @@ module.exports = api => {
     const isTest = api.env('test');
     const presets = [];
     const plugins = [
-        '@babel/plugin-proposal-optional-chaining' // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
+        ['@babel/plugin-proposal-optional-chaining', // https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining
+            {
+                loose: true // Need this for Storybook v6 beta.
+            }]
```
SachaZvetelman: Nit: indentation looks funny here.

packages/f-checkout/src/tenants/en-GB.js
```diff
@@ -28,6 +28,9 @@ export default {
     allergyText: 'If you or someone you’re ordering for has a food allergy or intolerance, click here',
 
     errorMessages: {
-        genericServerError: 'Something went wrong, please try again later'
+        errors: {
```
SachaZvetelman: Why do we have `errors` within `errorMessages`? :)

SachaZvetelman (changes requested):

packages/f-checkout/src/components/tests/Checkout.test.js
SachaZvetelman: @FayeCarter, I was actually talking about the data we're using in `setValue` (i.e. the phone number, the address, etc)

packages/f-checkout/babel.config.js
FayeCarter: Ahh didn't realise this was still in. This was an experiment to try and get babel to work. I'll remove it

packages/f-checkout/src/tenants/en-GB.js
FayeCarter: It's been a long day 😂

packages/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: Ahhhh that makes sense. Thanks

packages/f-checkout/src/components/Address.vue
FayeCarter: I was wrong 🥳

packages/f-checkout/src/components/Address.vue
```diff
@@ -6,58 +6,110 @@
...
             v-model="address.postcode"
             name="address-postcode"
             data-test-id="input-address-postcode"
-            :label-text="labels.postcode"
-            label-style="inline" />
+            :label-text="copy.labels.postcode"
+            label-style="inline">
+            <template #error>
+                <error-message
+                    v-if="!isAddressPostcodeValid"
```
SachaZvetelman: I'm thinking, should we rename this one `isAddressPostcodeEmpty` and then the one below would be `isAddressPostcodeValid`?
Up to you, really. My reasoning is that the first one is actually checking if it's empty or not, not if it's "valid".

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,39 +144,188 @@ export default {
...
+
+                await CheckoutServiceApi.submitCheckout(this.checkoutUrl, this.tenant, checkoutData, this.checkoutTimeout);
+
+                this.$emit(EventNames.CheckoutSuccess);
+            } catch (error) {
+                let thrownErrors = error;
+                if (error && error.response && error.response.data && error.response.data.errors) {
+                    thrownErrors = error.response.data.errors;
+                }
+                // TODO: Babel 'optional-chaining' plugin not working in storybook
```
SachaZvetelman: I guess we can remove this now?

packages/f-checkout/package.json
```diff
@@ -44,10 +44,14 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "axios": "0.20.0",
+    "@justeat/f-alert": "0.1.1",
     "@justeat/f-card": "0.6.0",
+    "@justeat/f-error-message": "0.2.0",
     "@justeat/f-form-field": "0.7.1",
-    "@justeat/f-services": "1.0.0"
+    "@justeat/f-services": "1.0.0",
```
xander-marjoram: We should probably update this to `1.3.0` soon

packages/f-checkout/src/components/Address.vue
```diff
@@ -6,58 +6,110 @@
...
             <form-field
                 v-model="address.city"
                 :class="$style['c-addressGroup-formField']"
                 name="address-city"
                 data-test-id="input-address-city"
-                :label-text="labels.city"
-                label-style="inline" />
+                :label-text="copy.labels.city"
+                label-style="inline">
+                <template #error>
```
xander-marjoram, line 35: Is it necessary to use the `template`? Would this work?
```vue
<error-message #error>
    v-if="!isAddressCityValid"
    data-test-id="error-address-city-empty">
    {{ copy.validationMessages.city.requiredError }}
</error-message>
```

xander-marjoram: I see you also have it below when passing in two errors, which makes more sense

packages/f-checkout/src/components/Address.vue
```diff
@@ -6,58 +6,110 @@
...
+import ErrorMessage from '@justeat/f-error-message';
+import '@justeat/f-error-message/dist/f-error-message.css';
 import FormField from '@justeat/f-form-field';
 import '@justeat/f-form-field/dist/f-form-field.css';
 
 export default {
-    components: { FormField },
+    components: { FormField, ErrorMessage },
+
+    inject: ['$v', 'copy'],
```
xander-marjoram: Would be good to get an explanation of this and why it's necessary for those of us who aren't familiar with provide/inject

FayeCarter: Yeah sure, I'll add something

packages/f-checkout/src/components/Address.vue
FayeCarter: Can do. I'd image it would be best to rename the others that I have for address and mobile number from 'Valid' to be 'Empty' as well?

packages/f-checkout/src/components/Address.vue
FayeCarter: Ah, I thought you needed the template tag for a named slot, I'll have a look

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Will update with design as a separate ticket

packages/f-checkout/src/components/Address.vue
```diff
@@ -6,58 +6,114 @@
...
             v-model="address.postcode"
             name="address-postcode"
             data-test-id="input-address-postcode"
-            :label-text="labels.postcode"
-            label-style="inline" />
+            :label-text="copy.labels.postcode"
+            label-style="inline">
+            <template #error>
+                <error-message
+                    v-if="!isAddressPostcodeEmpty"
```
xander-marjoram: This doesn't quite make semantic sense to me. It reads as "if postcode is **not** empty then show the error", shouldn't we be showing the error if the postcode is empty?

packages/f-checkout/src/components/Address.vue
```diff
@@ -6,58 +6,114 @@
...
+        isAddressCityEmpty () {
+            return !this.isFieldValid('city');
+        },
+
+        isAddressPostcodeEmpty () {
+            return !this.isFieldValid('postcode');
+        },
+
+        isAddressPostcodeValid () {
+            return !(this.$v.addressErrors.postcode.$dirty && !this.$v.addressErrors.postcode.isValidPostcode);
```
xander-marjoram: Can we simplify this to:
`!this.$v.addressErrors.postcode.$dirty || this.$v.addressErrors.postcode.isValidPostcode`?
I think it's more readable, you can see it means "untouched or actually valid".

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,39 +148,167 @@ export default {
...
+        },
+
+        isMobileNumberValid () {
+            /*
+            * Validation methods return true if the validation conditions
+            * have not been met and the field has been `touched` by a user.
+            * The $dirty boolean changes to true when the user has focused/lost
+            * focus on the input field.
+            */
+            const isMobileNumberValid = !(!this.$v.mobileNumber.required || !this.$v.mobileNumber.numeric || !this.$v.mobileNumber.minLength);
```
xander-marjoram: This can also be simplified:
`this.$v.mobileNumber.required && this.$v.mobileNumber.numeric && this.$v.mobileNumber.minLength`

packages/f-checkout/src/components/Address.vue
FayeCarter: Thanks

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,39 +148,167 @@ export default {
...
+            return !(this.$v.mobileNumber.$dirty && !isMobileNumberValid);
```
xander-marjoram: `!this.$v.mobileNumber.$dirty || isMobileNumberValid`

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Thank you!

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Brilliant, thanks

packages/f-checkout/src/components/Address.vue
FayeCarter: Good point

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +98,198 @@ describe('Checkout', () => {
...
+
+            beforeEach(() => {
+                CheckoutServiceApi.submitCheckout.mockClear();
+                CheckoutServiceApi.submitCheckout.mockImplementation(async () => Promise.resolve());
+
+                wrapper = mount(VueCheckout, { propsData });
+            });
+
+            afterEach(() => {
+                wrapper.destroy();
```
xander-marjoram: I'm happy for this to stay, but is this necessary if we're reassigning it for every test anyway?

FayeCarter: Tests still pass without it. I'm happy to remove it

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +98,194 @@ describe('Checkout', () => {
...
+
+            it('should not show error message or emit failure event when the address input fields are not populated', async () => {
+                // Arrange
+                wrapper.find('[data-test-id="input-mobile-number"]').setValue(mobileNumber);
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+
+                // Assert
+                expect(wrapper.vm.address.required).toBe(undefined);
```
xander-marjoram: I think this might be checking `data.address` rather than the validation, so this is likely to always be undefined.

packages/f-checkout/package.json
```diff
@@ -44,10 +44,15 @@
...
+    "@justeat/f-alert": "0.1.1",
+    "@justeat/f-button": "0.1.0",
     "@justeat/f-card": "0.6.0",
+    "@justeat/f-error-message": "0.2.0",
     "@justeat/f-form-field": "0.7.1",
-    "@justeat/f-services": "1.0.0"
+    "@justeat/f-services": "1.4.0",
+    "axios": "0.20.0",
+    "flush-promises": "1.0.2",
+    "vuelidate": "0.7.6"
```
ashleynolan, line 55: Not necessarily one for this PR, but we should definitely create a ticket to spike if we can make Vueliddate a `peerDependency` so that we don't bundle it into every component that needs validation (as registration also uses vuelidate for instance).

packages/f-checkout/src/components/Address.vue
ashleynolan: Not sure if this has been addressed, but why do you need the `template` tags here?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,39 +148,167 @@ export default {
...
+            return !this.$v.mobileNumber.$dirty || isMobileNumberValid;
+        },
+
+        isDeliveryMethod () {
+            return this.checkoutMethod === CHECKOUT_METHOD_DELIVERY;
         }
+    },
+
+    methods: {
+        async onFormSubmit () {
```
ashleynolan, line 235: This is quite a big method. Can it be split up so that it calls other smaller methods that do smaller things (instead of one big method that does all the things)? So imo the `try` should be in a separate method that handles the API call, and then the `catch` should be a separate method that handles the error state being applied (you could likely just put the inner code of the try/catch into different methods).

Would also be good to have some comments to explain what's happening (so I'm guessing the try/catch if trying to submit to an endpoint and if this fails, it catches the error?).

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,39 +148,167 @@ export default {
...
+            * Validation methods return true if the validation conditions
+            * have not been met and the field has been `touched` by a user.
+            * The $dirty boolean changes to true when the user has focused/lost
+            * focus on the input field.
+            */
+            const isMobileNumberValid = this.$v.mobileNumber.required && this.$v.mobileNumber.numeric && this.$v.mobileNumber.minLength;
+            return !this.$v.mobileNumber.$dirty || isMobileNumberValid;
+        },
+
+        isDeliveryMethod () {
```
ashleynolan: Not sure the name makes complete sense. Looks like it's checking for if the checkout method is `Delivery` in which case I'd call this `isDelivery` or `isCheckoutMethodDelivery` (if it should be more explicit).

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -99,39 +148,167 @@ export default {
...
+        }
+
+        return deliveryDetails;
     }
 };
 </script>
 
 <style lang="scss" module>
 $line-height                              : 16px;
+$checkout-width                           : 462px;
```
ashleynolan: Looks like a bit of a magic number – is there a reason it's exactly `462px` rather than just `460px`?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -150,21 +327,13 @@ $line-height                              : 16px;
         @include font-size(heading-s);
```
ashleynolan: Referring to the line above (don't need to do this as part of this PR, but should be done at some point).

Best to give the `h1` a class of `.c-checkout-title` as that way if the heading tag ever changes, the styles are less brittle and won't break.

packages/f-checkout/package.json
FayeCarter: Sure, I can set of a ticket for this if that's ok?

ashleynolan: Yep – that'd be good 👍

packages/f-checkout/src/components/Checkout.vue
FayeCarter: So this is the width that the checkout component is set to in consumerWeb. I was trying to get it as close to ConsumerWeb for now. Is that OK?

xander-marjoram (dismissed):

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Hey @ashleynolan, I've updated and created new methods/descriptions. Are these ok?

xander-marjoram (dismissed):

dandel10n (dismissed):
Looks good to me. I noticed that allerst are not responsive, but this will be covered in next ticket.

packages/f-checkout/src/components/Address.vue
```diff
@@ -6,58 +6,114 @@
...
-        return {
-            address: {
-                line1: null,
-                line2: null,
-                city: null,
-                postcode: null
-            }
-        };
+    methods: {
+        isFieldEmpty (field) {
```
ashleynolan, line 119: Super NIT – can you add some JSDoc comments to this function? (happy for this to happen in next PR though)

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -2,7 +2,15 @@
     <div
         :data-theme="theme"
         :class="$style['c-checkout']"
-        data-test-id='checkout-component'>
+        data-test-id="checkout-component">
+        <alert
+            v-if="genericErrorMessage"
+            :locale="locale"
```
ashleynolan: Not sure (as haven't looked at the alert component), but if it uses the same code as other components, it should just inherit the locale from the global `i18n` config that gets set, so you may not need to pass this through.

packages/f-checkout/src/components/Checkout.vue
ashleynolan: Yep – looks much better 👍 

Could you add JSDoc comments to the methods too? (we're trying to add these to anything added to the methods object in components going forward)

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -89,4 +98,201 @@ describe('Checkout', () => {
             });
         });
     });
+
+    describe('when form submitted', () => {
+        const mobileNumber = '07777777777';
+
+        const address = {
+            line1: 'Fleet Place House',
```
ashleynolan: I'd probably change this to be data that means as little as possible (as close to lorem ipsum style as you can while still passing the validations)

packages/f-checkout/src/components/Address.vue
```diff
@@ -6,58 +6,119 @@
...
+                </template>
+            </form-field>
 
             <form-field
                 v-model="address.line2"
                 :class="$style['c-addressGroup-formField']"
                 name="address-line-2"
                 data-test-id="input-address-line-2"
-                :label-text="labels.line2"
+                :label-text="$t.labels.line2"
```
SachaZvetelman: This doesn't look right. As far as I know, and I could be wrong, it should be `$t('labels.line2')`.
Same for all of the places where we're using `$t`.

packages/f-checkout/src/components/Address.vue
FayeCarter: Discussed this with Ash and Xander, template tags are needed for the named slots and allows standardisation where multiple error messages are needed

packages/f-checkout/src/components/Address.vue
FayeCarter: This is now updated 😄

SachaZvetelman (dismissed):

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -271,6 +276,9 @@ export default {
         },
 
         isFormValid () {
+            /*
+            * Check to see if any `Vulidate` calidation errors
```
xander-marjoram: Nit: typos `Vuelidate`, `validation`

I think I might have seen other instances of "vulidate" as well?

packages/f-checkout/package.json
```diff
@@ -44,10 +44,14 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
+    "@justeat/f-alert": "0.1.1",
+    "@justeat/f-button": "0.1.0",
     "@justeat/f-card": "0.6.0",
+    "@justeat/f-error-message": "0.2.0",
     "@justeat/f-form-field": "0.7.1",
+    "vuelidate": "0.7.6",
```
xander-marjoram: Nit: Can you move this dependency to the bottom please?

xander-marjoram (approved):

SachaZvetelman (approved):

PR #402, opened 2020-10-29
f-footer@4.3.0 - new country list in country selector

dandel10n:
### Changed
- New country list in country selector
- New round flag icons in country selector

Screen Shot 2020-10-29 at 17 00 17

[Image from Gyazo](https://gyazo.com/f15f22fce5a4e0c5944effe8badbbf21)


xander-marjoram (commented):
Are all of the country lists the same now? Might it be worth having the ordered list of icons/links in a separate file that we import to make changes easier in future?

xander-marjoram:
Also how does it look on mobile? Do we run into any scrolling issues now?

simonsandersje (dismissed):

dandel10n:
> Are all of the country lists the same now? Might it be worth having the ordered list of icons/links in a separate file that we import to make changes easier in future?

@xander-marjoram  It was the same before, now it is just bigger. I don't remember why we haven't done it like you suggest from the start. I will try now

dandel10n:
> Also how does it look on mobile? Do we run into any scrolling issues now?

Tested in storybook, didn't notice any issues. I added a short giff to the description

xander-marjoram (approved):

simonsandersje (approved):

PR #409, opened 2020-11-03
f-registration@v0.39.0

ross-gratton:
Added f-error-message to inline error messages in f-registration

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/f-registration/CHANGELOG.md
+++ b/packages/f-registration/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Latest (to be added to next release)
 ------------------------------
+*November 3, 2020*
+
+### Changed
+- Added f-error-message for inline registration page errors
+
 *October 28, 2020*
 
 ### Added
--- a/packages/f-registration/CHANGELOG.md
+++ b/packages/f-registration/CHANGELOG.md
@@ -4,7 +4,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-Latest (to be added to next release)
+v0.39.0
 ------------------------------
 *November 3, 2020*
 
--- a/packages/f-registration/src/components/Registration.vue
+++ b/packages/f-registration/src/components/Registration.vue
@@ -41,27 +41,21 @@
                     label-style="inlineNarrow"
                     @blur="formFieldBlur('firstName')">
                     <template #error>
-                        <p
+                        <error-message
                             v-if="shouldShowFirstNameRequiredError"
-                            :class="$style['o-form-error']"
-                            data-test-id='error-first-name-empty'>
-                            <warning-icon :class="$style['o-form-error-icon']" />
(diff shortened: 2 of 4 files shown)


packages/f-registration/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Latest (to be added to next release)
 ------------------------------
+*November 3, 2020*
```
ashleynolan, line 9: Should change the `Latest` line above to be the version number `v0.39.0`

packages/f-registration/src/components/Registration.vue
```diff
@@ -41,27 +41,21 @@
                     label-style="inlineNarrow"
                     @blur="formFieldBlur('firstName')">
                     <template #error>
-                        <p
+                        <error-message
```
ashleynolan, line 44: Can you delete the `.o-form-error` styles (etc) from the style block in this component now?

ross-gratton: Yes we can, I knew there was something I'd missed.

luke-andrew-parry (approved):

packages/f-registration/test-utils/component-objects/f-registration.component.js
```diff
@@ -10,19 +10,19 @@ const privacyPolicyLink = () => registrationComponent().$('[data-test-id="privac
 const cookiesPolicyLink = () => registrationComponent().$('[data-test-id="cookies-policy-link"]');
 
 // Validation errors
-const firstNameEmptyError = () => $('[data-test-id="error-first-name-empty"]');
```
kevinrodrigues, line 13: Should these have changed? I think the convention is to use `data-test-id` for test related setups?

xander-marjoram: @kevinrodrigues I've had a chat with Ross and we're going to address this in a follow-up ticket, looking to replace the selectors with [findComponent](https://vue-test-utils.vuejs.org/api/wrapper/findComponent.html).

xander-marjoram (approved):

kevinrodrigues (approved):

packages/f-registration/test-utils/component-objects/f-registration.component.js
ross-gratton: Having spoken with the QEs this approach will remain until a satisfactory alternative can be found. The issue with findComponent is that while it will work in Jest/Vue, there is no comparable function in WebDriver.io, so we're still left with no clear better method than to define an additional `data-test-title` attribute for component tests.

xander-marjoram: Ok. I think we should remove the prop from f-error-message because it's not necessary for accessing the element (you can just apply it directly like so: `<error-message data-test-id="" />`

I also do still think we should use findComponent wherever possible in Jest tests to simplify the tests and reduce some noise in the templates.

ross-gratton: I'll spin up a ticket for the f-error-message work and we should probably engage with the QEs to see if we can standardise the test approach.

PR #411, opened 2020-11-04
vue-i18n - Common wiring for globalisation within components (WIP)

tommcclean:
Not ready for reviews or merge. Prototype

To break out into 3 PR's
1. F-Globalisation module
2. Update Component generator to install above module
3. Incorporate changes into F-Checkout


lerna.json
```diff
@@ -2,7 +2,8 @@
   "version": "independent",
   "npmClient": "yarn",
   "packages": [
-    "packages/*"
+    "packages/*",
+    "utilities/*"
```
tommcclean, line 6: I actually think we should consider this a "service"

PR #412, opened 2020-11-04
f-checkout@0.8.0 Get necessary data from the CheckoutOrchestrator

pkedziora:
Displays data populated from GET endpoint

v0.8.0
*November 4, 2020*

### Added
- Checkout component calls server to get data

---

## UI Review Checks
Screenshot 2020-11-05 at 10 28 26
Screenshot 2020-11-05 at 10 28 58

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/f-checkout/package.json
+++ b/packages/f-checkout/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-checkout",
   "description": "Fozzie Checkout – Fozzie Checkout Component",
-  "version": "0.6.0",
+  "version": "0.8.0",
   "main": "dist/f-checkout.umd.min.js",
   "files": [
     "dist",
@@ -44,10 +44,11 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "axios": "0.20.0",
     "@justeat/f-card": "0.6.0",
     "@justeat/f-form-field": "0.7.1",
-    "@justeat/f-services": "1.0.0"
+    "@justeat/f-services": "1.0.0",
+    "axios": "0.20.0",
+    "axios-mock-adapter": "^1.19.0"
   },
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.1.1"
--- a/packages/f-checkout/package.json
+++ b/packages/f-checkout/package.json
@@ -48,7 +48,7 @@
     "@justeat/f-form-field": "0.7.1",
     "@justeat/f-services": "1.0.0",
     "axios": "0.20.0",
-    "axios-mock-adapter": "^1.19.0"
+    "axios-mock-adapter": "1.19.0"
   },
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.1.1"
--- a/packages/f-checkout/src/components/Checkout.vue
+++ b/packages/f-checkout/src/components/Checkout.vue
@@ -13,22 +13,24 @@
             :class="$style['c-card--dimensions']">
             <form action="post">
                 <form-field
-                    v-model="mobileNumber"
+                    v-model="customer.phoneNumber"
                     name="mobile-number"
                     data-test-id="input-mobile-number"
                     :label-text="copy.labels.mobileNumber"
                     label-style="inline" />
 
                 <address-block
-                    v-if="checkoutMethod === delivery"
+                    v-if="isDelivery"
                     :labels="copy.labels"
+                    :address="fulfillment.address"
                     data-test-id='address-block' />
 
                 <form-selector
-                    :order-method="checkoutMethod"
+                    :order-method="serviceType"
+                    :times="fulfillment.times"
                     data-test-id='selector' />
 
-                <form-selector />
+
                 <user-note data-test-id='user-note' />
                 <button
                     :class="[
@@ -38,7 +40,6 @@
                     data-test-id="allergy-button">
                     {{ copy.allergyText }}
                 </button>
-
                 <button
                     :class="[
                         $style['o-btn--payment'],
@@ -63,6 +64,7 @@ import AddressBlock from './Address.vue';
 import FormSelector from './Selector.vue';
 import UserNote from './UserNote.vue';
 import tenantConfigs from '../tenants';
+import CheckoutServiceApi from '../services/CheckoutServiceApi';
 
 export default {
     name: 'VueCheckout',
@@ -83,8 +85,19 @@ export default {
 
         checkoutMethod: {
             type: String,
-            default: 'Collection',
+            default: 'delivery',
             validator: value => (VALID_CHECKOUT_METHOD.indexOf(value) !== -1)
+        },
+
+        checkoutUrl: {
+            type: String,
+            required: true
+        },
+
+        getCheckoutTimeout: {
+            type: Number,
+            required: false,
+            default: 1000
         }
     },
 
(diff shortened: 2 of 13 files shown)


packages/f-checkout/stories/Checkout.stories.js
```diff
@@ -25,5 +26,6 @@ Checkout.parameters = {
 Checkout.storyName = 'f-checkout';
 
 Checkout.args = {
-    checkoutMethod: CHECKOUT_METHOD_COLLECTION
+    checkoutMethod: CHECKOUT_METHOD_COLLECTION,
+    checkoutUrl: 'http://all-checkoutorchestrator-qa1.cwa.je-labs.com/uk/checkout/12345'
```
DamianMullins: Do we want references to QA environments in an open source code? Sounds like something InfoSec would want us to purge?

pkedziora: @DamianMullins isn't this a private repo?

kevinrodrigues: It's in the process of being made public @pkedziora

DamianMullins: Which does worry me. It's very easy for things like this to slip through. I know iOS has a similar approach with modules in a mono-repo but at least theirs is private.

pkedziora: Fixed

DamianMullins: Now that this URL has been added in a PR it will always be visible even if we squash commits etc. There's no way to fully purge it without requesting Github to delete the content or creating a new repo and copying the code across (which means we lose all PRs, issues etc.

pkedziora: This url is hardly a secret

DamianMullins: Not in this case it isn't, but what if something sensitive gets through like some credentials?

pkedziora: And it's not available externally

packages/f-checkout/package.json
```diff
@@ -44,10 +44,11 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "axios": "0.20.0",
     "@justeat/f-card": "0.6.0",
     "@justeat/f-form-field": "0.7.1",
-    "@justeat/f-services": "1.0.0"
+    "@justeat/f-services": "1.0.0",
+    "axios": "0.20.0",
```
SachaZvetelman, line 50: @ashleynolan, we should actually be using `f-services` for this, right? Or is it correct to be importing `axios`?

packages/f-checkout/package.json
```diff
@@ -44,10 +44,11 @@
...
+    "axios-mock-adapter": "^1.19.0"
```
SachaZvetelman: Please remove the `^` so we can pin the version down.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -13,22 +13,24 @@
             :class="$style['c-card--dimensions']">
             <form action="post">
                 <form-field
-                    v-model="mobileNumber"
+                    v-model="customer.phoneNumber"
```
SachaZvetelman, line 16: If we're using `phoneNumber` instead of `mobileNumber`, we should rename the rest of the attribute values for this element. cc @FayeCarter

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -83,8 +85,19 @@ export default {
 
         checkoutMethod: {
             type: String,
-            default: 'Collection',
+            default: 'delivery',
```
SachaZvetelman, line 88: _I think_, @FayeCarter is extracting this to a `constant` somewhere. So when we merge Faye's changes, let's use that constant instead of using a magic string here.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -94,28 +107,89 @@ export default {
...
             copy: { ...localeConfig },
             theme,
-            firstName: 'firstName',
-            mobileNumber: null,
-            address: {
-                lineOne: null,
-                lineTwo: null,
-                city: null,
-                postcode: null
+            id: '',
```
SachaZvetelman: Maybe a silly question, but what is this `id`? Is it a `CheckoutId`? `BasketId`?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -94,28 +107,89 @@ export default {
...
 
     computed: {
         name () {
-            return (this.firstName.charAt(0).toUpperCase() + this.firstName.slice(1));
+            return (this.customer.firstName.charAt(0).toUpperCase() + this.customer.firstName.slice(1));
         },
 
         title () {
-            return `${this.name}, confirm your details`;
+            return this.name ? `${this.name}, confirm your details` : '';
```
SachaZvetelman: At this stage, we should always have a name. If we don't have a name it means it's a guest checkout (we have a separate ticket for that), so I don't think we need this change.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -94,28 +107,89 @@ export default {
...
+        },
+
+        isDelivery () {
+            return this.serviceType === CHECKOUT_METHOD_DELIVERY;
+        },
+
+        tenant () {
+            return {
```
SachaZvetelman: Nit: I'd probably assign this object to a `const` called `tenants` or something like that to improve readability a bit.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -94,28 +107,89 @@ export default {
...
+            this.customer = data.customer;
+            this.fulfillment.times = data.fulfillment.times;
+
+            this.fulfillment.address = {
+                line1: data.fulfillment.address.lines[0],
+                line2: data.fulfillment.address.lines[1],
+                city: data.fulfillment.address.lines[3],
+                postcode: data.fulfillment.address.postalCode
+            };
+            this.messages = data.messages;
```
SachaZvetelman, line 181: Very nit: add an empty line before this, please.

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -4,7 +4,7 @@ import VueCheckout from '../Checkout.vue';
 
 describe('Checkout', () => {
     it('should be defined', () => {
-        const propsData = {};
+        const propsData = { checkoutUrl: '/checkout/uk/12345' };
```
SachaZvetelman: We can extract `checkout/uk/12345` to a constant since we're using this in multiple places.

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -25,10 +26,11 @@ describe('Checkout', () => {
                 expect(selectorComponent.attributes('ordermethod')).toEqual(definedType);
             });
 
-            it('should display the address block if set to `Delivery`', () => {
+            it('should display the address block if set to `delivery`', () => {
                 // Arrange
                 const propsData = {
-                    checkoutMethod: 'Delivery'
+                    checkoutMethod: 'delivery',
```
SachaZvetelman, line 34: Same as above, let's try to use the constant that Faye is introducing in the future.

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -39,10 +41,11 @@ describe('Checkout', () => {
                 expect(addressBlock.exists()).toBe(true);
             });
 
-            it('should not display the address block if set to `Collection`', () => {
+            it('should not display the address block if set to `collection`', () => {
                 // Arrange
                 const propsData = {
-                    checkoutMethod: 'Collection'
+                    checkoutMethod: 'collection',
```
SachaZvetelman, line 49: Same as above.

packages/f-checkout/src/demo/checkoutMock.js
```diff
@@ -0,0 +1,11 @@
+import axios from 'axios';
+import MockAdapter from 'axios-mock-adapter';
+import checkoutJson from './checkout.json';
+
+
```
SachaZvetelman: Ultra nit: remove empty line please :D

SachaZvetelman (changes requested):
Probably something we should do on a separate PR (unless you want to do it in this PR, and I can help you with that), but we should introduce `Vuex` to be able to keep state in a centralised manner. That will help us reduce the number of props and data we need to inject into child components.

packages/f-checkout/package.json
pkedziora: @SachaZvetelman btw I haven't imported axios. It was already there, it just moved couple of lines down

packages/f-checkout/src/components/Checkout.vue
pkedziora: @SachaZvetelman it's a CheckoutId that is currently actually a BasketId . David mentioned that it will be it separate thing in the future. For now it maps to BasketId

packages/f-checkout/package.json
xander-marjoram: Please can you pin this version (remove `^`)

packages/f-checkout/src/components/Checkout.vue
SachaZvetelman: Can we please rename it to `basketId` for now?

pkedziora: @SachaZvetelman I think it should be checkoutId in that case as that's what the id is meant to represent

packages/f-checkout/package.json
pkedziora: Fixed

packages/f-checkout/src/demo/checkoutMock.js
pkedziora: Fixed

packages/f-checkout/src/components/tests/Checkout.test.js
pkedziora: Fixed

packages/f-checkout/src/components/Checkout.vue
pkedziora: Fixed

packages/f-checkout/src/components/Checkout.vue

packages/f-checkout/src/components/Checkout.vue

packages/f-checkout/src/components/Checkout.vue
pkedziora: Changed to checkoutId

packages/f-checkout/package.json
pkedziora: Fixed

pkedziora:
@SachaZvetelman I would prefer to do Vuex as a separate PR

packages/f-checkout/stories/Checkout.stories.js
DamianMullins: In any case the same applies to all of our open source code so I guess we just need to be vigilant.

PR #425, opened 2020-11-10
f-checkout@0.7.0 - Integrating with f-globalisation

SachaZvetelman:
v0.7.0
*November 11, 2020*

### Added
- Integration with `f-globalisation`.


packages/f-checkout/src/components/Checkout.vue
```diff
@@ -89,13 +91,9 @@ export default {
...
     data () {
-        const locale = globalisationServices.getLocale(tenantConfigs, this.locale, this.$i18n);
-        const localeConfig = tenantConfigs[locale];
-        const theme = globalisationServices.getTheme(locale);
-
         return {
-            copy: { ...localeConfig },
-            theme,
+            tenantConfigs,
+            theme: 'jet',
```
tommcclean: Could put this hardcoded at the top where it is used; to simplify data (I have done in the generator)

tommcclean (dismissed):

packages/f-checkout/package.json
```diff
@@ -44,10 +44,11 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "axios": "0.20.0",
     "@justeat/f-card": "0.6.0",
     "@justeat/f-form-field": "0.7.1",
-    "@justeat/f-services": "1.0.0"
+    "@justeat/f-globalisation": "0.1.0",
```
DamianMullins: Might not be one for this PR but I think most of these dependencies can be moved into `devDependencies` as they will be included in the final bundle anyway (rather than required at run-time) so consuming applications don't need to download them as well.

DamianMullins: This is the case for multiple packages, I think I need to write up an example of what I mean here so it's clearer what changes need to be made in the future.

DamianMullins (dismissed):

packages/f-checkout/package.json
SachaZvetelman: @DamianMullins, if you could write up an example of that, it would be really useful. We'll keep this as it is for now, but I think I agree with you, so it's something we might want to do across the board in the future.

DamianMullins: Sounds good, I'll do that hopefully this week and share it with everyone.

xander-marjoram (dismissed):

FayeCarter (dismissed):

tommcclean (dismissed):

xander-marjoram (dismissed):

tommcclean (dismissed):

tommcclean (approved):

siggerzz (approved):

PR #426, opened 2020-11-10
generator-component@1.7.0

tommcclean:
This Pull Request updates the Generator Component to consume the new F-Globalisation service, which wires up the localisation of messages within components.

**Changes to the generator that affect new components...**
- A prop called `locale` is added to component stories; so out of the box you can change languages in storybook
- A dependency to F-Globalisation is added
- Existing copy mechanism is replaced by Vue-I18n capabilities like `$t`

<hr>

_This is what a component looks like now out of the box._
image

### DIFF
--- a/packages/generator-component/generators/app/templates/stories/Skeleton.__stories__.js
+++ b/packages/generator-component/generators/app/templates/stories/Skeleton.__stories__.js
@@ -2,8 +2,14 @@
 // import {
 //     withKnobs, select, boolean
 // } from '@storybook/addon-knobs';
+import { select, boolean } from '@storybook/addon-knobs';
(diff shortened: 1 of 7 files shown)


packages/generator-component/generators/app/templates/stories/Skeleton.__stories__.js
```diff
@@ -2,9 +2,15 @@
 // import {
 //     withKnobs, select, boolean
```
siggerzz, line 3: This can be removed.

ashleynolan:
Upgrade note – I've had to clean up some of our git history. If you could rebase master onto your branch and push back up to this PR, that'd be appreciated 🙌

PR #427, opened 2020-11-10
f-globalisation@0.2.0 - Export a wrap of vue-i18n

tommcclean:
> Although components don't need to initialise Vue-I18n, as this is the concern of the web host; our component tests still require the package as they have no web host to install it.

This pull request imports and exports vue-i18n, this means that tests within components using F-Globalisation do not need to install it themselves, and we will always have the same version in use across component tests.

### Added
- Provided a wrapper around the vue-i18n module for component tests to import

### DIFF
--- a/services/f-globalisation/package.json
+++ b/services/f-globalisation/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-globalisation",
   "description": "Fozzie Globalisation – A utility which wires up vue-i18n within your Smart Component",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "main": "dist/f-globalisation.umd.min.js",
   "files": [
     "dist"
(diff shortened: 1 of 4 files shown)


SachaZvetelman (dismissed):

SachaZvetelman (approved):

kevinrodrigues (approved):

ashleynolan:
Just a note – the Danger checks will keep raising that issue because they don't recognise the `/services` folder yet. We'll need to update the danger script to make sure that it's updated for this new folder structure (here's one of the [references for instance](https://github.com/justeat/fozzie-components/blob/master/dangerfile.js#L16))

services/f-globalisation/package.json
```diff
@@ -34,13 +34,15 @@
   "browserslist": [
     "extends @justeat/browserslist-config-fozzie"
   ],
+  "dependencies": {
+    "vue-i18n": "8.0.0"
```
ashleynolan, line 38: QQ – (very delayed question) – Does this mean we're bundling the `vue-i18n`module as part of this service? Or are we expecting that it would be tree-shaken out when consumed if it's not used

/ cc @SachaZvetelman

PR #428, opened 2020-11-11
f-error-message@0.3.0 - Update Icon to display inline with text

FayeCarter:
### Changed
- Icon from `Warning Icon` to `Danger Icon`.

### Added
- Styling to display icon inline with text for multiline messages.
---

## UI Review Checks

**f-error-message**
| **Multiline message** | **Single line message** |
| -- | -- |
| image | image |

**Designs**
| **Multiline message** | **Single line message** |
| -- | -- |
image | image

### DIFF
--- a/packages/f-error-message/src/components/ErrorMessage.vue
+++ b/packages/f-error-message/src/components/ErrorMessage.vue
@@ -52,7 +52,6 @@ export default {
 .c-errorMessage-icon {
     position: absolute;
     width: 16px;
-    line-height: 20px;
     margin-top: 2px;
 }
 </style>
--- a/packages/f-error-message/src/components/ErrorMessage.vue
+++ b/packages/f-error-message/src/components/ErrorMessage.vue
@@ -52,6 +52,7 @@ export default {
     width: 16px;
     line-height: 20px;
     margin-right: spacing();
+    margin-top: 2px;
     float: left;
 }
 
--- a/packages/f-error-message/src/components/ErrorMessage.vue
+++ b/packages/f-error-message/src/components/ErrorMessage.vue
@@ -36,8 +36,8 @@ export default {
 </script>
 
 <style lang="scss" module>
-
 .c-errorMessage {
+    position: relative;
     color: $color-text--danger;
     @include font-size(base);
     margin-top: spacing();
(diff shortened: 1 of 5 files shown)


SachaZvetelman:
Is it just me or the icon and the test are not vertically aligned?

FayeCarter:
@Sacha, I went for it. This alignment has been adjusted 😄

simonsandersje (dismissed):

packages/f-error-message/src/components/ErrorMessage.vue
```diff
@@ -36,17 +38,22 @@ export default {
...
+}
+
 .c-errorMessage-icon {
     width: 16px;
-    height: 16px;
-    margin-right: spacing(x0.5);
+    line-height: 20px;
+    margin-right: spacing();
+    margin-top: 2px;
+    float: left;
```
ashleynolan: A potentially better way of doing this would be to use absolute positioning of the icon (and making the `c-errorMessage` `position: relative`.

I'd probably try and avoid `float` as much as possible as they've become a legacy way of doing layout (because they can be a bit flakey cross-browser in terms of how they get rendered).

ashleynolan: Would also probably mean you could get rid of the line-height too, which shouldn't be needed if you use position absolute (if it's easier to chat about on hangout, give me a shout)

FayeCarter: Brilliant, thanks. I've updated this now. If it's not right yes please to the hangout 😄

ashleynolan (approved):

simonsandersje (approved):

packages/f-error-message/src/components/ErrorMessage.vue
```diff
@@ -4,20 +4,22 @@
         :class="$style['c-errorMessage']"
```
kevinrodrigues, line 4: Not a blocker for this PR. Would we be adding the accessibility aria attribute to this component anytime soon? Currently using this component in f-searchbox and have added it to the component inline (aria-live="assertive") but would be good for it to come from here in future.

kevinrodrigues (approved):

packages/f-error-message/src/components/ErrorMessage.vue
FayeCarter: I'll do this as a separate PR. 😄

PR #429, opened 2020-11-11
fozzie-components@2.13.0 - Spike for adding data-test-ids via separate object #globalconfig

LTurns:
We currently hardcode 'data-test-ids' into each component. There is no set standard here and it can lead to unnecessary extra code and a lot of duplication, such as unit tests using `data-js-test` and component tests using `data-test-id`, leading to multiple IDs, which has been pulled up in a few PRs.

As a spike, we have been looking into adding a data-test-id object for each component, which can then be shared and imported in both unit and component tests. This way, we can standardise the naming convention of these ids and won't have to worry about changing them in each file if a component changes. It could be useful for improving consistency in the repo. In future, we could add this to the generator. 

 Would love to know peoples thoughts!


BanksySan:
Usually I'm all for extracting constants out, but in this case each ID is owned by the element it's on so I'm not seeing the benefit in adding that extra hop when you want to find out what the ID is.
I can also see that `test-data-ids.js` file getting bloated with dead ids over time, as they don't get deleted.  Currently, I think I prefer it as it is, _however_ it did occur to me that extracting them could solve a problem that's been bugging me for years with regards to SPAs.  Because there's no scoping in a HTML page there's no way to check for duplicate IDs, if all our IDs (test and application) were in two separate files then we _could_ check for duplicates which could save us allot of time chasing down some ninja bug.

packages/f-alert/test-utils/component-objects/data-test-ids.js
```diff
@@ -0,0 +1,12 @@
+export default {
+    id: 'alert-component',
+    headings: {
+        title: 'heading-title'
```
zlatinivanov-je, line 4: I like the idea! Will it make sense to include an abbreviation to the component name in the values? Hoping it will make them easier to identify when CI jobs fail or so. Is there such use case at all or is this only used just with in the component tests? 
Also how would we avoid possible duplication in values? Is that a potential problem?

zlatinivanov-je (commented):
There is a suggestion for a version update missing

SachaZvetelman:
I agree with most of @BanksySan's comment 🙂 
Personally, I'd stick to keeping the IDs with the element, but instead of having two IDs (one for unit tests and one for UI tests), I'd just keep one ID.

ross-gratton:
To add to this discussion, I would say that the value to this approach is that we have a single source of truth for test ids in each component. This allows the component, the unit tests and the component tests to all use the same source of the string for the selectors.
Plus having this as an exportable const would allow it to be imported into consuming applications should tests within that consuming application require selectors to access the component elements.

ashleynolan:
I think the general idea could be useful, but I think my main concern would be as @BanksySan mentioned – when parts of the component get deleted, test-ids could hang around and not get removed, which could lead to that file storing outdated info that's not used. I'm not sure how you could try and stop this with automated checks too (would need us to script something pretty custom I'd think as part of our PR checks).

Could we not standardise around `data-test-id` (and get rid of `data-js-test`) or is it not as simple as that?

PR #430, opened 2020-11-12
Alert box using f-button.

BanksySan:
**The Problem**

Buttons aren't being styled correctly in Storybook.

**Acceptance Criteria**

Styling is correctly encapsulated in the f-button component such that the buttons all render as expected.

There are no extra imports of style definitions by Storybook itself to patch any gaps.

**The Solution**

Extend the prop so that the configuration can accommodate the requirements of the components using the native HTML5 button.

Replace the button tags with the VueButton component.

### DIFF
--- a/packages/f-alert/src/components/Alert.vue
+++ b/packages/f-alert/src/components/Alert.vue
@@ -18,10 +18,11 @@
                 data-test-id="alert-heading">
                 {{ heading }}
             </h2>
-            <button
+            <vue-button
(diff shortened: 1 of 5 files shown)


xander-marjoram (changes requested):
Your branch name and PR title should avoid using internal IDs, but rather versions of components.

Also it looks like this PR should be 3 different PRs - a minor version bump for f-alert, f-button and f-checkout

PR #439, opened 2020-11-13
f-searchbox@v4.0.0 beta.2

kevinrodrigues:
This PR brings in the baseline for UK functionality. It should cover everything apart from emitting events (Will be done in another PR). Vuex needs some work, I couldn't get it to work correctly so need to read up abit on this. 

### Added
- Baseline for UK logic (without Events at the moment).
- Tests to cover changes.
- Searchbox with/without shell comfigs.
- Vuex (Needs work).
- `shouldSetCookies`
- `onSubmit`
- `shouldAutoPopulateAddress`
- `isShellHidden`
- `shouldClearAddressOnValidSubmit`
- Allow previous postcode (saved in cookie) to proceed.

### Changed
- Extracted out form specific components into their own component directory & ported over styles into those directories.

### Video

Note: The search clear button in the video will be styled soon.

https://www.loom.com/share/0dac3b8b1cc941a8996e65ea69e30b49


packages/f-searchbox/src/assets/scss/form.scss
```diff
@@ -5,115 +5,58 @@ $search-button-bgColor: $color-primary;
 $search-button-bgColor--hover: $orange--dark;
 $search-button-bgColor--active: $orange--darkest;
 
-.c-search {
-    background-color: transparent;
-    display: inline-block;
-    position: relative;
-    width: 100%;
-    max-width: 610px;
+.c-search-placeholder {
```
ashleynolan, line 12: More a question as to whether it may be good to put a comment here – should we try and get design to look at the searchbox input to see if it can be normalised with the other input boxes they've designed?

We could then probably remove a lot of the styling from the searchbox component and pull that in eventually.

packages/f-searchbox/src/assets/scss/form.scss
```diff
@@ -5,115 +5,58 @@ $search-button-bgColor: $color-primary;
...
-    &:hover {
-        background-color: $search-button-bgColor--hover;
-    }
-
-    &:active {
-        background-color: $search-button-bgColor--active;
-    }
+    border-radius: $border-radius 0 0 $border-radius;
+    @include font-size(body-s, false);
+    font-weight: $font-weight-base;
```
ashleynolan, line 51: Could use `@include font-size()` for this? The line-height should be the same then that comes out from that function.

packages/f-searchbox/src/assets/scss/form.scss
```diff
@@ -5,115 +5,58 @@ $search-button-bgColor: $color-primary;
...
+    line-height: 1.3;
+    position: relative;
```
ashleynolan: Again, more of a preference thing, but I tend to put `position` as the first thing defined under a selector as it impacts the layout directly usually.

packages/f-searchbox/src/assets/scss/form.scss
```diff
@@ -133,98 +76,37 @@ $search-button-bgColor--active: $orange--darkest;
...
-    // Add padding to input when new elements overlap field
+  // Add padding to input when new elements overlap field
     .has-streetNumber & {
-       padding-right: 60px;
+      padding-right: 60px;
     }
 
     @include media('>=narrow') {
-        height: 60px;
+      height: 60px;
```
ashleynolan: Can see `56px` and `60px` defined a couple of times (in the label and input) as heights – could you pull them out as variables so that they can be changed in one place?

packages/f-searchbox/src/assets/scss/form.scss
```diff
@@ -133,98 +76,37 @@ $search-button-bgColor--active: $orange--darkest;
...
     }
 
-    &:focus + .c-search-placeholder,
-    &.is-notEmpty + .c-search-placeholder {
-        @include font-size(caption);
-        transform: translateY(0);
+      &:focus + .c-search-placeholder,
```
ashleynolan: Not sure you need the extra spaces here?

packages/f-searchbox/src/assets/scss/form.scss
```diff
@@ -133,98 +76,37 @@ $search-button-bgColor--active: $orange--darkest;
...
+              transform: translateY(-110%);
+          }
+      }
 }
 
-.c-search-label {
-    border: 1px solid $grey--light;
-    display: flex;
-    width: 100%;
+.c-search-innerFields {
```
ashleynolan, line 108: Might be good to add a comment as to what sort of elements this applies to. Without looking at the markup, `innerFields` could be a lot of different things.

packages/f-searchbox/src/components/Form.vue
```diff
@@ -75,58 +64,167 @@ export default {
...
+            cuisine,
+            query,
+            formUrl,
+            queryString,
+            isCompressed,
+            clearAddressOnValidSubmit,
+            addressField,
+            setCookies,
+            onSubmit,
+            autoPopulateAddress
```
ashleynolan: Super small naming thing – should make this clearly `Boolean` in naming. So `shouldAutoPopulateAddress` or `isAddressAutoPopulated` or something along those lines?

packages/f-searchbox/src/components/Form.vue
```diff
@@ -75,58 +64,167 @@ export default {
...
     },
 
     data () {
         const {
-            address = '',
-            cuisine = '',
-            query = '',
-            queryString = '',
-            isCompressed = false
+            address,
```
ashleynolan, line 76: May be worth adding some comments to any of these that aren't immediately obvious (unless they're documented elsewhere)

packages/f-searchbox/src/components/Form.vue
```diff
@@ -75,58 +64,167 @@ export default {
...
+            cuisine,
+            query,
+            formUrl,
+            queryString,
+            isCompressed,
+            clearAddressOnValidSubmit,
```
ashleynolan: Same with this one – should be obvious from name it's a `Boolean` (so using `should` or `will` or something like that)

packages/f-searchbox/src/components/Form.vue
```diff
@@ -75,58 +64,167 @@ export default {
...
+            addressField,
+            setCookies,
```
ashleynolan: Same with this one (Boolean naming) – so `shouldSetCookies`?

packages/f-searchbox/src/components/Form.vue
```diff
@@ -75,58 +64,167 @@ export default {
...
+        hasLastSavedAddress () {
+            return this.lastAddress
+                    && this.address === this.lastAddress
+                    && this.store.state.isValid === true;
+        }
+    },
+
+    methods: {
+        /**
+         * Main submit handler that's responsible for submitting searches to SERP:
```
ashleynolan: Would change this to `Main submit handler that's responsible for submitting searches: for example to SERP` (as searchbox is used elsewhere, like on Menu I think?)

packages/f-searchbox/src/components/Form.vue
```diff
@@ -75,58 +64,167 @@ export default {
...
+        clearAddressValue (shouldClear) {
+            if (shouldClear) {
+                this.address = '';
+                this.store.commit('SET_IS_DIRTY', false);
+            }
+        }
+    },
+
+    mounted () {
+        this.lastAddress = this.config.locationFormat(getLastLocation());
```
ashleynolan, line 197: Probably useful to comment next to this that this is getting the `lastAddress` that has been stored for a user

packages/f-searchbox/src/components/formElements/FormLabelWrapper.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+    <div
+        :class="{
```
ashleynolan: Think the name of this component is a bit confusing – I assumed it was for a form label, but it's for a label with an input. If it's just for the main `FormSearchField`, I'd probably just call it that?

packages/f-searchbox/src/components/formElements/FormSearchButton.vue
```diff
@@ -0,0 +1,135 @@
...
+    <div>
+        <button
+            type="submit"
+            data-test-id="find-restaurants-button"
+            :aria-label="copy.buttonText"
+            :class="{
+                [$style['c-search-btn']]: true,
+                [$style['c-search-btn--compressed']]: isCompressed
+            }">
+            <div :class="$style['c-search-btn-icon']">
```
ashleynolan: Makes more sense as a `span` I think, as that will be `inline` rather than `block`

packages/f-searchbox/src/components/formElements/FormSearchButton.vue
```diff
@@ -0,0 +1,135 @@
...
+            default:  () => ({})
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+@import '../../assets/scss/form';
+
+.c-search-btn {
```
ashleynolan, line 45: Same as the input – would be good to at some point get styles from the `f-button` component here. Not a priority, but happy to hep work this out in the future as to the best way to do this.

packages/f-searchbox/src/components/tests/Base.test.js
```diff
@@ -0,0 +1,73 @@
...
+    it('should be defined', () => {
+        // Arrange
+        const propsData = {};
+        const wrapper = shallowMount(Base, { propsData });
+
+        // Act & Assert
+        expect(wrapper.exists()).toBe(true);
+    });
+
+    describe('Base `shell` configs', () => {
```
ashleynolan: Don't think you need Base here again, as tests will read `Base Base  "shell" configs`?

packages/f-searchbox/src/components/tests/Base.test.js
```diff
@@ -0,0 +1,73 @@
...
+        it('should set the correct `shell` element to `search-shell` when `hideShell` is not set', () => {
```
ashleynolan: Not sure how much of a pain it would be, but naming this to be obviously Boolean would be better = `isShellHidden`

packages/f-searchbox/src/components/tests/Form.test.js
```diff
@@ -0,0 +1,376 @@
...
+                        // Act
+                        wrapper.vm.submit(event);
+
+                        // Assert
+                        expect(spy).toHaveBeenCalledWith(true);
+                    });
+                });
+
+                describe('when `address` `isValid` is falsy', () => {
+                    it('should prevent default to stop the form from submitting a invalid search', () => {
```
ashleynolan: Super NIT – `an invalid search`

packages/f-searchbox/src/utils/helpers.js
```diff
@@ -18,7 +18,47 @@ const doesPostcodeMatchRegex = postcode => {
...
+    let expires = '';
+    if (days) {
+        const date = new Date();
+        date.setTime(date.getTime() + (days * millisecondsPerDay));
+        expires = `; expires=${date.toUTCString()}`;
+    }
+    document.cookie = `${name}=${(value || '') + expires}; path=/`;
+};
+
+const getLastLocation = () => window.document.cookie
```
ashleynolan, line 55: JSDoc comment please

packages/f-searchbox/src/utils/helpers.js
```diff
@@ -18,7 +18,47 @@ const doesPostcodeMatchRegex = postcode => {
     return postcodeRegex.test(postcode);
 };
 
+const normalisePostcode = postcode => (postcode ? postcode.toUpperCase().replace(/\s/g, '').trim() : '');
```
ashleynolan, line 25: JSDoc comment

ashleynolan (commented):
Nothing major – mostly naming related things, otherwise LGTM

packages/f-searchbox/src/assets/scss/form.scss
kevinrodrigues: Yup there's lots of styles in here and probably in the components themselves that aren't necessarily searchbox specific. I'll add a comment.

packages/f-searchbox/src/assets/scss/form.scss
kevinrodrigues: Won't that change the size of the font from the line above tho?

packages/f-searchbox/src/components/Form.vue
kevinrodrigues: Something I still need to do as part of this [ticket](https://justeattakeaway.atlassian.net/browse/LS-980), but the readme will define all of the `config` names here. I may update the read me after the google places work which will happen next :)

packages/f-searchbox/src/components/Form.vue
kevinrodrigues: Happy with changing this and the ones below just conscious that some consuming apps are probably already using these names already. I think these three are used by MenuWeb. cc @DamianMullins FYI - I'll update the read me with these new values so you know.

packages/f-searchbox/src/components/formElements/FormSearchButton.vue
kevinrodrigues: I did actually try to pull in the f-button component but struggled to get the styles to work as i wanted. I'll grab you in the next PR if that's ok.

packages/f-searchbox/src/components/tests/Base.test.js
kevinrodrigues: Happy to change this one too just down to where it's being used. AFAIK menu probably would be the one place the shell is hidden I think.

packages/f-searchbox/src/components/Form.vue
kevinrodrigues: Similar to above.

packages/f-searchbox/src/components/Form.vue
kevinrodrigues: Similar to above.

packages/f-searchbox/src/components/tests/Base.test.js
kevinrodrigues: Fixed 👍

packages/f-searchbox/src/components/Form.vue
kevinrodrigues: Have fixed all the names now 👍

packages/f-searchbox/src/components/Form.vue
kevinrodrigues: Fixed.

packages/f-searchbox/src/components/Form.vue
kevinrodrigues: Fixed.

packages/f-searchbox/package.json
```diff
@@ -42,19 +42,24 @@
...
   ],
   "dependencies": {
+    "@justeat/f-error-message": "0.3.1",
     "@justeat/f-services": "1.0.0",
     "@justeat/f-vue-icons": "1.8.0",
     "axios": "0.19.2"
   },
   "peerDependencies": {
-    "@justeat/browserslist-config-fozzie": ">=1.1.1"
+    "@justeat/browserslist-config-fozzie": ">=1.1.1",
```
DamianMullins: I think this can go into `devDependencies` as it's only used at build time?

packages/f-searchbox/package.json
```diff
@@ -42,19 +42,24 @@
...
+    "vuex": "3.5.1",
```
DamianMullins, line 51: If you don't want thid to be included in the final bundle you'll need to exclude it in the `vue.config` file [like this](https://github.com/justeat/fozzie-components/blob/29c158bda76866bf7ade5d05cad447ab2da94b9e/packages/f-mega-modal/vue.config.js#L17-L19).

kevinrodrigues: Yeah I did actually try this and got an error:

// Added to the `chainWebpack` method...
`config.externals({
            'vuex': 'vuex'
        });`

Error:

`Uncaught ReferenceError: vuex is not defined`

packages/f-searchbox/package.json
kevinrodrigues: Yeah i think so, this came from the generator I believe.

kevinrodrigues: Fixed.

packages/f-searchbox/package.json
DamianMullins: Is that when you're linking the project from another? Or when you run `f-searchbox` in its own?

kevinrodrigues: On its own which is the issue. :\

DamianMullins: Yeah, so it'd need the instance passed from somewhere. This is the issue we were discussing the other day isn't it?

On any case, I'm happy for this to remain as it is at the moment as it's not a straight-forward fix.

packages/f-searchbox/src/utils/helpers.js
```diff
@@ -18,7 +18,59 @@ const doesPostcodeMatchRegex = postcode => {
...
+
+/**
+ * Returns a location object that contains previous searches from stored cookies.
+ * Example return if a `je-location` cookie existed (from en-GB):
+ *
+ * {postcode: "AR511AR"}
+ *
+ * @returns Object
+ */
+const getLastLocation = () => window.document.cookie
```
DamianMullins, line 55: One for another time maybe, we might want to extract this into `f-services` as I've seen it used in a few places before? 

It has actually been [extracted into `f-utilities`](https://github.com/justeat/f-utilities/blob/master/src/js/modules/locationCookies/index.js) but not sure that's the best place for it.

packages/f-searchbox/src/store/searchbox.module.js
```diff
@@ -0,0 +1,27 @@
+import Vue from 'vue';
+import Vuex from 'vuex';
+
+Vue.use(Vuex);
+
+export default new Vuex.Store({
+    namespaced: true,
```
DamianMullins: You don't need to set this if you aren't using Vuex modules.

packages/f-searchbox/package.json
kevinrodrigues: Yup it would need it passed in so the standalone won't work without that hooked up. I'm still in two minds about using Vuex so it might end up getting removed if it's not really adding any real benefit, for the work i've done in this pr it wasn't really required that much too.

packages/f-searchbox/src/services/core/factory.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Callback service, empty for now, will be expanded upon
+ * when we add the other services e.g. google places in the next few prs.
+ *
+ * @param Service
+ */
+export default Service => {
+    Service();
```
DamianMullins, line 8: Is the plan to do something more with this function in the future? At the moment it looks like you can just call the function in-place without having this extra abstraction?

packages/f-searchbox/src/utils/helpers.js
kevinrodrigues: Yeah, I spoke to the web team about a few of these cookie helpers. There's a ticket (looks to be in progress) for creating a cookie service in `f-services`. I was planning on removing these as soon as that's released 👍

packages/f-searchbox/src/services/core/factory.js
kevinrodrigues: This will be expanded to handle the google places stuff too (in the next PR) 👍

packages/f-searchbox/src/components/Form.vue
```diff
@@ -75,58 +64,164 @@ export default {
...
+         * Main submit handler that's responsible for submitting searches: for example to SERP.
+         *
+         * 1. Checks address validation.
+         * 2. If the address is valid we submit the address.
+         * 3. Note: `configs` determine custom behaviour, e.g shouldSetCookies & shouldClearAddressOnValidSubmit etc see readme.
+         *
+         * @param e
+         */
+        async submit (e) {
+            this.store.commit('SET_IS_VALID', this.service.isValid(this.address));
```
SachaZvetelman, line 143: I'm thinking we'd probably want to have the component call an action, which inside might have some business logic (or not) and will call the necessary mutations.
Similarly, actions can be `async`. This function (`async submit`) is asynchronous, but there is no `await` inside.

I guess if we are not gonna have business logic or asynchronous logic, then we don't need actions.

kevinrodrigues: Yup that's what I want to do but i couldn't get the Vuex stuff to work correctly (Added a note in the PR description). The `await` can be removed but the next PR will be adding it again. :)

packages/f-searchbox/src/store/searchbox.module.js
kevinrodrigues: Fixed.

packages/f-searchbox/src/components/Form.vue
kevinrodrigues: Removed the `async` 👍

DamianMullins (dismissed):

packages/f-searchbox/src/assets/scss/form.scss
ashleynolan: Ah – ignore me. I think I misread the line as font-size when it's font-weight!

packages/f-searchbox/src/components/formElements/FormSearchButton.vue
kevinrodrigues: Managed to figure out why the f-button element wasn't working now, I'll add this change to my next PR, and yes would be good to try and remove most of these styles so it comes from that component instead 👍

packages/f-searchbox/src/components/Form.vue
```diff
@@ -1,70 +1,59 @@
...
+            {{ errorMessage }}
+        </error-message>
     </form>
 </template>
 
 <script>
-import { EyeglassIcon } from '@justeat/f-vue-icons';
+import ErrorMessage from '@justeat/f-error-message';
+import '@justeat/f-error-message/dist/f-error-message.css';
+import FormLabelWrapper from './formElements/FormSearchField.vue';
```
ashleynolan: Should update this to be `FormSearchField`

ashleynolan (approved):

kevinrodrigues:
@DamianMullins would you mind re-ticking this. Pushed up a small change.

DamianMullins (approved):

PR #443, opened 2020-11-16
f-checkout@0.9.0 Wire up the form with response from get checkout

pkedziora:
v0.9.0
*November 16, 2020*

### Added
- Checkout component calls server to get data

## UI Review Checks

Screenshot 2020-11-16 at 14 21 41
Screenshot 2020-11-16 at 14 21 51

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/f-checkout/package.json
```diff
@@ -52,7 +52,8 @@
     "@justeat/f-globalisation": "0.2.0",
     "@justeat/f-services": "1.4.0",
     "axios": "0.20.0",
-    "vuelidate": "0.7.6"
+    "vuelidate": "0.7.6",
+     "axios-mock-adapter": "1.19.0"
```
SachaZvetelman: Nit: indentation.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -118,7 +119,7 @@ export default {
     props: {
         checkoutMethod: {
             type: String,
-            default: 'Collection',
+            default: 'collection',
             validator: value => (VALID_CHECKOUT_METHOD.indexOf(value) !== -1)
         },
```
SachaZvetelman: We should remove this prop, actually. Happy to do it in a different ticket, if you think it's better, though.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -131,13 +132,19 @@ export default {
             type: Number,
             required: false,
             default: 1000
+        },
+
+        getCheckoutTimeout: {
+            type: Number,
+            required: false,
+            default: 1000
         }
```
SachaZvetelman, line 135: What's the advantage of passing the checkout timeout as a prop?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -147,7 +154,15 @@ export default {
...
             shouldDisableCheckoutButton: false,
-            delivery: CHECKOUT_METHOD_DELIVERY
+            delivery: CHECKOUT_METHOD_DELIVERY,
+            checkoutId: '',
+            isFulfillable: true,
+            times: [],
+            messages: [],
+            notes: [],
+            notices: [],
+            buttonText: 'Go to payment',
```
SachaZvetelman: This should be removed. It's no longer used. It's coming from the tenant files.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -147,7 +154,15 @@ export default {
             },
             genericErrorMessage: null,
             shouldDisableCheckoutButton: false,
-            delivery: CHECKOUT_METHOD_DELIVERY
+            delivery: CHECKOUT_METHOD_DELIVERY,
```
SachaZvetelman: We should get rid of this as well, if we have `serviceType`.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -261,6 +294,22 @@ export default {
             */
             this.$v.$touch();
             return !this.$v.$invalid;
+        },
+
+        mapResponse (data) {
+            this.checkoutId = data.id;
+            this.isFulfillable = data.isFulfillable;
+            this.firstName = data.customer.firstName;
```
SachaZvetelman: Do we always get a `customer`? If there are scenarios in which we don't, this will fail.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -261,6 +294,22 @@ export default {
...
+            this.mobileNumber = data.customer.phoneNumber;
+            this.times = data.fulfillment.times;
```
SachaZvetelman, line 305: Do we always get a `fulfillment`? If there are scenarios in which we don't, this will fail.

packages/f-checkout/src/components/Selector.vue
```diff
@@ -14,12 +14,13 @@
             id="time-selection"
             v-model="selectedTime"
             :class="$style['o-form-select-input']"
-            data-test-id="delivery-time">
+            data-test-id="delivery-time"
```
SachaZvetelman: This will be `collection-time` when it's a collection order. Should we update this in this PR even though you haven't modified this line?

packages/f-checkout/src/components/Selector.vue
```diff
@@ -43,6 +44,25 @@ export default {
...
+
+    methods: {
+        selectionChanged () {
+            this.times.forEach(el => { el.selected = false; });
+            const newSelected = this.times.find(t => t.from === this.selectedTime);
+            if (newSelected) {
+                newSelected.selected = true;
+            }
+        }
     }
```
SachaZvetelman, line 66: Could you please clarify why we need a `watch` here? 🙂

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -55,10 +56,10 @@ describe('Checkout', () => {
                 expect(selectorComponent.attributes('ordermethod')).toEqual(definedType);
             });
 
-            it('should display the address block if set to `Delivery`', () => {
+            it('should display the address block if set to `delivery`', () => {
                 // Arrange
                 const propsData = {
-                    checkoutMethod: 'Delivery',
+                    checkoutMethod: 'delivery',
```
SachaZvetelman: As I mentioned above, we'll remove this prop, so we might have to update these tests. Again, if you think it should be better done in a separate PR, I understand.

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -151,9 +152,25 @@ describe('Checkout', () => {
             postcode: 'EE1E 1EE'
         };
 
-        describe('if checkoutMethod set to `Collection`', () => {
+        function CheckoutGetData (serviceType) {
```
SachaZvetelman: This could be a `let` at the top of the file which is updated in a `beforeEach`.
You can then do something like:
```
let defaultCheckoutGetData = ...;
...
... // beforeEach and other magic.
...
const checkoutGetData = { ...defaultCheckoutGetData, serviceType };
```

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -162,6 +179,8 @@ describe('Checkout', () => {
             beforeEach(() => {
                 CheckoutServiceApi.submitCheckout.mockClear();
                 CheckoutServiceApi.submitCheckout.mockImplementation(async () => Promise.resolve());
+                CheckoutServiceApi.getCheckout.mockClear();
```
SachaZvetelman: You can move this to an `afterEach` so the `beforeEach` ends up being cleaner.

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -346,4 +367,65 @@ describe('Checkout', () => {
             });
         });
     });
+
+    describe('when form is loaded', () => {
+        function CheckoutGetData () {
```
SachaZvetelman: Not sure why we need to redefine this function here? I think it adds confusion since this function had already been defined above (but I suggested removing it actually, hehe)

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -346,4 +367,65 @@ describe('Checkout', () => {
...
+
+        const propsData = {
+            checkoutMethod: CHECKOUT_METHOD_DELIVERY,
+            checkoutUrl
+        };
+
+        describe('when request fails', () => {
+            let wrapper;
+
+            beforeEach(() => {
```
SachaZvetelman, line 399: I think we normally create a `beforeEach` when we have things that are reused in multiple tests. Otherwise, we just include it in the tests, since it's easier to read.

packages/f-checkout/stories/Checkout.stories.js
```diff
@@ -29,7 +32,7 @@ export const CheckoutComponent = () => ({
         }
     },
     template:
-        '<vue-checkout :checkoutMethod="checkoutMethod" :locale="locale" />'
+        '<vue-checkout :checkoutMethod="checkoutMethod" checkoutUrl="/checkout.json" :locale="locale" />'
```
SachaZvetelman: I'm not entirely sure what's the benefit of passing the `checkoutUrl` as a prop, and similarly, we probably need to ensure this works with non-relative paths (i.e. when we deploy this to other environments).

SachaZvetelman: One thing we should probably do is to provide a couple of different json files. One for delivery, one for collection. So then we provide a dropdown in storybook to choose between the two files?

SachaZvetelman (changes requested):

packages/f-checkout/stories/Checkout.stories.js
pkedziora: @SachaZvetelman That's what registration component does and it makes sense because how else would you pass this to a component if it's hosted outside of fozzie-components.
Having dropdown with file selection makes sense

packages/f-checkout/src/components/Checkout.vue
pkedziora: So that the calling page can configure the timeout

packages/f-checkout/src/components/Selector.vue
pkedziora: Watch is needed to update the selected item in the drop down when the GET request comes back

packages/f-checkout/package.json
pkedziora: Fixed

packages/f-checkout/src/components/Checkout.vue
pkedziora: Removed

packages/f-checkout/src/components/Checkout.vue
pkedziora: Removed

packages/f-checkout/src/components/Checkout.vue

packages/f-checkout/src/components/Checkout.vue
pkedziora: I think we can assume that fulfillment will always be there

packages/f-checkout/src/components/Selector.vue
pkedziora: Changed to fulfillment-time

packages/f-checkout/src/components/tests/Checkout.test.js
pkedziora: Fixed

packages/f-checkout/src/components/tests/Checkout.test.js
pkedziora: Fixed

packages/f-checkout/src/components/tests/Checkout.test.js
pkedziora: Removed

packages/f-checkout/src/components/tests/Checkout.test.js
pkedziora: I'm inclined to leave it as it's very likely more tests will be added here as the codebase grows

packages/f-checkout/src/components/Checkout.vue
pkedziora: Removed

packages/f-checkout/src/components/tests/Checkout.test.js
pkedziora: Removed

packages/f-checkout/src/demo/demo.js
```diff
@@ -18,7 +21,8 @@ new Vue({
     i18n,
     render: h => h(VueCheckout, {
         props: {
-            checkoutMethod: 'Delivery'
+            checkoutMethod: 'delivery',
```
SachaZvetelman: We can remove this now, right?

packages/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -33,62 +48,80 @@ describe('Checkout', () => {
...
+
                 // Act
                 const wrapper = shallowMount(VueCheckout, {
                     i18n,
                     localVue,
                     propsData
                 });
 
+                wrapper.setData(data);
+                await wrapper.vm.$nextTick();
```
SachaZvetelman, line 97: Do we need `$nextTick` in these tests?

pkedziora: yes, tests fail without it as the ui doesn't get a chance to update after data change

packages/f-checkout/src/demo/demo.js
pkedziora: Yep, well spotted. Removed

SachaZvetelman (approved):

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -131,13 +134,19 @@ export default {
...
+        getCheckoutTimeout: {
+            type: Number,
+            required: false,
+            default: 1000
         }
     },
 
     data () {
         return {
             tenantConfigs,
```
kevinrodrigues, line 140: Does `tenantConfigs` actually get used somewhere?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -94,7 +95,9 @@ import AddressBlock from './Address.vue';
 import FormSelector from './Selector.vue';
 import UserNote from './UserNote.vue';
 
-import { VALID_CHECKOUT_METHOD, CHECKOUT_METHOD_DELIVERY } from '../constants';
+import {
+    VALID_CHECKOUT_METHOD, CHECKOUT_METHOD_DELIVERY, CHECKOUT_METHOD_COLLECTION, TENANT_MAP
```
kevinrodrigues: Could you break these onto separate lines please.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -210,6 +239,16 @@ export default {
             this.$emit(EventNames.CheckoutSuccess);
         },
 
+        async getCheckout () {
+            try {
+                const result = await CheckoutServiceApi.getCheckout(this.checkoutUrl, this.tenant, this.getCheckoutTimeout);
+                this.$emit(EventNames.CheckoutGetSuccess);
```
kevinrodrigues, line 237: I assume this is listened to by the consuming app? Does the checkout component pass these listeners up? (Couldn't see it anywhere)

packages/f-checkout/src/components/Checkout.vue
tommcclean: It is used by the globalisation mixin, this is how the tenant configs get from a component and wired up to vue-i18n

https://www.npmjs.com/package/@justeat/f-globalisation#usage

packages/f-checkout/src/components/Checkout.vue
pkedziora: @kevinrodrigues this is outdated diff as this change is no longer there

packages/f-checkout/src/components/Checkout.vue
pkedziora: @kevinrodrigues calling component would be able to bind to those events. I assume the calling component can just bind to those with @ or v-on

packages/f-checkout/src/components/Checkout.vue
kevinrodrigues: Ah my bad!

packages/f-checkout/src/components/Checkout.vue
kevinrodrigues: Yeah should be able to provided that the component is passing those emits up when it's being used. Normally the base component or the component that's going to be used in the consuming app would need to set `v-on="$listeners"` so that the `@` can be used. This is how i've done it before but it might work differently in how you guys are building checkout so wanted to check.

kevinrodrigues (approved):
LGTM - only that emit question, could be looked at separately tho.

PR #446, opened 2020-11-17
fozzie-components@v2.17.0 - Consolidate data-test-ids and data-js-test attributes #globalconfig

LTurns:
### Changed

- Consolidated test ids - replaced `data-js-test` attribute with `data-test-id` where possible, to be shared by unit and component tests. The only test ids in the repo should now be either 'data-test-id' or 'data-theme-{component-name}', where required.


packages/f-header/test-utils/component-objects/f-header.component.js
```diff
@@ -9,9 +9,11 @@ exports.clickLoginLink = () => {
 }
 
 exports.clickOffersLink = () => {
-    offersLink().click();
+    offersLinkMobile().click();
 }
 
+//do one for offerslinkdesktop too
```
xander-marjoram: @LTurns Do we have a ticket for this or is this still a WIP?

packages/f-registration/test-utils/component-objects/f-registration.component.js
```diff
@@ -10,19 +10,19 @@ const privacyPolicyLink = () => registrationComponent().$('[data-test-id="privac
 const cookiesPolicyLink = () => registrationComponent().$('[data-test-id="cookies-policy-link"]');
 
 // Validation errors
-const firstNameEmptyError = () => $('[data-test-title="error-first-name-empty"]');
-const firstNameMaxLengthError = () => $('[data-test-title="error-first-name-maxlength"]');
+const firstNameEmptyError = () => $('[data-test-id="error-first-name-empty"]');
+const firstNameMaxLengthError = () => $('[data-test-id="error-first-name-maxlength"]');
 const firstNameInvalidError = () => $('[data-test-title="error-first-name-invalid"]');
```
xander-marjoram: I think you missed this one 😄

packages/f-header/test-utils/component-objects/f-header.component.js
LTurns: @xander-marjoram think I'll quickly do that now actually - should only be one small test. Thanks for reminding me!

packages/f-registration/test-utils/component-objects/f-registration.component.js
LTurns: whoops, thank youu!

ashleynolan:
I've added `#globalconfig` to your title so that it stops asking for CHANGELOG entries at package level

LTurns:
> I've added `#globalconfig` to your title so that it stops asking for CHANGELOG entries at package level

thanks Ash, will remember to do that next time! :)

packages/f-header/test-utils/component-objects/f-header.component.js
```diff
@@ -1,5 +1,6 @@
 const loginLink = () => $('[data-test-id="login-link"]');
-const offersLink = () => $('[data-test-id="offers-link]"');
+const offersLinkMobile = () => $('[data-test-id="offers-link-mobile]"');
+const offersLinkDesktop = () => $('[data-test-id="offers-link-desktop]"');
```
siggerzz: Is there a way we can consolidate this into a single selector that can be used for both mobile and desktop?

siggerzz (changes requested):
See comments!

packages/f-header/test-utils/component-objects/f-header.component.js
xander-marjoram: Could we have two data-test-ids on each link, one "offers-link-mobile/desktop" for the mobile/desktop tests and one "offers-link" for generic use?

minrwhite: Wouldn't the attribute starts-with selector `$('[data-test-id^="offers-link"]')` suffice for [the case where we need to match either]? Also the closing quotes and square-parens are in the wrong order above: 
```suggestion
const offersLinkDesktop = () => $('[data-test-id="offers-link-desktop"]');
```

packages/f-header/test-utils/component-objects/f-header.component.js
```diff
@@ -1,5 +1,6 @@
 const loginLink = () => $('[data-test-id="login-link"]');
-const offersLink = () => $('[data-test-id="offers-link]"');
+const offersLinkMobile = () => $('[data-test-id="offers-link-mobile]"');
```
minrwhite: This line also needs that change:
```suggestion
const offersLinkMobile = () => $('[data-test-id="offers-link-mobile"]');
```

minrwhite (approved):

siggerzz (approved):

PR #449, opened 2020-11-18
f-form-field@1.2.0 - Add dropdown module

FayeCarter:
### Added
- New `dropdown` module.

## UI Review Checks

image


packages/f-form-field/src/components/FormDropdown.vue
```diff
@@ -0,0 +1,71 @@
...
+        updateOption (event) {
+            this.$emit('update', event.target.value);
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+.c-formDropdown {
+    .c-formDropdown-icon {
```
minrwhite: It's better to not nest these, as we know it's under a `c-formDropdown` element because of the class naming scheme.

packages/f-form-field/src/components/FormDropdown.vue
```diff
@@ -0,0 +1,71 @@
...
+        width: 12px;
```
minrwhite: Wouldn't
```suggestion
        width: spacing(x1.5);
```
be better, or is there something I'm missing?

packages/f-form-field/src/components/FormDropdown.vue
```diff
@@ -0,0 +1,71 @@
...
+    methods: {
+        updateOption (event) {
+            this.$emit('update', event.target.value);
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+.c-formDropdown {
```
ashleynolan: If you aren't currently using the `c-formDropdown` class for anything atm, I'd just remove this nesting (as it'll just make your selectors more specific).

packages/f-form-field/src/components/FormField.vue
```diff
@@ -207,10 +240,13 @@ $form-input-fontSize                      : 'body-l';
         position: relative;
     }
 
+    .c-formField-input-padding {
+        padding: $form-input-padding; //convert padding to rem
```
ashleynolan: Can remove this comment – looks like it should've been deleted but has been left in

minrwhite: I'm pretty sure the comment doesn't have any relevance any more, please can you remove this? And also the comment from line 249 below.

packages/f-form-field/stories/FormField.stories.js
```diff
@@ -27,13 +27,17 @@ export const FormFieldComponent = () => ({
         },
         hasError: {
             default: boolean('hasError', false)
+        },
+        dropdownOptions: {
+            default: array('Dropdown Options', ['As soon as possible', 'Today in 5 minutes'], ',')
         }
+
```
minrwhite:
```suggestion
```

packages/f-form-field/src/components/FormField.vue
```diff
@@ -207,10 +240,13 @@ $form-input-fontSize                      : 'body-l';
         position: relative;
     }
 
+    .c-formField-input-padding {
```
ashleynolan: Looks like you've split this out so that it only applies to the text inputs?

I'd avoid using the word `padding`, incase in the future we ever need to add other styles to those class, so something like `c-formField-input--textField` might make sense?

packages/f-form-field/src/components/tests/FormDropdown.test.js
```diff
@@ -0,0 +1,46 @@
+import { shallowMount } from '@vue/test-utils';
+import FormDropdown from '../FormDropdown.vue';
+
+describe('FormDropdown', () => {
+    const dropdownOptions = ['test option'];
```
ashleynolan: I'd probably have a bigger array of options here, so it's more representative of a normal option list

packages/f-form-field/src/components/tests/FormDropdown.test.js
```diff
@@ -0,0 +1,46 @@
...
+    const propsData = { dropdownOptions };
+
+    it('should be defined', () => {
+        const wrapper = shallowMount(FormDropdown, { propsData });
+        expect(wrapper.exists()).toBe(true);
+    });
+
+    describe('props ::', () => {
+        describe('dropdownOptions ::', () => {
+            it('should populate `<option>` tags', () => {
```
ashleynolan, line 15: Worth having a test for what happens if no options are passed in?

packages/f-form-field/src/components/tests/FormField.test.js
```diff
@@ -54,19 +55,34 @@ describe('FormField', () => {
                 expect(formInput.attributes('type')).toBe(DEFAULT_INPUT_TYPE);
             });
 
-            it.each(VALID_INPUT_TYPES)('should set the type of form input element as expected if inputType=%p is specified', definedType => {
```
ashleynolan: I would probably split this constant out into two – keep `VALID_INPUT_TYPES` and then make a constant for `CUSTOM_INPUT_TYPES` and then change the prop validation to check against both of those arrays. Then you can use this value in the tests still (rather than duplicating the strings)

packages/f-form-field/src/components/tests/FormField.test.js
```diff
@@ -54,19 +55,34 @@ describe('FormField', () => {
...
+                'text', 'email', 'password', 'radio', 'checkbox'
+            ])('should set the type of form input element as expected if inputType=%p is specified', definedType => {
                 // Arrange
                 const propsData = {
                     inputType: definedType
                 };
 
                 // Act
-                const wrapper = shallowMount(FormField, { propsData });
+                const wrapper = mount(FormField, { propsData });
```
ashleynolan: What was the reason for needing to change this? (more out of interest)

ashleynolan (commented):
Few comments – nothing major though

packages/f-form-field/src/components/FormField.vue
FayeCarter: Removed this now 🙂

packages/f-form-field/src/components/FormField.vue
FayeCarter: Updated thanks

packages/f-form-field/src/components/tests/FormField.test.js
FayeCarter: @ashleynolan I've tried this. Is the way I've done the prop validation alright? https://github.com/justeat/fozzie-components/pull/449/files#diff-1e64645db2045486e265ff004518a181cb42a188fd12e744120faafb223af20fR102

packages/f-form-field/src/components/tests/FormField.test.js
FayeCarter: An experiment I forgot to take out. Thanks!

packages/f-form-field/src/components/FormField.vue
```diff
@@ -79,7 +99,7 @@ export default {
         inputType: {
             type: String,
             default: DEFAULT_INPUT_TYPE,
-            validator: value => (VALID_INPUT_TYPES.indexOf(value) !== -1) // The prop value must match one of the valid input types
+            validator: value => (VALID_INPUT_TYPES.includes(value) || (CUSTOM_INPUT_TYPES.includes(value)))// The prop value must match one of the valid input types
```
ashleynolan: Probably worth checking, but as we support IE11 currently, `includes` isn't supported unless we polyfill it. Not sure if babel handles this, but if not, best to just use `indexOf` as an alternative here.

FayeCarter: Updated, thanks

ashleynolan (approved):

packages/f-form-field/src/components/FormDropdown.vue
```diff
@@ -0,0 +1,69 @@
...
+    padding: spacing(x1.5) spacing(x2);
+    height: 100%;
+    width: 100%;
+
+    /* Remove default styling */
+    border: none;
+    appearance: none;
+    -webkit-appearance: none;
+    -moz-appearance: none;
+}
```
AlixRichards, line 68: Do we need to include the prefixes here? Or could we rely on something like auto-prefixer to take care of this - not sure if its included in the project, if not maybe something we could look to add in the future?

AlixRichards (approved):
Left a minor question 👍

packages/f-form-field/src/components/FormDropdown.vue
FayeCarter: Will update in the next PR 😄

PR #454, opened 2020-11-19
f-form-field@1.3.0 - Add focus state to text fields and dropdown

FayeCarter:
### Added
- Focus and Active styles for `text`, `password`, `dropdown`, and `email` inputs.

### Changed
- Inline label `font-weight` and `color`.

## UI Review Checks
**Inline label styles**
image

**Text fields and dropdown focus**
image

### DIFF
--- a/packages/f-form-field/src/components/FormField.vue
+++ b/packages/f-form-field/src/components/FormField.vue
@@ -26,7 +26,8 @@
                 :data-test-id="testId"
                 :class="[
                     $style['o-form-field'],
-                    $style['c-formField-input']
+                    $style['c-formField-input'],
+                    $style['c-formField-input-textFields--focus']
                 ]"
                 :dropdown-options="dropdownOptions"
                 @update="updateOption"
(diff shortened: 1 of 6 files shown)


packages/f-form-field/src/components/FormField.vue
```diff
@@ -186,6 +189,10 @@ export default {
 
         isDropdown () {
             return this.inputType === 'dropdown';
+        },
+
+        isTextField () {
```
kevinrodrigues: Is this checking the input `type` is set to `text`? Should it also exclude the other valid types, e.g email & password? If not then might be worth calling this `isInputField`?

FayeCarter: Yeah, it's checking it's anything but `radio` and `checkbox` so that's a much better name, thanks 🙂

AlixRichards (approved):
LGTM and thank you for revisting the prefixes 👍

kevinrodrigues (approved):

PR #457, opened 2020-11-20
f-checkout@0.10.0 - Updates to UI to match PIE design

FayeCarter:
UI for `f-checkout` updated to match PIE designs.

**Selector component will be updated with new `dropdown` module from `f-form-field` in next PR.
Inline label styles for Address line 1 and 2 will also need to be updated in another PR**

### Changed
- Updated`f-card`, `f-button`, `f-form-field`, `f-error-message` versions.
- Updated font sizes, padding and layout to match PIE design.

## UI Review Checks
| Empty Form | Filled out form | Errors |
| ---| --- | --- |
| image | image | image |

### DIFF
--- a/packages/f-checkout/src/components/Address.vue
+++ b/packages/f-checkout/src/components/Address.vue
@@ -7,7 +7,7 @@
         <fieldset :class="$style['c-address-group']">
             <form-field
                 v-model="address.line1"
-                :class="$style['c-address-group-formField']"
+                :class="$style['c-address-formField']"
                 name="address-line-1"
                 data-test-id="input-address-line-1"
                 :label-text="$t('labels.line1')"
@@ -25,7 +25,7 @@
 
             <form-field
                 v-model="address.line2"
-                :class="$style['c-address-group-formField']"
+                :class="$style['c-address-formField']"
                 name="address-line-2"
                 data-test-id="input-address-line-2"
                 :label-text="$t('labels.line2')"
@@ -151,7 +151,7 @@ $address-weight-bold     : $font-weight-bold;
     border: none;
     @include font-size($address-fontSize);
 
-    .c-address-group-formField {
+    .c-address-formField {
         margin-bottom: -17px;
 
         &:focus-within,
--- a/packages/f-checkout/src/components/Address.vue
+++ b/packages/f-checkout/src/components/Address.vue
@@ -16,7 +16,7 @@
                 <template #error>
                     <error-message
                         v-if="!isAddressLine1Complete"
-                        :class="$style['c-address-group-error']"
+                        :class="$style['c-address-error']"
                         data-test-id="error-address-line1-incomplete">
                         {{ $t('validationMessages.addressLine1.requiredError') }}
                     </error-message>
@@ -34,7 +34,6 @@
 
         <form-field
             v-model="address.city"
-            :class="$style['c-address-group-formField']"
             name="address-city"
             data-test-id="input-address-city"
             :label-text="$t('labels.city')"
@@ -134,15 +133,15 @@ export default {
 </script>
 
 <style lang="scss" module>
-$address-group-colour          : $color-text;
-$address-group-fontSize        : 'body-s';
-$address-group-weight-bold     : $font-weight-bold;
+$address-colour          : $color-text;
+$address-fontSize        : 'body-s';
+$address-weight-bold     : $font-weight-bold;
 
 .c-address-label {
     display: block;
-    color: $address-group-colour;
-    @include font-size($address-group-fontSize);
-    font-weight: $address-group-weight-bold;
+    color: $address-colour;
+    @include font-size($address-fontSize);
+    font-weight: $address-weight-bold;
     margin: spacing(x2) 0 spacing();
 }
 
@@ -150,7 +149,7 @@ $address-group-weight-bold     : $font-weight-bold;
     margin-bottom: spacing(x4);
     padding: 0;
     border: none;
-    @include font-size($address-group-fontSize);
+    @include font-size($address-fontSize);
 
     .c-address-group-formField {
         margin-bottom: -17px;
@@ -163,7 +162,7 @@ $address-group-weight-bold     : $font-weight-bold;
     }
 }
 
-.c-address-group-error {
+.c-address-error {
     margin-bottom: spacing(x5);
 }
 </style>
--- a/packages/f-checkout/src/components/Address.vue
+++ b/packages/f-checkout/src/components/Address.vue
@@ -1,36 +1,38 @@
 <template>
     <div data-test-id='address-component'>
-        <div :class="$style['c-address-group']">
-            <p
+        <fieldset :class="$style['c-address-group']">
+            <legend
                 :class="$style['c-address-label']">
                 {{ $t('labels.addressGroup') }}
-            </p>
-            <form-field
-                v-model="address.line1"
-                :class="$style['c-address-formField']"
-                name="address-line-1"
-                data-test-id="input-address-line-1"
-                :label-text="$t('labels.line1')"
-                label-style="inline"
-                :has-error="isAddressLine1Empty">
-                <template #error>
-                    <error-message
-                        v-if="isAddressLine1Empty"
-                        :class="$style['c-address-error']"
-                        data-test-id="error-address-line1-empty">
-                        {{ $t('validationMessages.addressLine1.requiredError') }}
-                    </error-message>
-                </template>
-            </form-field>
+            </legend>
+            <div>
+                <form-field
+                    v-model="address.line1"
+                    :class="$style['c-address-formField']"
+                    name="address-line-1"
+                    data-test-id="input-address-line-1"
+                    :label-text="$t('labels.line1')"
+                    label-style="inline"
+                    :has-error="isAddressLine1Empty">
+                    <template #error>
+                        <error-message
+                            v-if="isAddressLine1Empty"
+                            :class="$style['c-address-error']"
+                            data-test-id="error-address-line1-empty">
+                            {{ $t('validationMessages.addressLine1.requiredError') }}
+                        </error-message>
+                    </template>
+                </form-field>
 
-            <form-field
-                v-model="address.line2"
-                :class="$style['c-address-formField']"
-                name="address-line-2"
-                data-test-id="input-address-line-2"
-                :label-text="$t('labels.line2')"
-                label-style="inline" />
-        </div>
+                <form-field
+                    v-model="address.line2"
+                    :class="$style['c-address-formField']"
+                    name="address-line-2"
+                    data-test-id="input-address-line-2"
+                    :label-text="$t('labels.line2')"
+                    label-style="inline" />
+            </div>
+        </fieldset>
 
         <form-field
             v-model="address.city"
--- a/packages/f-checkout/src/components/Address.vue
+++ b/packages/f-checkout/src/components/Address.vue
@@ -5,33 +5,31 @@
                 :class="$style['c-address-label']">
                 {{ $t('labels.addressGroup') }}
             </legend>
-            <div>
-                <form-field
-                    v-model="address.line1"
-                    :class="$style['c-address-formField']"
-                    name="address-line-1"
-                    data-test-id="input-address-line-1"
-                    :label-text="$t('labels.line1')"
-                    label-style="inline"
-                    :has-error="isAddressLine1Empty">
-                    <template #error>
-                        <error-message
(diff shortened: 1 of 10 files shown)


SachaZvetelman:
@FayeCarter, in my opinion, the placeholders for the address fields look a bit weird.
1) They look like a normal input rather than a placeholder.
2) Shouldn't they say "Line 1" and "Line 2"? Adding "address" might be redundant.
It's more of a question for the designers if you've implemented this following their designs.

packages/f-checkout/src/components/Address.vue
```diff
@@ -107,6 +114,10 @@ export default {
 
         isAddressPostcodeValid () {
             return !this.$v.addressValidations.postcode.$dirty || this.$v.addressValidations.postcode.isValidPostcode;
+        },
+
+        isAddressPostcodeFieldValid () {
```
kevinrodrigues: Shouldn't we be using what we have already? i.e the one above (`isAddressPostcodeValid`) but with the extra empty check instead?

packages/f-checkout/src/components/Address.vue
```diff
@@ -124,15 +135,36 @@ export default {
...
     border: none;
-    @include font-size(body-s);
+    @include font-size($addressGroup-fontSize);
 
     .c-addressGroup-formField {
         margin-bottom: -17px;
+
+        &:focus-within,
+        &:active {
+            z-index: 9999;
```
kevinrodrigues: Can we use the z-index mapper from fozzie here?

FayeCarter: Didn't know this existed 😂 . Thanks

packages/f-checkout/src/components/Address.vue
FayeCarter: I've updated this now 🙂

FayeCarter:
@SachaZvetelman I'll update the placeholder styles in another PR for `f-form-field` is that OK?

kevinrodrigues (dismissed):

packages/f-checkout/src/components/Address.vue
```diff
@@ -124,15 +134,36 @@ export default {
 </script>
 
 <style lang="scss" module>
+$addressGroup-colour          : $grey--darkest; // Text colour of form labels
```
simonsandersje: Better to use $color-text IMO, in case the colours are updated:
https://github.com/justeat/fozzie-colour-palette/blob/18dc9fd565de01ab0a88c84db522c5bd0c5d828b/src/scss/index.scss#L67

packages/f-checkout/src/components/UserNote.vue
```diff
@@ -31,38 +31,37 @@ export default {
...
 <style lang="scss" module>
-$note-input-borderRadius                  : 3px;
-$note-input-borderWidth                   : 1px;
-$note-input-borderColour                  : $grey--light;
-$note-textColour                          : $grey--dark;
-$line-height                              : 17.5px;
+$userNote-textArea-borderRadius                  : 3px;
+$userNote-textArea-borderWidth                   : 1px;
+$userNote-textArea-borderColour                  : $grey--light;
+$userNote-textColour                             : $grey--darkest;
```
simonsandersje: Better to use $color-text IMO, in case the colours are updated:
https://github.com/justeat/fozzie-colour-palette/blob/18dc9fd565de01ab0a88c84db522c5bd0c5d828b/src/scss/index.scss#L67

packages/f-checkout/src/components/UserNote.vue
```diff
@@ -31,38 +31,37 @@ export default {
...
 
 <style lang="scss" module>
-$note-input-borderRadius                  : 3px;
-$note-input-borderWidth                   : 1px;
-$note-input-borderColour                  : $grey--light;
-$note-textColour                          : $grey--dark;
-$line-height                              : 17.5px;
+$userNote-textArea-borderRadius                  : 3px;
+$userNote-textArea-borderWidth                   : 1px;
+$userNote-textArea-borderColour                  : $grey--light;
```
simonsandersje, line 36: Better to use $color-disabled (or define a new border-light colour?):
https://github.com/justeat/fozzie-colour-palette/blob/18dc9fd565de01ab0a88c84db522c5bd0c5d828b/src/scss/index.scss#L85

simonsandersje (commented):
Couple of super-minor comments

packages/f-checkout/src/components/Address.vue
```diff
@@ -1,17 +1,23 @@
 <template>
     <div data-test-id='address-component'>
+        <label
+            :class="$style['c-addressGroup-label']">
```
ashleynolan: The naming here is a bit off now. `c-addressGroup-label` implies that `label` is a child element of `c-addressGroup`, but that element is on line 7 adjacent to it.

So the name should be `c-addressGroupLabel` (or alternatively the `div` on line 2 could be called `c-address` and then this class would be `c-address-label` and the fieldset on line 7 would be `c-address-group`.

Let me know if that makes sense. I think changi g it to c-address might make the most sense from a component point of view, especially as the component is called Address.vue too (but you'll need to change all the `c-addressGroup…` classes to `c-address…` classes)

packages/f-checkout/src/components/Address.vue
```diff
@@ -93,46 +100,70 @@ export default {
...
         isAddressPostcodeValid () {
-            return !this.$v.addressValidations.postcode.$dirty || this.$v.addressValidations.postcode.isValidPostcode;
+            return (!this.$v.addressValidations.postcode.$dirty || this.$v.addressValidations.postcode.isValidPostcode) && this.isAddressPostcodeComplete;
         }
     },
 
     methods: {
         /*
-        * Returns true if `field` has been touched and if it is empty
+        * Returns true if `field` has been touched and if it is complete
```
ashleynolan: Is `complete` the same as `valid` here? If so, might be worth changing to be `valid` as we use that terminology quite often.

packages/f-checkout/src/components/UserNote.vue
```diff
@@ -31,38 +31,37 @@ export default {
 </script>
 
 <style lang="scss" module>
-$note-input-borderRadius                  : 3px;
-$note-input-borderWidth                   : 1px;
-$note-input-borderColour                  : $grey--light;
-$note-textColour                          : $grey--dark;
-$line-height                              : 17.5px;
+$userNote-textArea-borderRadius                  : 3px;
```
ashleynolan: PIE radius for all components is 2px now, so worth updating this (can just use `$border-radius` variable I think).

packages/f-checkout/src/components/UserNote.vue
```diff
@@ -31,38 +31,37 @@ export default {
...
+$userNote-textArea-borderWidth                   : 1px;
+$userNote-textArea-borderColour                  : $color-disabled;
```
ashleynolan: Seems weird to link the disabled colour to this border colour – is the textArea disabled?

packages/f-checkout/src/components/Address.vue
FayeCarter: Hey @ashleynolan , I've updated this now to use `address-group`. Does it look better now?

packages/f-checkout/src/components/Address.vue
FayeCarter: I shouldn't have changed this one from empty I think. It should just be checking that nothing has been typed into the field

packages/f-checkout/src/components/UserNote.vue
FayeCarter: This is now back at grey--lighter

packages/f-checkout/src/components/Address.vue
```diff
@@ -1,56 +1,63 @@
...
+                        data-test-id="error-address-line1-incomplete">
                         {{ $t('validationMessages.addressLine1.requiredError') }}
                     </error-message>
                 </template>
             </form-field>
 
             <form-field
                 v-model="address.line2"
-                :class="$style['c-addressGroup-formField']"
+                :class="$style['c-address-group-formField']"
```
ashleynolan: You could just make this `c-address-formField` if you want. Would it be useful to go through the naming scheme at some point, as get it's not totally obvious coming into it, but can try and explain how I use it?

packages/f-checkout/src/components/Address.vue
```diff
@@ -93,46 +100,70 @@ export default {
...
     methods: {
         /*
-        * Returns true if `field` has been touched and if it is empty
+        * Returns true if `field` has been touched and if it is still empty
         * The $dirty boolean changes to true when the user has focused/lost
         * focus on the input field.
         */
-        isFieldEmpty (field) {
-            return this.$v.addressValidations[field].$dirty && !this.$v.addressValidations[field].required;
+        isFieldComplete (field) {
```
ashleynolan: Sorry - not sure if confused my last comment – I just meant the method name itself (not just the comment 😆 )

Just wondering what `complete` means and if it's interchangeable with `valid` here?

FayeCarter: Ahhh I see, so the naming has changed a few times trying to find the one that is the most readable 😂 . I have `Valid` and `Complete` being used for separate types of checks. So at the moment I'm using:
`isValid` - for things like the postcode to see if the postcode is an actual postcode.
`isComplete` - to check that the field isn't empty.
Do you think these are ok, or could is there something more clear that I could use?

packages/f-checkout/src/components/Address.vue
FayeCarter: Yeah that would be great, thanks. I keep thinking I understand it and then a new one comes along 😂

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -348,53 +350,35 @@ export default {
 
 <style lang="scss" module>
 $line-height                              : 16px;
-$checkout-width                           : 460px;
+$checkout-width                           : 596px;
```
ashleynolan, line 353: Might be worth a check with design, but should this be the same width as other pages like registration? That uses the `isPageContentWrapper` prop to make it `472px` wide, but not sure if there was a reason for making this much wider than that

packages/f-checkout/src/components/Address.vue
FayeCarter: Updated to `isEmpty` and `isValid`

packages/f-checkout/src/components/Checkout.vue
FayeCarter: I'll check this with design. I'm aiming to get feedback from them when this PR is merged so I will check with them that this width is correct. Is that OK?

simonsandersje (dismissed):

kevinrodrigues (dismissed):

packages/f-checkout/src/components/Address.vue
ashleynolan: Looks like it's still complete?

ashleynolan (changes requested):

packages/f-checkout/src/components/Address.vue
```diff
@@ -1,56 +1,62 @@
 <template>
     <div data-test-id='address-component'>
-        <fieldset :class="$style['c-addressGroup']">
+        <label
```
minrwhite: This seems an odd use of `<label>` to me, as it's not associated with an input and might confuse screen readers. [The accessibility guide for this element on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Headings) would suggest that it might be better to use a `<legend>` element for this purpose, but that must be within the fieldset. I'd look at moving the opening `<fieldset>` tag prior to this and replacing the existing fieldset with a `<div>` for styling - let me know if you want a chat around this.

ashleynolan (approved):

AlixRichards (approved):
Looks good :thumbsup:

I have a minor question - which is likely due to me not knowing the designs :) 

There is a different left hand padding/margin/spacing in the text fields compared to the dropdown. To me this looks a little odd when I look at the form as a whole. Do you know if this is by design/there are plans to align the spacing in the future maybe?

minrwhite (approved):

FayeCarter:
> Looks good 👍
> 
> I have a minor question - which is likely due to me not knowing the designs :)
> 
> There is a different left hand padding/margin/spacing in the text fields compared to the dropdown. To me this looks a little odd when I look at the form as a whole. Do you know if this is by design/there are plans to align the spacing in the future maybe?

@AlixRichards, The dropdown will be updated in the next PR. This dropdown is the old design, I decided to update it separately because I think it's going to take me some time 🙂

PR #461, opened 2020-11-24
f-checkout@0.11.0 - Adding Vuex to f-checkout.

SachaZvetelman:
v0.11.0
*November 25, 2020*

### Added
- Vuex modules to make state management cleaner and clearer.


packages/f-checkout/src/components/Checkout.vue
```diff
@@ -214,32 +211,62 @@ export default {
...
-            };
+        ...mapActions('checkout', [
+            'getCheckoutDetails',
+            'postCheckoutDetails'
+        ]),
 
-            if (this.isCheckoutMethodDelivery) {
-                checkoutData.address = this.address;
-            }
+        async postCheckout () {
```
ashleynolan: Could you include a JSDoc comment for this method.

Also, at first glance to have methods called `postCheckout` and a `postCheckoutDetails` – what's the difference between them (could it be called out in their name to differentiate them?)

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -214,32 +211,62 @@ export default {
...
+                this.$emit(EventNames.CheckoutFailure, thrownErrors);
+            }
         },
 
         async getCheckout () {
             try {
-                const result = await CheckoutServiceApi.getCheckout(this.checkoutUrl, this.tenant, this.getCheckoutTimeout);
-                this.$emit(EventNames.CheckoutGetSuccess);
-                this.mapResponse(result.data);
+                await this.getCheckoutDetails({
```
ashleynolan: Know you didn't add this method, but could you add a JSDoc comment to it if you know what it does?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -140,23 +141,8 @@ export default {
     data () {
         return {
             tenantConfigs,
-            firstName: '',
```
ashleynolan, line 143: Massive 👍 for this – really cleans it up

ashleynolan (commented):
Looks good – few small minor comments, otherwise LGTM

packages/f-checkout/src/demo/demo.js
```diff
@@ -1,4 +1,5 @@
 import Vue from 'vue';
```
kevinrodrigues, line 1: Be good to document this (if it's not already) as a nice way to pass through instances of something e.g vuex. Just stolen this too :P Thanks @SachaZvetelman

SachaZvetelman: Sure thing!
This `demo.js` thing came from a different PR. _I think_, @tommcclean was gonna update the `generator-component` to generate demo files this way. Although I might be wrong!

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -206,40 +203,78 @@ export default {
...
+        if (!this.$store.hasModule('checkout')) {
+            this.$store.registerModule('checkout', checkoutModule);
+        }
+    },
+
+    /*
+        TODO: in the future, we should actually try to deregister modules.
+        However, right now, given we might have several instances of the same component, we don't want to remove the module for all of them.
+        We tried generating a dynamic module name (so we could add and remove a module per component),
+        but we couldn't get it to work. It needs more investigation when this is really needed, not now.
```
minrwhite, line 219: I'd be interested to know what the issues that were encountered here were, in order to influence the investigation down the road.

minrwhite (commented):
Great first stab at this!

Would it be worth checking within the install function in `src/index.js` whether Vuex is already installed, and dynamically requiring in at that point if necessary, or are we comitted to always requiring the consuming code to supply an instance of Vue that has Vuex pre-attached?

SachaZvetelman:
@benwhite-deltas, the problem with that `install` bit is that even if you install Vuex in there, you still cannot provide the new `Vuex.Store` to your component, which is why we've decided to let "smart components" (i.e. the primary page including the different components, such as Storybook, or CoreWeb) add the store. Then each of the child components (such as this one, `f-checkout`) will dynamically register the module.
So, I think we're good for now. Unless someone else has a different opinion/suggestion.

packages/f-checkout/src/components/Checkout.vue
SachaZvetelman: We didn't spend a lot of time on this since it's not a must-have right now, but in a nutshell, we tried generating a dynamic key for the module (e.g. `checkout-123`) using a computed property. The problem was, if I recall correctly, that we wanted to dynamically register the module before that computed property was evaluated, so it was failing. But basically, this needs more research.

packages/f-checkout/src/components/tests/Selector.test.js
```diff
@@ -1,23 +1,119 @@
...
+
+        // Assert
+        expect(selectorLabel.html()).toMatchSnapshot();
+    });
+
     describe('data ::', () => {
         describe('selectedTime ::', () => {
             // TODO: Component is working correctly but the tests do not work with `scss modules`
 
             xit('should add class to display label above option when not null', async () => {
```
kevinrodrigues, line 107: Should this be x'd out?

SachaZvetelman: Not sure. It was there already so I'll leave it x'd for now. We'll review these tests soon anyway 😄

packages/f-checkout/src/components/tests/Selector.test.js
```diff
@@ -1,23 +1,119 @@
...
+            propsData
+        });
+
+        const selectorLabel = wrapper.find('[data-test-id="fulfillment-time-label"]');
+
+        // Assert
+        expect(selectorLabel.html()).toMatchSnapshot();
+    });
+
+    it('should show the delivery label when the `serviceType` is `collection`', async () => {
```
minrwhite: Should this be
```suggestion
    it('should show the collection label when the `serviceType` is `collection`', async () => {
```

kevinrodrigues (dismissed):

packages/f-checkout/src/store/checkout.module.js
```diff
@@ -0,0 +1,117 @@
+import axios from 'axios';
+
+export default {
+    namespaced: true,
+
+    state: () => ({
+        id: '',
+        serviceType: '',
```
minrwhite, line 8: Does this, in combination with the ternary in orderMethod in the `Selector.vue` file mean that the default will be shown as collection? Should we be explicitly setting a default here instead, so as not to risk submitting an invalid order method via the postCheckout action?

SachaZvetelman: This is actually set immediately as the API returns all the information (including the type of order). We could set a default here, although I don't think it will give us a big benefit, because it will be overridden pretty much straight-away.

Please, do correct me here if I'm wrong, though!

minrwhite (dismissed):

minrwhite (dismissed):

minrwhite (approved):

kevinrodrigues (approved):

PR #462, opened 2020-11-24
fozzie-components@v2.19.0 - Added webdriver tests for f-header and video-reporting #globalconfig

LTurns:
### Added
- Additional webdriver tests for f-header.
- Allure reporter in the package.json script of each component folder (so it can be called more easily).
- Video reporter in wdio.conf (for errors only - so they are shown via recording in allure report).

A separate beforeEach hook is not needed in every wedriverIO test - instead it can be called directly from the wdio.conf. I have changed this and it hasn't caused any issues - tests still run in the same format as before. Happy to discuss this if people prefer to call the url in the tests directly 👍


packages/f-footer/test/specs/component/f-footer.component.spec.js
```diff
@@ -32,6 +26,9 @@ describe('f-footer component tests', () => {
 
         //Assert
         expect(browser.getUrl()).toContain("https://play.google.com/");
+
+        //Act
+        browser.back();
```
siggerzz: Do we need to go back after we've already done our assert?

packages/f-header/test-utils/component-objects/f-header.component.js
```diff
@@ -26,3 +27,6 @@ exports.clickHelpLink = () => {
 }
 
 exports.isLogoDisplayed = () => headerLogo().isDisplayed();
```
siggerzz: Can we use the same isDisplayedInViewport here too?

packages/f-header/test/specs/component/f-header.component.spec.js
```diff
@@ -1,14 +1,56 @@
...
     it('should display the f-header component', () => {
-
         // Assert
         expect(HeaderComponent.isLogoDisplayed()).toBe(true);
     });
-});
+
+    it('should alter visibility of navbar depending on window size', () => {
+        // Act
+        browser.setWindowSize(500, 500)
```
siggerzz, line 11: Are we trying to emulate mobile here? If so, it might be worth to try and enter emulation mode instead somehow (we can look into this!)

packages/f-header/test/specs/component/f-header.component.spec.js
```diff
@@ -1,14 +1,56 @@
...
+
+         // Assert
+         expect(HeaderComponent.isNavIconVisible()).toBe(true);
+         expect(HeaderComponent.isHelpLinkDisplayed()).toBe(false); 
+         expect(HeaderComponent.isLoginLinkDisplayed()).toBe(false);
+
+        // Act
+         browser.setWindowSize(1000, 1000);
```
siggerzz, line 19: ditto (but for desktop)

siggerzz (changes requested):
Few comments to address, but otherwise looking good!

packages/f-header/test-utils/component-objects/f-header.component.js
LTurns: yep can change that

packages/f-header/test/specs/component/f-header.component.spec.js
LTurns: yeah that was the aim, that sounds good!

siggerzz (dismissed):

wdio.conf.js
```diff
@@ -222,8 +228,10 @@ exports.config = {
     /**
      * Function to be executed after a test (in Mocha/Jasmine).
      */
-    // afterTest: function(test, context, { error, result, duration, passed, retries }) {
-    // },
+    afterTest: () => {
+        browser.url('http://localhost:8080');
+        browser.maximizeWindow();
+    },
```
minrwhite:
```suggestion
    }
```

packages/f-header/test-utils/component-objects/f-header.component.js
```diff
@@ -1,28 +1,52 @@
 const loginLink = () => $('[data-test-id="login-link"]');
-const offersLinkMobile = () => $('[data-test-id="offers-link-mobile"]');
-const offersLinkDesktop = () => $('[data-test-id="offers-link-desktop"]');
+const offersLink = () =>$$('[data-test-id="offers-link"]');
```
minrwhite: Typo on the double `$`?

packages/f-button/package.json
```diff
@@ -32,6 +32,7 @@
     "lint": "vue-cli-service lint",
     "lint:fix": "yarn lint --fix",
     "lint:style": "vue-cli-service lint:style",
+    "report": "cd ../.. && yarn report",
```
minrwhite, line 35: For each of these packages, is this simply a convenience script for CLI access to the allure report when within a package?

packages/f-header/test-utils/component-objects/f-header.component.js
LTurns: it's actually part of the syntax in webdriver - lets you grab multiple elements with the same id

packages/f-button/package.json
LTurns: yeah exactly, seemed to make things easier than having to cd out of folders before running the report 👍

packages/f-header/test-utils/component-objects/f-header.component.js
minrwhite: Cool, that makes sense - in that case, please keep the space in between the `=>` and the first `$` then, to make it clear and consistent.

LTurns: ooh yes thank you - will do that now

minrwhite (approved):

JoshuaNg2332 (approved):

PR #463, opened 2020-11-25
f-vue-icons@v1.12.0

kevinrodrigues:
### Added

- Geo outline icon.
- Geo fill icon.

### DIFF
--- a/packages/f-vue-icons/src/components/GeoFillIcon.js
+++ b/packages/f-vue-icons/src/components/GeoFillIcon.js
@@ -1,0 +1,15 @@
+export default {
+    name: 'GeoFillIcon',
+
+    props: {
+    },
+
+    functional: true,
+
+    render (h, ctx) {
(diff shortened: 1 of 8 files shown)


packages/f-vue-icons/src/components/GeoFillIcon.js
```diff
@@ -0,0 +1,15 @@
+export default {
+    name: 'GeoFillIcon',
+
+    props: {
```
matthewhardern, line 4: do we need an empty props?

AlixRichards (approved):

simonsandersje (approved):

packages/f-vue-icons/src/components/GeoFillIcon.js
kevinrodrigues: Seems to be rendered out by default. This is compiled by the `yarn build` command.

PR #464, opened 2020-11-25
f-services@1.5.0 - Add phone validation

FayeCarter:
### Added
- Phone Number validation method.

### DIFF
--- a/packages/f-services/src/validations.js
+++ b/packages/f-services/src/validations.js
@@ -36,7 +36,7 @@ const isValidPostcode = postcode => POSTCODE_REGEX.test(postcode);
  */
 const meetsCharacterValidationRules = value => /^[\u0060\u00C0-\u00F6\u00F8-\u017Fa-zA-Z-' ]*$/.test(value);
 
-const PHONE_REGEX = { 'en-GB': /^(\+(44))?[0-9]{10}|(\+(44))?[0-9]{11}$/ };
+const PHONE_REGEX = { 'en-GB': /^(\+(44))?[0-9]{10,11}$/ };
 
 const isValidPhoneNumber = (number, locale) => PHONE_REGEX[locale].test(number);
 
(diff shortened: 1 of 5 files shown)


packages/f-services/src/validations.js
```diff
@@ -36,8 +36,13 @@ const isValidPostcode = postcode => POSTCODE_REGEX.test(postcode);
  */
 const meetsCharacterValidationRules = value => /^[\u0060\u00C0-\u00F6\u00F8-\u017Fa-zA-Z-' ]*$/.test(value);
 
+const PHONE_REGEX = /^(\+(44))?[0-9]{10}|(\+(44))?[0-9]{11}$/;
```
simonsandersje: Is this only being used in the UK?

FayeCarter: For now, I think it's just going to be in checkout which is just UK at the moment. It will definitely need to be changed for everywhere else in the world though. 
I didn't know if it was best to start with UK and add other countries later or if it should all be done at once?

simonsandersje (dismissed):

packages/f-services/src/validations.js
FayeCarter: This has been updated and a ticket added add other tenants.

packages/f-services/src/validations.js
```diff
@@ -36,8 +36,13 @@ const isValidPostcode = postcode => POSTCODE_REGEX.test(postcode);
  */
 const meetsCharacterValidationRules = value => /^[\u0060\u00C0-\u00F6\u00F8-\u017Fa-zA-Z-' ]*$/.test(value);
 
+const PHONE_REGEX = { 'en-GB': /^(\+(44))?[0-9]{10}|(\+(44))?[0-9]{11}$/ };
```
minrwhite: You might find that the `|` operator here includes `^` and `$` as part of the alternate patterns. In other words, the pattern will accept anything that *starts* with 10 digits (optionally preceded by `+44`), or anything that *ends* with 11 digits (also optionally preceded by `+44`)?

I think this could be a range, as aside from the line end/line start markers I can't see anything different between the patterns otherwise?
```suggestion
const PHONE_REGEX = { 'en-GB': /^(\+(44))?[0-9]{10,11}$/ };
```

FayeCarter: Sounds good to me, thanks! I've updated it 🙂

minrwhite (approved):

simonsandersje (approved):

PR #473, opened 2020-11-30
f-checkout@v0.12.0 - Add dropdown module to selector

FayeCarter:
### Changed
- Selector component to use `f-form-field` dropdown component.
- Button version.


packages/f-checkout/src/components/Selector.vue
```diff
@@ -48,85 +27,38 @@ export default {
...
+            /*
+            * Update all fulfillment.times.selected to false
+            * Update chosen fulfillment.times.selected to true
+            */
+            this.fulfillment.times.forEach(time => {
+                if (time.label.text === e) {
+                    time.selected = true;
+                } else {
+                    time.selected = false;
+                }
```
minrwhite: Could simplify this to:
```suggestion
                time.selected = time.label.text === e;
```

packages/f-checkout/src/components/Selector.vue
```diff
@@ -48,85 +27,38 @@ export default {
...
 <style lang="scss" module>
-$form-label-colour                        : $grey--dark;
-$form-input-colour                        : $color-text;
-$form-input-bg                            : $white;
-$form-input-borderRadius                  : 3px;
-$form-input-borderWidth                   : 1px;
-$form-input-borderColour                  : $grey--light;
-$form-input-borderColour--focus           : $grey--dark;
-
 .o-form-select {
```
minrwhite: For consistency, it would be better to name the style something like:
```suggestion
.c-selector {
```

packages/f-checkout/src/components/tests/Selector.test.js
```diff
@@ -70,74 +81,73 @@ describe('Selector', () => {
...
                 const wrapper = shallowMount(Selector, {
-                    store: createStore(),
+                    store: createStore({ ...defaultState }),
                     i18n,
                     localVue,
                     propsData
                 });
 
-                const selector = wrapper.find("[data-test-id='form-select']");
+                wrapper.vm.selectionChanged('time 1');
```
minrwhite, line 147: This is the `// Act`, in this test, no?
```suggestion
                // Act
                wrapper.vm.selectionChanged('time 1');
```

packages/f-checkout/src/components/tests/Selector.test.js
```diff
@@ -70,74 +81,73 @@ describe('Selector', () => {
...
             });
+        });
+    });
 
-            it('should remove class to display label centrally when null', () => {
-                // Arrange & Act
+    describe('methods ::', () => {
+        describe('selectionChanged', () => {
+            it('should update `fulfillment.times` time to be selected', () => {
+                // Arrange && Act
```
minrwhite:
```suggestion
                // Arrange
```

packages/f-checkout/src/components/Selector.vue
FayeCarter: 🤦‍♀️ Thanks

packages/f-checkout/src/components/Selector.vue
```diff
@@ -48,85 +27,34 @@ export default {
...
-    watch: {
-        'fulfillment.times' (newValue) {
-            const selected = newValue.find(t => t.selected);
-            if (selected) {
-                this.selectedTime = selected.from;
-            }
+        },
+
+        fulfillmentTimes () {
+            /*
```
kevinrodrigues: Could you change these comments to follow jsDoc style ones please (so they also sit above the method name too rather than inside the method)

minrwhite (dismissed):

packages/f-checkout/src/components/Selector.vue
```diff
@@ -1,42 +1,21 @@
...
+        :class="$style['c-selector']"
+        input-type="dropdown"
+        :label-text="orderMethod"
+        :dropdown-options="fulfillmentTimes"
+        @input="selectionChanged" />
 </template>
 
 <script>
 import { mapState } from 'vuex';
+import FormField from '@justeat/f-form-field';
```
kevinrodrigues: I think it's worth calling this the form element it represents so at glance it's easier to work out the type of element we're using? E.g `FormDropDownSelector` or something

packages/f-checkout/src/components/Selector.vue
```diff
@@ -48,85 +27,34 @@ export default {
...
-            if (newSelected) {
-                newSelected.selected = true;
-            }
+        selectionChanged (e) {
+            /*
+            * Update all fulfillment.times.selected to false
+            * Update chosen fulfillment.times.selected to true
+            */
+            this.fulfillment.times.forEach(time => {
+                time.selected = time.label.text === e;
```
kevinrodrigues: Is this line correct? Seems to be comparing the `time.label.text` to the event object that Vue returns?

FayeCarter: @kevinrodrigues Yeah, it's testing that if it finds a time that matches the event object it will change the fulfillment.times object so that that time has a selected value of true then changing the rest to false.
https://github.com/justeat/fozzie-components/pull/473#discussion_r533218552 (not sure if that link helps my explanation 🙂 )

kevinrodrigues: Might just be me here but i still don't get it :P To me i can't see how this would ever return true as it's comparing it to the event object `InputEvent: {'contains things like preventDefault etc'}` Again could be me not reading it or understanding it correctly.

FayeCarter: Ah I've just realised why it doesn't contain the preventDefault stuff. It comes from an emitted event from `f-form-field` that just sends the time. `e` is probably not the best name for the variable. I'll change it

kevinrodrigues: Yeah that now makes sense, sorry partially me not knowing the the code fully. I think if we name it something else it would make it easier to follow. I'd also add a param to the jsDoc above so we know what value we're expecting here e.g `@param {time}` :)

packages/f-checkout/src/components/Selector.vue
```diff
@@ -48,85 +27,34 @@ export default {
...
-.o-form-select--float {
-    .o-form-select-label {
-        @include font-size(body-s);
-        top: 15px;
-    }
-
-    .o-form-select-input {
-        padding-top: spacing(x3);
-    }
+.c-selector {
```
ashleynolan: I would make this name more specific to being the checkout selector, as the styling is specific to that. So calling it `c-checkout-selector` (I'm assuming there's a `c-checkout` class in the parent component)

packages/f-checkout/src/components/tests/Selector.test.js
```diff
@@ -10,6 +10,17 @@ const localVue = createLocalVue();
 localVue.use(VueI18n);
 localVue.use(Vuex);
 
+const fulfillmentTimes = [
```
ashleynolan, line 13: If you're setting this multiple times in different files (as the same thing), you could put it into the constants file.

packages/f-checkout/src/components/Selector.vue
FayeCarter: @kevinrodrigues I've updated the description now too. Does it make more sense like that?

packages/f-checkout/src/constants.js
```diff
@@ -13,3 +13,14 @@ export const TENANT_MAP = {
...
+export const FULFILLMENT_TIMES = [
+    {
+        from: 'beginning',
+        label: {
+            text: 'time 1'
+        },
+        selected: false,
+        to: 'end'
+    }
+];
```
SachaZvetelman: I don't think this should live here, to be honest. This file has constants for the whole component, whereas these fulfillment times are for testing purposes only. I'd move these to the test file.

packages/f-checkout/src/constants.js
```diff
@@ -13,3 +13,14 @@ export const TENANT_MAP = {
     'it-IT': 'it',
     'nb-NO': 'no'
 };
+
+export const FULFILLMENT_TIMES = [
+    {
+        from: 'beginning',
```
SachaZvetelman: `from` and `to` normally will contain a datetime. I'd stick to that, to make the test look as real as possible (e.g. `"from": "2020-01-01T00:00+00:00"`)

SachaZvetelman (changes requested):

packages/f-checkout/src/components/Selector.vue
```diff
@@ -48,85 +27,36 @@ export default {
...
-            this.fulfillment.times.forEach(el => { el.selected = false; });
-            const newSelected = this.fulfillment.times.find(t => t.from === this.selectedTime);
-            if (newSelected) {
-                newSelected.selected = true;
-            }
+        /**
+        * Update all fulfillment.times.selected to false
+        * Update chosen fulfillment.times.selected to true
+        *
+        * @param {string} selectedTime The time emmited when dropdown value is changed.
```
kevinrodrigues: minor, typo `emitted`

packages/f-checkout/src/components/Selector.vue
kevinrodrigues: Yup looks good now 👍

packages/f-checkout/src/constants.js
SachaZvetelman: cc @ashleynolan, since I now realise he suggested the opposite thing 😬.

packages/f-checkout/src/components/Selector.vue
FayeCarter: Thanks, I thought I'd survived a full PR without typos then 😂

packages/f-checkout/src/components/tests/Selector.test.js
FayeCarter: Sacha has mentioned that as it's only used for testing it would be better in the test files. I'm pretty sure I agree as it will make it easier to understand testing. Is that OK? https://github.com/justeat/fozzie-components/pull/473#discussion_r533434507

packages/f-checkout/src/components/Selector.vue
FayeCarter: Thanks

SachaZvetelman (approved):

packages/f-checkout/src/components/tests/Selector.test.js
```diff
@@ -70,74 +81,73 @@ describe('Selector', () => {
...
 
-            it('should remove class to display label centrally when null', () => {
-                // Arrange & Act
+    describe('methods ::', () => {
+        describe('selectionChanged', () => {
+            it('should update `fulfillment.times` time to be selected', () => {
+                // Arrange
                 const wrapper = shallowMount(Selector, {
-                    store: createStore(),
+                    store: createStore({ ...defaultState }),
```
kevinrodrigues, line 140: Do you need to pass in the default state here as it's already being set at the top?

kevinrodrigues (approved):
LGTM  - one question, could be looked at in another pr tho

PR #476, opened 2020-11-30
f-content-cards@2.2.0-beta.8 💾 Bumps `@justeat/f-metadata` version

minrwhite:
f-content-cards@v2.2.0-beta.8
*November 30, 2020*

### Changed
- Bumped `@justeat/f-metadata` version


dandel10n (commented):
Should here be a yarn.lock file too?

minrwhite:
I don't think so, @dandel10n  - the same version of the repo includes the version of `f-metadata` we've updated to, with no other dependency changes. I've run yarn install, and the lockfile doesn't change. Technically the `yarn.lock` file could have changed beteween this and the previous PR, to include the older version of `f-metadata`, but this PR would have removed it again.

dandel10n (approved):

DamianMullins (approved):

PR #477, opened 2020-11-30
f-services@1.6.0 - Changed getFormValidationState function to return the full path of a nested object.

SachaZvetelman:
v1.6.0
*December 1, 2020*

### Added
- `getDeepObjectByPath` to `utilities.js` to traverse an object using a given path.

### Changed
- `getFormValidationState` function to return the full path of a nested object.

### DIFF
--- a/packages/f-services/src/validations.js
+++ b/packages/f-services/src/validations.js
@@ -1,11 +1,4 @@
-/**
- * Returns a deep object by traversing following the provided path.
- *
- * @param {Object} obj The object to traverse.
- * @param {Array} path The path used to traverse the object.
- * @return {Object} The object at the end of the path.
- */
-const getDeepObjectByPath = (obj, path) => path.reduce((o, key) => (o && o[key] ? o[key] : null), obj);
+import utilities from './utilities';
 
 /**
  * Returns an object containing arrays of the names of valid and invalid validation rules.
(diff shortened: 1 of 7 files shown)


DamianMullins (dismissed):

packages/f-services/src/validations.js
```diff
@@ -1,19 +1,31 @@
+/**
+ * Returns a deep object by traversing following the provided path.
+ *
+ * @param {Object} obj The object to traverse.
+ * @param {Array} path The path used to traverse the object.
+ * @return {Object} The object at the end of the path.
+ */
+const getDeepObjectByPath = (obj, path) => path.reduce((o, key) => (o && o[key] ? o[key] : null), obj);
```
kevinrodrigues: test for this method?

packages/f-services/src/validations.js
```diff
@@ -1,19 +1,31 @@
...
+    flattenedParams.forEach(param => {
+        const deepObjectProperty = getDeepObjectByPath($v, param.path);
+        const propertyFullPath = param.path.join('.');
 
-    Object.keys(fields).forEach(key => {
-        if ($v[key].$invalid) {
-            invalidFields.push(key);
-        } else {
-            validFields.push(key);
+        if (deepObjectProperty.$invalid && invalidFields.indexOf(propertyFullPath) === -1) {
```
kevinrodrigues, line 19: It's probably either worth extracting this logic out to another method so it can be commented on or adding a comment here as it's quite hard to scan and see what's being checked here.

packages/f-services/src/validations.js
kevinrodrigues: Hadn't realised that this method wasn't exposed. For testing purposes and seeing as it's a key piece of inner functionality, i think exposing it might be a good idea here, that way if this method ever changes & breaks it will flag up in the tests. Happy to leave as is so it can be thought about later.

kevinrodrigues (approved):

DamianMullins (approved):

PR #481, opened 2020-12-01
f-header@v4.3.0 - Added prop to allow skip link to be disabled.

DamianMullins:
### Changed

- Added prop to allow skip link to be disabled.


dandel10n:
What is a use case for this?

simonsandersje (approved):

DamianMullins:
On menu we already have a skip link, so at the moment we have two! Also, I've been looking into how to handle skip links in a SPA with regards to accessibility and we are going to use [`vue-skip-to`](https://github.com/vue-a11y/vue-skip-to) to handle announcing route changes and setting user focus to the correct place. Currently screen reader users do not know that they have changed route and their focus is not set correctly.

If the outcome in menu is something sharable then I'll either add it into this package; create a new package; or share what we did and how to handle it across projects.

dandel10n (approved):

PR #484, opened 2020-12-01
f-metadata@3.0.0-beta.7 - adding the beginning of logging functionality to the f-metadata library

matthewhardern:
Beginning of adding logging functionality to f-metadata. Approach here is to allow multiple logging callbacks.

### DIFF
--- a/packages/f-metadata/src/BrazeDispatcher.js
+++ b/packages/f-metadata/src/BrazeDispatcher.js
@@ -153,7 +153,7 @@ class BrazeDispatcher {
         }
 
         Object.keys(loggerCallbacks).forEach(key => {
-            this.loggerCallbacks.push(loggerCallbacks[key]);
+            this.loggerCallbacks.push((new LogService(loggerCallbacks[key])));
         });
 
         window.dataLayer = window.dataLayer || [];
@@ -303,7 +303,7 @@ class BrazeDispatcher {
     logger (type, logMessage, payload) {
         this.loggerCallbacks.forEach(callback => {
             // eslint-disable-next-line no-unused-expressions
-            (new LogService(callback))[type]?.(logMessage, payload);
+            callback[type]?.(logMessage, payload);
         });
     }
 }
--- a/packages/f-metadata/src/BrazeDispatcher.js
+++ b/packages/f-metadata/src/BrazeDispatcher.js
@@ -101,7 +101,7 @@ class BrazeDispatcher {
 
         this.dispatcherOptions = null;
 
-        this.loggerCallbacks = [];
+        this.loggerCallbackInstances = [];
         this.cardsCallbacks = [];
         this.groupedCardsCallback = [];
         this.inAppMessageClickEventCallbacks = [];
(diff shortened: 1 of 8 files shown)


packages/f-metadata/src/BrazeDispatcher.js
```diff
@@ -149,6 +165,15 @@ class BrazeDispatcher {
             throw new Error('attempt to reinitialise appboy with different parameters');
         }
 
+        // checks for logging callbacks
+        if (loggerCallbacks.length > 0) {
+            logger.bind(this);
```
minrwhite: Should we be saving, or using, the resulting function somewhere?

packages/f-metadata/src/BrazeDispatcher.js
```diff
@@ -165,14 +152,9 @@ class BrazeDispatcher {
...
-        // checks for logging callbacks
-        if (loggerCallbacks.length > 0) {
-            logger.bind(this);
-            loggerCallbacks.filter(callback => typeof callback === 'function')
-                .forEach(callback => {
-                    this.loggerCallbacks.push(callback);
-                });
-        }
+        Object.keys(loggerCallbacks).forEach(key => {
+            this.loggerCallbacks.push(loggerCallbacks[key]);
```
minrwhite: Is it worth wrapping each callback preemptively using `LogService` here rather than within `logger()`, so we're not instantiating a new LogService each time we log something?

matthewhardern: good shout

minrwhite (dismissed):

simonsandersje (dismissed):

packages/f-metadata/src/tests/BrazeDispatcher.test.js
```diff
@@ -5,6 +5,7 @@ import 'core-js/modules/es.object.from-entries';
 
 import isAppboyInitialised from '../utils/isAppboyInitialised';
 import ContentCards from '../services/contentCard.service';
+// import { LogService } from '../services/logging/logging.service';
```
kevinrodrigues: Should this be here?

matthewhardern: no have removed thanks for the spot

simonsandersje (approved):

minrwhite (approved):

packages/f-metadata/src/BrazeDispatcher.js
```diff
@@ -286,6 +293,19 @@ class BrazeDispatcher {
...
+    /**
+     * Logger for braze dispatcher, utilizes a logging class to call the call back based on type
+     * @param type - Function name being called on the class ( should be in ['info', 'warn', 'error'])
+     * @param {string} logMessage - Describe what happened
+     * @param {object} [payload={}] - Any additional properties you want to add to the logs
+     */
+    logger (type, logMessage, payload) {
+        this.loggerCallbackInstances.forEach(instance => {
+            // eslint-disable-next-line no-unused-expressions
+            instance[type]?.(logMessage, payload);
```
dandel10n, line 306: This bit is braking storybook unfortunately

PR #489, opened 2020-12-02
storybook@v0.18.0 - Add Cookie Universal to Context

tommcclean:
> This PR adds Cookie Universal to context in Storybook. It already exists in Context within CoreWeb.

- This means the method signatures are identical between StoryBook and CoreWeb as we use the same library, and other web hosts can use the same package which provides both SSR and CSR implementations
- Cookie Universal is isomorphic; however Storybook isn't, so only the client implementation is placed into context in Storybook
- **The PR establishes a pattern I intend to use** to update the I18n wiring within StoryBook, and add logger wiring too!
- In the very rare situations where a component is exported native for non vue websites; you can just import cookie universal in your component and add it to context.

**Advantages**

- Using an Isomorphic third party package within Web hosts gives us consistency and doesn't affect/increase bundle sizes for components
- Components don't need to install or manage package versions, simply rely on context
- Components will never be using different versions with different signatures and features!
- The library provides all the latest features like maxage, samesite and secure options
- This particular library has been in use in CoreWeb for some time, as it has a Nuxt official wrapper

**Example**
> Lets say we want to read a cookie in F-Button on mounted and write it to the console. All you need to do is
```
mounted () {
  const myName = this.$cookies.get('my-name');
  console.log(myName);
}
```

**What about tests?**
- Its a context value, and you don't want to test that cookies are actually set, thats not your problem
- If you REALLY want to add cookies in tests, you can import it as a devDependency and use it that way, I would argue you should never do this though.
- Simply mock the context by adding the cookie context with the methods you use, for example...

```
const context = {
  $cookies: {
    get: jest.fn()
  }
};
```

**What about other microsites?**
- If they use a component which relies on a context value; that microsite will also need to implement the context value, its pretty simple though; simply look at the context folder in storybook for the wiring you can steal. If a component relies on $cookies or i18n or $logger and the web host running it doesn't implement that, it will error until you wire it up.

**The pattern**
> The pattern means that instead of importing each individual concern into each individual component as it is needed; we rely on the website itself to be configured in a specific way. When this works components can use a wide range of features with zero wiring and zero package weight - with the core features even being available while developing in storybook. It means components work in the same way, not with subtly different implementations depending on who did it.

### DIFF
--- a/packages/storybook/CHANGELOG.md
+++ b/packages/storybook/CHANGELOG.md
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.18.0
+------------------------------
+*December 2, 2020*
+
+### Changed
+- Add Cookie Universal into Context so Mono Repo components can do this.$cookies.get('my-cookie-name');
+
 
 v0.17.0
 ------------------------------
--- a/packages/storybook/CHANGELOG.md
+++ b/packages/storybook/CHANGELOG.md
@@ -7,7 +7,7 @@ v0.18.0
 ------------------------------
 *December 2, 2020*
 
-### Changed
+### Added
 - Add Cookie Universal into Context so Mono Repo components can do this.$cookies.get('my-cookie-name');
 
 
--- a/packages/storybook/config/storybook/preview.js
+++ b/packages/storybook/config/storybook/preview.js
@@ -1,5 +1,9 @@
 import Vue from 'vue';
 import VueI18n from 'vue-i18n';
+
+// Setup Context as a Web Host (I18n, Cookies etc)
+import '../../context/index';
+
 import { addDecorator } from '@storybook/vue';
 import { ENGLISH_LOCALE } from '../../constants/globalisation';
 
--- a/packages/storybook/context/cookie.context.js
+++ b/packages/storybook/context/cookie.context.js
@@ -1,0 +1,6 @@
+import Vue from 'vue';
+import Cookie from 'cookie-universal';
+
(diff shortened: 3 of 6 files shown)


xander-marjoram (dismissed):

kevinrodrigues:
Have we spoken to Valtec about this approach? For f-searchbox, if I understand it correctly they would need to do some work to configure it, i.e inject a context?

packages/storybook/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.18.0
+------------------------------
+*December 2, 2020*
+
+### Changed
```
kevinrodrigues: Added

tommcclean:
> Have we spoken to Valtec about this approach? For f-searchbox, if I understand it correctly they would need to do some work to configure it, i.e inject a context?

@kevinrodrigues - I haven't, i don't have a contact for that, do you know who I should talk too?

kevinrodrigues:
> > Have we spoken to Valtec about this approach? For f-searchbox, if I understand it correctly they would need to do some work to configure it, i.e inject a context?
> 
> @kevinrodrigues - I haven't, i don't have a contact for that, do you know who I should talk too?

Sean Wheeler & Tim Hacker are the Principles there, they would be the best people to talk to. I don't think this would be an issue for them to add it btw, I just think we should let them know about it so we can implement this in all components fully rather than be blocked on one single component like searchbox.

kevinrodrigues (approved):
LGTM - Provided other third party apps are ok with the injection.

tommcclean:
Spoken to Param who is happy with this approach and will work with us on SearchBox when that is required `Let me know and we can coordinate accordingly`

xander-marjoram (approved):

PR #490, opened 2020-12-02
f-checkout@v0.19.0 - webdriver tests for Selector and UserNote components

LTurns:
### Added
- Webdriver IO tests for dropdown in `Selector` component.
- Webdriver tests for note-field character limit in `UserNote`

### Changed
- `f-form-field` version.
- `data-test-id` references in tests.
- Loglevel to `silent` in wdio.conf


packages/f-checkout/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.15.0
+------------------------------
+*December 2, 2020*
+
+### Added
+- Wedriver IO tests for dropdown in `Selector` component.
```
xander-marjoram: Webdriver 😄

LTurns: hahahaa thankss

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -6,7 +6,11 @@ const userNoteInput = () => $('[data-test-id="user-note"] textarea');
 
 const allergenLink = () => $('[data-test-id="allergy-button"]');
 
-const fulfillmentTimeDropdown = () => $('[data-test-id="fulfillment-time"]');
+// Dropdown
+
+const dropdownSelector = () => $('[data-test-id="form-select"]');
+
+const dropdownOptions = () => $$("select option");
```
xander-marjoram: Is it worth making this selector more specific, in case there are ever more dropdowns added?

packages/f-checkout/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-checkout",
   "description": "Fozzie Checkout – Fozzie Checkout Component",
-  "version": "0.14.0",
+  "version": "0.15.0",
```
xander-marjoram: Do we have a precedent for minor version bumps when there are no functional code changes? (Wondering if this should be a patch, but don't let this block the PR)

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -21,7 +25,14 @@ exports.inputs = {
 exports.waitForCheckoutComponent = () => checkoutComponent().waitForExist();
 exports.isCheckoutComponentDisplayed = () => checkoutComponent().isDisplayed();
 exports.isAllergenLinkDisplayed = () => allergenLink().isDisplayed();
+exports.isDropdownDisplayed = () => dropdownSelector().isDisplayed();
 
+exports.selectOption = () => $$("select option");
+exports.showDropdownOption = (index) => {
```
xander-marjoram: The name of this function doesn't seem quite right, maybe something like `getDropdownOptionText`?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -34,4 +34,20 @@ describe('f-checkout component tests', () => {
...
+        expect(CheckoutComponent.isDropdownDisplayed()).toBe(true);
+
+        // Act
+        CheckoutComponent.selectOption().forEach((element) => {
+            element.click();
+            browser.pause(1000);
+        });
+        
+        // Assert
+        expect(CheckoutComponent.showDropdownOption(0)).toBe(' As soon as possible ');
```
xander-marjoram: Is the whitespace padding around the string correct?

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
LTurns: Yeah I think that makes sense, it's very general the moment - will change it to "time-dropdown" to be more specific 👍

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -34,4 +34,20 @@ describe('f-checkout component tests', () => {
...
         CheckoutComponent.submit();
     });
+
+    it('should display times in ascending order, with default text "As soon as possible" showing first', () => {
+        // Assert
+        expect(CheckoutComponent.isDropdownDisplayed()).toBe(true);
+
+        // Act
+        CheckoutComponent.selectOption().forEach((element) => {
+            element.click();
```
xander-marjoram: What does clicking on each element test? That it is clickable? If so, that might be worth pulling out into a test of its own

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
xander-marjoram: Just to clarify I mean the CSS selector (although you can update the variable name too if you like)
```suggestion
const orderTimeDropdownOptions = () => $$('[data-test-id="form-select"] select option');
```
Would this work?

LTurns: Ah okay I see, yep happy to change this - much clearer

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -21,7 +25,14 @@ exports.inputs = {
...
+    const element = dropdownOptions();
+   
+     return element[index].getText();
```
SachaZvetelman: Nit: indentation seems off.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -34,4 +34,20 @@ describe('f-checkout component tests', () => {
...
+            browser.pause(1000);
```
SachaZvetelman: 1 second seems to be a very long pause. Is this something we must have?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
LTurns: So I've just checked and the spaces are there for every option in the dropdown. I wonder if it's intentional or not - Will double check with Faye 👍

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -34,4 +34,20 @@ describe('f-checkout component tests', () => {
         CheckoutComponent.selectFulfillmentTime('As soon as possible');
         CheckoutComponent.submit();
     });
+
+    it('should display times in ascending order, with default text "As soon as possible" showing first', () => {
+        // Assert
+        expect(CheckoutComponent.isDropdownDisplayed()).toBe(true);
```
SachaZvetelman: Any particular reason to have this `Assert` here?
In this case, I think it can be at the bottom with all the other asserts.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
LTurns: Nope happy to take this away 👍

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
LTurns: Yep exactly, I put it in here originally as an extra visual to show the user going down the list in ascending order - I could put it in a separate test though, would be equally valid and might break things down a bit

packages/f-checkout/package.json
LTurns: Not that I'm aware of, but @siggerzz might know - should I change the version to 0.14.1? It's quite a small PR with no big changes so happy to change 👍

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
LTurns: @xander-marjoram this is being solved now by Ash and Faye. Thank you for flagging! Think I've learned a good testing lesson here 👍

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -28,10 +28,47 @@ describe('f-checkout component tests', () => {
...
-        CheckoutComponent.selectFulfillmentTime('As soon as possible');
+        CheckoutComponent.inputUserNote(usernote);
+        CheckoutComponent.selectOrderTime('As soon as possible');
         CheckoutComponent.submit();
     });
+
+    it('should display times in ascending order, with default text "As soon as possible" showing first', () => {
+        // Assert
+        expect(CheckoutComponent.isOrderTimeDropdownDisplayed()).toBe(true);
+        // expect(CheckoutComponent.getOrderTimeOptionText(0)).toBe('As soon as possible');
```
Festivejelly: Remove commented out code

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -12,7 +12,7 @@ describe('f-checkout component tests', () => {
         expect(CheckoutComponent.isAllergenLinkDisplayed()).toBe(true);
     });
 
-    forEach(Object.keys(CheckoutComponent.inputs)).it('should display all fields', key => {
+    forEach(Object.keys(CheckoutComponent.inputs)).it.skip('should display all fields', key => {
```
Festivejelly: Why are we skipping this test?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
LTurns: @Festivejelly I just commented these out while I'm waiting for a small change in f-dropdown - there were some extra spaces in the dropdown text, but this should be fixed this morn so I'll uncomment and rerun the tests then 👍

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
LTurns: My bad, just added these to save time when running locally, will un-skip!

The 'submit checkout form' test has been skipped for awhile though because it's not quite finished - might wait until your PR is merged before I tackle that one, want to see how you've structure the consts in the page object 👍

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -17,11 +19,15 @@ exports.inputs = {
     addressCity: () => $('[data-test-id="input-address-city"]'),
     addressPostcode: () => $('[data-test-id="input-address-postcode"]')
 };
+exports.orderTimeDropdownOptions = () => $$('[data-test-id="form-select"] select option');
```
siggerzz: Can we rename this `getOrderTimeDropdownOptions`?

siggerzz: Having said that, it might be worth rethinking this. What're your thoughts @Festivejelly?

We shouldn't be making the elements accessible from the spec.

LTurns: Yeah I get you - seems odd to just export the html directly, not sure how else we can iterate over it in the forEach though. Could break it down so it's similar to export.inputs?

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -44,12 +50,22 @@ exports.submitCheckoutForm = addressInfo => {
...
+};
+
+/** 
+ * @description
+ * The time of the order should increase when a higher index is applied.
  */
-exports.selectFulfillmentTime = fulfillmentTimeText => {
-    fulfillmentTimeDropdown().selectByVisibleText(fulfillmentTimeText);
+exports.getOrderTimeOptionText = (index) => {
+    const element = orderTimeDropdownOptions();
```
siggerzz: This should be `elements` as it's an array.

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -58,8 +74,9 @@ exports.selectFulfillmentTime = fulfillmentTimeText => {
  *
  * @param {String} userNote The user note value to be entered
  */
-exports.inputUserNote = userNote => {
+exports.inputUserNoteLength = userNote => {
     userNoteInput().setValue(userNote);
+    return userNoteInput().getValue().length
```
siggerzz: Nit: Semicolon

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -69,3 +86,8 @@ exports.inputUserNote = userNote => {
 exports.submit = () => {
     goToPaymentButton().click();
 };
+
+exports.getCharacterLength = (userNote) => {
```
siggerzz: Should we rename this function to `getUserNoteLength`?

Alternatively, put this in a helper file where we can make this function more generic/reusable by passing in any selector and import it into this component object. Thoughts @Festivejelly?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -28,10 +28,47 @@ describe('f-checkout component tests', () => {
...
+
+        CheckoutComponent.selectOrderTime(' As soon as possible ');
+    });
+
+    it('should prevent user from writing a note of over 200 characters', () => {
+        // Arrange 
+        const usernote = 'A'
+        const maxUserNote = usernote.repeat(300);
+
+        console.log('HEYYY', CheckoutComponent.inputUserNoteLength(maxUserNote))
```
siggerzz: This can be removed.

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
siggerzz: Can we also add a JSDoc description to this function if we're going to keep it?

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
siggerzz: Can we add JSDoc description?

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
siggerzz: Can we add JSDoc description?

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
siggerzz: Can we add JSDoc description?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -28,10 +28,47 @@ describe('f-checkout component tests', () => {
...
+
+        // Assert
+        expect(CheckoutComponent.userNoteMaxCharacterCount()).toEqual('200');
+        expect(CheckoutComponent.inputUserNoteLength(maxUserNote)).toEqual(200)
+    });
+
+    it.skip('should enable a user to submit without adding a note', () => {
+        // Act
+        browser.refresh();
```
siggerzz: What's this refresh for?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
LTurns: hahahaha yep sorry, a lot of these things I'm still editing while I'm waiting for f-dropdown to be merged, hence some weird notes and comments

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
LTurns: can be removed - will do that now!

LTurns: Don't think this test can be finished until the button leads to a new page, so we can grab the new url and show that it submits

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
LTurns: yep will do @siggerzz. Which functions require JSDoc descriptions? Is it any function that's more complex than the normal .isDisplayed() or .getText()?

LTurns: I really like the idea of having a helper file with generic functions - would be so useful for future tests 👍

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v2.21.2
```
ashleynolan: I just merged version 2.22.0, so this will need to bump to v2.22.1 I think if you rebase/merge with master again

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -67,31 +83,41 @@ exports.populateCheckoutForm = addressInfo => {
...
- *
- * @param {String} userNote The user note value to be entered
+ * Sets the value of the user note and grabs the length of characters.
+ * 
+ * @param {Object} addressInfo
+ * @param {String} addressInfo.note The user's extra note
  */
-exports.inputUserNote = userNote => {
-    userNoteInput().setValue(userNote);
+exports.getUserNoteLength = addressInfo => {
```
siggerzz: I think we should only make our functions do one thing, in this case returning the user note length.

If we need to fill our the user note, it should be done in another function that's called before this one in the spec.


e.g

```
// f-checkout.spec.js

.....

CheckoutComponent.inputUserNote('foo');

var noteLength = CheckoutComponent.getUserNoteLength(); // returns: 3

expect(noteLength).toBe(3))
```

We'll also need to alter the JSDoc as a result 👍

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -67,31 +83,41 @@ exports.populateCheckoutForm = addressInfo => {
...
+    fields.userNote.input().setValue(addressInfo.note);;
+    return userNoteInput().getValue().length;
 };
-
 /**
  * @description
- * Submits the checkout form.
+ * Clicks 'Go to payment' to submit the form.
```
siggerzz: I think this JSDoc should remove the reference to 'Go to payment' as the copy could always get updated in future.

I think 'Submits the checkout form' is sufficient

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -10,21 +11,82 @@ describe('f-checkout component tests', () => {
         // Assert
         expect(CheckoutComponent.isAllergenLinkDisplayed()).toBe(true);
     });
+    
+    forEach(['mobileNumber', 'addressLine1', 'addressLine2', 'addressCity', 'addressPostcode'])
```
siggerzz: Tests on lines 15-28 can be deleted as I moved them to another spec file

siggerzz (changes requested):
Just a few minor comments!

FayeCarter (dismissed):

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v2.22.1
```
xander-marjoram, line 6: Do you need to bump the version in package.json too?

LTurns: done 👍

siggerzz (approved):
I still think the selectors could do with some work, but happy to merge for now 👍

FayeCarter (approved):

PR #494, opened 2020-12-03
f-searchbox@v4.0.0beta.7

kevinrodrigues:
PR allows keyboard selection of suggestions. We should reuse the basis of this for all API consumption. Further work will need to happen to allow this to be more flexible, for now it's tailored to google places.

### Added
- Geo icons svg colour.
- `copyOverride` prop to allow copy changes.
- Further tenant specific tests.
- Allow keyboard selection.

### Fixed
- Allow selection of pre-populated address when submit is hit.

### Changed
- Move `@justeat/browserslist-config-fozzie` to peerDependencies

### Video:

https://www.loom.com/share/2c22ba58dfa14166a20cc90755a5537c

## Browsers Tested

- [x] Chrome (latest)


packages/f-searchbox/src/components/tests/Form.test.js
```diff
@@ -113,6 +113,34 @@ describe('`Form`', () => {
                 });
 
                 describe('when `hasLastSavedAddress` is `truthy`', () => {
+                    it('should `preventDefault` so that we can post a payload correctly', () => {
```
minrwhite, line 117: Love the `so that ...` explaination here!

packages/f-searchbox/src/components/tests/Form.test.js
```diff
@@ -436,15 +464,14 @@ describe('`Form`', () => {
...
-                        callback: getLastLocation,
                         event
                     };
 
                     // Act
                     wrapper.vm.searchPreviouslySavedAddress(event);
 
                     // Assert
-                    expect(spy).toHaveBeenCalledWith(searchPayload);
+                    expect(spy).toHaveBeenCalledWith(searchPayload, getLastLocation());
```
minrwhite: I'm not sure we should be invoking `getLastLocation()` as part of the test - IMO we should mock it, set the result to be returned, and assert that the spy is called with that?

packages/f-searchbox/src/services/search.services.js
```diff
@@ -56,71 +55,69 @@ const search = ({
...
+
+                search({}, payload);
+                return payload;
+            } else if (
             // if street required but no street number available
             // display street number field
-            location.street
+                location.street
             && requiredFields.includes('street')
             && missingFields.includes('streetNumber')
```
minrwhite: I think it reads cleaner if the indentation is aligned, if the linter doesn't complain?
```suggestion
                && requiredFields.includes('street')
                && missingFields.includes('streetNumber')
```

packages/f-searchbox/src/components/tests/Form.test.js
kevinrodrigues: Yup you're right. Not sure why i left it like that. Ill update soon 👍

kevinrodrigues: Fixed.

packages/f-searchbox/src/services/search.services.js
kevinrodrigues: Fixed

minrwhite (dismissed):

packages/f-searchbox/src/tenants/tests/en-IE.test.js
```diff
@@ -3,26 +3,41 @@ import Service from '../../services/core';
...
-            const service = Service(tenant.default.service);
+        describe('when address is valid', () => {
+            it('should return `true` for populated addresses', () => {
+                // Arrange
+                const service = Service(tenant.default.service);
 
-            // Act
-            const result = service.isValid('     ');
+                // Act
+                const result = service.isValid('Cleric Beast');
```
AlixRichards, line 37: I don't think I understand this bit `service.isValid('Cleric Beast');` - where does `Cleric Beast` comes from?

AlixRichards (dismissed):
Just left a minor question for my own understanding

packages/f-searchbox/src/tenants/tests/en-IE.test.js
kevinrodrigues: It's just a mock bit of data for the `isValid` call, anything string related was required. Happy to update it to something else?

kevinrodrigues: Have changed it now

minrwhite (approved):

packages/f-searchbox/src/tenants/tests/en-IE.test.js
AlixRichards: Cool that's what I wanted to check really - could it just be any string 👍

AlixRichards (approved):

PR #495, opened 2020-12-03
f-content-cards@2.2.0-beta.9 - logging callbacks

matthewhardern:
A small PR to add a callback to the new f-metadata `loggerCallbacks`. Added tests and changes to package json to include new f-metadata

### DIFF
--- a/packages/f-content-cards/src/components/ContentCards.vue
+++ b/packages/f-content-cards/src/components/ContentCards.vue
@@ -329,7 +329,7 @@ export default {
                     handleContentCardsGrouped: this.metadataContentCardsGrouped
                 },
                 loggerCallbacks: {
-                    logger: this.handleLogging()
+                    logger: this.handleLogging
                 }
             })
             .then(dispatcher => {
@@ -545,7 +545,10 @@ export default {
         },
 
         handleLogging () {
-
+            // eslint-disable-next-line func-names
+            return function (type, logMessage, payload) {
+                if (this.$logger) this.$logger[type](logMessage, payload);
+            };
         }
     }
 };
--- a/packages/f-content-cards/src/components/ContentCards.vue
+++ b/packages/f-content-cards/src/components/ContentCards.vue
@@ -550,9 +550,9 @@ export default {
          */
         handleLogging (logger) {
             // eslint-disable-next-line func-names
-            return ((type, logMessage, payload) => {
+            return (type, logMessage, payload) => {
                 if (logger && logger[type]) logger[type](logMessage, null, payload);
-            });
+            };
         }
     }
 };
--- a/packages/f-content-cards/src/components/ContentCards.vue
+++ b/packages/f-content-cards/src/components/ContentCards.vue
@@ -544,10 +544,14 @@ export default {
                 this.testId && `ContentCard-${this.testId}-${index}`;
         },
 
+        /**
+         * Handles logging from f-metadata (callback)
+         * @returns {function(*, *=, *=): void}
+         */
         handleLogging () {
             // eslint-disable-next-line func-names
             return function (type, logMessage, payload) {
-                if (this.$logger) this.$logger[type](logMessage, payload);
+                if (this.$logger && this.$logger[type]) this.$logger[type](logMessage, null, payload);
             };
         }
     }
--- a/packages/f-content-cards/src/components/ContentCards.vue
+++ b/packages/f-content-cards/src/components/ContentCards.vue
@@ -327,6 +327,9 @@ export default {
                 callbacks: {
                     handleContentCards: this.metadataContentCards,
                     handleContentCardsGrouped: this.metadataContentCardsGrouped
+                },
+                loggerCallbacks: {
+                    logger: this.handleLogging()
                 }
             })
             .then(dispatcher => {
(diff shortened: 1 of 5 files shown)


packages/f-content-cards/src/components/tests/ContentCards.test.js
```diff
@@ -851,4 +851,60 @@ describe('ContentCards', () => {
...
+
+            // Act
+            const loggingCallback = instance.vm.handleLogging(instance.vm.$logger);
+            loggingCallback(loggingType, testMessage, testPayload);
+            await instance.vm.$nextTick();
+
+            // Assert
+            expect(instance.vm.$logger.logInfo).toHaveBeenCalledWith(testMessage, null, testPayload);
+        });
+        it('should NOT return a function with the correct logging parameters when callback is called with incorrect logging type', async () => {
```
kevinrodrigues: Should this read:

`should NOT return a function when the callback is called with incorrect logging type`

kevinrodrigues (dismissed):

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -539,6 +542,17 @@ export default {
...
                 this.testId && `ContentCard-${this.testId}-${index}`;
+        },
+
+        /**
+         * Handles logging from f-metadata (callback)
+         * @returns {function(*, *=, *=): void}
+         */
+        handleLogging (logger) {
+            // eslint-disable-next-line func-names
+            return function (type, logMessage, payload) {
```
DamianMullins: Do you need to use the `function` keyword here?

matthewhardern: As far as I am aware yes, as I need to return a function to be the callback, essentially using closure to provide the logger to the callback, with `handleLogging` being the outer function that provides it.

matthewhardern: although open to better ideas

DamianMullins: You're not referencing `this` anywhere so not sure the context matters? I could be wrong but an arrow function might suffice.

matthewhardern: oh sorry misread what you were saying, yes happy to swap, as no use of `this` keyword.

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -539,6 +542,17 @@ export default {
...
+                if (logger && logger[type]) logger[type](logMessage, null, payload);
+            };
```
minrwhite: So you're meaning something like this, @DamianMullins? I don't see a problem with this, unless theres something else that precluded this, @matthewhardern ?
```suggestion
            return (type, logMessage, payload) => {
                if (logger && logger[type]) logger[type](logMessage, null, payload);
            };
```

packages/f-content-cards/src/components/tests/ContentCards.test.js
matthewhardern: good point thanks

packages/f-content-cards/src/components/ContentCards.vue
matthewhardern: changed to arrow function

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -539,6 +542,17 @@ export default {
...
+        },
+
+        /**
+         * Handles logging from f-metadata (callback)
+         * @returns {function(*, *=, *=): void}
+         */
+        handleLogging (logger) {
+            // eslint-disable-next-line func-names
+            return ((type, logMessage, payload) => {
+                if (logger && logger[type]) logger[type](logMessage, null, payload);
```
DamianMullins: Just thinking, it might be a little easier to follow the code if you wrap the conditional code down onto a new line? It's more consistent with the style used elsewhere also.

```suggestion
                if (logger && logger[type]) {
                    logger[type](logMessage, null, payload);
                }
```

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -539,6 +542,17 @@ export default {
...
                 this.testId && `ContentCard-${this.testId}-${index}-${groupIndex}` :
                 this.testId && `ContentCard-${this.testId}-${index}`;
+        },
+
+        /**
+         * Handles logging from f-metadata (callback)
+         * @returns {function(*, *=, *=): void}
+         */
+        handleLogging (logger) {
+            // eslint-disable-next-line func-names
```
DamianMullins: Can remove this now.

packages/f-content-cards/src/components/ContentCards.vue
DamianMullins: Exactly @benwhite-deltas

matthewhardern: changed as agreed, thought @DamianMullins meant remove the function lol, misread

packages/f-content-cards/src/components/ContentCards.vue
matthewhardern: agreed and resolved

DamianMullins (approved):

kevinrodrigues (approved):

PR #505, opened 2020-12-07
#trivial CircleCI tweaks

xander-marjoram:
Revert tag to SHA, because not all commits are tagged.

Not sure if the attachments works, but I _think_ it's the only way to get the background colour.


.circleci/config.yml
```diff
@@ -34,44 +34,46 @@ commands:
...
-              "blocks": [
-                {
-                  "type": "section",
-                  "text": {
-                    "type": "mrkdwn",
-                    "text": "Your job *${CIRCLE_JOB}* has failed ⚠️"
-                  },
-                  "fields": [
-                    {
+              "attachments": [
```
xander-marjoram, line 37: I recommend hiding whitespace changes to view this file

tommcclean (approved):

siggerzz (approved):

PR #507, opened 2020-12-07
f-alert@0.4.0 - vue-button

dandel10n:
### Changed
- Regular button changed to vue button component
- Renamed VueAlert to FAlert

### Updated
- `f-vue-icons` to the latest version

Screen Shot 2020-12-07 at 11 21 56

### Changed
- Renamed VueButton to FButton

### DIFF
--- a/packages/f-alert/src/components/Alert.vue
+++ b/packages/f-alert/src/components/Alert.vue
@@ -18,19 +18,21 @@
                 data-test-id="alert-heading">
                 {{ heading }}
             </h2>
-            <button
+            <vue-button
                 v-if="isDismissible"
                 type="button"
-                :class="[$style['c-alert-dismiss'], 'o-btn o-btn--icon']"
+                :class="[$style['c-alert-dismiss']]"
+                button-type="icon"
+                button-size="xsmall"
                 data-test-id="alert-dismiss"
-                @click="dismiss">
+                @click.native="dismiss">
                 <cross-icon
                     :class="[$style['c-alert-dismiss-icon']]"
                 />
                 <span class="is-visuallyHidden">
                     {{ copy.dismissAlertText }}
                 </span>
-            </button>
+            </vue-button>
         </div>
         <div
             :class="$style['c-alert-content']">
@@ -48,7 +50,9 @@ import {
     WarningIcon
 } from '@justeat/f-vue-icons';
 import { globalisationServices } from '@justeat/f-services';
+import VueButton from '@justeat/f-button';
 import tenantConfigs from '../tenants';
+import '@justeat/f-button/dist/f-button.css';
 
 export default {
     name: 'VueAlert',
@@ -57,7 +61,8 @@ export default {
         DangerIcon,
         InfoIcon,
         SuccessIcon,
-        WarningIcon
+        WarningIcon,
+        VueButton
     },
     props: {
         locale: {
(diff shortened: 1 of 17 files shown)


FayeCarter (dismissed):

packages/f-alert/src/components/Alert.vue
```diff
@@ -48,7 +50,9 @@ import {
     WarningIcon
 } from '@justeat/f-vue-icons';
 import { globalisationServices } from '@justeat/f-services';
+import VueButton from '@justeat/f-button';
```
ashleynolan: Super NIT – I'd probably call this `ButtonComponent` as that's how we've started to name these imports elsewhere

dandel10n: I changed it here and in f-button readme file :ok_hand:

FayeCarter (dismissed):

packages/f-alert/src/components/Alert.vue
matthewhardern: Question - Does this naming convention break the vue style guides https://vuejs.org/v2/style-guide/#Base-component-names-strongly-recommended ?

matthewhardern (dismissed):
Have a question but can be resolved in later PR

packages/f-alert/src/components/Alert.vue
DamianMullins: Looking at the guide `FButton` might be a good choice?

FayeCarter (dismissed):

matthewhardern (dismissed):

packages/f-alert/src/components/Alert.vue
```diff
@@ -48,16 +50,19 @@ import {
     WarningIcon
 } from '@justeat/f-vue-icons';
 import { globalisationServices } from '@justeat/f-services';
+import FButton from '@justeat/f-button';
```
kevinrodrigues, line 53: Out of interest why not call this something like `DismissAlertButton` or similar. Would help with understanding at a glance what the button represents.

matthewhardern: Because if the use case arrose for two buttons to be required then this would invalidate that naming convention

kevinrodrigues: Ah that's my bad i just realised this is a component and not a consuming one. Sorry!

kevinrodrigues (dismissed):

FayeCarter (dismissed):

packages/f-alert/src/components/Alert.vue
ashleynolan: Can we just call this `ButtonComponent`? Would make it nice and generic that way (unless people think that's not a good name)

DamianMullins: `FButton` looked to [match one of the recommended naming schemes](https://vuejs.org/v2/style-guide/#Base-component-names-strongly-recommended)?

AlixRichards (approved):

FayeCarter (approved):

packages/f-alert/src/components/Alert.vue
ashleynolan: Didn't even know that this was a thing!

I'd suggest if we want to go with a consistent approach for including components, we change this across the board, as we don't do this anywhere else currently.

I think I prefer using `BaseXXXXX` for our component names as seems more generic. @DamianMullins Thoughts?

PR #510, opened 2020-12-07
f-mega-modal@0.3.0 - vue-button

dandel10n:
### Added
- `f-vue-icons` to use cross icon in the modal

### Changed
- Regular close button changed to vue-button

### Removed
- Hardcoded buttons from storybook file

Screen Shot 2020-12-07 at 11 38 02

### DIFF
--- a/packages/f-mega-modal/package.json
+++ b/packages/f-mega-modal/package.json
@@ -36,8 +36,6 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-button": "0.4.0",
-    "@justeat/f-vue-icons": "1.13.0"
   },
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.1.1",
@@ -50,6 +48,8 @@
     "@vue/cli-plugin-unit-jest": "4.4.6",
     "@vue/test-utils": "1.0.3",
     "body-scroll-lock": "3.0.3",
-    "node-sass-magic-importer": "5.3.2"
+    "node-sass-magic-importer": "5.3.2",
+    "@justeat/f-button": "0.4.0",
+    "@justeat/f-vue-icons": "1.13.0"
   }
 }
--- a/packages/f-mega-modal/package.json
+++ b/packages/f-mega-modal/package.json
@@ -49,7 +49,7 @@
     "@vue/test-utils": "1.0.3",
     "body-scroll-lock": "3.0.3",
     "node-sass-magic-importer": "5.3.2",
-    "@justeat/f-button": "0.4.0",
+    "@justeat/f-button": "0.4.1",
     "@justeat/f-vue-icons": "1.13.0"
   }
 }
--- a/packages/f-mega-modal/package.json
+++ b/packages/f-mega-modal/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-mega-modal",
   "description": "Fozzie Mega Modal – A Vue.js modal component",
-  "version": "0.2.1",
+  "version": "0.3.0",
   "main": "dist/f-mega-modal.common.js",
   "files": [
     "dist"
@@ -35,6 +35,10 @@
   "browserslist": [
     "extends @justeat/browserslist-config-fozzie"
   ],
+  "dependencies": {
+    "@justeat/f-button": "0.4.0",
+    "@justeat/f-vue-icons": "1.13.0"
+  },
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.1.1",
     "body-scroll-lock": "3.x"
--- a/packages/f-mega-modal/src/components/MegaModal.vue
+++ b/packages/f-mega-modal/src/components/MegaModal.vue
@@ -38,8 +38,7 @@
                         data-test-id="close-modal"
                         @click.native="close">
                         <cross-icon
-                            :class="[$style['c-megaModal-closeIcon']]"
-                        />
+                            :class="[$style['c-megaModal-closeIcon']]" />
                         <span class="is-visuallyHidden">
                             {{ closeButtonCopy }}
                         </span>
--- a/packages/f-mega-modal/src/components/MegaModal.vue
+++ b/packages/f-mega-modal/src/components/MegaModal.vue
@@ -27,25 +27,23 @@
                 <slot
                     v-if="hasCloseButton"
                     name="close-button">
-                    <button
+                    <vue-button
                         type="button"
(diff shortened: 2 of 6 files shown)


packages/f-mega-modal/stories/MegaModal.stories.js
```diff
@@ -2,6 +2,7 @@ import { withKnobs, boolean, text } from '@storybook/addon-knobs';
 import { withA11y } from '@storybook/addon-a11y';
 import MegaModal from '../src/components/MegaModal.vue';
 
+
```
FayeCarter: Sorry, extra line here?

packages/f-mega-modal/package.json
```diff
@@ -35,6 +35,10 @@
   "browserslist": [
     "extends @justeat/browserslist-config-fozzie"
   ],
+  "dependencies": {
+    "@justeat/f-button": "0.4.0",
```
DamianMullins: I believe these can go into `devDependencies` as they are baked in at build time and aren't needed by the consuming application.

packages/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -27,25 +27,23 @@
...
-                            :class="$style['c-megaModal-closeIcon']"
-                            viewBox="0 0 9 9"
-                            xmlns="http://www.w3.org/2000/svg">
-                            <path d="M4.33 4.74L.83 8.25a.29.29 0 0 1-.41-.4l3.5-3.52L.43.83a.29.29 0 1 1 .4-.41l3.51 3.5L7.85.43a.29.29 0 0 1 .4.4l-3.51 3.5 3.51 3.52a.29.29 0 0 1-.4.4L4.33 4.74z" />
-                        </svg>
-
+                        @click.native="close">
+                        <cross-icon
+                            :class="[$style['c-megaModal-closeIcon']]"
+                        />
```
DamianMullins: NIT: Closing tag should be on line above.

packages/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -55,9 +53,16 @@
 </template>
 
 <script>
+import { CrossIcon } from '@justeat/f-vue-icons';
 import { disableBodyScroll, enableBodyScroll } from 'body-scroll-lock';
+import VueButton from '@justeat/f-button';
+import '@justeat/f-button/dist/f-button.css';
```
DamianMullins, line 58: What happens when the modal is used in an application where the button CSS is already included? Is the CSS duplicated?

ashleynolan: We haven't done much digging around this yet (as far as I know). Would be worth testing this out with bundle analyser though to see if webpack is smart enough to remove the duplicate CSS though.

DamianMullins: I can do a quick test on menu.

DamianMullins: I pulled both this branch and the f-buttons CSS into menu and the CSS is duplicated. 

One way to resolve this is to only pull in the CSS in the consuming application (extra step required), that way the CSS is only included once. 

I guess that would also break the StoryBook styles so would have to look at importing the CSS in the story config file maybe?

DamianMullins: Also, I see a weird dotted outline on the close button? Should that be there? Looks like some sort of updated outline style? 

Screenshot 2020-12-07 at 13 20 39

I [added a new bit of CSS into Fozzie recently](https://github.com/justeat/fozzie/pull/315) which hides focus styles when not required - maybe the same needs to be done in the button components?

dandel10n: @damian I checked that dotted outline is in the latest designs for :focus state for all the buttons now

dandel10n: About css import - I will write a ticket to investigate it and change for all the components, because now we import styles for all of them this way.

DamianMullins: I think the focus style should only show for non-mouse focus. I'll chase that up and change in the button component if so.

packages/f-mega-modal/package.json
dandel10n: Interesting, I am trying to understand why we have justeat packages as dependencies not dev dependencies for all the components

packages/f-mega-modal/stories/MegaModal.stories.js
dandel10n: :heavy_check_mark: Removed

packages/f-mega-modal/src/components/MegaModal.vue
dandel10n: :heavy_check_mark:

DamianMullins: I have started a write-up on the differences and need to get that finished, but in essence `dependencies` are things you need to use when the component is run, and `devDependencies` are things which are baked in when you build the package and aren't required when the component is run i.e. icons.

packages/f-mega-modal/package.json
DamianMullins: I have started a write-up on the differences and need to get that finished, but in essence `dependencies` are things you need to use when the component is run i.e. the scroll locking package for this modal, and `devDependencies` are things which are baked in when you build the package and aren't required when the component is run i.e. icons.

dandel10n: My understanding was that in dev dependencies should be packages that are only needed for local development and testing. [Npm docs ](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file) suggest the same.

DamianMullins: Yeah, that's right, in this case the mega-modal is compiled locally and the result is pushed to npm meaning that the compiled bundles have everything (apart from the scroll package) they need at that point.

Leaving the icons in doesn't really affect the consuming application but it means that you'll be downloading resources you're never using i.e. the icons.

dandel10n: Sold :smile:

packages/f-mega-modal/src/components/MegaModal.vue
DamianMullins: @dandel10n Let me know when you have a ticket to track the CSS changes, and let me know if I can help with that in any way.

DamianMullins (dismissed):

DamianMullins (dismissed):

kevinrodrigues (dismissed):

kevinrodrigues (approved):

DamianMullins (approved):

PR #513, opened 2020-12-07
f-checkout@0.17.0 - Add collection prop tests

siggerzz:
v0.17.0
*December 7, 2020*

### Added
- Component tests for 'Collection' prop variant.

### Changed
- Split component tests into separate spec files


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "root",
-  "version": "2.21.0",
+  "version": "2.21.1",
```
xander-marjoram, line 3: I don't think you need to bump this unless you're also bumping it in the changelog

siggerzz: An entry was made to the changelog 👍

xander-marjoram: But the version was not bumped, you can just replace `Latest (to be added to next release)` with `2.21.1` and that'll be groovy

siggerzz: done!

packages/f-checkout/test/specs/component/f-checkout-collection.component.spec.js
```diff
@@ -0,0 +1,35 @@
+import forEach from 'mocha-each';
+import DemoControls from '../../../test-utils/component-objects/demo-controls';
+import CheckoutComponent from '../../../test-utils/component-objects/f-checkout.component';
+
+describe('f-checkout "collection" component tests', () => {
+    forEach(['mobileNumber'])
+    .it.only('each fields error message should be displayed', field => {
```
xander-marjoram: Just checking that the `.only` here isn't preventing any of the other tests from being run?

xander-marjoram (dismissed):

xander-marjoram (approved):

JoshuaNg2332 (approved):

PR #514, opened 2020-12-07
f-form-field@v1.6.0 - Grouped fields

FayeCarter:
### Added
- Grouped fields selection and styles.

### DIFF
--- a/packages/f-form-field/src/components/FormField.vue
+++ b/packages/f-form-field/src/components/FormField.vue
@@ -73,7 +73,7 @@ import {
     DEFAULT_INPUT_TYPE,
     VALID_INPUT_TYPES,
     VALID_LABEL_STYLES,
-    VALID_GROUP_POSITION,
+    VALID_GROUP_POSITIONS,
     MOBILE_WIDTH
 } from '../constants';
 
@@ -133,7 +133,7 @@ export default {
         groupPosition: {
             type: String,
             default: 'middle',
-            validator: value => (VALID_GROUP_POSITION.indexOf(value) !== -1) // The prop value must match one of the valid input types
+            validator: value => (VALID_GROUP_POSITIONS.indexOf(value) !== -1)
         }
     },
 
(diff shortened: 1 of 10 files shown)


packages/f-form-field/src/components/FormField.vue
```diff
@@ -4,7 +4,8 @@
         :class="[
             $style['c-formField'], {
                 [$style['c-formField--invalid']]: hasError
-            }
+            },
+            ( isGrouped ? $style[`c-formField--grouped${groupPosition}`] : $style[`c-formField--unGrouped`] )
```
ashleynolan: Do you need an ungrouped class? Would think that would just be the default state

FayeCarter: I created it because I couldn't work out a better way to keep the styling that added margin between non grouped form fields that sit next to each other. 
The styling on the default `c-formField` class, before I started, meant that the middle and bottom grouped fields would have a top margin too.
Very open to changing this though 😄. I'm sure there's gotta be a better way.

ashleynolan: I would probably split the `c-formField--grouped` and group position into two separate classes. May be easier to explain on a hangout, but have `isGrouped` control adding `c-formField--grouped` (which can remove the margin), and then add a `c-formfield--topGroup` or something like that to the element based on the `groupPosition` setting.

ashleynolan: Shall we have a hangout on this tomorrow? May be better to sow me what you're trying to do with it.

FayeCarter: Yeah, that'd be great thank you 😄

simonsandersje (dismissed):

packages/f-form-field/src/components/FormField.vue
```diff
@@ -3,7 +3,8 @@
         :data-theme-formfield="theme"
         :class="[
             $style['c-formField'], {
-                [$style['c-formField--invalid']]: hasError
+                [$style['c-formField--invalid']]: hasError,
+                [$style[`c-formField--grouped`]]: isGrouped
```
kevinrodrigues: Minor but can use a normal string here? '' instead of ``?

FayeCarter: 👍 I'll update this now

kevinrodrigues (approved):

AlixRichards (approved):

PR #515, opened 2020-12-07
generator@v1.9.0 - Update Template to Support F-Globalisation and F-Development-Context

tommcclean:
> Superpower the generator

**Changed**
- Updated demo to use new format script which uses `f-development-context` to be externalised meaning OOTB `$logger`, `$cookies`, `I18n` and `Vuex`!
- Wire up `F-Globalisation` within the Skeleton and tests so you can use `<i18n>`, `$t('message')` and `this.i18n`
- Fix an existing issue that results in the generator not executing (somebody removed yeoman by accident)
- Fix an issue with the test script - (somebody removed it by accident)

I have tested this by loading the generator component in...
- Storybook (and playing with Locale)
- Yarn Build
- Yarn Test
- Yarn Demo


packages/generator-component/generators/app/templates/src/components/Skeleton.vue
```diff
@@ -1,34 +1,32 @@
 <template>
     <div
-        :data-theme="theme"
+        data-theme="jet"
```
ashleynolan: Can delete this line now – it's not needed anymore.

ashleynolan (commented):
Looks good – one tiny comment

packages/generator-component/generators/app/templates/src/components/Skeleton.vue
tommcclean: Cheers, will do

DamianMullins (approved):

tommcclean:
Paused for now due to Storybook Modifications

PR #516, opened 2020-12-08
f-checkout@v0.18.0 -  Update `f-services` validations and remove Spanish locale

FayeCarter:
### Changed
- `f-services` version.
- Postcode validation.

### Removed
- Spanish Locale from storybook.

### DIFF
--- a/packages/f-checkout/src/components/Checkout.vue
+++ b/packages/f-checkout/src/components/Checkout.vue
@@ -328,6 +328,10 @@ export default {
         */
         isValidPhoneNumber () {
             return validations.isValidPhoneNumber(this.customer.mobileNumber, this.$i18n.locale);
+        },
+
+        isValidPostcode () {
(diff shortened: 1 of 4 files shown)


dandel10n (approved):

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -328,6 +328,14 @@ export default {
         */
         isValidPhoneNumber () {
             return validations.isValidPhoneNumber(this.customer.mobileNumber, this.$i18n.locale);
+        },
+
+        /*
```
AlixRichards, line 333: Do you need to add test for this?

AlixRichards (approved):

packages/f-checkout/src/components/Checkout.vue
FayeCarter: Hey! There are tests in place already to make sure the validations work correctly. I've also done tests for the validations in f-services. 🙂

PR #521, opened 2020-12-09
f-checkout@0.21.0 - Updated f-checkout with component tests for mobile number and title field

JoshuaNg2332:
v0.20.0
*December 9, 2020*

### Added
- Webdriver test for phone number format in `f-checkout-delivery`.
- Mobile number field title into `Checkout.vue`.
- Webdriver test for mobile number form field title.


packages/f-checkout/src/components/Checkout.vue
```diff
@@ -24,6 +24,7 @@
                 @submit.prevent="onFormSubmit">
                 <form-field
                     v-model="customer.mobileNumber"
+                    :title="$t('validationMessages.mobileNumber.requiredError')"
```
SachaZvetelman: Not sure I understand this. It's a form field but we're passing the translation of a "required" error?

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -63,6 +63,7 @@ exports.isAllergenLinkDisplayed = () => allergenLink().isDisplayed();
 exports.isOrderTimeDropdownDisplayed = () => orderTimeDropdown().isDisplayed();
 exports.userNoteMaxCharacterCount = () => userNoteInput().getAttribute('maxlength');
 exports.clickPaymentButton = () => goToPaymentButton().click();
+exports.getAttributeValue = fieldName => fields[fieldName].input().getAttribute('title');
```
SachaZvetelman: I'd rename this to `getTitleAttribute` or make it more generic and pass the attribute name as a parameter and call it `getAttributeValueByName` or similar.

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
```diff
@@ -22,4 +22,28 @@ describe('f-checkout "delivery" component tests', () => {
         // Assert
         expect(CheckoutComponent.isFieldDisplayed(field)).toBe(true);
     });
+
+    forEach(['mobileNumber'])
```
SachaZvetelman: There's no need for the `forEach` if it's only one element.

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
```diff
@@ -22,4 +22,28 @@ describe('f-checkout "delivery" component tests', () => {
...
+    .it('should have a title matching the error on the field', field => {
+        // Arrange
+        const mobileNumber = 'Your phone number should be at least 10 characters long and shouldn\'t contain letters or special characters';
+
+        // Assert
+        expect(CheckoutComponent.getAttributeValue(field, 'title')).toBe(mobileNumber);
+    });
+
+    forEach(['mobileNumber'])
```
SachaZvetelman: Same as above.

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
```diff
@@ -22,4 +22,28 @@ describe('f-checkout "delivery" component tests', () => {
         // Assert
         expect(CheckoutComponent.isFieldDisplayed(field)).toBe(true);
     });
+
+    forEach(['mobileNumber'])
+    .it('should have a title matching the error on the field', field => {
+        // Arrange
+        const mobileNumber = 'Your phone number should be at least 10 characters long and shouldn\'t contain letters or special characters';
```
SachaZvetelman: Please try to find a more meaningful name for this variable, since it looks like it contains a mobile number, but in reality, it doesn't 🙂

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
```diff
@@ -22,4 +22,28 @@ describe('f-checkout "delivery" component tests', () => {
...
+
+        // Assert
+        expect(CheckoutComponent.getAttributeValue(field, 'title')).toBe(mobileNumber);
+    });
+
+    forEach(['mobileNumber'])
+    .it('should display an error message when a number format is incorrectly entered', field => {
+        // Arrange
+        const wrongNumber = {
```
SachaZvetelman: Same as above.

SachaZvetelman (changes requested):

packages/f-checkout/src/components/Checkout.vue
JoshuaNg2332: Correct me if I'm wrong @Festivejelly. But this is for when screen readers pass over the field, they know what the format of the field is. The "required" error part is pulled from the tenant which is the same copy as the error message that we are then passing as the title of the actual field itself.

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
JoshuaNg2332: @SachaZvetelman Are you highlighting the name of the variable as well to change? Or just the `forEach`.

SachaZvetelman: Both :)

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -44,19 +43,19 @@ describe('f-checkout component tests', () => {
     });
 
     it.only('should prevent user from writing a note of over 200 characters', () => {
```
kevinrodrigues: Won't this stop the rest of the `IT` statements running in this file? i.e looks like jest will only run the spec for this and not the other tests. Was this accidently left in?

JoshuaNg2332: Looks like this was accidently left in from a previous PR. Will fix this.

packages/f-checkout/src/components/Checkout.vue
SachaZvetelman: Just to clarify, I'm in favour of passing a title so that it's screenreader-friendly, but passing a required error message as the title looks incorrect to me.

JoshuaNg2332: @SachaZvetelman Just looked into this and had a chat with @Festivejelly and the scope of this ticket was to match this to prod and to test that it's the same as prod. But Nick did mention that maybe now is a good time to improve this design.

SachaZvetelman: Sure, could you (or @Festivejelly) point me in the right direction? I cannot see this in the existing Production version. Thanks!

JoshuaNg2332: @SachaZvetelman I can at least see it on prod in the dom when I go to checkout and check the inspect element if this is what you mean?

JoshuaNg2332:
After chats with @SachaZvetelman , we've decided to not include the title changes to this PR as it doesn't make sense to have this message as the title of the form field. We will revisit this in a separate PR. As such, the title changes and the tests associated with that have been removed.

packages/f-checkout/test/specs/component/f-checkout-collection.component.spec.js
```diff
@@ -3,16 +3,15 @@ import DemoControls from '../../../test-utils/component-objects/demo-controls';
 import CheckoutComponent from '../../../test-utils/component-objects/f-checkout.component';
 
 describe('f-checkout "collection" component tests', () => {
-    forEach(['mobileNumber'])
-    .it('each fields error message should be displayed', field => {
+    it('each fields error message should be displayed', () => {
```
kevinrodrigues: Minor but could you change this to: `should display 'mobileNumber' error message when collection method is set` (unsure if that's the correct scenario description wise)

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
```diff
@@ -17,9 +17,22 @@ describe('f-checkout "delivery" component tests', () => {
...
 
-    forEach(['mobileNumber'])
-    .it('should display the mandatory fields', field => {
+    it('should display the mandatory fields', () => {
         // Assert
-        expect(CheckoutComponent.isFieldDisplayed(field)).toBe(true);
+        expect(CheckoutComponent.isFieldDisplayed('mobileNumber')).toBe(true);
+    });
+
+    it('should display an error message when a number format is incorrectly entered', () => {
```
kevinrodrigues: Is the reverse of this covered? i.e when the number is valid it should not display the error message?

JoshuaNg2332: Yup. Its been covered [here](https://github.com/justeat/fozzie-components/blob/master/packages/f-checkout/test/specs/component/f-checkout.component.spec.js#L62-L80) and will be enabled soon.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -44,19 +43,19 @@ describe('f-checkout component tests', () => {
...
 
         // Act
-        CheckoutComponent.inputUserNote(addressInfo)
+        CheckoutComponent.inputUserNote(addressInfo);
         var noteLength = CheckoutComponent.getUserNoteLength();
 
         // Assert
         expect(CheckoutComponent.userNoteMaxCharacterCount()).toEqual('200');
-        expect(noteLength).toBe(200)
+        expect(noteLength).toBe(200);
```
kevinrodrigues: Out of interest why do we need the expect above? It seems to be mapped to essentially the same thing as this check. I think this expect alone is fine.

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
kevinrodrigues: That has no `expects` in that block you linked to.

JoshuaNg2332: There is nothing currently to expect on as there isn't an API in place however, this will soon be enabled and the field will be indirectly tested.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
JoshuaNg2332: Good point. I've removed these now. This could also be refactored into unit test at some point.

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
kevinrodrigues: Ok, I still think this test needs the reverse tho because that particular test is looking for a correct submission rather than the scenario around showing an error message, it doesn't look for the error message in the same way as this one does. This one is more granular.

kevinrodrigues (dismissed):

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
```diff
@@ -17,9 +17,22 @@ describe('f-checkout "delivery" component tests', () => {
         expect(CheckoutComponent.doesInputFieldExist(field)).toBe(true);
     });
 
-    forEach(['mobileNumber'])
-    .it('should display the mandatory fields', field => {
```
LTurns: does this mobileNumber test need to be separate from the fields test above it?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -33,30 +32,28 @@ describe('f-checkout component tests', () => {
...
-        const userNote = 'A'
+    it('should prevent user from writing a note of over 200 characters', () => {
+        // Arrange
+        const userNote = 'A';
         const addressInfo = {
             note: userNote.repeat(300)
         };
 
         // Act
-        CheckoutComponent.inputUserNote(addressInfo)
```
LTurns, line 59: do we not need these functions? I quite like the input from the user as well as getting the maxlength attribute, but happy to discuss! :)

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
JoshuaNg2332: Ok, I've added that extra test now and also renamed the tests so its a little clearer.

kevinrodrigues (dismissed):

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
JoshuaNg2332: Had a chat with Kevin and Ben about this and the expect is already testing for it to be 200 and also we're looking at revisiting this as a unit test.

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
JoshuaNg2332: Looking into this, I believe these are 2 separate tests. One is checking it is displayed and the other is if it exists in the dom. Doing a bit of digging now to see why it's only the mobileNumber is being tested.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
LTurns: ahh okay cool, sounds good! Might need to remove the functions from the component object too if we're getting rid, or this could be tackled separately if it's becoming a unit test 👍

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -35,7 +35,6 @@
                         </error-message>
                     </template>
                 </form-field>
-
```
DamianMullins, line 38: Not one to change now necessarily, but it's best to keep whitespace between elements like this as it makes the file easier to scan - especially as components get larger it can be hard to pick out where sibling elements sit alongside each other.

DamianMullins (dismissed):

siggerzz (dismissed):

siggerzz (dismissed):

kevinrodrigues (dismissed):

siggerzz (approved):

kevinrodrigues (approved):

SachaZvetelman (approved):

PR #522, opened 2020-12-09
f-registration@v0.44.0 - Add loginURL Prop to support returnURL

tommcclean:
**Problem**
- When using the "Already have an account?" link, returnURL is lost; so navigating from Offers -> Registration -> Login and signing in, means the user won't get back to offers.

### Added
- 'loginUrl' prop so that returnUrl is not lost when clicking 'Already on Just Eat' as it can be passed in by the parent

### Changed
- Updated storybook to include missing props
- Make 'showLoginLink' a required prop

### Removed
- Unsupported EU tenants
- Obselete localisation message for loginUrl

image

### DIFF
--- a/packages/f-registration/CHANGELOG.md
+++ b/packages/f-registration/CHANGELOG.md
@@ -3,6 +3,21 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.44.0
+------------------------------
+*December 9, 2020*
+
+### Added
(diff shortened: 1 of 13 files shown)


tommcclean:
> Fails
> 🚫	
> ⬆️ This PR should include a [`SEMVER`](https://docs.npmjs.com/getting-started/semantic-versioning) version bump for **f-registration**, so that these packages can be published once merged.
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [68485d4](https://github.com/justeat/fozzie-components/commit/68485d44e73ecf8655439ef12c43e036c91c71e2)

uhm, it does thanks

tommcclean:
My fork got lost by the public/private :(

PR #532, opened 2020-12-10
fozzie-components@2.24.0 - Update component README's #globalconfig

siggerzz:
This PR updates the README's for components and removes references to the old demo files.


ashleynolan:
Out of interest – is there any way of running this standalone file, like we do currently `yarn demo`? Only reason I ask is that we found an issue in `storybook`/`webpack` by testing using the demo file (to do with shared CSS being generated in storybook that was incorrect), so it may be useful to be able to spin up a separate view like this in the future, until we fix this storybook issue.

siggerzz:
As already discussed with @ashleynolan, Storybook is now the only mechanism to view components.

LTurns (dismissed):
Just need to add that when running component tests, you need to change terminals before doing 'yarn test-component:chrome' but I can add that separately once this is merged. :)

JoshuaNg2332 (approved):

LTurns (approved):

PR #536, opened 2020-12-11
f-checkout@0.26.0 - Add tests for vuex store module

FayeCarter:
### Added
- Tests for store `checkout.module.js`.


simonsandersje (dismissed):

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+    ]
+};
+
+const state = CheckoutModule.state();
+const { updateState } = CheckoutModule.mutations;
+const { getCheckout, postCheckout } = CheckoutModule.actions;
+
+describe('CheckoutModule', () => {
+    describe('mutations ::', () => {
```
kevinrodrigues, line 43: I like the use of `::` 👍

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,125 @@
...
+                updateState(state, DELIVERY_RESPONSE);
+
+                // Assert
+                expect(state).toEqual(expectedState);
+            });
+        });
+    });
+
+    describe('actions ::', () => {
+        const commit = jest.fn();
```
kevinrodrigues: These should go inside a `beforeEach` I feel?

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,125 @@
...
+        const axiosMock = new MockAdapter(axios);
+
+        const payload = {
+            url: 'http://localhost/account/checkout',
+            tenant,
+            timeout
+        };
+
+        describe('getCheckout ::', () => {
+            const config = {
```
kevinrodrigues: Same as above?

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }],
+    messages: [{
+        type: 'warning',
+        message:
+            {
+                text: 'Please hurry, the restaurant is closing soon'
+            }
+    }
+    ]
```
SachaZvetelman: Nit: please move this to the line above.

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,125 @@
+import axios from 'axios';
+import MockAdapter from 'axios-mock-adapter';
```
kevinrodrigues: Do you need this? With axios i've normally spy'd on the instance and returned a mocked response using jest itself.

e.g

`axios.get = jest.fn(() => Promise.resolve({ data: 'Some data' }));`

or

`spyOn(axios, 'get').mockImpletmentation(() => Promise.resolve({ data: 'some data' }))`

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,125 @@
...
+};
+
+const state = CheckoutModule.state();
+const { updateState } = CheckoutModule.mutations;
+const { getCheckout, postCheckout } = CheckoutModule.actions;
+
+describe('CheckoutModule', () => {
+    describe('mutations ::', () => {
+        describe('updateState ::', () => {
+            it('should update state with delivery response.', () => {
```
SachaZvetelman, line 49: What happens when there's no customer data?
And when there's no address?

We need to ensure our tests cover all the possible paths within the `updateState` function 🙂

packages/f-checkout/src/store/tests/checkout.module.test.js
FayeCarter: Thanks 🙂

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,184 @@
...
+            postcode: ''
+        }
+    },
+    notes: [],
+    isFulfillable: true,
+    notices: [],
+    messages: []
+};
+
+const expectedState = {
```
kevinrodrigues: There's a mixture of double and single quotes in this object. You can escape single quotes with a backslash.

FayeCarter: Thanks

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }]
+};
+
+const { updateState } = CheckoutModule.mutations;
+const { getCheckout, postCheckout } = CheckoutModule.actions;
+let state = CheckoutModule.state();
+
+describe('CheckoutModule', () => {
+    it(('should create default state when initialised.'), () => {
```
kevinrodrigues: Seems to be an extra set of braces here, should only need the one set?

`it('Something', () => {})` instead of `it(('Something'), ()=> {})`

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,184 @@
...
+    });
+
+    describe('mutations ::', () => {
+        describe('updateState ::', () => {
+            beforeEach(() => {
+                state = defaultState;
+            })
+
+            it('should update state with delivery response.', () => {
+                // Arrange && Act
```
kevinrodrigues: Minor but this one is more of just an `Act`.

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,184 @@
...
+                updateState(state, checkoutDelivery);
+
+                // Assert
+                expect(state).toEqual(expectedState);
+            });
+
+            it('should leave customer state empty if no customer data is returned from the API.', () => {
+                // Arrange && Act
```
kevinrodrigues: Can you split this out like:

```javascript
// Arrange
checkoutDelivery.customer = null;

// Act
updateState(state, checkoutDelivery);

File: packages/f-checkout/src/store/tests/checkout.module.test.js [hunk 11]
```diff
@@ -0,0 +1,184 @@
...
+                checkoutDelivery.customer = null;
+                updateState(state, checkoutDelivery);
+
+                // Assert
+                expect(state.customer).toEqual(defaultState.customer);
+            });
+
+            it('should leave address state empty if no address data is returned from the API.', () => {
+                // Arrange && Act
```
kevinrodrigues: Same as above please

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,184 @@
...
+            commit = jest.fn();
+        });
+
+        describe('getCheckout ::', () => {
+            beforeEach(() => {
+                axios.get = jest.fn(() => Promise.resolve({ data: checkoutDelivery }));
+            });
+
+            it('should get the checkout details from the backend and call `updateState` mutation.', async () => {
+                // Arrange && Act
```
kevinrodrigues: Same, this one is just an `Act`

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,184 @@
...
+                await getCheckout({ commit }, payload);
+
+                // Assert
+                expect(axios.get).toHaveBeenCalled();
```
kevinrodrigues: Should this one be a `toHaveBeenCalledWith` statement so we can verify it was passed the correct arguments? e.g the `url` from the payload & the `config`.

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,184 @@
...
+            beforeEach(() => {
+                axios.post = jest.fn(() => Promise.resolve({ status: 200 }));
+            });
+
+            it('should post the checkout details to the backend.', async () => {
+                // Arrange && Act
+                await postCheckout({ commit }, payload);
+
+                // Assert
+                expect(axios.post).toHaveBeenCalled();
```
kevinrodrigues: Similar thing regarding the `toHaveBeenCalledWith()`

packages/f-checkout/src/store/tests/checkout.module.test.js
FayeCarter: Oh yeah, sure 👍

kevinrodrigues (dismissed):

kevinrodrigues (dismissed):

packages/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -0,0 +1,237 @@
...
+            state = defaultState;
+        });
+
+        describe('updateState ::', () => {
+            it('should update state with delivery response.', () => {
+                // Act
+                updateState(state, checkoutDelivery);
+
+                // Assert
+                expect(state).toEqual(expectedState);
```
DamianMullins: You _could_ use a snapshot here which would allow you to avoid typing out the large expected variable above? i.e. `expect(state).toMatchSnapshot()`.

FayeCarter: Oh that's so much better, thanks 🙂

kevinrodrigues (approved):

DamianMullins (approved):

PR #538, opened 2020-12-14
f-checkout@v0.33.0 - Add login link to checkout.

BanksySan:
---

Add a message to the Checkout, similar to the one in Registration that checks that the correct user is signed in and offers them a link if not.

---

## UI Review Checks

Added message

## Browsers Tested

- [X] Chrome (latest) (87.0.4280.88 (Official Build) (64-bit))
- [ ] Internet Explorer 11

- Firefox 77.0b1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,12 +20,20 @@ v3.2.0
 - Fixes danger package checks.
 
 
-v3.1.0
+v3.1.1
 ------------------------------
 *January 4, 2020*
 
 ### Fixed
 - `bundlewatch` config updated in line with new structure.
+
+
+v3.1.0
+------------------------------
+*January 04, 2021*
+
+### Changed
+- Added a link to the checkout page to allow the user to correct who is logged in.
 
 
 v3.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,20 +20,12 @@ v3.2.0
 - Fixes danger package checks.
 
 
-v3.1.1
+v3.1.0
 ------------------------------
 *January 4, 2020*
 
 ### Fixed
 - `bundlewatch` config updated in line with new structure.
-
-
-v3.1.0
-------------------------------
-*January 04, 2021*
-
-### Changed
-- Added a link to the checkout page to allow the user to correct who is logged in.
 
 
 v3.0.0
--- a/packages/components/organisms/f-checkout/CHANGELOG.md
+++ b/packages/components/organisms/f-checkout/CHANGELOG.md
@@ -3,6 +3,15 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+v0.33.0
+------------------------------
+*January 6, 2021*
+
+### Added
+- Link to the checkout page to allow the user to correct who is logged in.
+
+
 v0.32.0
 ------------------------------
 *January 5, 2021*
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -30,7 +30,7 @@
                         :href="loginUrl"
                         data-test-id="switch-user-link"
                         @click="onVisitLoginPage">
-                        {{ $t('switchUserText', { name: name }) }}
+                        {{ $t('switchUserText', { name }) }}
                     </a>
                 </p>
                 <form-field
@@ -141,10 +141,12 @@ export default {
             required: false,
             default: 1000
         },
+
         authToken: {
             type: String,
             default: ''
         },
+
         loginUrl: {
             type: String,
             required: true
@@ -247,6 +249,7 @@ export default {
             'getAvailableFulfilment',
             'setAuthToken'
         ]),
+
         onVisitLoginPage () {
             this.$emit(EventNames.CheckoutVisitLoginPage);
         },
@@ -424,7 +427,6 @@ $checkout-padding                         : spacing(x5) 100px;
   text-align: center;
 
   a {
-    // TODO: check default link styles in PIE and update fozzie // (should be able to remove these styles then)
     text-decoration: none;
     font-weight: $font-weight-bold;
 
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -22,14 +22,13 @@
                 type="post"
                 :class="$style['c-checkout-form']"
                 @submit.prevent="onFormSubmit">
-                <p
+                <a
+                    :href="loginUrl"
                     :class="[
-                        $style['c-checkout-switchUserLink']
+                        $style['c-checkout-link']
                     ]"
-                    data-test-id="create-account-login-link"
-                    @click="onVisitLoginPage">
-                    <a :href="loginUrl">Not {{ name }}? Click here.</a>
-                </p>
+                    data-test-id="account-login-link"
+                    @click="onVisitLoginPage">Not {{ name }}? Click here.</a>
                 <form-field
                     v-model="customer.mobileNumber"
                     name="mobile-number"
@@ -417,7 +416,7 @@ $line-height                              : 16px;
 $checkout-width                           : 596px;
 $checkout-padding                         : spacing(x5) 100px;
 
-.c-checkout-switchUserLink {
+.c-checkout-link {
   text-align: center;
 
   a {
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -28,8 +28,12 @@
                     ]">
                     <a
                         :href="loginUrl"
-                        data-test-id="account-login-link"
-                        @click="onVisitLoginPage">Not {{ name }}? Click here.</a>
+                        data-test-id="switch-user-link"
+                        @click="onVisitLoginPage">
+                        <i18n
+                            path="switchUserText">
+                            <template #name>{{ name }}</template>
+                        </i18n></a>
                 </p>
                 <form-field
                     v-model="customer.mobileNumber"
@@ -246,7 +250,7 @@ export default {
             'setAuthToken'
         ]),
         onVisitLoginPage () {
-            this.$emit(EventNames.VisitLoginPage);
+            this.$emit(EventNames.CheckoutVisitLoginPage);
         },
 
         /**
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -145,7 +145,6 @@ export default {
         loginUrl: {
             type: String,
             required: true,
-            default: ''
         }
     },
 
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -423,20 +423,6 @@ $line-height                              : 16px;
 $checkout-width                           : 596px;
 $checkout-padding                         : spacing(x5) 100px;
 
-.c-checkout-link {
-  text-align: center;
-
-  a {
-    text-decoration: none;
-    font-weight: $font-weight-bold;
-
-    &:hover,
-    &:focus {
-      text-decoration: underline;
-    }
-  }
-}
-
 .c-checkout {
     margin: auto;
     font-family: $font-family-base;
@@ -464,5 +450,19 @@ $checkout-padding                         : spacing(x5) 100px;
     .c-checkout-submitButton {
         margin: spacing(x4) 0 spacing(x0.5);
     }
+}
+
+.c-checkout-link {
+  text-align: center;
+
+  a {
+    text-decoration: none;
+    font-weight: $font-weight-bold;
+
+    &:hover,
+    &:focus {
+      text-decoration: underline;
+    }
+  }
 }
 </style>
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -22,6 +22,14 @@
                 type="post"
                 :class="$style['c-checkout-form']"
                 @submit.prevent="onFormSubmit">
+                <p
+                    :class="[
+                        $style['c-checkout-switchUserLink']
+                    ]"
+                    data-test-id="create-account-login-link"
+                    @click="visitLoginPage">
+                    <a :href="loginUrl">Not {{ name }}? Click here.</a>
+                </p>
                 <form-field
                     v-model="customer.mobileNumber"
                     name="mobile-number"
@@ -130,9 +138,13 @@ export default {
             required: false,
             default: 1000
         },
-
         authToken: {
             type: String,
+            default: ''
+        },
+        loginUrl: {
+            type: String,
+            required: true,
             default: ''
         }
     },
@@ -233,6 +245,9 @@ export default {
             'getAvailableFulfilment',
             'setAuthToken'
         ]),
+        visitLoginPage () {
+            this.$emit(EventNames.VisitLoginPage);
+        },
 
         /**
          * Submit the checkout details while emitting events to communicate its success or failure.
@@ -402,6 +417,21 @@ export default {
 $line-height                              : 16px;
 $checkout-width                           : 596px;
 $checkout-padding                         : spacing(x5) 100px;
+
+.c-checkout-switchUserLink {
+  text-align: center;
+
+  a {
+    // TODO: check default link styles in PIE and update fozzie // (should be able to remove these styles then)
+    text-decoration: none;
+    font-weight: $font-weight-bold;
+
+    &:hover,
+    &:focus {
+      text-decoration: underline;
+    }
+  }
+}
 
 .c-checkout {
     margin: auto;
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -27,7 +27,7 @@
                         $style['c-checkout-switchUserLink']
                     ]"
                     data-test-id="create-account-login-link"
-                    @click="visitLoginPage">
+                    @click="onVisitLoginPage">
                     <a :href="loginUrl">Not {{ name }}? Click here.</a>
                 </p>
                 <form-field
@@ -144,7 +144,7 @@ export default {
         },
         loginUrl: {
             type: String,
-            required: true,
+            required: true
         }
     },
 
@@ -244,7 +244,7 @@ export default {
             'getAvailableFulfilment',
             'setAuthToken'
         ]),
-        visitLoginPage () {
+        onVisitLoginPage () {
             this.$emit(EventNames.VisitLoginPage);
         },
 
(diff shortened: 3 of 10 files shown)


packages/f-checkout/stories/Checkout.stories.js
```diff
@@ -32,11 +32,14 @@ export const CheckoutComponent = () => ({
...
             default: select('Checkout Url', [deliveryUrl, collectionUrl], deliveryUrl)
+        },
+        loginUrl: {
+            default: text('Login Url', '/login')
         }
     },
     store: new Vuex.Store({}),
     template:
-        '<vue-checkout :checkoutUrl="checkoutUrl" :locale="locale" />'
+            '<vue-checkout :checkoutUrl="checkoutUrl" :locale="locale"  :login-url="loginUrl"/>'
```
tommcclean:
```suggestion
            '<vue-checkout :checkout-url="checkoutUrl" :locale="locale"  :login-url="loginUrl"/>'
```

Not your fault; but this existing prop should be hyphenated

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -22,6 +22,14 @@
                 type="post"
                 :class="$style['c-checkout-form']"
                 @submit.prevent="onFormSubmit">
+                <p
```
tommcclean: QUESTION: Do we think it would be a good idea to try and extract some of the markup out of this component into smaller dumb components?

I don't mean to reuse between registration and checkout, but a small component within the F-Checkout package just to keep the main component concise and readable.

Just a question because I wonder the same thing in F-Registration where our main component ends up quite large

ie: `packages/f-checkout/src/components/LogoutLink.vue`

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -124,6 +132,11 @@ export default {
             type: Number,
             required: false,
             default: 1000
+        },
+        loginUrl: {
+            type: String,
+            required: true,
+            default: ''
```
tommcclean: Default shouldn't be needed if it is required

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -374,6 +389,21 @@ $line-height                              : 16px;
 $checkout-width                           : 596px;
 $checkout-padding                         : spacing(x5) 100px;
 
+.c-checkout-switchUserLink {
+  text-align: center;
+
+  a {
+    // TODO: check default link styles in PIE and update fozzie // (should be able to remove these styles then)
```
tommcclean: Should add a ticket number to todo's

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -22,6 +22,14 @@
                 type="post"
                 :class="$style['c-checkout-form']"
                 @submit.prevent="onFormSubmit">
+                <p
+                    :class="[
+                        $style['c-checkout-switchUserLink']
+                    ]"
+                    data-test-id="create-account-login-link"
+                    @click="visitLoginPage">
```
dandel10n: Why we have click event on p tag and not on an actual link?

packages/f-checkout/src/components/Checkout.vue
kevinrodrigues: Probably not for open source repos tho.

packages/f-checkout/src/components/Checkout.vue
BanksySan: I think a logout/login component is a good idea.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -221,7 +234,9 @@ export default {
             'getCheckout',
             'postCheckout'
         ]),
-
+        visitLoginPage () {
```
kevinrodrigues: worth naming this `onVisitLoginPage`?

packages/f-checkout/src/components/Checkout.vue
kevinrodrigues: Agree here.

packages/f-checkout/src/components/Checkout.vue
BanksySan: I didn't actually add that, but I completely agree.

packages/f-checkout/src/components/Checkout.vue
kevinrodrigues: Could you add a unit test for this change please? (Could be done in another PR)

packages/f-checkout/src/components/Checkout.vue
kevinrodrigues: I feel using a p tag as a clickable element is a bit strange here? Could we use just the anchor here? Would help with tabbing or keyboard accessibility too (around the click event trigger)

packages/f-checkout/src/components/Checkout.vue
dandel10n: Can this TODO be completed here? Because you copied it from f-registration component as I can see, which is not great

dandel10n (commented):
Could you please resolve fozzie bot comment + few notes from me

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -22,6 +22,14 @@
                 type="post"
                 :class="$style['c-checkout-form']"
                 @submit.prevent="onFormSubmit">
+                <p
+                    :class="[
+                        $style['c-checkout-switchUserLink']
+                    ]"
+                    data-test-id="create-account-login-link"
+                    @click="visitLoginPage">
+                    <a :href="loginUrl">Not {{ name }}? Click here.</a>
```
SachaZvetelman: @BanksySan, could you ensure this is coming from a translation file? So that we don't have to change it in the future.

packages/f-checkout/src/components/Checkout.vue
SachaZvetelman: Yup, please don't add ticket numbers here.

packages/f-checkout/src/components/Checkout.vue
BanksySan: I'll get this checked in and approved, then create a ticket for that.  It'd be a trivial thing for anyone familiar with Vue, but as I'm a newbie it might take longer,

packages/f-checkout/src/components/Checkout.vue
ashleynolan: Can you add a JSDoc comment for this method

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -374,6 +389,21 @@ $line-height                              : 16px;
 $checkout-width                           : 596px;
 $checkout-padding                         : spacing(x5) 100px;
 
+.c-checkout-switchUserLink {
```
ashleynolan: Not sure if this could just be called `c-checkout-link` if it's the only link being styled specifically on checkout currently?

packages/f-checkout/src/components/Checkout.vue
ashleynolan: I think at minimum, we probably just need to create a ticket to resolve this with design (so to check the default link design and make sure we have it styled correctly in fozzie).

packages/f-checkout/src/components/Checkout.vue
BanksySan: @ashleynolan There isn't JDocs on the other methods.

SachaZvetelman: What are the guidelines for JSDocs, if we have some? Because, personally, I don't think we should write JSDocs on every single function or method we write. cc @ashleynolan

kevinrodrigues: I don't think there are any guidelines set in stone at the moment, apart from just adding it to all methods we introduce / modify. My two cents tho, I did question the value when added them to f-searchbox at the start (mostly for the smaller methods) but after doing so I do really see the value in having them for a few reasons:

1. Anyone (including your future self) can pick up or understand the methods easily if there's a simple line explaining what it is used for. This particular example here is just a simple emit but a doc around why this emit is being used would be nice to have particularly for the next person that picks it up, otherwise you'd have to navigate the code base to figure out why it's emitting, who's listening to it & what happens after the emit has been listened to, you'd potentially have to do this anyway but it makes it that much easier to understand IMO.

2. Methods tend to expand and grow, if there's a base line comment then the next PR usually gets picked up if you've forgotten to add or update the comment (i've seen this on my own PRs).

SachaZvetelman: I partially agree, @kevinrodrigues. I normally find writing JSDocs or any other documentation for methods, in general, is particularly useful when those methods are publicly consumed, or in some cases when we're doing some really complex logic that requires clarification.
If we **need** to document every method to make it clear or readable, it means either we're over-documenting or we're not writing code that is simple enough.
However, if the team feels we should be adding JSDoc for every method and computed property, I'm personally happy with that, as long as it's been thought and discussed 🙂

kevinrodrigues: I feel it's mostly around making everything readable for `everyone`. Something readable/understandable to one person might not be to someone else (even a simple emit like this, if you've not used Vue) :) Have to also remember unit tests are there as further documentation too but i've actually found this to be quite a nice additional layer on top of those tests. 😸

ashleynolan: So the main guideline is as @kevinrodrigues said that if we're writing or touching methods, then we should be adding JSDOc comments. Kev did a good job of why I think they're good, but a couple of other things:

> If we need to document every method to make it clear or readable, it means either we're over-documenting or we're not writing code that is simple enough.

I don't necessarily agree with the opinion that if you write simple code you don't need comments. Comments shouldn't just describe what you're doing, but any caveats around how it interacts with other parts of the app.

---

> I feel it's mostly around making everything readable for everyone. Something readable/understandable to one person might not be to someone else (even a simple emit like this, if you've not used Vue)

Definitely agree with this.  Also, adding comments to all methods will get people into the habit of writing better comments (and is easier to enforce than if we said only certain methods of different sizes). I think if we ever get to the point where we think we're over-documenting code, I think that would be a great problem to have, but I think we're miles away from that atm 😄

BanksySan: To me, the existence of a comment means that the code isn't readable and should be corrected.  There are occasions when this is impossible/unfeasible though but to me those are:

* A complex, optimised algorithm.
* An unintuitive fix for some bug, the comment preventing someone from trying to correct it only to hit the same bug.
* Public consumption, if we must.

ashleynolan: >To me, the existence of a comment means that the code isn't readable and should be corrected

I agree (broadly) with this when it's inside a function (on a line of code), but JSDoc comments are there to provide clear documentation via comments. They can be used to generate actual documentation if needed and also provide info via code editors when relevant plugins are used.

From my experience, codebases where comment structure is mandated are much easier to understand and traverse than those without any comments.

ashleynolan (dismissed):
Marking this as do not merge. This PR will need refactoring once the folder structure is changed (probably easier to make a copy of the changes you have made and re-apply them to the new locations, as merging will likely give you a bunch of conflicts)

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -22,6 +22,17 @@
...
                 @submit.prevent="onFormSubmit">
+                <p
+                    :class="[
+                        $style['c-checkout-link']
+                    ]">
+                    <a
+                        :href="loginUrl"
+                        data-test-id="switch-user-link"
+                        @click="onVisitLoginPage">
+                        {{ $t('switchUserText', { name: name }) }}
```
SachaZvetelman: I think we can simplify this to `{ name }`.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -130,10 +141,13 @@ export default {
             required: false,
             default: 1000
         },
-
         authToken: {
             type: String,
             default: ''
+        },
+        loginUrl: {
```
SachaZvetelman, line 150: Ultra nit: add an empty line, please, so it's consistent with the other props.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -234,6 +248,9 @@ export default {
             'getAvailableFulfilment',
             'setAuthToken'
         ]),
+        onVisitLoginPage () {
```
SachaZvetelman, line 253: Same as above.

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -18,9 +21,10 @@ describe('Checkout', () => {
     allure.feature('Checkout');
     const checkoutUrl = 'http://localhost/checkout';
     const checkoutAvailableFulfilmentUrl = 'http://localhost/checkout/fulfilment';
+    const loginUrl = 'http://dummy-login.example.com';
 
     it('should be defined', () => {
-        const propsData = { checkoutUrl, checkoutAvailableFulfilmentUrl };
+        const propsData = { checkoutUrl };
```
SachaZvetelman: Are you sure you need to remove the existing URL being passed? This doesn't look right.

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -32,6 +36,27 @@ describe('Checkout', () => {
         expect(wrapper.exists()).toBe(true);
     });
 
+    it('should show the login link.', () => {
```
SachaZvetelman: Hyper mega ultra nit: please remove the `.` at the end (again, to keep it consistent) 😄

SachaZvetelman: Not sure, but should we place this test within a `describe` so it's grouped like the rest of the tests?

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -82,6 +107,25 @@ describe('Checkout', () => {
             // Assert
             expect(registerModuleSpy).not.toHaveBeenCalled();
         });
+
+        it('should show emit VisitLoginPage event when login link is clicked.', () => {
```
SachaZvetelman:
```suggestion
        it('should emit the `VisitLoginPage` event when the login link is clicked', () => {
```

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -32,6 +36,27 @@ describe('Checkout', () => {
         expect(wrapper.exists()).toBe(true);
     });
 
+    it('should show the login link.', () => {
+        const dummyName = 'Dave';
```
FayeCarter: I think there should already be a name that you can access in `defaultState.customer.firstName`

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -32,6 +36,27 @@ describe('Checkout', () => {
...
+        const propsData = { checkoutUrl, loginUrl };
+        const wrapper = shallowMount(VueCheckout, {
+            i18n,
+            store: createStore(),
+            localVue,
+            propsData,
+            computed: {
```
FayeCarter: Not sure if you'll need this if the comment above worked

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
BanksySan: No, good spot.

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v3.1.0
```
dandel10n: `f-checkout` changelog should be updated, not the root one + package.json version + yarn.lock change shouldn't be committed

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -405,6 +425,20 @@ $line-height                              : 16px;
 $checkout-width                           : 596px;
 $checkout-padding                         : spacing(x5) 100px;
 
+.c-checkout-link {
```
dandel10n: Could you please move this class after `.c-checkout` to maintain the structure

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -32,18 +32,30 @@ describe('Checkout', () => {
...
+            localVue,
+            propsData
+        });
+
+        // Act
+        const loginLink = wrapper.find("[data-test-id='switch-user-link']");
+
+        // Assert
+        expect(loginLink.exists()).toBe(true);
+        expect(loginLink.text()).toBe(`Not ${dummyName}? Click here.`);
```
dandel10n:
```suggestion
        expect(loginLink.text()).toBe(`Not ${defaultState.customer.firstName}? Click here.`);
```

packages/components/organisms/f-checkout/src/event-names.js
```diff
@@ -4,12 +4,14 @@ const CheckoutGetSuccess = 'checkout-get-success';
 const CheckoutGetFailure = 'checkout-get-failure';
 const CheckoutAvailableFulfilmentGetSuccess = 'checkout-available-fulfilment-get-success';
 const CheckoutAvailableFulfilmentGetFailure = 'checkout-available-fulfilment-get-failure';
+const CheckoutVisitLoginPage = 'checkout-visit-login-page';
```
dandel10n, line 7: This need to be added to the README.md events description along with ccs class you've added:

https://github.com/justeat/fozzie-components/blob/master/packages/components/organisms/f-checkout/README.md#events
https://github.com/justeat/fozzie-components/blob/master/packages/components/organisms/f-checkout/README.md#css-classes

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
BanksySan: There were a few reasons:

- The default state is a shared fixture, so creating a clone of the `firstName` ensures that it's not getting changed inside the test and then causing other tests to fail.
- Making it a `const` doubly ensures that the value isn't accidently changed.
- Using a named token makes its purpose explicit (i.e. it's not a real name, just a stub).

SachaZvetelman: I second @dandel10n's suggestion. The `defaultState` should be anyway provided into each test, so changing it in one test shouldn't have an impact on another test. I'd favour reusing existing available data, personally.

dandel10n (dismissed):

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
BanksySan: @SachaZvetelman @FayeCarter 

If you guys want to grab me tomorrow I'll show you the why I prefer it the way it is.

SachaZvetelman: cc @dandel10n ☝️

packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -24,7 +33,7 @@ v0.31.0
 ### Added
 - New `clearCheckoutForm` function added to clear the fields in tests in the `f-checkout-component.js` file.
 
-*December 30, 2020*
+  *December 30, 2020*
```
dandel10n: I know that it is not you change but should we have this date at all here?

FayeCarter (approved):

dandel10n (approved):

PR #540, opened 2020-12-14
#trivial - Add tests to helper.js file.

kevinrodrigues:
Adds tests to helper.js. I have not included the cookie helpers as we will strip these out in favour of the universal cookie helper which should itself contain tests.

### Added
- Tests to helper.js.

### Changed
- Test helper directory `tests` > `testHelpers`.


packages/f-searchbox/src/utils/tests/helpers.test.js
```diff
@@ -0,0 +1,145 @@
...
+
+                    // Assert
+                    expect(result).toEqual({
+                        Apophis: '99942',
+                        isNearEath: true
+                    });
+                });
+            });
+
+            describe('AND the string provided is invalid i.e contains invalid JSO syntax', () => {
```
AlixRichards: There is a tiny typo here ^

AlixRichards (approved):

packages/f-searchbox/src/utils/tests/helpers.test.js
kevinrodrigues: Just spotted that earlier. Had already pushed it up :)

simonsandersje (approved):

PR #542, opened 2020-12-14
f-checkout@0.25.0 Add support for auth token

pkedziora:
Added authToken property that is used to pass the jwt api authentication token to call APIs.
The token is stored in vuex.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/f-checkout/src/components/Checkout.vue
```diff
@@ -124,6 +124,12 @@ export default {
             type: Number,
             required: false,
             default: 1000
+        },
+
+        authToken: {
+            type: String,
+            required: false,
```
kevinrodrigues: Do you need to define this line if it's not `required`?

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -221,6 +232,7 @@ export default {
             'getCheckout',
             'postCheckout'
         ]),
+        ...mapMutations('checkout', ['updateAuth']),
```
kevinrodrigues: Ideally we should call an action that in turn invokes a mutation. (Action called in the component > mutation invoked in the store)

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -193,10 +199,15 @@ export default {
...
             await this.loadCheckout();
+        },
+
+        authToken () {
+            this.updateAuth({ authToken: this.authToken });
         }
     },
 
     async mounted () {
+        this.updateAuth({ authToken: this.authToken });
```
kevinrodrigues: Should be an action call. You can pass that a payload in the same way and then the store will commit the mutation instead.

packages/f-checkout/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.25.0
+------------------------------
+*December 14, 2020*
+
+### Changed
```
kevinrodrigues: Minor `### Added`

pkedziora: Fixed

packages/f-checkout/src/components/Checkout.vue

packages/f-checkout/src/components/Checkout.vue

packages/f-checkout/src/components/Checkout.vue
pkedziora: Probably not, but I think it making it explicit improves readability. Also other non required props set it explicitly

kevinrodrigues (dismissed):

packages/f-checkout/src/components/Checkout.vue
kevinrodrigues: Might be worth reviewing it at some point later, by default they're false (if setting a default) so it might have just been added to the others unintentionally. (Other components aren't doing this)

DamianMullins: `required` attribute is not needed here if you are setting a default value. We use one or the other in all of our components.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -193,10 +199,15 @@ export default {
     watch: {
         async checkoutUrl () {
             await this.loadCheckout();
+        },
+
+        authToken () {
+            this.setAuthToken({ authToken: this.authToken });
```
DamianMullins: If you are only passing a single value then you can pass it directly rather than as an object property.

DamianMullins (approved):

packages/f-checkout/src/store/checkout.module.js
```diff
@@ -112,6 +124,11 @@ export default {
             state.isFulfillable = isFulfillable;
             state.notices = notices;
             state.messages = messages;
+        },
+
+        updateAuth: (state, { authToken }) => {
+            state.authToken = authToken;
+            state.isLoggedIn = !!authToken;
```
kevinrodrigues, line 131: Would be good to split this out into it's own mutation/action in future (the `setAuthToken` could call that mutation in your case), also other components can set a logged in/out state individually in future.

kevinrodrigues (approved):

packages/f-checkout/src/store/checkout.module.js
DamianMullins: Maybe convert it into a getter if it depends on the `authToken` value?

DamianMullins: Or maybe not if the state can be set individually as you say @kevinrodrigues.

kevinrodrigues: Yup that's a better idea!

kevinrodrigues: I think both ways could potentially work, i think either way it would still need setting individually tho.

PR #546, opened 2020-12-15
f-checkout@0.31.0 - Storybook component tests

JoshuaNg2332:
v0.31.0
*December 23, 2020*

### Changed
- Updated skipped tests to work with Storybook component tests.

### Added
- New `clearCheckoutForm` function added to clear the fields in tests in the `f-checkout-component.js` file.


packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -37,8 +37,8 @@ describe('f-checkout component tests', () => {
         // Waiting for route here, so we can grab redirect url and show form submits.
```
minrwhite, line 37: Should we be asserting something here?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -62,7 +62,7 @@ describe('f-checkout component tests', () => {
         expect(CheckoutComponent.userNoteMaxCharacterCount()).toEqual('200');
     });
 
-    it.skip('should enable a user to submit without adding a note', () => {
+    it('should enable a user to submit without adding a note', () => {
```
minrwhite, line 65: And should we assert in this test also?

packages/f-checkout/test/specs/component/f-checkout-collection.component.spec.js
```diff
@@ -1,18 +1,15 @@
 import forEach from 'mocha-each';
-import DemoControls from '../../../test-utils/component-objects/demo-controls';
 import CheckoutComponent from '../../../test-utils/component-objects/f-checkout.component';
 
 describe('f-checkout "collection" component tests', () => {
-      beforeEach(() => {
+    before(() => {
```
minrwhite, line 5: Can I ask, why the change from 'beforeEach()'?

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
JoshuaNg2332: We're not asserting anything here yet as we don't have the URLs to test this just yet.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
JoshuaNg2332: Same as above. We need the URLs to assert stuff here.

packages/f-checkout/test-utils/component-objects/f-checkout-selectors.js
```diff
@@ -1,3 +1,5 @@
+export const KNOB_BUTTON = '[id$="tabbutton-knobs"]';
+export const KNOB_CHECKOUT_DROPDOWN = '[name="Checkout Url"]';
```
SachaZvetelman, line 2: I'd add a `data-test-id` to this element and use it instead of the name.

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -18,11 +20,13 @@ const checkoutComponent = () => $(CHECKOUT_COMPONENT);
 
 const orderTimeDropdown = () => $(ORDER_TIME_DROPDOWN);
 const orderTimeDropdownOptions = () => $$(ORDER_TIME_DROPDOWN_OPTIONS);
```
SachaZvetelman, line 22: I know you haven't modified this line, but is this `$$` correct?

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -47,15 +51,20 @@ const fields = {
...
         typeError: () => $(FIELDS.addressPostcode.typeError)
-    }, 
+    },
     userNote: {
         input: () => $(FIELDS.userNote.input),
         error: ''
     }
 };
 
+exports.changeToCollectionMethod = () => {
```
SachaZvetelman: I'd personally make this function more generic so it will also work to change to `delivery` as well if we needed. Not a must-have right now, tho.

packages/f-checkout/test/specs/component/f-checkout-collection.component.spec.js
```diff
@@ -21,18 +18,12 @@ describe('f-checkout "collection" component tests', () => {
...
 
     forEach(['addressLine1', 'addressLine2', 'addressCity', 'addressPostcode'])
-    .it.skip('address fields should not exist', field => {
-        // Arrange
-        DemoControls.selectCheckoutMethod('collection');
-
+    .it('address fields should not exist', field => {
         // Assert
-        expect(CheckoutComponent.doesInputFieldExist(field)).toBe(false);
+        expect(CheckoutComponent.isFieldVisible(field)).toBe(false);
```
SachaZvetelman: I'm a bit confused between the name of the test and the assertion. A field could exist without being visible.
The assertion probably should be to check the element exists, rather than if it's visible or not. 😄

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
```diff
@@ -2,14 +2,14 @@ import forEach from 'mocha-each';
 import CheckoutComponent from '../../../test-utils/component-objects/f-checkout.component';
 
 describe('f-checkout "delivery" component tests', () => {
-    beforeEach(() => {
+    before(() => {
```
SachaZvetelman, line 5: Same as above from Ben.

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
```diff
@@ -18,17 +18,17 @@ describe('f-checkout "delivery" component tests', () => {
...
+    it('should display the mandatory fields', () => {
         // Assert
         expect(CheckoutComponent.isFieldDisplayed('mobileNumber')).toBe(true);
     });
 
-    it.skip('should display a "mobileNumber" error message when a number format is incorrectly entered', () => {
+    it('should display a "mobileNumber" error message when a number format is incorrectly entered', () => {
         // Arrange
         const addressDetails = {
             mobileNumber: '+4512345678911'
```
SachaZvetelman, line 35: I know this was added on a previous PR, but this is not necessarily a wrong number format. I believe this would be valid in Denmark, so we should either change the test name, the value of `mobileNumber` or something else.
For example, we could be existing that for UK, it should display an error if the number doesn't start with `+44`, although you need to be careful since a valid UK number without `+44` is still...valid, hehe. Feel free to address this in a separate PR.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -48,8 +48,8 @@ describe('f-checkout component tests', () => {
         expect(CheckoutComponent.getOrderTimeOptionText(2)).toBe('Monday 00:30');
     });
 
-    it.skip('should prevent user from writing a note of over 200 characters', () => {
-        // Arrange 
+    it('should prevent user from writing a note of over 200 characters', () => {
```
SachaZvetelman, line 51: I don't remember, so this comment may be irrelevant, but do we show an error message saying the max limit is 200 characters for notes when users reach it? If we do, we should have a test for that.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -37,8 +37,8 @@ describe('f-checkout component tests', () => {
         // Waiting for route here, so we can grab redirect url and show form submits.
     });
 
-    it.skip('should display times in ascending order, with default text "As soon as possible" showing first', () => {
-        // Act 
+    it('should display times in ascending order, with default text "As soon as possible" showing first', () => {
+        // Act
         CheckoutComponent.selectOrderTime('As soon as possible');
```
SachaZvetelman, line 42: More of a general question for all of us to think about rather than a specific comment on this PR.
Going forward, how are we planning to test each tenant shows the right text? This test would automatically fail for non-English tenants. Not a big concern now, since it's the UK first, but I'd like to at least think about how we want to tackle this.
cc @Festivejelly, @xander-marjoram, @siggerzz, @LTurns.

SachaZvetelman (changes requested):

packages/f-checkout/test/specs/component/f-checkout-collection.component.spec.js
SachaZvetelman: One more tiny comment on this. We should stick to the same naming convention for our functions. It should be `doesFieldExist` rather than `doesInputFieldExist`, right? (again, not your PR, but still).

packages/f-checkout/test/specs/component/f-checkout-collection.component.spec.js
LTurns: this was me actually - 'before' is the mocha equivalent of beforeAll in jest. Since we only want to go to the checkout url and click on collection once, it made more sense to change it so the page didn't keep reloading and switching back and fourth to delivery before every test - some tests were failing/passing inconsistently due to the time spent reloading the page. Hope that helps! :)

packages/f-checkout/test/specs/component/f-checkout-collection.component.spec.js
JoshuaNg2332: Yeah, so I've looked into this and for some reason, and we can't check the field exists because when you try to do that after switching the checkout method, the fields are still visible to the tests in the DOM when switched from delivery to collection. So I've had to use the function `.isDisplayedInViewport()` to get this test to pass. We're trying to test that it can be seen in this instant.

Agreed. This is something from a previous PR that has a sister function next to it called `doesErrorMessageExist`. I can make this change.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
JoshuaNg2332: Just checked this on storybook. This doesn't allow the user to go over 200 characters. So there isn't an error message for that.

packages/f-checkout/test/specs/component/f-checkout-collection.component.spec.js
SachaZvetelman: I'm not sure I fully understood, but it's probably me 😬 
Do you need some time from one of the Engineers to help to see if it's possible to assert the field exists instead, or not really?

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
JoshuaNg2332: @LTurns Correct me if I'm wrong, but is this so that we can allow the selector to choose something within the selector?

packages/f-checkout/test/specs/component/f-checkout-collection.component.spec.js
JoshuaNg2332: I'm not sure. As me and @LTurns tested this for a while to get this working. So in this test, we're checking that the fields are actually **not** visible (as opposed to it not existing in the dom). Happy to jump on a call when you are free to explain my thinking here.

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
LTurns: yep, we use $$ to fetch multiple elements - it treats them like an array so is useful for lists and dropdowns. Means you can write syntax like $$(dropdown)[0] to grab the first option for example 👍

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
LTurns: My first thought would be to do something similar to what we've done here using the storybook knob button - but instead of changing the URL, we change the tenant and then write a test suite to check the text is correct. Not sure if it's the most scalable option though if we have 20 tenants, but useful for testing these in isolation with storybook. Happy to look into doing a spike for this in future 👍

packages/f-checkout/test/specs/component/f-checkout-collection.component.spec.js
JoshuaNg2332: Ok, so some derping around and got the original, `doesFieldExist`(formally `doesInputFieldExist` 😉 ) test working as expected. So rather than ditching the visibility test, will add this as it's own seperate test so it actually tests for visibility on the page as opposed to visible in the DOM or it existing.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
Preet-Sandhu: Looked at this with Lizzie. Would be good to do a spike and explore different options.

packages/f-checkout/test-utils/component-objects/f-checkout-selectors.js
Preet-Sandhu: +1

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
JoshuaNg2332: Ticket created for this.

packages/f-checkout/test-utils/component-objects/f-checkout-selectors.js
JoshuaNg2332: @SachaZvetelman Do you know how to add `data-test-id` to a storybook element? This isn't a `data-test-id` for the component but for storybook itself and we're not sure you can do that as it's sorta the same as adding a `data-test-id` to a webpage not hosted by us.

JoshuaNg2332: So looking through the documentation, we couldn't find anything with regards to setting attributes on the generated HTML so we believe this can't be done.

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
JoshuaNg2332: Agreed with the above and chatting to Lizzie about this earlier, we could have the tenants called into the component.js file and the tests referring to the different tenants based on the locale setting. A spike would probably the best to explore options.

siggerzz: My preference would be to avoid asserting on translation strings, and instead assert on something else instead. (perhaps that the dropdown option has a `value` attribute of 'ASAP'? (This will require some changes to the dropdown vue component though).

We can then test translations via visual regression instead as it'll be less expensive / maintenance. From my own experience in other repos, I've just found asserting on text to be fragile.

Happy to look at this in the new year though 👍 


(I'm going now, bye! 🏃‍♂️💨)

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -47,15 +51,27 @@ const fields = {
...
 };
 
+/**
+ * @description
+ * Changes checkout page to reflect checkout method to either delivery or collection depending on index given.
+ *
+ * @param {Number} index The index of the drop down option
+ */
+
+exports.changeCheckoutPage = index => {
```
siggerzz: I'd rename this to `changeCheckoutMethod` just to make it a bit clearer, then do `knobCheckoutDropdown().selectByText(text);` instead.

Then in the spec, you'll call it as: 

`CheckoutComponent.changeCheckoutMethod('Delivery');`

packages/f-checkout/test/specs/component/f-checkout.component.spec.js
JoshuaNg2332: I've created a ticket to initiate a spike for this on the board. 👍🏼

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
JoshuaNg2332: Looked into this and we can't call it by `'Delivery'` as the text used uses, `/checkout-delivery.json`. Would you want that as the name?

DamianMullins (dismissed):

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -87,6 +103,23 @@ exports.populateCheckoutForm = addressInfo => {
     fields.addressPostcode.input().setValue(addressInfo.postcode);
     fields.userNote.input().setValue(addressInfo.note);
 };
+
+exports.clearCheckoutForm = () => {
+    exports.waitForCheckoutComponent();
+    fields.mobileNumber.input().setValue([' ', "\uE003"]);
```
SachaZvetelman: I'd extract `setValue([' ', "\uE003"])` into a separate function that can be reused across the board. Otherwise, it's not really clear what we're doing here for those who are unaware of the issue.
Similarly, I'd extract that unicode backspace into a `const` that shows what it is 🙂

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -47,15 +51,27 @@ const fields = {
...
 };
 
+/**
+ * @description
+ * Changes checkout page to reflect checkout method to either delivery or collection depending on index given.
+ *
+ * @param {Number} index The index of the drop down option
+ */
+
+exports.changeCheckoutMethod = index => {
```
SachaZvetelman: Is it possible to pass `delivery` or `collection` instead? Since people might not know which index belongs to each of them. Alternatively, let's add something like `index 0 = delivery, index 1 = collection` to the JSDoc above.

packages/f-checkout/test/specs/component/f-checkout-collection.component.spec.js
```diff
@@ -1,38 +1,35 @@
...
+        CheckoutComponent.changeCheckoutMethod(1);
         browser.switchToFrame(0);
         CheckoutComponent.waitForCheckoutComponent();
     });
 
-    it.skip('should display "mobileNumber" error message when collection method is set', () => {
+    it('should display "mobileNumber" error message when collection method is set and number is incorrect', () => {
         // Arrange
-        DemoControls.selectCheckoutMethod('collection');
+        const incorrectMobileNumber = {
```
SachaZvetelman: We're calling this `addressDetails` in other tests, maybe we should stick to the same.

packages/f-checkout/test-utils/component-objects/f-checkout.component.js
JoshuaNg2332: Ben suggested a similar thing with passing `delivery` or `collection` and using a `.selectByText` above but I've looked into this and the text i pass would have be `/checkout-delivery.json` as thats the text used in storybook. I'm not to sure of an elegant way of implementing this. 🤔 
Just looked into labelling the knobs and broke my local ver of storybook 😅  in the process. I want to revisit this in a separate PR. In the meantime, I'll add this comment into the JSDoc as recommended.

SachaZvetelman: Well, you could always do something like:
```
exports.changeCheckoutMethod = method => {
    const file = `checkout-${method}.json`;
    ...
}
```

ashleynolan (approved):

packages/f-checkout/test/specs/component/f-checkout-delivery.component.spec.js
```diff
@@ -2,55 +2,56 @@ import forEach from 'mocha-each';
...
         };
 
         // Act
         CheckoutComponent.populateCheckoutForm(addressDetails);
-        CheckoutComponent.submit();
+        CheckoutComponent.goToPayment();
 
         // Assert
         expect(CheckoutComponent.isFieldErrorDisplayed('mobileNumber')).toBe(true);
     });
```
Preet-Sandhu, line 44: Is this test duplicated above? 

it('should display "mobileNumber" error message when collection method is set and number is incorrect', () =>

JoshuaNg2332: This is supposed to be a seperate test that tests the formatting of a number. So a number has been entered, but not a supported format (e.g. not a Tenant phone code we support). The other test above tests the error message appears if the field is blank.

ashleynolan (changes requested):
Marking this as do not merge. This PR will need refactoring once the folder structure is changed (probably easier to make a copy of the changes you have made and re-apply them to the new locations, as merging will likely give you a bunch of conflicts)

siggerzz (approved):

JoshuaNg2332:
This PR is being closed as a new PR was opened to address the changes to the Fozzie folder structure that this PR needed to incorporate. 
This PR is now defunct with an updated version of it found [here](https://github.com/justeat/fozzie-components/pull/562).

PR #547, opened 2020-12-16
WIP:f-content-cards@v3.0.0-beta.1 - first attempt after POC of creating a scopeSlot base cards component …

matthewhardern:
As future projects are going to require us to build additional cards and changes into the f-content-cards library we have decided to go down a new route with content cards and instead now have multiple layouts that provide different stylistic choice for their use case. Example here is that we will have a layout for the For you section that will be different to the post order section. 

Why are we doing this? Easier integration for the developers consuming the component, less props and configuration means easier integration. Second is bundle size the more we stuff into this component the more bloated and heavy it gets, the aim is to reduce bundle size when being used, so that only what is needed ends up in the consuming application. Third single responsibility, this makes sure for example that Post order layout does not need to be affected by changes in For you layout.

The change should also make implementing new layouts a lot easier, this being the reason for the render-less component, it provides a base framework to design a new layout around. 

This WIP is the first of a few phases to get this new change into a place that can be fully utilised. This is demonstrated by the importing of `FourYouLayout.vue` into `ContentCards.vue` . This for now will be like this, although redundant will enable us to quickly swap this version 2 out for version 3. A later PR and a new phase will modify -f-content-cards to act as a layout library exporting multiple layouts, where the layout that is required can be imported. This in turn will remove  `ContentCards.vue`  and the auto install of the component. 

Apologies for the size of the PR, as it’s the first round it could not really be helped, plus a lot is just refactoring work and moving tests around.


packages/f-content-cards/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-content-cards",
   "description": "Fozzie Content Cards",
-  "version": "2.2.0-beta.9",
+  "version": "3.0.0-beta.1",
```
ashleynolan, line 4: QQ – will v2 be released prior to this change? Just wondering when beta v2 stops and beta v3 begins 😄

packages/f-content-cards/src/components/BaseContentCards.vue
```diff
@@ -0,0 +1,313 @@
+<script>
```
ashleynolan, line 1: If this component is intended to just be script, I'd probably add a comment at the top just describing what it's for and that that is the case. Guessing this is to reuse common methods between layouts?

packages/f-content-cards/src/components/layouts/ForYouLayout.vue
```diff
@@ -0,0 +1,161 @@
+<template>
```
ashleynolan, line 1: I think if you're going down the route of having these defined layouts, you definitely will need to document how these layouts are structured somewhere (as looking from the outside, I have no idea how the `ForYouLayout` is different from the `PostOrderLayout`).

Also, is there any eventuality that one of these layouts may be needed elsewhere? Say the `PostOrderLayout` and a future Homepage layout are exactly the same – will we end up with duplicated layouts, or trying to reuse one layout in a place where the name doesn't match?

packages/f-content-cards/src/components/BaseContentCards.vue
DamianMullins, line 1: Does this file need to be a `.vue` file? A pure JS module might be a better approach here?

packages/f-content-cards/src/components/layouts/_tests/PostOrderLayout.test.js
```diff
@@ -0,0 +1,18 @@
+describe('PostOrderLayout', () => {
+    allure.feature('Post Order Layout');
+    // it('should emit a custom content callback when mounted', () => {
```
DamianMullins, line 3: Is this file yet to be implemented?

packages/f-content-cards/package.json
matthewhardern: V2 has been released now, so yes I will rebase from v2.2.0.

packages/f-content-cards/src/components/BaseContentCards.vue
matthewhardern: @DamianMullins made a fair point could just be a js script. This essentially provides common methods around the retrieval of cards from braze. Now I could have used a mixin, but I am not a fan for things like this, they hide away what is going on, not great when dealing with critical parts such as pulling in data. They also have the issue of naming collapses, running the risk that anyone implementing a layout could accidentally override a key function in the system.

matthewhardern: agreed will make the change

packages/f-content-cards/src/components/layouts/_tests/PostOrderLayout.test.js
matthewhardern: no post order layout has yet to be implemented, this is a start from the POC. I can remove, but helpful when trying to understand what we are doing.

packages/f-content-cards/src/components/layouts/ForYouLayout.vue
matthewhardern: @ashleynolan as discussed separately, docs will be provided to explain use cases. The naming is open for debate, I can agree with your points I will go away and have a think on this one.

matthewhardern:
closing due to change in thinking

PR #548, opened 2020-12-16
f-globalisation@1.0.0 - Adding support for dateTimeFormats.

SachaZvetelman:
v1.0.0
*December 17, 2020*

### Added
- Support for `dateTimeFormats`, ensuring backwards compatibility.

### DIFF
--- a/services/f-globalisation/package.json
+++ b/services/f-globalisation/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-globalisation",
   "description": "Fozzie Globalisation – A utility which wires up vue-i18n within your Smart Component",
-  "version": "0.3.0",
+  "version": "1.0.0",
   "main": "dist/f-globalisation.umd.min.js",
   "files": [
     "dist"
--- a/services/f-globalisation/package.json
+++ b/services/f-globalisation/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-globalisation",
   "description": "Fozzie Globalisation – A utility which wires up vue-i18n within your Smart Component",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "main": "dist/f-globalisation.umd.min.js",
   "files": [
     "dist"
--- a/services/f-globalisation/src/mixins/globalisation.mixin.vue
+++ b/services/f-globalisation/src/mixins/globalisation.mixin.vue
@@ -26,13 +26,8 @@ export default {
         setupLocale (locale, applyLocale = false) {
             const localeConfig = this.tenantConfigs[locale];
 
-            const messages = localeConfig.messages || localeConfig; // Backwards compatibility if a tenant file only has messages without the new structure.
-
-            this.$i18n.setLocaleMessage(locale, messages);
-
-            if (localeConfig.dateTimeFormats) {
-                this.$i18n.setDateTimeFormat(locale, localeConfig.dateTimeFormats);
-            }
+            this.$i18n.setLocaleMessage(locale, localeConfig.messages);
+            this.$i18n.setDateTimeFormat(locale, localeConfig.dateTimeFormats);
 
             if (applyLocale) {
                 this.$i18n.locale = locale;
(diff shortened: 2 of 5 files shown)


services/f-globalisation/src/mixins/globalisation.mixin.vue
```diff
@@ -25,7 +25,14 @@ export default {
     methods: {
         setupLocale (locale, applyLocale = false) {
             const localeConfig = this.tenantConfigs[locale];
-            this.$i18n.setLocaleMessage(locale, localeConfig);
+
+            const messages = localeConfig.messages || localeConfig; // Backwards compatibility if a tenant file only has messages without the new structure.
+
+            this.$i18n.setLocaleMessage(locale, messages);
+
+            if (localeConfig.dateTimeFormats) {
```
BanksySan: Is it valid for `dateTimeFormats` to be missing from the `localeConfig`?

services/f-globalisation/src/mixins/tests/globalisation.mixin.test.js
```diff
@@ -15,7 +15,12 @@ const defaultData = {
             test: 'Test message (EN)'
         },
         [ALTERNATIVE_LOCALE]: {
```
BanksySan, line 22: Why do we need the `[Foo_Bar]` syntax here?  `ABC_DEF` is a valid JavaScript name.

BanksySan (dismissed):

BanksySan:
> I'd like some opinions on this PR. Given it's a rather new service, I'm more inclined towards actually making it a major release rather than a minor with backwards compatibility, to simplify the code and tests. But I did it this way just in case someone is against that idea 🙂

I prefer a major release rather than having soon to be obsolete code in there.

services/f-globalisation/src/mixins/tests/globalisation.mixin.test.js
SachaZvetelman: If you don't use the brackets, your object's property will be the variable name. Using the brackets, it evaluates the variable so your object's property will be the variable's value.

xander-marjoram (dismissed):

DamianMullins:
@SachaZvetelman What are `dateTimeFormats` and where are they used?

SachaZvetelman:
@DamianMullins, `vue-i18n` supports `dateTimeFormats` so that you can customise your datetimes based on your tenant. In this package, `f-globalisation`, we pretty much setup `vue-i18n` dynamically, so we want to make sure it also supports this `dateTimeFormats` functionality 🙂 
In terms of its usage, we'll start using it in `f-checkout`, for example.

DamianMullins (dismissed):
@SachaZvetelman Thanks, I haven't used them before, look forward to seeing them in the wild!

services/f-globalisation/src/mixins/tests/globalisation.mixin.test.js
BanksySan: That's it.  Thanks.

services/f-globalisation/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-globalisation",
   "description": "Fozzie Globalisation – A utility which wires up vue-i18n within your Smart Component",
-  "version": "0.2.0",
+  "version": "1.0.0",
```
DamianMullins, line 4: Any idea why we have packages in a `services` folder rather than the `packages` folder? And why you would add to one and not the other?

BanksySan (dismissed):

services/f-globalisation/package.json
SachaZvetelman: This is something we introduced with @tommcclean recently, to differentiate between visual and usable components (`packages`) and services these components will use (`services`).
And the idea is to solve common problems with these services (such as i18n). Makes sense? @DamianMullins

xander-marjoram (dismissed):

DamianMullins (dismissed):

BanksySan (dismissed):

xander-marjoram (approved):

SachaZvetelman:
Merging since the only change after the 3 approvals was the CHANGELOG date.

PR #551, opened 2020-12-21
f-checkout@0.27.0 - Retrieving fulfilment information from a separate endpoint/file.

SachaZvetelman:
v0.27.0
*December 21, 2020*

### Added
- Retrieving fulfilment information from a separate endpoint/file.

### Removed
- Watchers for Storybook. Replaced it with an object `:key` containing the different props.


packages/f-checkout/src/components/tests/Selector.test.js
```diff
@@ -143,12 +211,55 @@ describe('Selector', () => {
...
+            // Assert
+            expect(selectionChangedSpy).toHaveBeenCalledWith('As soon as possible');
+        });
+
+        it('should not call `selectionChanged` when there are no fulfilment times', () => {
+            // Arrange & Act
+            const selectionChangedSpy = jest.spyOn(Selector.methods, 'selectionChanged');
+
+            shallowMount(Selector, {
+                store: createStore({ ...defaultState }),
```
FayeCarter: Do you need `defaultState` here or is it added here? https://github.com/justeat/fozzie-components/pull/551/files#diff-01e360fd0493be83e7b73038652b2a4d1f3d7d39dd323d44dedb1ea6a2cbf3b5R63

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -200,14 +205,18 @@ export default {
             await this.loadCheckout();
         },
 
+        async checkoutAvailableFulfilmentUrl () {
```
DamianMullins: ~~Is this used anywhere? It appears to match the new prop you added?~~

I should expand the file before commenting — can see it's part of a watcher 😄

minrwhite: Are we expecting this to change post-mount on an integrated instance of the component? If this is to support the prop changing in StoryBook only, you can change the story to include the url prop as part of the value passed to `:key` to avoid having to set a watcher within the component itself.

packages/f-checkout/src/store/checkout.module.js
```diff
@@ -84,6 +91,30 @@ export default {
             const response = await axios.post(url, data, config);
         },
 
+        /**
+         * Get the fulfilment details from the backend and update the state.
+         *
+         * @param {Object} commit - Automatically handled by Vuex to be able to commit mutations.
+         * @param {Object} payload - Parameter with the different configurations for the request.
+         */
+        getAvailableFulfilment: async ({ commit }, payload) => {
```
DamianMullins: If you're not using the `payload` variable anywhere in this action you could restructure the variables in place? 

```suggestion
        getAvailableFulfilment: async ({ commit }, { url, tenant, timeout }) => {
```

packages/f-checkout/src/store/checkout.module.js
```diff
@@ -108,7 +139,7 @@ export default {
                 state.customer.mobileNumber = customer.phoneNumber;
             }
 
-            state.fulfilment.times = fulfilment.times;
+            state.fulfilment.time = fulfilment.time;
```
DamianMullins, line 136: nested object properties in Vuex store state are a bit of an anti-pattern as they can cause issues with reactivity - typically state data should be normalised and flattened if possible i.e. `state.customerFirstName` etc.

I can see you haven't introduced this as part of this PR though so might be one for discussion outside of this PR?

packages/f-checkout/src/components/Selector.vue
```diff
@@ -31,25 +37,37 @@ export default {
...
+                times.push(formattedTime);
+            });
+
+            return times;
+        }
+    },
+
+    mounted () {
+        if (this.fulfilmentTimes.length > 0) {
+            this.selectionChanged(this.fulfilmentTimes[0]);
```
matthewhardern: are fulfillment times ordered / are you intending to set the first available fulfillment time here? If so is there anyway that `this.fulfilmentTimes[0]` may not be the first available fulfillment time?

packages/f-checkout/src/store/checkout.module.js
SachaZvetelman: Good one. Yes, let's take this outside this PR, but something worth discussing, for sure.

packages/f-checkout/src/components/Selector.vue
```diff
@@ -31,25 +37,37 @@ export default {
...
+
+                times.push(formattedTime);
+            });
+
+            return times;
+        }
+    },
+
+    mounted () {
+        if (this.fulfilmentTimes.length > 0) {
```
minrwhite: Could there be a possible race condition here between the mounting of this component and the fetching of fulfilment times from the service, meaning that if the fulfilment times are not initially available, no default value is set?

packages/f-checkout/src/components/Selector.vue
SachaZvetelman: If we get an empty response from the API, or not even empty, but no available fulfilment times and `isAsapAvailable` set to `false`, this could potentially fail. Given I have no clue how possible that is, but I reckon it can happen, I decided to be extra cautious.
Unless someone is against this, hehe.

matthewhardern: Not against just asking if the response is ordered and if not should we be checking if the time in first array element is not the earliest available slot? Because if it is not, for example `['19:00', '17:00', '18:00']` then we may be setting a later time when an earlier one is available?

SachaZvetelman: Good question, @matthewhardern.
@pkedziora, do you know if the API will return these in order? I have the feeling I've heard somewhere they are, but it would be good to confirm. If we don't know, I'll mark this with a `TODO` so we check this in Jan when everyone is back.

pkedziora: @SachaZvetelman I don't know 100% but I think it's reasonable to assume they will be in order.

packages/f-checkout/src/components/Checkout.vue
```diff
@@ -195,19 +200,9 @@ export default {
...
-    watch: {
-        async checkoutUrl () {
-            await this.loadCheckout();
-        },
-
-        authToken () {
-            this.setAuthToken(this.authToken);
-        }
-    },
-
```
pkedziora: Any reason why this section is removed?

SachaZvetelman: Yes. I believe these were introduced to support Storybook changing the URLs and values, correct?
I've replaced that with a `:key` in the story file, so we don't need these watchers. But if there's a scenario I'm missing, please let me know.

pkedziora: @SachaZvetelman this will be needed by CoreWeb as well. Both checkoutUrl and authToken can change any time after the component is mounted

SachaZvetelman: Kept the `authToken` one for now.

pkedziora (dismissed):

FayeCarter (approved):

packages/f-checkout/stories/Checkout.stories.js
```diff
@@ -33,13 +35,17 @@ export const CheckoutComponent = () => ({
...
+            default: select('Available Fulfilment Url', [checkoutAvailableFulfilmentUrl], checkoutAvailableFulfilmentUrl)
+        },
         authToken: {
             default: text('Auth token', '')
         }
     },
     store: new Vuex.Store({}),
-    template:
-        '<vue-checkout :checkoutUrl="checkoutUrl" :locale="locale" :authToken="authToken" />'
+    template: '<vue-checkout'
```
DamianMullins, line 46: Minor thing, and can be done in another PR, but you could use template strings here which would allow you to set each prop on a new line?

DamianMullins (approved):

PR #553, opened 2020-12-22
f-checkout@v0.28.0 - Add tests to `Address` component

FayeCarter:
### Added
- Tests for `Address` component.


packages/f-checkout/src/components/tests/Address.test.js
```diff
@@ -0,0 +1,281 @@
...
+                $v
+            })
+        });
+
+        // Assert
+        expect(wrapper.exists()).toBe(true);
+    });
+
+    describe('computed ::', () => {
+        const isFieldEmptySpy = jest.spyOn(Address.methods, 'isFieldEmpty');
```
SachaZvetelman, line 106: You can get some test pollution here. You probably want to put this in a `beforeEach`, and also do:
```
afterEach(() => {    
  jest.clearAllMocks();
});
```

packages/f-checkout/src/components/tests/Address.test.js
```diff
@@ -0,0 +1,281 @@
...
+                    propsData,
+                    provide: () => ({
+                        $v
+                    })
+                });
+
+                // Assert
+                expect(isFieldEmptySpy).toHaveBeenCalledWith('postcode');
+            });
+        });
```
SachaZvetelman: Could this be simplified and placed in an `it.each`?

packages/f-checkout/src/components/tests/Address.test.js
```diff
@@ -0,0 +1,281 @@
...
+
+        describe('isAddressPostcodeValid ::', () => {
+            it('should return `false` if postcode field has been touched, is not Valid and is not empty', async () => {
```
SachaZvetelman: Ultra nit: `valid` instead of `Valid`.

packages/f-checkout/src/components/tests/Address.test.js
```diff
@@ -0,0 +1,281 @@
...
+                })
+            });
+
+            it('should return `false` if `field` has not been touched and input is required', async () => {
+                // Act
+                $v.addressValidations[field].$dirty = false;
+                $v.addressValidations[field].required = true;
+
+                // Assert
+                expect(wrapper.vm.isFieldEmpty('line1')).toEqual(false);
```
SachaZvetelman: You can use `field` here instead of `line1`.

packages/f-checkout/src/components/tests/Address.test.js
```diff
@@ -0,0 +1,281 @@
...
+        // Assert
+        expect(wrapper.exists()).toBe(true);
+    });
+
+    describe('computed ::', () => {
+        const isFieldEmptySpy = jest.spyOn(Address.methods, 'isFieldEmpty');
+        let wrapper;
+
+        describe('isAddressLine1Empty ::', () => {
+            it('should call `isFieldEmpty` with argument `line1`', async () => {
```
DamianMullins: I don't think you need the `async` keyword for these tests as you're not `await`ing anything?

packages/f-checkout/src/components/tests/Address.test.js
FayeCarter: Ooh, nice. I'll give it a go

packages/f-checkout/src/components/tests/Address.test.js
```diff
@@ -0,0 +1,281 @@
...
+                    store: createStore(),
+                    localVue,
+                    propsData,
+                    provide: () => ({
+                        $v
+                    }),
+                    computed: {
+                        isAddressPostcodeEmpty () {
+                            return false;
+                        }
```
DamianMullins: You can use an arrow function here to keep things concise. I think we were looking to use the shorthand format for computed properties in unit tests to keep them consistent.

```suggestion
                        isAddressPostcodeEmpty: () => false
```

Same for the tests below.

DamianMullins (dismissed):

packages/f-checkout/src/components/tests/Address.test.js
```diff
@@ -0,0 +1,200 @@
...
+                expect(wrapper.vm.isAddressPostcodeValid).toEqual(expected);
+            });
+        });
+    });
+
+    describe('methods ::', () => {
+        describe('isFieldEmpty ::', () => {
+            const field = 'line1';
+
+            const wrapper = shallowMount(Address, {
```
SachaZvetelman: This should be in a `beforeEach`.

DamianMullins (dismissed):

DamianMullins (approved):

SachaZvetelman (approved):

PR #558, opened 2021-01-04
f-checkout@0.35.0 - vuex mutation for data update

dandel10n:
### Changed
- Replaced direct state manipulations with mutations and actions

### Added
- Tests to cover new methods, actions and mutations

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/Address.vue
+++ b/packages/components/organisms/f-checkout/src/components/Address.vue
@@ -6,7 +6,8 @@
                 {{ $t('labels.addressGroup') }}
             </legend>
             <form-field
-                v-model="fulfilment.address.line1"
+                :value="fulfilment.address.line1"
+                @input="updateAddressLine1"
                 :class="$style['c-address-formField']"
                 name="address-line-1"
                 :label-text="$t('labels.line1')"
@@ -24,7 +25,8 @@
             </form-field>
 
             <form-field
-                v-model="fulfilment.address.line2"
+                :value="fulfilment.address.line2"
+                @input="updateAddressLine2"
                 :class="$style['c-address-formField']"
                 name="address-line-2"
                 :label-text="$t('labels.line2')"
@@ -33,7 +35,8 @@
         </fieldset>
 
         <form-field
-            v-model="fulfilment.address.city"
+            :value="fulfilment.address.city"
+            @input="updateAddressCity"
             name="address-city"
             :label-text="$t('labels.city')"
             :has-error="isAddressCityEmpty">
@@ -47,7 +50,8 @@
         </form-field>
 
         <form-field
-            v-model="fulfilment.address.postcode"
+            :value="fulfilment.address.postcode"
+            @input="updateAddressPostcode"
             name="address-postcode"
             :label-text="$t('labels.postcode')"
             :has-error="!isAddressPostcodeValid">
@@ -72,7 +76,7 @@ import ErrorMessage from '@justeat/f-error-message';
 import '@justeat/f-error-message/dist/f-error-message.css';
 import FormField from '@justeat/f-form-field';
 import '@justeat/f-form-field/dist/f-form-field.css';
-import { mapState } from 'vuex';
+import { mapState, mapActions } from 'vuex';
 
 export default {
     components: { FormField, ErrorMessage },
@@ -115,6 +119,9 @@ export default {
     },
 
     methods: {
+        ...mapActions('checkout', [
+            'updateFulfilmentAddress'
+        ]),
         /*
         * Returns true if `field` has been touched and if it is still empty
         * The $dirty boolean changes to true when the user has focused/lost
@@ -122,6 +129,29 @@ export default {
         */
         isFieldEmpty (field) {
             return this.$v.addressValidations[field].$dirty && !this.$v.addressValidations[field].required;
+        },
+        /*
+        * Dispatches map action `updateFulfilmentAddress` to update input fields values in vuex
+        */
+        updateAddressLine1 (value) {
+            this.updateFulfilmentAddress({
+                "line1": value
+            })
+        },
+        updateAddressLine2 (value) {
+            this.updateFulfilmentAddress({
+                "line2": value
+            })
+        },
+        updateAddressCity (value) {
+            this.updateFulfilmentAddress({
+                "city": value
+            })
+        },
+        updateAddressPostcode (value) {
+            this.updateFulfilmentAddress({
+                "postcode": value
+            })
         }
     }
 };
--- a/packages/components/organisms/f-checkout/src/components/Address.vue
+++ b/packages/components/organisms/f-checkout/src/components/Address.vue
@@ -138,16 +138,19 @@ export default {
                 "line1": value
             })
         },
+
         updateAddressLine2 (value) {
             this.updateFulfilmentAddress({
                 "line2": value
             })
         },
+
         updateAddressCity (value) {
             this.updateFulfilmentAddress({
                 "city": value
             })
         },
+
         updateAddressPostcode (value) {
             this.updateFulfilmentAddress({
                 "postcode": value
--- a/packages/components/organisms/f-checkout/src/components/Address.vue
+++ b/packages/components/organisms/f-checkout/src/components/Address.vue
@@ -7,13 +7,13 @@
             </legend>
             <form-field
                 :value="fulfilment.address.line1"
-                @input="updateAddressDetails('line1', $event)"
                 :class="$style['c-address-formField']"
                 name="address-line-1"
                 :label-text="$t('labels.line1')"
                 label-style="inline"
                 is-grouped
-                :has-error="isAddressLine1Empty">
+                :has-error="isAddressLine1Empty"
+                @input="updateAddressDetails('line1', $event)">
                 <template #error>
                     <error-message
                         v-if="isAddressLine1Empty"
@@ -26,20 +26,20 @@
 
             <form-field
                 :value="fulfilment.address.line2"
-                @input="updateAddressDetails('line2', $event)"
                 :class="$style['c-address-formField']"
                 name="address-line-2"
                 :label-text="$t('labels.line2')"
                 is-grouped
-                label-style="inline" />
+                label-style="inline"
+                @input="updateAddressDetails('line2', $event)" />
         </fieldset>
 
         <form-field
             :value="fulfilment.address.city"
-            @input="updateAddressDetails('city', $event)"
             name="address-city"
             :label-text="$t('labels.city')"
-            :has-error="isAddressCityEmpty">
+            :has-error="isAddressCityEmpty"
+            @input="updateAddressDetails('city', $event)">
             <template #error>
                 <error-message
                     v-if="isAddressCityEmpty"
@@ -51,10 +51,10 @@
 
         <form-field
             :value="fulfilment.address.postcode"
-            @input="updateAddressDetails('postcode', $event)"
             name="address-postcode"
             :label-text="$t('labels.postcode')"
-            :has-error="!isAddressPostcodeValid">
+            :has-error="!isAddressPostcodeValid"
+            @input="updateAddressDetails('postcode', $event)">
             <template #error>
                 <error-message
                     v-if="isAddressPostcodeEmpty"
@@ -134,7 +134,7 @@ export default {
         * Dispatches map action `updateFulfilmentAddress` to update input fields values in vuex
         */
         updateAddressDetails (key, event) {
-            this.updateFulfilmentAddress({ [key]: event })
+            this.updateFulfilmentAddress({ [key]: event });
         }
     }
 };
--- a/packages/components/organisms/f-checkout/src/components/Address.vue
+++ b/packages/components/organisms/f-checkout/src/components/Address.vue
@@ -133,8 +133,8 @@ export default {
         /*
         * Dispatches map action `updateFulfilmentAddress` to update input fields values in vuex
         */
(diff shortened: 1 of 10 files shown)


packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -265,15 +265,9 @@ describe('Checkout', () => {
...
-                const state = {
-                    ...defaultState,
-                    serviceType: CHECKOUT_METHOD_DELIVERY
-                };
-
-                // Act
+                // Arrange and Act
                 const wrapper = shallowMount(VueCheckout, {
-                    store: createStore(state),
+                    store: createStore({...defaultState, serviceType: CHECKOUT_METHOD_DELIVERY}),
```
dandel10n: `serviceType` is `CHECKOUT_METHOD_DELIVERY` in defaultState but it is specifically added here both for visibility and in case it is changed in default values later

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -284,15 +278,9 @@ describe('Checkout', () => {
...
-                const state = {
-                    ...defaultState,
-                    serviceType: CHECKOUT_METHOD_COLLECTION
-                };
-
-                // Act
+                // Arrange and Act
                 const wrapper = shallowMount(VueCheckout, {
-                    store: createStore(state),
+                    store: createStore({...defaultState, serviceType: CHECKOUT_METHOD_COLLECTION}),
```
dandel10n: `serviceType` is `CHECKOUT_METHOD_DELIVERY` in defaultState but it is specifically added here both for visibility and in case it is changed in default values later

FayeCarter (dismissed):

packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -115,13 +119,42 @@ export default {
...
         * The $dirty boolean changes to true when the user has focused/lost
         * focus on the input field.
         */
         isFieldEmpty (field) {
             return this.$v.addressValidations[field].$dirty && !this.$v.addressValidations[field].required;
+        },
+        /*
+        * Dispatches map action `updateFulfilmentAddress` to update input fields values in vuex
+        */
+        updateAddressLine1 (value) {
```
SachaZvetelman: Very minor, but if you change `value` to `line1`, you can then simplify your call to: `this.updateFulfilmentAddress({ line1 });`

packages/components/organisms/f-checkout/src/components/tests/Address.test.js
```diff
@@ -145,5 +145,83 @@ describe('Address', () => {
                 expect(wrapper.vm.isFieldEmpty(field)).toEqual(false);
             });
         });
+
+        describe('updateAddressLine1', () => {
+            it('should be called with new input value on user input', async () => {
+                // Arrange
+                jest.clearAllMocks();
```
SachaZvetelman: Can we extract this to an `afterEach` instead?

packages/components/organisms/f-checkout/src/components/tests/Address.test.js
```diff
@@ -145,5 +145,83 @@ describe('Address', () => {
...
+
+                // Assert
+                expect(updateAddressLine1Spy).toHaveBeenCalledWith(newValue);
+            })
+        });
+
+        describe('updateAddressCity', () => {
+            it('should be called with new input value on user input', async () => {
+                // Arrange
+                jest.clearAllMocks();
```
SachaZvetelman: Same as above.

packages/components/organisms/f-checkout/src/components/tests/Address.test.js
```diff
@@ -145,5 +145,83 @@ describe('Address', () => {
                 expect(wrapper.vm.isFieldEmpty(field)).toEqual(false);
             });
         });
+
```
SachaZvetelman, line 148: We are missing a test for `updateAddressLine2`, right?

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -517,7 +524,24 @@ describe('Checkout', () => {
...
+                            ...defaultState,
+                            serviceType: CHECKOUT_METHOD_DELIVERY,
+                            fulfilment: {
+                            ...defaultState.fulfilment,
+                            address: {
+                                ...defaultState.fulfilment.address,
+                                city: ''
+                            }
+                        }}),
+                    i18n,
```
SachaZvetelman: Indentation seems a bit off here.

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -528,7 +552,24 @@ describe('Checkout', () => {
...
+                            ...defaultState,
+                            serviceType: CHECKOUT_METHOD_DELIVERY,
+                            fulfilment: {
+                            ...defaultState.fulfilment,
+                            address: {
+                                ...defaultState.fulfilment.address,
+                                postcode: ''
+                            }
+                        }}),
+                    i18n,
```
SachaZvetelman: And here.

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -540,7 +581,21 @@ describe('Checkout', () => {
...
+                            ...defaultState,
+                            serviceType: CHECKOUT_METHOD_DELIVERY,
+                            fulfilment: {
+                            ...defaultState.fulfilment,
+                            address: {
+                                ...defaultState.fulfilment.address,
+                                postcode: '?!hdb-se'
+                            }
+                        }}),
+                    i18n,
```
SachaZvetelman: And here.

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -555,7 +610,21 @@ describe('Checkout', () => {
...
+                            ...defaultState,
+                            serviceType: CHECKOUT_METHOD_DELIVERY,
+                            fulfilment: {
+                            ...defaultState.fulfilment,
+                            address: {
+                                ...defaultState.fulfilment.address,
+                                postcode: 'EC4M 7R'
+                            }
+                        }}),
+                    i18n,
```
SachaZvetelman: Here :D

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -950,6 +1019,27 @@ describe('Checkout', () => {
...
+                    localVue,
+                    propsData
+                });
+                const newNumber = '+447111111112'
+
+                // Act
+                await wrapper.find('[data-test-id="formfield-mobile-number-input"]').setValue(newNumber);
+                await wrapper.vm.$nextTick();
+                // Assert
```
SachaZvetelman, line 1146: Nit: add empty line, please 🙂

packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -38,7 +38,12 @@ const defaultState = {
     isLoggedIn: false
 };
 
-const { updateState, updateAuth, updateAvailableFulfilment } = CheckoutModule.mutations;
+const { 
+    UPDATE_STATE,
+    UPDATE_AUTH,
+    UPDATE_AVAILABLE_FULFILMENT_TIMES,
+    UPDATE_FULFILMENT_ADDRESS,
+    UPDATE_MOBILE_NUMBER } = CheckoutModule.mutations;
```
SachaZvetelman:
```
const { 
    UPDATE_STATE,
    UPDATE_AUTH,
    UPDATE_AVAILABLE_FULFILMENT_TIMES,
    UPDATE_FULFILMENT_ADDRESS,
    UPDATE_MOBILE_NUMBER 
} = CheckoutModule.mutations;

File: packages/components/organisms/f-checkout/src/components/Address.vue (hunk 3 above)
Comment (kevinrodrigues) [inline packages/components/organisms/f-checkout/src/components/Address.vue]: You could have one method here and pass a second param for the property i.e "Line1, Line2, city".

Comment (SachaZvetelman) [reply packages/components/organisms/f-checkout/src/components/Address.vue]: This ☝️

File: packages/components/organisms/f-checkout/src/components/tests/Address.test.js (hunk 4 above)
Comment (dandel10n) [reply packages/components/organisms/f-checkout/src/components/tests/Address.test.js]: I added it here because there is a before each function with the wrapper on line 109 for other tests and I needed to specify the wrapper here after spyOn() so to be safe added clearAllMocks to start with. Does it make sense?

File: packages/components/organisms/f-checkout/src/components/Address.vue (hunk 3 above)
Comment (dandel10n) [reply packages/components/organisms/f-checkout/src/components/Address.vue]: @kevinrodrigues I am not sure how to call this method then. At the moment it is called on every input field like this `:value="fulfilment.address.line1" @input="updateAddressLine1"` etc

Comment (kevinrodrigues) [reply packages/components/organisms/f-checkout/src/components/Address.vue]: You should be able to pass along the `$event` as the first param then in the second param pass along Line1 etc You will then have access to pass along just that property to the call for `updateFulfilmentAddress`

e.g `:value="fulfilment.address.line1" @input="updateAddressLine($event, 'Line1')"`

Method call:

```javascript
updateAddressLine(event, value) {
  this.updateFulfilmentAddress({
                value
  })
}
```
Not tested it but something along those lines should reduce it down to one method call. Might also be worth making the method call to a generic name as it it currently tied to the thing it is doing, e.g UpdateAddressLineX, maybe `updateAddressDetails`?

packages/components/organisms/f-checkout/src/components/tests/Address.test.js
SachaZvetelman: Hmmm, I'm not sure I understand why extracting this into an `afterEach` would impact negatively :)

packages/components/organisms/f-checkout/src/components/tests/Address.test.js
dandel10n: Fixed

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
dandel10n: Done

packages/components/organisms/f-checkout/src/components/tests/Address.test.js
dandel10n: @SachaZvetelman I rewrote this so now should be all good :)

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/src/components/Address.vue
zlatinivanov-je: Passing in `$event` at the time of trigger, will give access to the element e.g. `$event.target` and you would not even need the second param. 

```javascript
updateAddress(e) {
  this.updateFulfilmentAddress({ e.target.name: e.target.value })
}
```

zlatinivanov-je:
What is the reason for this change? Swapping the `model` approach with "manual" state update.

dandel10n:
> What is the reason for this change? Swapping the `model` approach with "manual" state update.

@zlatinivanov-je You should never manipulate the state directly.  As state should only be mutated by mutations in the module. Here is [docs ref](https://vuex.vuejs.org/guide/forms.html) for this.

packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -115,13 +119,22 @@ export default {
...
         * focus on the input field.
         */
         isFieldEmpty (field) {
             return this.$v.addressValidations[field].$dirty && !this.$v.addressValidations[field].required;
+        },
+        /*
+        * Dispatches map action `updateFulfilmentAddress` to update input fields values in vuex
+        */
+        updateAddressDetails (key, event) {
+            this.updateFulfilmentAddress({ [key]: event })
```
kevinrodrigues: I think this will pass the whole event object to this key now? I think what you want is: `[key]: key` now?

kevinrodrigues: Will this not pass the whole event object now? I think you will need to supply the value here from the event possibility?

SachaZvetelman (dismissed):

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/src/components/Address.vue
dandel10n: The event in this case is a string because form-field components [emits the input](https://github.com/justeat/fozzie-components/blob/cae0417e2cc188b47a30a0176b3d491aa7d0cf12/packages/components/atoms/f-form-field/src/components/FormField.vue#L216) event and passes along only the value:
```
updateValue (event) {
}
```

dandel10n: @zlatinivanov-je because of the above :point_up_2: your suggestion unfortunately wouldn't work

kevinrodrigues: Is it worth renaming this param to `value` then? Just so it's a little clear that the input value is being passed along like this? so instead of `[key]: event` it can be `[key]: value`

dandel10n: Sure, will do it now

FayeCarter (approved):

kevinrodrigues (approved):

PR #560, opened 2021-01-04
f-header@v4.4.0 - Updated feature tests with selector.js file

LTurns:
Copied these changes over from my old PR to avoid merge conflicts with the new folder structure - 

v4.4.0
*January 7, 2021*

### Changed
- Updated Webdriver test-utils to include `selector` file.
- Refactored component tests.

I've updated the before hook to take in a new storybook URL which can then be edited to include add-ons. Happy to discuss this if people prefer the original way, but it seemed to make testing slightly easier

### DIFF
--- a/packages/components/organisms/f-header/CHANGELOG.md
+++ b/packages/components/organisms/f-header/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v4.4.0
 ------------------------------
-*January 6, 2021*
+*January 7, 2021*
 
 ### Changed
 - Updated Webdriver test-utils to include `selector` file.
--- a/packages/components/organisms/f-header/CHANGELOG.md
+++ b/packages/components/organisms/f-header/CHANGELOG.md
@@ -3,7 +3,6 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-
 v4.4.0
 ------------------------------
 *January 4, 2021*
--- a/packages/components/organisms/f-header/CHANGELOG.md
+++ b/packages/components/organisms/f-header/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v4.4.0
 ------------------------------
-*January 4, 2021*
+*January 6, 2021*
 
 ### Changed
 - Updated Webdriver test-utils to include `selector` file.
--- a/packages/components/organisms/f-header/CHANGELOG.md
+++ b/packages/components/organisms/f-header/CHANGELOG.md
@@ -4,8 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-Latest (add to next release)
+v4.4.0
 ------------------------------
+*January 4, 2021*
+
+### Changed
+- Updated Webdriver test-utils to include `selector` file.
+- Refactored component tests.
+
 *December 30, 2020*
 
 ### Changed
--- a/packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
+++ b/packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
@@ -28,12 +28,11 @@ const navigation = {
     } 
 }
 const mobileNavigationBar = () => $(MOBILE_NAVIGATION_BAR);
-const mobileOffersIcon = () => {
-    return navigation.offers.icon().filter(element => element.getAttribute('class').includes('u-showBelowMid'));
-};
-const webOffersIcon = () => {
-    return navigation.offers.icon().filter(element => element.getAttribute('class').includes('u-showAboveMid'));
-};
+const mobileOffersIcon = () => 
+    navigation.offers.icon().filter(element => element.getAttribute('class').includes('u-showBelowMid'));
+    
+const webOffersIcon = () =>
+    navigation.offers.icon().filter(element => element.getAttribute('class').includes('u-showAboveMid'));
 
 // Functions
 
--- a/packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
+++ b/packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
@@ -31,8 +31,11 @@ const mobileNavigationBar = () => $(MOBILE_NAVIGATION_BAR);
 const mobileOffersIcon = () => {
     return navigation.offers.icon().filter(element => element.getAttribute('class').includes('u-showBelowMid'));
 };
+const webOffersIcon = () => {
+    return navigation.offers.icon().filter(element => element.getAttribute('class').includes('u-showAboveMid'));
+};
 
-//Functions
+// Functions
 
 exports.waitForHeader = () => headerComponent().waitForExist();
 exports.isFieldLinkDisplayed = fieldName => navigation[fieldName].link().isDisplayedInViewport();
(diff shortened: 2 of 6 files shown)


packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
```diff
@@ -1,57 +1,58 @@
...
+const headerComponent = () => $(HEADER_COMPONENT);
+
+// Logo
+const headerLogo = () => $(HEADER_LOGO);
+
+// Navigation
+const navigation = {
+    offers: {
+        link: () => $(NAVIGATION.offers.link), 
+        icon: () => $$(NAVIGATION.offers.link)
```
xander-marjoram, line 18: Should this be `icons` rather than `icon`?

siggerzz: I don't think so, there's only one offer icon in f-header 👍🏼

siggerzz (dismissed):

packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
```diff
@@ -1,57 +1,58 @@
...
+        link: () => $(NAVIGATION.delivery.link), 
+    }, 
+    userAccount: {
+        link: () => $(NAVIGATION.userAccount.link), 
+    } 
+}
+const mobileNavigationBar = () => $(MOBILE_NAVIGATION_BAR);
 const mobileOffersIcon = () => {
-    return offersLink().filter(element => element.getAttribute('class').includes('u-showBelowMid'));
+    return navigation.offers.icon().filter(element => element.getAttribute('class').includes('u-showBelowMid'));
```
xander-marjoram: I would recommend using `element.classList.contains('u-showBelowMid')` to prevent a false positive.

If you do `element.getAttribute('class').includes('u-showBelowMid')` that would also return `true` for an element with the class `u-showBelowMidTEST`

packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
```diff
@@ -1,57 +1,58 @@
...
-};
+exports.waitForHeader = () => headerComponent().waitForExist();
+exports.isFieldLinkDisplayed = fieldName => navigation[fieldName].link().isDisplayedInViewport();
+exports.isLogoDisplayed = () => headerLogo().isDisplayedInViewport();
+exports.isMobileNavigationBarVisible = () => mobileNavigationBar().isDisplayedInViewport();
 
 exports.isMobileOffersIconDisplayed = () => {
     const element = mobileOffersIcon();
 
     return element.length === 1 && element[0].isDisplayedInViewport();
```
xander-marjoram, line 47: Is there a test case for `element.length > 1`?

packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
```diff
@@ -1,57 +1,58 @@
...
+const mobileNavigationBar = () => $(MOBILE_NAVIGATION_BAR);
 const mobileOffersIcon = () => {
-    return offersLink().filter(element => element.getAttribute('class').includes('u-showBelowMid'));
+    return navigation.offers.icon().filter(element => element.getAttribute('class').includes('u-showBelowMid'));
 };
 
-const navigationOffersLink = () => {
-    return offersLink().filter(element => element.getAttribute('class').includes('u-showAboveMid'));
-};
+//Functions
```
xander-marjoram: Nit: space after `//`

packages/components/organisms/f-header/CHANGELOG.md
```diff
@@ -3,9 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-
-Latest (add to next release)
+v4.4.0
 ------------------------------
+*January 4, 2021*
```
xander-marjoram: Nit: maybe update to January 6 for if this gets merged tomorrow

packages/components/organisms/f-header/test/specs/component/f-header.component.spec.js
```diff
@@ -1,85 +1,68 @@
...
-
-        // Act
-        // resize to mobile
-        browser.setWindowSize(500, 500);
-
-        expect(HeaderComponent.isNavigationOffersLinkDisplayed()).toBe(false);
-        expect(HeaderComponent.isMobileOffersIconDisplayed()).toBe(true);
-
+    forEach(['offers', 'help', 'delivery', 'userAccount'])
+    .it.skip('should show navigation fields when mobile-navigation has been opened', field => {
```
xander-marjoram, line 77: Should these still be skipped?

packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
LTurns: not at the moment but happy to add one in - we're still seeing the random extra offers icon at the top when in desktop mode so this test might fail until that's sorted

packages/components/organisms/f-header/test/specs/component/f-header.component.spec.js
LTurns: ooh yes! So I noticed that when we we're in mobile mode, the navigation isn't showing 'delivery', but it is on the live site - do you know if this is intentional? Test is currently failing because it's expecting all the fields to be showing :)

packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
LTurns: I can't seem to get this to work sadly - it might not like the $$ syntax from Webdriver? Or maybe I'm just not with it this morning! Tried a few different options and it keeps returning false 🤦

packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
LTurns: Instead of doing `element.length > 1`, I re-added the webOffersIcon function so that we can test which icons should be showing at which window size. That assert isn't passing at the moment due to the random unhidden mobile icon, but hopefully it makes it clearer! :)

packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js
```diff
@@ -1,57 +1,67 @@
...
+        link: () => $(NAVIGATION.delivery.link), 
+    }, 
+    userAccount: {
+        link: () => $(NAVIGATION.userAccount.link), 
+    } 
+}
+const mobileNavigationBar = () => $(MOBILE_NAVIGATION_BAR);
 const mobileOffersIcon = () => {
-    return offersLink().filter(element => element.getAttribute('class').includes('u-showBelowMid'));
+    return navigation.offers.icon().filter(element => element.getAttribute('class').includes('u-showBelowMid'));
```
kevinrodrigues: You can remove the return and the braces here so it reads:

```javascript
const mobileOffersIcon = () => 
   navigation.offers.icon()...

File: packages/components/organisms/f-header/test-utils/component-objects/f-header.component.js [hunk 8]
```diff
@@ -1,57 +1,67 @@
...
 };
-
-const navigationOffersLink = () => {
-    return offersLink().filter(element => element.getAttribute('class').includes('u-showAboveMid'));
+const webOffersIcon = () => {
+    return navigation.offers.icon().filter(element => element.getAttribute('class').includes('u-showAboveMid'));
```
kevinrodrigues: Similar to above.

packages/components/organisms/f-header/test/specs/component/f-header.component.spec.js
```diff
@@ -1,85 +1,85 @@
...
-    it.skip('should display the "For You" icon for mobile', () => {
+    it('should change url when help-link is clicked', () => {
         // Act
-        browser.setWindowSize(500, 500);
+        browser.setWindowSize(1000, 1000);
+        HeaderComponent.clickHelpLink();
 
         // Assert
-        expect(HeaderComponent.isMobileOffersIconDisplayed()).toBe(true);
+        expect(browser.getUrl()).toContain("/help");
```
kevinrodrigues: `'/help'`

siggerzz (dismissed):

FayeCarter (approved):

kevinrodrigues (approved):

PR #561, opened 2021-01-05
f-content-cards@v3.0.0-beta.1  - Update content cards to become a renderless component

matthewhardern:
In order to both reducing the ongoing complexity around implementing and the maintainability of content cards component, after a POC it was decided that we would now move to an approach where content cards purely served as a scoped slot component that provided all the functions around content cards and delivering the cards array, but be agnostic in terms of style. This approach will now enable us to implement different styles of the content cards being displayed without bloating this component. 

The changes included in this PR represent discussion had at the end of 2020 on the approach, removing the previous editions concept of layouts in the component, leaving this now entirely up to the consumer. For an insight into the discussion please see #547 now closed pull request.

I have updated the unit tests accordingly and removed no longer needed tests ( example grouping tests are now gone, as this would be handled by the consumer ).

I will be submitting a future PR with some major changes to the story book stories, I have not included this as part of this PR due to its size. Also attached to this next PR will be an update to the readme.


packages/components/organisms/f-content-cards/src/components/cardTemplates/GroupHeaderCard.vue
```diff
@@ -0,0 +1,30 @@
...
+        {{ title }}
+    </h2>
+</template>
+
+<script>
+export default {
+    name: 'GroupHeaderCard',
+    props: {
+        title: {
+            default: '',
```
kevinrodrigues: Do you need a default if it's required?

packages/components/organisms/f-content-cards/src/components/cardTemplates/GroupHeaderCard.vue
```diff
@@ -0,0 +1,30 @@
...
+            required: true,
+            type: String
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+    .c-contentCards--group-title {
```
kevinrodrigues: Minor but could you tab this in and line break between selectors please.

packages/components/organisms/f-content-cards/src/components/cardTemplates/GroupHeaderCard.vue
matthewhardern: agreed, will change

packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -224,36 +114,13 @@ export default {
         cards (current, previous) {
             this.$emit('get-card-count', current.length);
 
-            if (current.length && (current.length !== previous.length) && !this.groupCards) {
+            if (current.length && (current.length !== previous.length)) {
```
ashleynolan, line 117: Not specific to this line, but quickly scanning this, it's not super obvious what `current` and `previous` represent. Would be worth updating the JSDoc comment to explain what it holds (I'm guessing it a current set of content cards being displayed vs a previous set, but I wouldn't be 100% sure) and I wouldn't know what the context of current compared with previous would be (previous page load?)

ashleynolan (approved):
Looks good – one small comment but other than that LGTM

packages/components/organisms/f-content-cards/src/components/ContentCards.js
matthewhardern: Happy to update the comment, are you ok with that update being in the docs PR following this?

kevinrodrigues (approved):

packages/components/organisms/f-content-cards/src/components/ContentCards.js
ashleynolan: Yeh, absolutely 👍

PR #563, opened 2021-01-05
#trivial - Prevent vuex store from being overridden

tommcclean:
@dandel10n  spotted that the dev tools in storybook showed mutations, but did not show the actual state.

Investigating this showed that we were setting up VUEX state in two places, so the state setup in the Checkout story was being overridden. By making sure state is only setup once we can then see state in the dev tools.

There is one downside to this approach; but we don't have any use case for it right now, so I haven't tried to cover it. 

Because VUEX is being setup by the story of a component; it means that components won't be able to use the same state as other components which might be something we want to do in future. We can tackle that later. I have opted to let the story setup the state, and not do it globally in storybook.

image


dandel10n (dismissed):

kevinrodrigues (dismissed):

ashleynolan (dismissed):

tommcclean:
Brought in the changes from master which fix the broken builds

dandel10n (approved):

ashleynolan (approved):

PR #564, opened 2021-01-05
f-checkout@v0.32.0 - add further tests for `Checkout` methods and computed values

FayeCarter:
### Changed
- Tests for `Checkout` component methods and computed values.

### Added
- `it-IT` tenant file.

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
+++ b/packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
@@ -14,6 +14,33 @@ const localVue = createLocalVue();
 localVue.use(VueI18n);
 localVue.use(Vuex);
 
+const $v = {
+    customer: {
+        mobileNumber: {
+            $dirty: false,
+            isValidPhoneNumber: false
+        }
+    },
+    fulfilment: {
+        address: {
+            city: {
+                $dirty: false,
+                required: true
+            },
+            line1: {
+                $dirty: false,
+                required: false
+            },
+            postcode: {
+                $dirty: false,
+                required: true,
+                isValidPostcode: false
+            }
+        }
+    },
+    $touch: jest.fn()
+};
+
 describe('Checkout', () => {
     allure.feature('Checkout');
     const checkoutUrl = 'http://localhost/checkout';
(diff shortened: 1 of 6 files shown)


xander-marjoram (dismissed):

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -619,38 +754,210 @@ describe('Checkout', () => {
...
+                // Assert
+                expect(submitCheckoutSpy).toHaveBeenCalled();
+            });
+
+            it('should call `handleErrorState` if `submitCheckout` returns an error', async () => {
+                // Arrange
+                const handleErrorStateSpy = jest.spyOn(VueCheckout.methods, 'handleErrorState');
+                const error = new Error('errorMessage');
+
+                submitCheckoutSpy.mockImplementation(async () => { throw error; });
```
kevinrodrigues: Do you need the async here? The mock should be enough to just return the error thrown i would have thought? (Could be wrong tho)

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -619,38 +754,210 @@ describe('Checkout', () => {
...
                     i18n,
                     localVue,
                     propsData
                 });
 
+                // Act
+                wrapper.vm.isValidPhoneNumber();
+
                 // Assert
                 expect(isValidPhoneNumberSpy).toBeCalledWith(defaultState.customer.mobileNumber, i18n.locale);
```
kevinrodrigues: Not your change but Is this call intentional? It's the same as `toHaveBeenCalledWith`

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -619,38 +754,210 @@ describe('Checkout', () => {
...
+                        $v
+                    }
+                });
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+
+                // Assert
+                expect(isFormValidSpy).toBeCalled();
+                expect(getFormValidationStateSpy).toBeCalledWith($v);
```
kevinrodrigues: Similar to the question earlier about `toHaveBeenCalledWith`?

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -619,38 +754,210 @@ describe('Checkout', () => {
...
+                    mocks: {
+                        $v
+                    }
+                });
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+
+                // Assert
+                expect(isFormValidSpy).toBeCalled();
```
kevinrodrigues: toHaveBeenCalled?

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -619,38 +754,210 @@ describe('Checkout', () => {
...
+            beforeEach(() => {
+                isFormValidSpy = jest.spyOn(VueCheckout.methods, 'isFormValid');
+                submitCheckoutSpy = jest.spyOn(VueCheckout.methods, 'submitCheckout');
+            });
+
+            afterEach(() => {
+                jest.clearAllMocks();
+            });
+
+            it('should emit `CheckoutFailure` with validation state if form is invalid', async () => {
```
kevinrodrigues, line 832: Maybe for another PR but i think we could split this test out. There's a few expects in this particular IT statement which indicate it could be broken down a little. Or maybe the first two spy checks aren't really needed?

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: Changed it, thanks

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: Yeah it works fine removed, thanks 😄

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
FayeCarter: Ahh good point. I've removed the spy checks 👍

kevinrodrigues (approved):

simonsandersje (approved):

packages/components/organisms/f-checkout/src/components/tests/Checkout.test.js
```diff
@@ -619,38 +754,208 @@ describe('Checkout', () => {
...
+                });
+
+                // Act
+                await wrapper.vm.onFormSubmit();
+
+                // Assert
+                expect(submitCheckoutSpy).toHaveBeenCalled();
+            });
+
+            it('should call `handleErrorState` if `submitCheckout` returns an error', async () => {
```
matthewhardern, line 888: This could lead to brittle tests, as this function emits an event should we not be testing that event is emitted not that the method has been called?

FayeCarter: Hey, I've added separate tests for the `handleErrorState` method here https://github.com/justeat/fozzie-components/pull/564/files/c0878b006e1f5a028798930ae043e48ec38e27d1#diff-0a066c99de868a4e439b79ef7378df68709dee1bcd9f9fe97a6ff67d389ad11dR682 that tests that the event is emitted. Is this OK?

PR #565, opened 2021-01-05
f-searchbox@v4.0.0 beta.13

kevinrodrigues:
Fixed NO & DK from not submitting correctly as the attributes on the component were wrong.` Postcode > Where`

Key Attribute required: 

Screen Shot 2021-01-05 at 1 53 21 PM

Was:

Screen Shot 2021-01-05 at 1 54 06 PM

### DIFF
--- a/packages/components/molecules/f-searchbox/src/components/formElements/FormSearchField.vue
+++ b/packages/components/molecules/f-searchbox/src/components/formElements/FormSearchField.vue
@@ -50,6 +50,20 @@ export default {
     components: {
         FormSearchInnerFieldWrapper
     },
+
+    directives: {
+        /**
+         * Extract object attributes and set them on the input component.
+         *
+         * */
+        setAttributes (element, directive) {
+            for (let i = 0, keys = Object.keys(directive.value); i < keys.length; i++) {
+                const key = keys[i];
+                element.setAttribute(key, directive.value[key]);
+            }
+        }
+    },
+
     props: {
         errorMessage: {
             type: [Boolean, String],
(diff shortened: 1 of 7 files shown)


simonsandersje (dismissed):

packages/components/molecules/f-searchbox/src/components/formElements/FormSearchField.vue
```diff
@@ -50,6 +50,20 @@ export default {
...
         FormSearchInnerFieldWrapper
     },
+
+    directives: {
+        /**
+         * Extract object attributes and set them on the input component.
+         *
+         * */
+        setAttributes (element, directive) {
+            for (let i = 0, keys = Object.keys(directive.value); i < keys.length; i++) {
```
minrwhite: Wouldn't [`Object.entries()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries)`.forEach()` work here?

kevinrodrigues: It would. Ill add that 👍

kevinrodrigues: Fixed.

kevinrodrigues: Double checked storybook by running build and serve locally and works fine with this change.

simonsandersje (approved):

matthewhardern (approved):

PR #566, opened 2021-01-05
fozzie-components@v3.2.0 – Fixing danger checks

ashleynolan:
### Fixed
- Fixes danger package checks.

### DIFF
--- a/dangerfile.js
+++ b/dangerfile.js
@@ -7,6 +7,13 @@ const isGlobalConfigUpdate = bodyAndTitle.includes('#globalconfig'); // turns of
 if (!isTrivial) {
     const failedChangelogs = [];
     const failedVersionBumps = [];
+    const packageDirectories = [
+        'packages/components/atoms/',
+        'packages/components/molecules/',
+        'packages/components/organisms/',
+        'packages/services/',
+        'packages/tools/'
+    ];
     const modifiedFiles = danger.git.modified_files;
 
     // Get an array of files in the root monorepo that have changed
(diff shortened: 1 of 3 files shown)


siggerzz (approved):

minrwhite (approved):

matthewhardern (approved):

dangerfile.js
```diff
@@ -7,14 +7,25 @@ const isGlobalConfigUpdate = bodyAndTitle.includes('#globalconfig'); // turns of
...
 
     // Get an array of files in the root monorepo that have changed
     const modifiedRootFiles = modifiedFiles.filter(filepath => !filepath.startsWith('packages/') && filepath !== 'yarn.lock');
 
     // Get an array of packages that have changed, then make that array unique
     const modifiedPackages = modifiedFiles.filter(filepath => filepath.startsWith('packages/'))
-        .map(filepath => `${filepath.split('/')[1]}/${filepath.split('/')[2]})`);
+        .map(filepath => {
+            const directoryMatch = packageDirectories.filter(subDirectory => filepath.includes(subDirectory));
+            return (filepath.split(directoryMatch)[1]) // splits the string based on the matching directory
```
xander-marjoram, line 26: I think `directoryMatch` will be an array which I expect won't work as an argument of `split`.

ashleynolan: I've tested it on a few PRs (from command line) and works ok.

Imagine it's because it always returns an Array of only one entry, so JS probably does some of it's weird magic when using the array string to split with. I've setup a [JS Fiddle](https://jsfiddle.net/vLko9qrb/) which shows it works in a similar scenario too.

PR #569, opened 2021-01-06
f-content-cards@v2.2.1 - changes to fix grouping headers when 0 cards

matthewhardern:
Minor change to existing group headers on V2.2.0 to make sure the headers only show if sub card count is greater than 0

### DIFF
--- a/packages/f-content-cards/src/components/ContentCards.vue
+++ b/packages/f-content-cards/src/components/ContentCards.vue
@@ -6,7 +6,7 @@
         <template v-if="groupCards">
             <template v-for="({ title, cards: subCards, id: groupId }, groupIndex) in cardsGrouped">
                 <h2
-                    v-if="groupId"
+                    v-if="groupId && subCards.length > 0"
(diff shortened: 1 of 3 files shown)


zlatinivanov-je (approved):

kevinrodrigues (approved):

packages/f-content-cards/src/components/ContentCards.vue
```diff
@@ -6,7 +6,7 @@
         <template v-if="groupCards">
             <template v-for="({ title, cards: subCards, id: groupId }, groupIndex) in cardsGrouped">
                 <h2
-                    v-if="groupId"
+                    v-if="groupId && subCards.length > 0"
```
xander-marjoram, line 9: Minor: Is it worth having this inside a computed property?

matthewhardern: would have to be a method, as subcards is a child of the main group, so would need passing as a variable I would think, although may be wrong

PR #570, opened 2021-01-06
f-checkout@v0.34.1 - Updated mobile number tests

JoshuaNg2332:
v0.34.1
*January 6, 2021*

### Changed
- Mobile number tests naming convention for `Checkout` component to make clear differences in tests.
- Mobile number test for `Checkout` component to an unused number format.
- Updated the location of the mobile number tests.

### DIFF
--- a/packages/components/organisms/f-checkout/test/specs/component/f-checkout.component.spec.js
+++ b/packages/components/organisms/f-checkout/test/specs/component/f-checkout.component.spec.js
@@ -37,7 +37,7 @@ describe('f-checkout component tests', () => {
         // Waiting for route here, so we can grab redirect url and show form submits.
     });
 
-    it('should display a "mobileNumber" error message when an unsupported country code is used in the mobile number', () => {
+    it('should display a "mobileNumber" error message when an unsupported country code is used in the mobile number field', () => {
         // Arrange
         const addressDetails = {
             mobileNumber: '+8112345678911'
--- a/packages/components/organisms/f-checkout/test/specs/component/f-checkout.component.spec.js
+++ b/packages/components/organisms/f-checkout/test/specs/component/f-checkout.component.spec.js
@@ -37,6 +37,34 @@ describe('f-checkout component tests', () => {
         // Waiting for route here, so we can grab redirect url and show form submits.
     });
 
+    it('should display a "mobileNumber" error message when an unsupported country code is used in the mobile number', () => {
+        // Arrange
+        const addressDetails = {
+            mobileNumber: '+8112345678911'
+        };
+
+        // Act
+        CheckoutComponent.populateCheckoutForm(addressDetails);
+        CheckoutComponent.goToPayment();
+
+        // Assert
+        expect(CheckoutComponent.isFieldErrorDisplayed('mobileNumber')).toBe(true);
+    });
+
+    it('should not display a "mobileNumber" error message when a number is formatted with a supported country code', () => {
+        // Arrange
+        const addressDetails = {
+            mobileNumber: '+4412345678911'
+        };
+
+        // Act
+        CheckoutComponent.populateCheckoutForm(addressDetails);
+        CheckoutComponent.goToPayment();
+
+        // Assert
+        expect(CheckoutComponent.isFieldErrorDisplayed('mobileNumber')).toBe(false);
+    });
+
     it('should display times in ascending order, with default text "As soon as possible" showing first', () => {
         // Act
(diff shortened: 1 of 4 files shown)


packages/components/organisms/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -37,6 +37,34 @@ describe('f-checkout component tests', () => {
         // Waiting for route here, so we can grab redirect url and show form submits.
     });
 
+    it('should display a "mobileNumber" error message when an unsupported country code is used in the mobile number', () => {
```
kevinrodrigues: `should display a "mobileNumber" error message when an unsupported country code is used in the mobile number field`

packages/components/organisms/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -37,6 +37,34 @@ describe('f-checkout component tests', () => {
...
+
+        // Act
+        CheckoutComponent.populateCheckoutForm(addressDetails);
+        CheckoutComponent.goToPayment();
+
+        // Assert
+        expect(CheckoutComponent.isFieldErrorDisplayed('mobileNumber')).toBe(true);
+    });
+
+    it('should not display a "mobileNumber" error message when a number is formatted with a supported country code', () => {
```
kevinrodrigues, line 54: May have asked this before so apologies. Do we cover the empty field scenarios too?

JoshuaNg2332: Haha thats ok. Yup. Thats covered already in other tests. 😀

kevinrodrigues (dismissed):

dandel10n (commented):
Question: we have some test to check that mobile number format is correct in checkout.test.js and here in spec. Should they be all in one place?

kevinrodrigues:
> Question: we have some test to check that mobile number format is correct in checkout.test.js and here in spec. Should they be all in one place?

I agree. Ideally we should have it in a single place. It's quite hard to follow what's covered otherwise.

JoshuaNg2332:
> > Question: we have some test to check that mobile number format is correct in checkout.test.js and here in spec. Should they be all in one place?
> 
> I agree. Ideally we should have it in a single place. It's quite hard to follow what's covered otherwise.

My PR is for the component tests that tests that the formats activate the error messages in a browser rather than them being the unit tests. So that's why they are in separate place.

dandel10n:
> > > Question: we have some test to check that mobile number format is correct in checkout.test.js and here in spec. Should they be all in one place?
> > 
> > 
> > I agree. Ideally we should have it in a single place. It's quite hard to follow what's covered otherwise.
> 
> My PR is for the component tests that tests that the formats activate the error messages in a browser rather than them being the unit tests. So that's why they are in separate place.

Does this mean that we can remove checks that error message is shown when format is not right from unit tests? Sorry is my question is silly, I just try to understand. It is not clear for me what we cover by unit tests and what we cover by component tests because it feels like we sometimes test the same thing just in a bit different formats.

kevinrodrigues:
> > > Question: we have some test to check that mobile number format is correct in checkout.test.js and here in spec. Should they be all in one place?
> > 
> > 
> > I agree. Ideally we should have it in a single place. It's quite hard to follow what's covered otherwise.
> 
> My PR is for the component tests that tests that the formats activate the error messages in a browser rather than them being the unit tests. So that's why they are in separate place.

Maybe something to look at in the future but i think unit tests alone should be enough to handle both scenarios here unless i'm missing something? Unit tests should be able to cover both the error messages appearing and field type checks too.

JoshuaNg2332:
> > > > Question: we have some test to check that mobile number format is correct in checkout.test.js and here in spec. Should they be all in one place?
> > > 
> > > 
> > > I agree. Ideally we should have it in a single place. It's quite hard to follow what's covered otherwise.
> > 
> > 
> > My PR is for the component tests that tests that the formats activate the error messages in a browser rather than them being the unit tests. So that's why they are in separate place.
> 
> Maybe something to look at in the future but i think unit tests alone should be enough to handle both scenarios here unless i'm missing something? Unit tests should be able to cover both the error messages appearing and field type checks too.

Agreed. We're setting out a meeting to discuss these boundaries so we can have this discussion.

kevinrodrigues (dismissed):

FayeCarter (dismissed):

siggerzz (approved):

LTurns (approved):

PR #571, opened 2021-01-07
f-searchbox@v4.0.0beta.14

kevinrodrigues:
PR introduces Loqate to searchbox. This covers the initial call to Loqate & displaying results.

### Added
- Loqate core functionality, initial Axios call & display of results.
- Tests to cover Loqate functionality.

### Screenshot
Screen Shot 2021-01-07 at 1 47 36 PM


packages/components/molecules/f-searchbox/src/components/Form.vue
```diff
@@ -143,13 +146,13 @@ export default {
         /**
          * Display API suggestions in component: `form-search-suggestions`:
          *
-         * 1. Service layer should contain `autocomplete`.
+         * 1. Service layer should contain `autocomplete` or `isFullAddressSearchEnabled`.
```
ashleynolan:
```suggestion
         * 1. Service layer should contain `isAutocompleteEnabled ` or `isFullAddressSearchEnabled`.
```

kevinrodrigues: This one is actually correct as it maps to an option called `autocomplete`. I have updated the comment so it's a little clearer.

dandel10n (dismissed):

packages/components/molecules/f-searchbox/src/components/Form.vue
```diff
@@ -30,7 +30,8 @@
                 :is-compressed="isCompressed" />
         </div>
 
-        <form-search-suggestions
+        <component
+            :is="setSuggestionType()"
```
DamianMullins: Do you need the parentheses here?

packages/components/molecules/f-searchbox/src/components/formElements/FormFullAddressSearchSuggestions.vue
```diff
@@ -0,0 +1,199 @@
+<template>
+    <div :class="$style.shell">
+        <button
+            v-for="(item, index) in getAddressItems"
+            :key="index"
```
DamianMullins: Best to give the key a value which is more likely to be unique i.e. ``:key="`${index}_ getAddressItem`"``

packages/components/molecules/f-searchbox/src/components/Form.vue
kevinrodrigues: It was a method but probably should have been a computed prop so updated it now 👍

packages/components/molecules/f-searchbox/src/components/formElements/FormFullAddressSearchSuggestions.vue
kevinrodrigues: Yup, fixed.

packages/components/molecules/f-searchbox/src/components/Form.vue
DamianMullins: Ah yeah, that makes sense 👍🏻

packages/components/molecules/f-searchbox/src/store/searchbox.module.js
```diff
@@ -174,6 +185,27 @@ export default {
...
+         * @param suggestions
+         */
+        [SET_PARTIAL_ADDRESS_SUGGESTIONS]: (state, suggestions) => {
+            state.suggestions = suggestions.Items.map(({
+                Description,
+                Text,
+                Id,
+                Type
+            }) => ({
+                Description,
```
DamianMullins, line 204: You _could` map these into camel case? Or would that cause issues elsewhere?

DamianMullins (dismissed):

packages/components/molecules/f-searchbox/src/store/searchbox.module.js
kevinrodrigues: WIll take a look in the next PR if that's ok. I think i did try that in the old searchbox but it was a while back so can't remember if it worked ok or not :D

dandel10n (dismissed):

DamianMullins (approved):

dandel10n (approved):

zlatinivanov-je:
All good to me. Just missed out on adding a review. 👍

PR #575, opened 2021-01-08
f-searchbox@v4.0.0-beta.15 - Fix accessibility errors.

kevinrodrigues:
### Fixed
- Accessibility warnings for street number input field & Powered by Google alt tag.

### DIFF
--- a/packages/components/molecules/f-searchbox/src/tenants/en-IE.js
+++ b/packages/components/molecules/f-searchbox/src/tenants/en-IE.js
@@ -4,6 +4,7 @@ const copy = {
     fieldPlaceholder: 'E.g. 21 Stonewall Drive, Saggart Abbey, Dublin 24, D24 TF12',
     headlineSubtitle: 'Find restaurants delivering right now, near you',
     headlineTitle: 'Tuck into a takeaway today',
+    streetNumberLabel: 'Location needs to be more specific',
     errors: {
         ADDRESS_EMPTY: 'Please enter a location',
         NO_SUGGESTION_FOUND: 'No suggestions found',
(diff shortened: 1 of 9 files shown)


packages/components/molecules/f-searchbox/src/tenants/en-IE.js
```diff
@@ -4,6 +4,7 @@ const copy = {
     fieldPlaceholder: 'E.g. 21 Stonewall Drive, Saggart Abbey, Dublin 24, D24 TF12',
     headlineSubtitle: 'Find restaurants delivering right now, near you',
     headlineTitle: 'Tuck into a takeaway today',
+    streetNumberLabel: 'Location needs to be more specific',
```
xander-marjoram: Is this only shown when the user has already entered an address? It seems to convey a different meaning than the other translations.

kevinrodrigues: Correct. I noticed that too, at the moment if the street number is required then the error displayed is mapped to the`MORE_SPECIFIC_SEARCH` copy, so i've just reused that for the label as that would be the message displayed when the street number is required.

xander-marjoram (dismissed):

xander-marjoram (approved):

packages/components/molecules/f-searchbox/src/tenants/en-IE.js
kevinrodrigues: Looked and there was a more specific copy available for the street number so have updated 👍

simonsandersje (approved):

PR #578, opened 2021-01-11
#trivial Update component versions

xander-marjoram:
Update components to get around storybook bug.

### DIFF
--- a/packages/components/molecules/f-searchbox/CHANGELOG.md
+++ b/packages/components/molecules/f-searchbox/CHANGELOG.md
@@ -4,7 +4,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-Latest (roll into next release)
+Latest (add to next release)
 ------------------------------
 *January 11, 2021*
 
--- a/packages/components/molecules/f-searchbox/CHANGELOG.md
+++ b/packages/components/molecules/f-searchbox/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 Latest (add to next release)
 ------------------------------
-*January 11, 2021*
+*January 12, 2021*
 
 ### Changed
 - Update axios version for security advisory
--- a/packages/components/molecules/f-searchbox/CHANGELOG.md
+++ b/packages/components/molecules/f-searchbox/CHANGELOG.md
@@ -3,6 +3,15 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+Latest (roll into next release)
+------------------------------
+*January 11, 2021*
+
+### Changed
+- Update axios version for security advisory
+
+
 v4.0.0-beta.16
 ------------------------------
 *January 11, 2021*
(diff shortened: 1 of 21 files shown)


packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -58,16 +67,12 @@ v0.31.0
 
 ### Changed
 - Updated skipped tests to work with Storybook component tests.
+- Updated config for latest `sass-loader`.
```
DamianMullins, line 70: Was this moved because the changes were actually made on the 5th?

packages/components/molecules/f-searchbox/CHANGELOG.md
```diff
@@ -34,7 +43,7 @@ v4.0.0-beta.13
 *January 05, 2021*
 
 ### Fixed
-- DK & NO field options - provides component with custom attributes, 
+- DK & NO field options - provides component with custom attributes,
```
DamianMullins: More of an observation; it looks like the line below should be pulled up onto this one.

DamianMullins (dismissed):

packages/components/organisms/f-checkout/CHANGELOG.md
xander-marjoram: @DamianMullins I moved it because we don't seem to have a convention of having multiple dates for a single version

DamianMullins:
@xander-marjoram not sure if it was caused by this PR but the build is failing due to an integration test failure in f-registration.

packages/components/molecules/f-searchbox/CHANGELOG.md
xander-marjoram: Updated, and fixed some more entries

dandel10n (dismissed):

DamianMullins (dismissed):

packages/components/organisms/f-registration/src/components/Registration.vue
```diff
@@ -34,7 +34,6 @@
                 <form-field
                     v-model="firstName"
                     name="firstName"
-                    data-test-id="input-first-name"
```
xander-marjoram, line 37: These can be added again if necessary but there are data-test-ids inside the form-field component which we can use to access the input element directly rather than its parent.

DamianMullins (dismissed):

packages/components/organisms/f-registration/CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+Latest (add to next release)
+------------------------------
+*January 12, 2021*
+
+### Changed
+- Update axios version for security advisory.
+- Use latest version of `f-button`.
```
minrwhite: This version already seems to be bumped, as there's no change to this line in the package.json.

yarn.lock
```diff
@@ -5584,27 +5557,20 @@ axios-mock-adapter@1.19.0:
...
 
-axios@0.19.2, axios@^0.19.0, axios@^0.19.2:
+axios@0.21.1:
+  version "0.21.1"
+  resolved "https://registry.yarnpkg.com/axios/-/axios-0.21.1.tgz#22563481962f4d6bde9a76d516ef0e5d3c09b2b8"
+  integrity sha512-dKQiRHxGD9PPRIUNIWvZhPTPpl1rf/OxTYKsqKUDjBwYylTvV7SjSHJb9ratfyzM6wCdLCOYLzs73qpg5c4iGA==
+  dependencies:
+    follow-redirects "^1.10.0"
+
+axios@^0.19.0, axios@^0.19.2:
```
minrwhite, line 5567: Do we know where these are still coming in from? What's the result of 'yarn why' for these package versions?

xander-marjoram: The dependencies are `bundlewatch`, which we're using the latest version of, and `chromedriver`, which would be a major version bump so we'll leave this for now.

minrwhite (approved):

dandel10n (approved):

DamianMullins (approved):

PR #593, opened 2021-01-13
DNM - Run build for affected packages

siggerzz:
Just playing around...


siggerzz:
Found a better solution using lerna

PR #607, opened 2021-01-18
#Trivial Correct f-mega-modal README formatting.

BanksySan:
Correct the heading levels as they jumped from level 1 to level 3.

### DIFF
--- a/packages/components/molecules/f-mega-modal/README.md
+++ b/packages/components/molecules/f-mega-modal/README.md
@@ -1,4 +1,3 @@
-
 <div align="center">
   <h1>f-mega-modal</h1>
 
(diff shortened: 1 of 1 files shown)


DamianMullins (dismissed):

kevinrodrigues (dismissed):

packages/components/molecules/f-mega-modal/README.md
```diff
@@ -17,7 +16,7 @@
 
 ## Usage
 
-### Installation
+## Installation
```
ashleynolan, line 19: This looks correct to me as-is. Line 17 is `h2`, and these are then `h3`?

xander-marjoram (changes requested):
I agree with Ash, I think it was correct already.

ashleynolan (dismissed):
Just request comment before it can be merged

packages/components/molecules/f-mega-modal/README.md
BanksySan: Yup.  The **Usage** being a level two confused me.  I've made it level 1.

xander-marjoram (approved):

kevinrodrigues (approved):

packages/components/molecules/f-mega-modal/README.md
ashleynolan: There's already an `h1` up on line 3, so I think it was still correct previously

PR #612, opened 2021-01-18
f-checkout@v0.40.0 - add `Guest` component validations

FayeCarter:
### Added
- `validations` mixin.
- `Guest` component validations.
- `VALIDATIONS` constant.

### Changed
- `updateAddressDetails` and `updateCustomerDetails` store actions.


minrwhite (dismissed):
Good solution for reusing the validation functionality 👍

packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -13,7 +13,7 @@
                 label-style="inline"
                 is-grouped
                 :has-error="isAddressLine1Empty"
-                @input="updateAddressDetails('line1', $event)">
+                @input="updateFulfilmentDetails('Address', 'line1', $event)">
```
dandel10n: Is there a way to show that `updateFulfilmentDetails ` is a method from `validations.mixin` somehow without searching through the whole project? It just a bit confusing when you try to understand where this method came from.

xander-marjoram: I do agree, but it's partly the idea of the mixin to act as if the method was defined as part of the component. I think it's perhaps something we need to get used to, where if it's not listed in the file and isn't explicitly imported then it's probably from a mixin.

xander-marjoram: I suppose one idea is prefixing the method names with `mixin_` or something

dandel10n (dismissed):

packages/components/organisms/f-checkout/src/components/Address.vue
dandel10n: Yeah, I quickly searched for the solution and basically it is what you said @xander-marjoram - we need to get used to it :grin:

packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -91,23 +96,23 @@ export default {
...
         /*
         * Validation methods return true if the validation conditions
         * have not been met and the field has been `touched` by a user.
         * The $dirty boolean changes to true when the user has focused/lost
         * focus on the input field.
         */
-
         isAddressLine1Empty () {
-            return this.isFieldEmpty('line1');
+            return this.isFieldEmpty('address', 'line1');
```
xander-marjoram: Does the casing matter here? In some places it's "Address"

packages/components/organisms/f-checkout/src/mixins/validations.mixin.js
```diff
@@ -0,0 +1,27 @@
...
+        },
+
+        /**
+        * Dispatches action `updateAddressDetails` or `updateCustomerDetails` to update input fields values in Vuex store.
+        *
+        * @param component {string} The component being updated. (`Address` or `Guest`)
+        * @param field {string} The field to be being updated. e.g. `firstName`, `line1`
+        * @param value {string} The new value.
+        */
+        updateFulfilmentDetails (component, field, value) {
```
xander-marjoram: Not sure if I've missed something but was there an issue with just dispatching the action directly?

FayeCarter: I couldn't call them directly on @input because it needed the field and value to be sent.
It worked fine as a method in each component but it was very similar code being written twice

packages/components/organisms/f-checkout/src/components/Address.vue
FayeCarter: Have added suggested `VALIDATIONS` to constants file

packages/components/organisms/f-checkout/src/components/Guest.vue
```diff
@@ -77,18 +78,18 @@ export default {
...
-            return this.isFieldEmpty('guest', 'lastName');
+            return this.isFieldEmpty(VALIDATIONS.guest, 'lastName');
         },
 
         /*
         * Checks that email field is not empty and is a valid email address.
         */
         isEmailValid () {
-            return !this.isFieldEmpty('guest', 'email') && this.$v.guestValidations.email.email;
+            return !this.isFieldEmpty(VALIDATIONS.guest, 'email') && this.$v[VALIDATIONS.guest].email.email;
```
xander-marjoram, line 92: Is `.email.email` correct?

FayeCarter: The first email is the email field validations.
The second email is the Vuelidate validation for emails

packages/components/organisms/f-checkout/src/mixins/validations.mixin.js
FayeCarter: Ignore me, really didn't need it in the end. It's removed now 😆

dandel10n (approved):

minrwhite (approved):

packages/components/organisms/f-checkout/src/components/Guest.vue
ashleynolan: Is it worth adding a comment saying this? Not sure it's obvious (as just saw this myself and wondered what it was)

PR #613, opened 2021-01-19
f-header@4.6.0 - Add country selector for both mobile and desktop view

dandel10n:
New country selector for header:

Desktop view:
Screen Shot 2021-01-18 at 16 58 03
Screen Shot 2021-01-18 at 16 58 37
Screen Shot 2021-01-18 at 17 02 27

Mobile view:
Screen Shot 2021-01-19 at 07 44 14
Screen Shot 2021-01-19 at 07 43 58


packages/components/organisms/f-header/src/components/FlagIcon.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+    <component :is="iconComponent" />
+</template>
+
+<script>
+import {
+    FlagAtRoundIcon as FlagIconAt,
```
DamianMullins, line 7: One thing it would be interesting to keep an eye on is the size of this package following the addition of more icons; I recently looked into why the f-footer is so large for such a small component and it was because of the icons which are all bundled up regardless of whether or not they are used in the current tenant (I tested by removing the icons and it went from ~108kb down to ~9kb which is a massive difference).

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,262 @@
+<template>
+    <li
+        :class="['c-nav-list-item has-sublist', {
+            'open': countrySelectorIsOpen
+        }]"
+        v-on="isBelowMid ? null : { mouseover: openCountrySelector, mouseleave: closeCountrySelector }"
+        @keyup.esc="closeCountrySelector">
+        <a
```
kevinrodrigues: This one should be a button with type set to `button`. `<button type='button'>`

dandel10n: Oh, yes, I just copied it from other nav tabs to start with and forgot to change :+1:

packages/components/organisms/f-header/src/components/FlagIcon.vue
ashleynolan: I wonder if there would be a way of lazy-loading in these icons rather than including them inline, as I imagine it'll be a very small subset of users that will ever hover over or click the country selector. 

Either that, or codesplitting this part of the component so that it loads it in only when it's requested (there's a section in the vue docs on this called [async components](https://vuejs.org/v2/guide/components-dynamic-async.html#Async-Components), that may be worth a look)

packages/components/organisms/f-header/README.md
```diff
@@ -1,7 +1,7 @@
 <div align="center">
 <h1>f-header</h1>
 
-<img width="125" alt="Fozzie Bear" src="../../bear.png" />
+<img width="125px" alt="Fozzie Bear" src="../../bear.png" />
```
ashleynolan: Width should be unitless here I think?

packages/components/organisms/f-header/package.json
```diff
@@ -40,12 +40,13 @@
...
   "dependencies": {
-    "@justeat/f-services": "1.3.0",
-    "@justeat/f-vue-icons": "1.4.0"
+    "@justeat/f-services": "1.3.0"
   },
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.1.1",
-    "@justeat/f-trak": ">=0.6.0"
+    "@justeat/f-trak": ">=0.6.0",
+    "@justeat/f-vue-icons": "2.3.0",
```
ashleynolan: `f-vue-icons` is likely still a dependency rather than a `peerDependency` as it will include the icons at build time?

ashleynolan: May even not need to be included here, as it would be covered by the mono-repo dependencies due to it being run at build time?

packages/components/organisms/f-header/package.json
```diff
@@ -40,12 +40,13 @@
...
+    "@justeat/f-button": "0.4.1"
```
ashleynolan: Same for the button I would think here too –  not sure you need to include it here, as it'll get built into the module at compile time.

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,288 @@
+<template>
+    <li
+        :class="['c-nav-list-item has-sublist', {
+            'open': countrySelectorIsOpen
```
ashleynolan: Should name this `is-open`

```suggestion
            'is-open': countrySelectorIsOpen
```

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,288 @@
...
+        }]"
+        data-test-id="countrySelector-toggle"
+        v-on="isBelowMid ? null : { mouseover: openCountrySelector, mouseleave: closeCountrySelector }"
+        @keyup.esc="closeCountrySelector">
+        <f-button
+            button-type="icon"
+            class="c-nav-list-text c-country-selector-list-item"
```
ashleynolan: `c-country-selector-list-item` doesn't really fit the naming atm.

Looking at the current structure, the base list item (line 2) should have a `c-countrySelector` class on it, and then this would be `c-countrySelector-btn` or something I would imagine?

```suggestion
            class="c-nav-list-text c-countrySelector-btn"
```

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,288 @@
...
+                        class="c-nav-popoverList-header-button"
+                        aria-label="Go back to main menu"
+                        @click.native="closeCountrySelector">
+                        <arrow-icon class="c-nav-popoverList-go-back-icon" />
+                    </f-button>
+
+                    <h3>{{ selectYourCountryText }}</h3>
+                </header>
+
+                <ul class="c-country-selector-list">
```
ashleynolan:
```suggestion
                <ul class="c-countrySelector-list">
```

packages/components/organisms/f-header/package.json
dandel10n: Oh, I wanted to move it to dev dependencies

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,288 @@
...
+                    :country-code="currentCountryKey"
+                    class="c-countrySelector-flag c-countrySelector-flag--current" />
+            </span>
+            <span class='c-countrySelector-title'>
+                {{ selectYourCountryText }}
+            </span>
+        </f-button>
+
+        <div :class="['c-nav-panel', { 'is-visible': countrySelectorIsOpen }]">
+            <div class="c-nav-popoverList c-nav-popoverList--countrySelector">
```
ashleynolan, line 29: Out of interest – does this reuse the styling and functionality of the other popover list? If so, it may be worth either:

1. Extracting out the component if it's similar enough, or
2. extracting out the styles into `common.scss` so that we aren't duplicating styles between here and the other popoverList styling

packages/components/organisms/f-header/src/components/FlagIcon.vue
DamianMullins: I did look at whether it was possible to use `import(` inside of an npm package and whilst it _did_ generate an extra file for the icons that file then contains _all_ of the icons rather than just the ones referenced so I didn't take it any further.

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -206,63 +206,53 @@
...
-                        {{ accountLogout.text }}
+                        {{ copy.accountLogout.text }}
                     </a>
                 </li>
+
+                <country-selector
+                    v-if="showCountrySelector"
+                    ref="countrySelector"
+                    :is-below-mid="isBelowMid"
+                    :current-country-key="copy.currentCountryKey"
```
ashleynolan: Rather than setting these as individual props, I'd probably just pass through the `copy` object so that it's used consistently across components (and it's clear then in the child component that it's coming from `copy.` in the parent)

ashleynolan: This also matches how you've changed `Navigation.vue` to work as well.

ashleynolan (dismissed):
Few comments from me

packages/components/organisms/f-header/package.json
DamianMullins: dev deps should be fine as consuming applications won't need to the icons packages as it's already bundled up.

packages/components/organisms/f-header/src/components/FlagIcon.vue
dandel10n: @DamianMullins @asherstoppard I will add a separate ticket for this under engineering excellence

packages/components/organisms/f-header/src/components/CountrySelector.vue
dandel10n: @ashleynolan it reuses some styles. I can make a separate ticket for this if you don't mind because we have a tight timeline for this to go live as part of "new old homepage". All works as expected so it shouldn't block us from releasing it.

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,287 @@
+<template>
+    <li
+        :class="['c-countrySelector c-nav-list-item has-sublist', {
+            'is-open': countrySelectorIsOpen
+        }]"
+        data-test-id="countrySelector-toggle"
+        v-on="isBelowMid ? null : { mouseover: openCountrySelector, mouseleave: closeCountrySelector }"
```
minrwhite, line 7: Love this way of optionally/conditionally binding event handlers! 👩‍🍳

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,287 @@
...
+        @keyup.esc="closeCountrySelector">
+        <f-button
+            :tabindex="isBelowMid && !navIsOpen ? -1 : 0"
+            button-type="icon"
+            class="c-nav-list-text c-countrySelector-btn"
+            :aria-expanded="!isBelowMid && countrySelectorIsOpen ? 'true' : 'false'"
+            :aria-haspopup="isBelowMid ? false : true"
+            :aria-label="copy.changeCurrentCountry"
+            @click="onCountrySelectorToggle"
+            v-on="isBelowMid && !countrySelectorIsOpen ? { blur: closeNav, focus: openNav } : null">
```
minrwhite: The `isBelowMid && !countrySelectorIsOpen` condition is effectively repeated twice here, as one is the inverse of the other - can you put that into a single `computed` state property?

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,287 @@
...
+        }]"
+        data-test-id="countrySelector-toggle"
+        v-on="isBelowMid ? null : { mouseover: openCountrySelector, mouseleave: closeCountrySelector }"
+        @keyup.esc="closeCountrySelector">
+        <f-button
+            :tabindex="isBelowMid && !navIsOpen ? -1 : 0"
+            button-type="icon"
+            class="c-nav-list-text c-countrySelector-btn"
+            :aria-expanded="!isBelowMid && countrySelectorIsOpen ? 'true' : 'false'"
+            :aria-haspopup="isBelowMid ? false : true"
```
xander-marjoram:
```suggestion
            :aria-haspopup="!isBelowMid"
```

packages/components/organisms/f-header/src/components/Header.vue
```diff
@@ -144,6 +143,9 @@ export default {
 
         headerBackgroundClass () {
             return this.headerBackgroundTheme === 'highlight' ? 'c-header--highlightBg' : '';
+        },
+        showOffersLinkWithContent () {
```
minrwhite, line 149: For consistency with the rest of the computed properties here:
```suggestion

        showOffersLinkWithContent () {
```

packages/components/organisms/f-header/src/components/Header.vue
```diff
@@ -115,6 +110,10 @@ export default {
         userInfoUrl: {
             type: String,
             default: '/api/account/details'
+        },
+        showCountrySelector: {
```
minrwhite, line 115: For consistency:
```suggestion

        showCountrySelector: {
```

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -320,6 +303,10 @@ export default {
         headerBackgroundTheme: {
             type: String,
             default: 'white'
+        },
+        showCountrySelector: {
+            type: Boolean,
+            default: false
```
minrwhite, line 310: Do we need to have a default here, given that this component (Navigation) is only consumed internally to the Header module, and this prop is directly overriden by the prop we're passing the consuming component, or *it's* default if we don't set one?

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,287 @@
...
+        </f-button>
+
+        <div :class="['c-nav-panel', { 'is-visible': countrySelectorIsOpen }]">
+            <div class="c-nav-popoverList c-nav-popoverList--countrySelector">
+                <header class="c-nav-popoverList-header">
+                    <f-button
+                        button-type="icon"
+                        button-size="xsmall"
+                        class="c-nav-popoverList-header-button"
+                        aria-label="Go back to main menu"
```
xander-marjoram: Do we have copy for this for all tenants?

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -320,6 +303,10 @@ export default {
         headerBackgroundTheme: {
             type: String,
             default: 'white'
+        },
+        showCountrySelector: {
```
minrwhite, line 308: For consistency:
```suggestion

        showCountrySelector: {
```

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,287 @@
...
+            this.countrySelectorIsOpen = true;
+        }
+
+    }
+};
+</script>
+
+<style lang="scss">
+
+$contrySelector-icon-color  : $color-secondary;
```
xander-marjoram: Typo in these variable names

packages/components/organisms/f-header/src/components/Header.vue
xander-marjoram, line 149: Do we need any tests around this computed property?

packages/components/organisms/f-header/src/tenants/en-AU.js
```diff
@@ -49,5 +49,8 @@ export default {
     offers: {
         text: 'For you',
         url: '/offers'
-    }
+    },
+    selectYourCountryText: 'Select your country',
+    currentCountryKey: 'au',
+    changeCurrentCountry: 'You are on the AU website, click here to change'
```
minrwhite: Is AU approved copy here?
```suggestion
    changeCurrentCountry: 'You are on the Australian website, click here to change'
```

packages/components/organisms/f-header/src/tenants/en-NZ.js
```diff
@@ -49,5 +49,8 @@ export default {
     offers: {
         text: 'For you',
         url: '/offers'
-    }
+    },
+    selectYourCountryText: 'Select your country',
+    currentCountryKey: 'nz',
+    changeCurrentCountry: 'You are on the NZ website, click here to change'
```
minrwhite: Is NZ approved copy here?
```suggestion
    changeCurrentCountry: 'You are on the New Zealandic website, click here to change'
```

packages/components/organisms/f-header/stories/header.stories.js
```diff
@@ -44,12 +44,24 @@ export const HeaderComponent = () => ({
...
+    template: `
+        <vue-header
+            :userInfoProp="userInfoProp"
+            :showOffersLink="showOffersLink"
+            :showHelpLink="showHelpLink"
+            :locale="locale"
+            :headerBackgroundTheme="headerBackgroundTheme"
+            :showDeliveryEnquiry="showDeliveryEnquiry"
+            :showLoginInfo="showLoginInfo"
+            :showCountrySelector="showCountrySelector" />`
```
minrwhite: It might be necessary to put the locale into the `:key` here to force rerender of the component when changing the locale knob, as we're using it in order to generate initial state in `data()`:
```suggestion
            :showCountrySelector="showCountrySelector"
            :key="locale" />`
```

packages/components/organisms/f-header/src/components/CountrySelector.vue
dandel10n: Hmm, @benwhite-deltas I can't see how to put them into a single property...

packages/components/organisms/f-header/src/components/Navigation.vue
dandel10n: I think, if you don't make prop to be required it is a good practice to specify a default value to be safe if it is removed in future from the parent or you decide to extract this component.

packages/components/organisms/f-header/src/components/CountrySelector.vue
dandel10n: Oh, actually we don't have translations for this. I will add it in english for all the tenants for now and will ask for translations

packages/components/organisms/f-header/src/components/Navigation.vue
minrwhite: That's fair - I mention because we got caught out in one instance where the externally exposed prop was false by default, but the default in the internal component was true - the prop ended up resolving to true when the external prop was not set, essentially overriding the supposed default. Happy to accept as-is based on your reasoning.

packages/components/organisms/f-header/src/tenants/en-AU.js
dandel10n: I took it from a provided translations file

packages/components/organisms/f-header/src/tenants/en-NZ.js
dandel10n: The same, we were provided with these versions for these tenants

packages/components/organisms/f-header/src/components/Navigation.vue
DamianMullins: The [`vue/require-default-prop` lint rule](https://eslint.vuejs.org/rules/require-default-prop.html) will (or at least _should_) complain if you don't provide a default value as it's best practice to provide one, or alternatively, mark the prop as `required`.

packages/components/organisms/f-header/src/components/CountrySelector.vue
minrwhite: Ah, sorry @dandel10n, I was misremembering my truth tables here. I'd hoped you could do something like:
```
            :aria-expanded="isWideAndExpanded ? 'true' : 'false'"
            :aria-haspopup="isBelowMid ? false : true"
            :aria-label="copy.changeCurrentCountry"
            @click="onCountrySelectorToggle"
            v-on="!isWideAndExpanded ? { blur: closeNav, focus: openNav } : null">
```
But actually `!(isBelowMid && !countrySelectorIsOpen)` == `!isBelowMid || countrySelectorIsOpen` (as per DeMorgan's Theorem). Might still be worth turning these separate values into computed properties to help readers of the code understand the semantics of the conditions, e.g. `isWideViewAndSelectorExpanded` and `isNarrowViewAndSelectorRetracted`, for example.

packages/components/organisms/f-header/src/components/Navigation.vue
minrwhite: That `required` tip is good to know, @DamianMullins - given the issues we had before with the For You link I'd err on the side of required props over defaults for components internal to a module, but that might just be my preference.

DamianMullins: I'd only suggest using `required` when you really want users to explicitly set a value for the prop — it could get a little annoying having to set values for things you don't know/care about rather than having a sensible default value. 

I think setting a default value of false here and allowing the user to opt-in themselves seemed like a good way to go.

packages/components/organisms/f-header/src/components/Header.vue
dandel10n: :heavy_check_mark:

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -10,11 +10,11 @@
             :tabindex="isBelowMid && !navIsOpen ? -1 : 0"
             button-type="icon"
             class="c-nav-list-text c-countrySelector-btn"
-            :aria-expanded="!isBelowMid && countrySelectorIsOpen ? 'true' : 'false'"
-            :aria-haspopup="isBelowMid ? false : true"
+            :aria-expanded="countrySelectorIsOpenOnDesktopView ? 'true' : 'false'"
```
minrwhite, line 13: 💯

minrwhite (dismissed):

packages/components/organisms/f-header/src/components/CountrySelector.vue
SachaZvetelman: Would the following work?
`:aria-expanded="countrySelectorIsOpenOnDesktopView"`

DamianMullins: Does the value not have to explicitly be "true" or "false"?

minrwhite: It does IIRC - the element needs to have the aria attribute present at element intiaisation for screen readers to be aware of it and react to it changing, and as per https://vuejs.org/v2/guide/syntax.html#Attributes :

> In the case of boolean attributes, where their mere existence implies true, v-bind works a little differently. In this example:
> 
>       <button v-bind:disabled="isButtonDisabled">Button</button>
>
> If isButtonDisabled has the value of null, undefined, or false, **the disabled attribute will not even be included** in the rendered <button> element.

DamianMullins: Could change this line to `:aria-expanded="countrySelectorIsOpenOnDesktopView.toString()"` instead then?

DamianMullins: Or just leave it as it was 😄

minrwhite (dismissed):

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,294 @@
...
+            type: Boolean,
+            default: false
+        },
+        copy: {
+            type: Object,
+            default: () => ({})
+        },
+        openNav: {
+            type: Function,
+            default: () => (() => {})
```
xander-marjoram: Minor: Would `() => {}` work just as well?

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -409,6 +398,7 @@ export default {
     methods: {
         onNavToggle () {
             this.navIsOpen = !this.navIsOpen;
+            this.$refs.countrySelector.closeCountrySelector();
```
xander-marjoram: Does this throw an error if the country selector is not present?

dandel10n: Good question! I will check now

packages/components/organisms/f-header/src/tenants/nb-NO.js
```diff
@@ -56,8 +56,8 @@ export default {
         url: '/account/logout',
         gtm: 'click_logout'
     },
-    offers: {
-        text: '%OFFERS%',
-        url: '/offers'
-    }
+    selectYourCountryText: 'Velg land',
```
xander-marjoram: Should these be inside their own object, i.e., `countrySelector.selectYourCountryText`?

packages/components/organisms/f-header/src/components/CountrySelector.vue
dandel10n: :heavy_check_mark:

packages/components/organisms/f-header/src/components/Navigation.vue
dandel10n: I wrapped it into a condition:+1:

minrwhite (approved):

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -0,0 +1,294 @@
...
+.c-nav-list-item .c-countrySelector-btn {
+    border-bottom: none;
+}
+
+.c-countrySelector-country {
+    padding: 0;
+    white-space: nowrap;
+    margin-bottom: 0;
+
+    &:hover {
```
kevinrodrigues, line 256: Do we need a focus state here too?

kevinrodrigues (approved):

xander-marjoram (approved):

packages/components/organisms/f-header/src/components/CountrySelector.vue
dandel10n: Default focus styles look fine there I think. I will double check and fix it in next PR if needed

PR #616, opened 2021-01-19
DNM: Run checks against changed packages #globalconfig

siggerzz:
This PR fixes the following:

- Run checks against all packages for `master` builds
- Run checks against changed packages for PR branches
- Skip the build if no packages have changed (root level)
- Fixed storybook deploy job

Build that mimics master functionality - https://app.circleci.com/pipelines/github/justeat/fozzie-components/4015/workflows/ee585ad7-7a31-4417-baba-52f1f64fb8f5

Build that mimics global config - https://app.circleci.com/pipelines/github/justeat/fozzie-components/4017/workflows/53f7368d-aff1-4359-abc3-2ea3cf3e4ccb


.circleci/config.yml
```diff
@@ -14,18 +14,53 @@ commands:
...
+                  # Runs 'npx lerna list --since origin/master'
+                  # Pipes the output to 'wc' (word count)
+                  # Returns true if no packages have changed (output of 0)
+                  # Returns false if packages have changed (output > 0)
+                  # Halts the job if true 
+                  #
+                  # Note: Ideally, this would be in an env var
+                  # but they can't be shared across Circle Jobs so we run this command before every job.
+                - run: |
+                  if echo $( (( $(npx lerna list --since origin/master | wc -l) == 0 )) && echo "true" || echo "false" ); then
```
siggerzz, line 33: Runs 'npx lerna list --since origin/master'
                  Pipes the output to 'wc' (word count)
                  Returns true if no packages have changed (output of 0)
                  Returns false if packages have changed (output > 0)
                  Halts the job if true 
                  
                  Note: Ideally, this would be in an env var
                  but they can't be shared across Circle Jobs so we run this command before every job.

PR #617, opened 2021-01-19
#trivial Reworking f-mega-modal README.md.

BanksySan:
Remove overly verbose wording.
Remove `No-Vue` section as this will be put into a common location.

### DIFF
--- a/packages/components/molecules/f-mega-modal/README.md
+++ b/packages/components/molecules/f-mega-modal/README.md
@@ -1,9 +1,11 @@
-<div align="center">
-  <h1>f-mega-modal</h1>
+<div style="text-align: center">
 
-  <img width="125px" alt="Fozzie Bear" src="../../../../bear.png" />
+# f-mega-modal
 
-  <p>A Vue.js modal component</p>
+<img width="125" alt="Fozzie Bear" src="../../../../bear.png" />
+
+Vue.js modal component
+
 </div>
 
 ---
@@ -13,9 +15,9 @@
 [![Coverage Status](https://coveralls.io/repos/github/justeat/f-mega-modal/badge.svg)](https://coveralls.io/github/justeat/f-mega-modal)
 [![Known Vulnerabilities](https://snyk.io/test/github/justeat/f-mega-modal/badge.svg?targetFile=package.json)](https://snyk.io/test/github/justeat/f-mega-modal?targetFile=package.json)
 
-# Usage
+## Usage
 
-## Installation
+### Installation
 
 ```sh
 yarn add @justeat/f-mega-modal
@@ -25,7 +27,7 @@ yarn add @justeat/f-mega-modal
 npm install @justeat/f-mega-modal
 ```
 
-### Vue Applications
+#### Vue Applications
 
 > Styles have to be imported separately.
 
@@ -40,7 +42,7 @@ export default {
 }
 ```
 
-### Webpack
+#### Webpack
 
 Import the component dynamically to separate the `mega-modal` bundle from the main `bundle.client.js`.
 
@@ -55,9 +57,9 @@ export default {
 }
 ```
 
-## Configuration
+### Configuration
 
-### Props
+#### Props
 
 | Prop  | Type  | Default | Description |
 | ----- | ----- | ------- | ----------- |
@@ -74,7 +76,7 @@ export default {
 | `close-on-blur` | `Boolean` | `true` | Controls whether or not to close the modal when the user clicks outside of the modal. |
 | `close-button-copy` | `String` | `"Close modal"` | Sets the hidden text value for the close button which is used by screen-readers. |
 
-### CSS Classes
+#### CSS Classes
 
 | Class | Description |
 | ----- | ----------- |
@@ -85,27 +87,27 @@ export default {
 | `c-megaModal-document--scrollable` | Can be used to target the modal document element when it is scrollable. |
 | `c-megaModal-closeBtn` | Can be used to target the modal close button element. |
 
-### Events
+#### Events
 
 | Event | Description |
 | ----- | ----------- |
 | `open` | This event is emitted when the modal is opened. |
 | `close` | This event is emitted when the modal is closed. |
 
-## Development
+### Development
 
 ```sh
-$ git clone git@github.com:justeat/fozzie-components.git
-$ cd fozzie-components
-$ yarn
+git clone git@github.com:justeat/fozzie-components.git
+cd fozzie-components
+yarn
 ```
 
-### Running storybook
+#### Running storybook
 
 From the **root** directory run:
 
 ```sh
-$ yarn storybook:serve
+yarn storybook:serve
 ```
 
-Serve storybook at [http://localhost:6006](http://localhost:6006).
+This serves storybook at [http://localhost:6006](http://localhost:6006).
--- a/packages/components/molecules/f-mega-modal/README.md
+++ b/packages/components/molecules/f-mega-modal/README.md
@@ -44,7 +44,7 @@ export default {
 
 #### Webpack
 
-Import the component dynamically to separate the `mega-modal` bundle from the main `bundle.client.js`.
+Import the component dynamically to separate the _mega-modal_ bundle from the main `bundle.client.js`.
 
 ```js
 import '@justeat/f-mega-modal/dist/f-mega-modal.css';
--- a/packages/components/molecules/f-mega-modal/README.md
+++ b/packages/components/molecules/f-mega-modal/README.md
@@ -1,7 +1,7 @@
 <div align="center">
   <h1>f-mega-modal</h1>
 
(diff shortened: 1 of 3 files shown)


DamianMullins (dismissed):

dandel10n (dismissed):

packages/components/molecules/f-mega-modal/README.md
```diff
@@ -1,7 +1,7 @@
 <div align="center">
   <h1>f-mega-modal</h1>
 
-  <img width="125" alt="Fozzie Bear" src="../../bear.png" />
+  <img width="125px" alt="Fozzie Bear" src="../../../../bear.png" />
```
ashleynolan: `width` should be unitless here

packages/components/molecules/f-mega-modal/README.md
```diff
@@ -30,7 +27,7 @@ npm install @justeat/f-mega-modal
 
 ### Vue Applications
 
-You can import it in your Vue SFC like this (please note that styles have to be imported separately):
+> Styles have to be imported separately.
```
ashleynolan, line 32: I would change this to be: `Please note that styles have to be imported separately to the rest of the component.`

packages/components/molecules/f-mega-modal/README.md
```diff
@@ -54,54 +53,12 @@ export default {
         MegaModal: () => import(/* webpackChunkName: "mega-modal" */ '@justeat/f-mega-modal')
     }
 }
-
-```
-
-### Non-Vue Applications
-
-This module can be ran as a micro front-end for applications that don't make use of the Vue framework.
```
ashleynolan, line 62: Is there a plan to migrate this to a separate doc somewhere?

packages/components/molecules/f-mega-modal/README.md
```diff
@@ -54,54 +53,12 @@ export default {
...
-    </script>
-</body>
-</html>
 ```
 
 ## Configuration
 
 ### Props
 
-`f-mega-modal` has a number of props that allow you to customise its functionality.
```
ashleynolan, line 101: I think this reads better as a README with this description (know it's a bit more verbose, but think it makes it more "friendly" as a README.

packages/components/molecules/f-mega-modal/README.md
```diff
@@ -119,10 +76,6 @@ The props that can be defined are as follows:
 
 ### CSS Classes
 
-The modal has its own styles which are scoped to the component using CSS modules to prevent conflicts with existing styles on the page.
```
ashleynolan, line 122: Same here as above – I'd keep this description in, especially as it does provide context to how the classes are made available.

packages/components/molecules/f-mega-modal/README.md
```diff
@@ -139,59 +92,20 @@ In addition to this, the modal exposes some classes which you can target in your
 | `open` | This event is emitted when the modal is opened. |
 | `close` | This event is emitted when the modal is closed. |
 
-You can add event listeners for these like so
```
ashleynolan, line 142: What's the reason for deleting this block?

packages/components/molecules/f-mega-modal/README.md
```diff
@@ -139,59 +92,20 @@ In addition to this, the modal exposes some classes which you can target in your
...
-      // Do stuff here
-    }
-  }
-}
-</script>
-```
-
 ## Development
 
-Start by cloning the repository and installing the required dependencies:
```
ashleynolan, line 170: Think this is useful for context of what's going on below (it's straightforward to us, but may not be for someone new to JE).

packages/components/molecules/f-mega-modal/README.md
```diff
@@ -139,59 +92,20 @@ In addition to this, the modal exposes some classes which you can target in your
...
-
 From the **root** directory run:
 
 ```sh
 $ yarn storybook:serve
 ```
 
-This will build and serve storybook at [http://localhost:6006](http://localhost:6006).
-
+Serve storybook at [http://localhost:6006](http://localhost:6006).
```
ashleynolan: I think this made more sense before potentially (as it now implies that you have to do something else to serve storybook at that URL, rather than the command above doing it).

ashleynolan: Maybe just changing this to `This serves storybook…`

packages/components/molecules/f-mega-modal/README.md
BanksySan: I think that's strongly implied by the structure.
I removed all the verbose/chatty language so that there's less text to compare and keep synced.

For example:  _Please not that_ is redundant as we wouldn't have written it if we didn't want the reader to note it and _to the rest of the component_ is redundant as the component is explicitly the subject of the document.

I think that unless some important information has been lost then we should keep the word count to a minimum.

packages/components/molecules/f-mega-modal/README.md
BanksySan: Yes, it'll be put into the root doc.

packages/components/molecules/f-mega-modal/README.md
BanksySan: I think it's totally redundant, properties are only there because they change functionality.

I understand that it's friendlier, but it's just more words to copy and copy and copy any nobody would think that it was rude without it.

packages/components/molecules/f-mega-modal/README.md
BanksySan: This is true for all Vue components, there are scoped and public properties.  The reader doesn't need to be reminded of it on every readme.  It could go in the root documentation if it should go anywhere, though this document isn't an instruction of how to use Vue it's just about how to use this specific document.

packages/components/molecules/f-mega-modal/README.md
BanksySan: If someone needed to know how to subscribe to an event then they'd look at Vue documentation, they wouldn't look for it here.

If we do want to put instructions on how to wire up event listeners then it should be in the root, but in every component's readme.

packages/components/molecules/f-mega-modal/README.md
BanksySan: If someone didn't realise that they had to clone the repo then they wouldn't be reading this documentation anyhow.

packages/components/molecules/f-mega-modal/README.md
ashleynolan: I kind of see the `Please note` as bringing special attention to the fact that it's separate to the component import (as when explaining the component setup to people, this isn't always obvious). Just reads a little better to me like that.

packages/components/molecules/f-mega-modal/README.md
ashleynolan: I don't have a strong opinion either way especially – but I would say that there's value in making a README more `human` and providing descriptions/intros to sections (it also breaks up the doc when reading through it – makes it less like a dry spec and more like a manual).

In reality, we don't copy this sort of structural text as it's auto-generated through the generator.

@justeat/ui-admins Any thoughts on this either way?

packages/components/molecules/f-mega-modal/README.md
ashleynolan: This won't be in every component. Not all components have styles that are exposed in this way as in the modal.

packages/components/molecules/f-mega-modal/README.md
ashleynolan: I think if we want to remove it, we should replace it with a doc in storybook replacing it before we do so, so we can reference it from here.

Can you make a ticket to do this and put this back in until then?

packages/components/molecules/f-mega-modal/README.md
ashleynolan: If that's the case, we don't need the lines below either then, as that's what the bash commands are showing.

Tbh I think this should be pulled out into storybook docs, but I think we should keep it until we replace it with those docs (or we create a JIRA explaining the docs we need to create).

packages/components/molecules/f-mega-modal/README.md
DamianMullins: If it's auto-generated I'd say leave it, otherwise, I'm don't mind either way really. I do agree that it's better to have readmes which are more friendly in tone as it makes the project more approachable for first-time users and contributors.

BanksySan: Ah.  I was looking at these as specs, not a manual.  If it should be a manual than being more human is correct.

packages/components/molecules/f-mega-modal/README.md
BanksySan: We don't document the internal styles though.  A sentence effectively saything that there's styles that you're not able to use is more likely to be confusing than helpful.

BanksySan:
This method of changing docs is generating quite a bit of noise, so I'll try a different approach.

PR #618, opened 2021-01-20
f-consumer-oidc@v1.0.0 - Migration

tommcclean:
> Not really looking for a code review, I am migrating this service "as is", there are things in there i would like to tidy up over time, but that time isn't now.

- Migrated je-consumer-oidc to mono repo and renamed f-consumer-oidc
- Restarted versioning at 1.0.0
- Excluded anything we didn't want to migrate; which can be hosted elsewhere

### DIFF
--- a/packages/services/f-consumer-oidc/README.md
+++ b/packages/services/f-consumer-oidc/README.md
@@ -1,0 +1,61 @@
+This library allows consumer microsites to obtain tokens for interacting with HTTP APIs via Smart Gateway.
+
+## Usage
+
+To install directly in a microsite:
+
+```
+npm install @justeat/f-consumer-oidc js-cookie
+```
+
+This library has a required `peerDependency` on `js-cookie` (to help reduce the bundle size).
+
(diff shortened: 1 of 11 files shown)


DamianMullins (dismissed):

kevinrodrigues (dismissed):

packages/services/f-consumer-oidc/README.md
```diff
@@ -0,0 +1,61 @@
...
+    .catch((error) => /* ... */);
+```
+
+* `OidcClient` is an instance of the `oidc-client` npm package which needs to be injected in from the consuming application.
+
+  The client is injected in this way to avoid users downloading the client twice; once from the consuming application and then again from the iframe web page.
+* `optionalSettings` can override the settings that initialise the `UserManager` object from the [`oidc-client`](https://github.com/IdentityModel/oidc-client-js/wiki#usermanager) library.
+* `force` is an optional flag that can be provided to force the signin process to begin, regardless of the current state (defaults to false).
+
+For more detail see flow diagram at bottom of this document.
```
gareththackeray: You've removed this?

tommcclean: Going to put them in the WIKI, i missed that line though; ill remove/update it, ta

kevinrodrigues (approved):

DamianMullins (approved):

PR #621, opened 2021-01-20
f-searchbox@v4.0.0-beta.19 - Refactor address to read from state & al…

kevinrodrigues:
### Changed
- `Address` so it comes from state instead of config.
- `created` hook to `beforeCreate` to resolve Vuex module registration issues with v-model.

### Removed
- `Address` reference from tenant config.

### Added
- Ability to set address via Vuex action `setAddress`.
- Tests to cover changes.
- `Loqate` - Allow user to continue with a suggested address.
- `Loqate` - Allow user to pick their address from street level results.

### Todo
- Hide dropdown when final selection happens (Next PR)
- Fix flicker between dropdowns (Next PR)

### Allow full address selection:
https://www.loom.com/share/d44fe136d2094580beba254f6f4a4537

### Allow continue with address selection:
https://www.loom.com/share/566c52a29ec64fba9e78df60229e04d6


ashleynolan (dismissed):

packages/components/molecules/f-searchbox/src/store/searchbox.module.js
```diff
@@ -234,13 +260,78 @@ export default {
...
+         * @param state
+         * @param fullAddressDetails
+         */
+        [SET_FULL_ADDRESS_DETAILS]: (state, fullAddressDetails) => {
+            state.fullAddressDetails = fullAddressDetails.Items.reduce((seed, val) => {
+                if (!seed.some(s => s.Id === val.Id)) {
+                    seed.push(val);
+                }
+                return seed;
+            }, []).map(({
```
AlixRichards, line 279: All LGTM - but I've been reading over this for a little while and I don't think I fully have in my head what's happening.

I can see its for removing the duplicates but I think maybe because I don't understand what a `seed` is I'm struggling a little. I don't think it needs any changes but maybe we can have a quick chat 👍

AlixRichards (dismissed):

packages/components/molecules/f-searchbox/src/store/searchbox.module.js
kevinrodrigues: Sure, So seed here could be called anything (can change it if that helps?), here it's just represents each item in the `fullAddressDetails`. `val` represents each object in that array. 

Seed:
```javascript
[{Id: "GB|RM|ENG|7AA-RM10", Type: "Postcode", Text: "RM10 7AA", Highlight: "0-4"}]
```

Val:

```javascript
{Id: "GB|RM|ENG|7AA-RM10", Type: "Postcode", Text: "RM10 7AA", Highlight: "0-4"}
```

It will go through the full `fullAddressDetails` list and remove all of the duplicates based on the ID & feed that back to the map method. So if `fullAddressDetails` ended up having two objects with the same ID the reduce method would output:

Before reduce:
```javascript
[
 {Id: "GB|RM|ENG|7AA-RM10", Type: "Postcode", Text: "RM10 7AA", Highlight: "0-4"},
 {Id: "GB|RM|ENG|7AA-RM10", Type: "Postcode", Text: "RM10 7AA", Highlight: "0-4"},
 {Id: "GB|RM|ENG|7AA-AR51", Type: "Postcode", Text: "RM10 7AA", Highlight: "0-4"}
]
```

After reduce:
```javascript
[
 {Id: "GB|RM|ENG|7AA-RM10", Type: "Postcode", Text: "RM10 7AA", Highlight: "0-4"},
 {Id: "GB|RM|ENG|7AA-AR51", Type: "Postcode", Text: "RM10 7AA", Highlight: "0-4"}
]`
```
Let me know if that makes sense? :)

AlixRichards: Yes, that makes 💯  sense, really clear explanation thank you!

AlixRichards (approved):

ashleynolan (approved):

kevinrodrigues:
Screenshot 2021-01-21 at 17 30 20

kevinrodrigues:
Screen Shot 2021-01-22 at 8 20 31 AM

PR #622, opened 2021-01-20
f-button@v0.5.0 -  add navigation link button

FayeCarter:
### Added
- `Action` component for standard buttons.
- `Link` component for links styled as buttons.


packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -28,6 +33,10 @@ export default {
         isFullWidth: {
             type: Boolean,
             default: false
+        },
+        href: {
```
ashleynolan: You shouldn't need to define href as a prop, as it's just a standard attribute (so it should be applied automatically onto the component when set).

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -1,21 +1,26 @@
...
 </template>
 
 <script>
+import ActionButton from './Action.vue';
+import LinkButton from './Link.vue';
 
 export default {
     name: 'FButton',
-    components: {},
+    components: { ActionButton, LinkButton },
```
ashleynolan:
```suggestion
    components: {
        ActionButton,
        LinkButton
     },
```

packages/components/atoms/f-button/stories/Button.stories.js
```diff
@@ -20,9 +20,12 @@ export const ButtonComponent = () => ({
         },
         isFullWidth: {
             default: boolean('isFullWidth', false)
+        },
+        href: {
```
ashleynolan, line 27: Not sure the best way to test this in storybook, but this should just be applied as a standard attribute (may be worth looking up if it's possible to link attributes up to knobs or not).

packages/components/atoms/f-button/src/components/Link.vue
```diff
@@ -0,0 +1,7 @@
+<template>
+    <a
+        :href="$attrs.href"
+        target="_blank">
```
dandel10n: I don't think we need to rewrite a default target prop here

ashleynolan (commented):
Couple of small comments – nothing major though, looks good.

dandel10n (commented):
Looks good, one comment

packages/components/atoms/f-button/src/components/Link.vue
xander-marjoram: Do we need `href` passed through as well?

FayeCarter: 👍  I've removed both

packages/components/atoms/f-button/src/components/Button.vue
FayeCarter: 👍 removed, thanks

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -8,4 +8,81 @@ describe('Button', () => {
...
+                // Assert
+                expect(wrapper.vm.buttonSizeClassname).toEqual('XSmall');
+            });
+        });
+
+        describe('componentType :: ', () => {
+            describe('when `href` prop is an empty string :: :: ', () => {
+                const propsData = { href: '' };
+
+                it('should return `action`', () => {
```
kevinrodrigues: `Action`

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -8,4 +8,81 @@ describe('Button', () => {
...
+                    // Assert
+                    expect(wrapper.find('[data-test-id="action-button-component"]').exists()).toBeTruthy();
+                });
+            });
+
+            describe('when `href` prop is not an empty string :: :: ', () => {
+                const link = 'http://www.just-eat.co.uk';
+                const propsData = { href: link };
+
+                it('should return `link`', () => {
```
kevinrodrigues: `Link`?

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -8,4 +8,81 @@ describe('Button', () => {
...
+                    // Act
+                    const wrapper = mount(FButton, { propsData });
+
+                    // Assert
+                    expect(wrapper.find('[data-test-id="action-button-component"]').exists()).toBeTruthy();
+                });
+            });
+
+            describe('when `href` prop is not an empty string :: :: ', () => {
+                const link = 'http://www.just-eat.co.uk';
```
kevinrodrigues: Should go in a `BeforeEach`

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -8,4 +8,81 @@ describe('Button', () => {
...
+                const wrapper = shallowMount(FButton, { propsData });
+
+                // Assert
+                expect(wrapper.vm.buttonSizeClassname).toEqual('XSmall');
+            });
+        });
+
+        describe('componentType :: ', () => {
+            describe('when `href` prop is an empty string :: :: ', () => {
+                const propsData = { href: '' };
```
kevinrodrigues: beforeEach

packages/components/atoms/f-button/src/components/Action.vue
```diff
@@ -0,0 +1,5 @@
+<template>
+    <button>
```
kevinrodrigues: Probably missed the change somewhere but is it possible to set this to `button=type` by default if the button component is used? To avoid setting it at component level.

FayeCarter: 👍  thanks

packages/components/atoms/f-button/src/components/_tests/Button.test.js
FayeCarter: changed to `componentType` of `link-button` after chat :)

packages/components/atoms/f-button/src/components/_tests/Button.test.js
FayeCarter: changed to `componentType` of `action-button` after chat :)

packages/components/atoms/f-button/src/components/Action.vue
```diff
@@ -0,0 +1,5 @@
+<template>
+    <button type="button">
```
kevinrodrigues: You probably want to set this dynamically? You could pass through an option to allow attributes to be passed through and then you could dynamically set them via a directive? Maybe make use of the prop on the button.vue component that exposes the $attrs already? Happy for this to be looked in another pr btw.

kevinrodrigues (dismissed):

packages/components/atoms/f-button/src/components/Action.vue
FayeCarter: @kevinrodrigues I _think_ I've got this working now 🤞

kevinrodrigues (approved):

simonsandersje (approved):

PR #625, opened 2021-01-21
f-cookie-banner@0.1.0 - First commit

simonsandersje:
---

Initial commit of the f-cookie-banner component

Legacy banner for UK:
image

New banner for UK (Desktop):
image

New banner for UK (Tablet):
image

New banner for UK (Mobile):
image

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [x] Internet Explorer 11

- [x] Firefox (latest)
- [x] Edge


packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
+<template>
+    <div
+        ref="cookieBanner"
+        :class="[$style['c-cookieBanner'], $style['c-cookieBanner-overlay'], { [$style['c-cookieBanner--hidden']]: hideBanner }]"
+        attr="data-cookie-consent-overlay"
```
SachaZvetelman: What is this `attr` exactly? Maybe a stupid question from my side.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
...
+        :aria-hidden="isHidden">
```
SachaZvetelman: Do we need a `hideBanner` and an `isHidden`?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
...
+        <div
+            v-if="!legacyBanner"
+            :class="$style['c-cookieBanner-card']">
+            <div
+                id="dialog1"
```
SachaZvetelman: We probably need a more descriptive `id` here.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
...
+                </p>
+                <p
+                    :class="$style['c-cookieBanner-text']">
+                    {{ copy.text3 }}
+                    <a
+                        :href="copy.linkHref"
+                        :class="$style['c-cookieBanner-link']"
+                        target="_blank">
+                        {{ copy.linkText }}</a>{{ copy.text4 }}
+                </p>
```
SachaZvetelman: This can be simplified with the `<i18n>` directive. Contact either @FayeCarter or me for more help on this, if you need.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
...
+                    data-button-accept
+                    is-full-width
+                    @click.native="acceptActions">
+                    {{ copy.buttonText2 }}
+                </button-component>
+                <button-component
+                    tabindex="0"
+                    role="button"
+                    button-type="ghost"
+                    data-button-nonaccept
```
SachaZvetelman: I'm not sure about this either. I see you're following this pattern but I'm struggling to understand it. Did you mean to add `data-test-id` here?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
...
+            </div>
+
+            <div :class="$style['c-cookieBanner-CTA']">
+                <button-component
+                    tabindex="0"
+                    role="button"
+                    data-button-accept
+                    is-full-width
+                    @click.native="acceptActions">
+                    {{ copy.buttonText2 }}
```
SachaZvetelman: Please create more descriptive variables here 🙂

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
...
+                        </a>
+                    </p>
+                    <button
+                        class="c-cookieWarning-btn"
+                        data-test-id="cookieBanner-close-button"
+                        aria-label="Close"
+                        @click="hideLegacyBanner" />
+                </div>
+            </div>
+        </div>
```
SachaZvetelman: It's probably worth extracting this into a separate component (within `f-cookie-banner`)

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
...
+            default: 7776000
+        }
+    },
+
+    data () {
+        const locale = globalisationServices.getLocale(tenantConfigs, this.locale, this.$i18n);
+        const localeConfig = tenantConfigs[locale];
+        const theme = globalisationServices.getTheme(locale);
+        const hideBanner = false;
+        const legacyBanner = false;
```
SachaZvetelman: I think we can remove these and just add them directly to the `return`.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
...
+    },
+
+    watch: {
+        isHidden (newVal) {
+            if (newVal) {
+                this.hideBanner = true;
+            } else {
+                this.hideBanner = false;
+            }
+        },
```
SachaZvetelman: Not fully sure why we need the watcher here, but this could be simplified:
```
isHidden (newVal) {
  this.hideBanner = !!newVal;
}

File: packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue [hunk 10]
```diff
@@ -0,0 +1,410 @@
...
+        showLegacyBanner (newVal) {
```
SachaZvetelman, line 140: Same here.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
...
+        /**
+         * Set focus to the cookie consent banner title for accessibility
+         */
+        focusOnTitle () {
+            this.$refs.cookieBannerHeadingFocus.focus();
+        },
+        /**
+         * hide legacy banner
+         */
+        hideLegacyBanner () {
```
SachaZvetelman: If we have this method, then we should call it in more places. We have several `this.hideBanner = true` instead of calling this function.

packages/components/organisms/f-cookie-banner/src/components/_tests/CookieBanner.test.js
```diff
@@ -0,0 +1,198 @@
...
+                propsData
+            });
+
+            // Assert
+            expect(wrapper.exists()).toBe(true);
+        });
+    });
+
+    describe('methods', () => {
+        describe('isNotExcluded() should be false for cookies that will not be deleted', () => {
```
xander-marjoram: You should just be able to have `'isNotExcluded()'` here

packages/components/organisms/f-cookie-banner/src/components/_tests/CookieBanner.test.js
```diff
@@ -0,0 +1,198 @@
...
+                    propsData
+                });
+
+                // Assert
+                expect(wrapper.vm.isNotExcluded(cookieName)).toBe(expected);
+            });
+        });
+
+        describe('dataLayerPush()', () => {
+            it('should push `consentLevel` to the dataLayer', () => {
```
xander-marjoram: Nit: `consentLevel` doesn't look like it's a variable name that's used here

packages/components/organisms/f-cookie-banner/src/components/_tests/CookieBanner.test.js
```diff
@@ -0,0 +1,198 @@
...
+            });
+        });
+
+        describe('checkCookieBannerCookie()', () => {
+            it.each([
+                [false, ''],
+                [false, 'random value'],
+                [true, 'full'],
+                [true, 'necessary']
+            ])('should show/hide banner according to `je-cookieConsent` value', (
```
xander-marjoram: Can you try and word it so `cookieValue` is part of the test name? If ones of the tests were to fail, currently it would say something like:

```
should show/hide banner according to `je-cookieConsent` value
Expected: true
But was: false
```
and it's not clear which test case it was

packages/components/organisms/f-cookie-banner/src/components/_tests/CookieBanner.test.js
```diff
@@ -0,0 +1,198 @@
...
+
+                wrapper.vm.setLegacyCookieBannerCookie();
+
+                // Assert
+                expect(cookieSpy).toHaveBeenCalledWith(payloadName, payloadValue, payloadObj);
+            });
+        });
+
+        describe('acceptActions()', () => {
+            it('should set the banner consent cookie to `full`, push `full` to dataLayer and hide the banner', () => {
```
xander-marjoram: This sounds like three test cases

packages/components/organisms/f-cookie-banner/src/components/_tests/CookieBanner.test.js
```diff
@@ -0,0 +1,198 @@
...
+
+                // Assert
+                expect(cookieSpy).toHaveBeenCalledWith('full');
+                expect(dataLayerSpy).toHaveBeenCalledWith('full');
+                expect(wrapper.vm.hideBanner).toBe(true);
+            });
+        });
+
+        describe('nonAcceptActions()', () => {
+            it('should set the banner consent cookie to `necessary`, push `necessary` to dataLayer, remove unnecessary cookies, resend GTM events and hide the banner', () => {
```
xander-marjoram: This sounds like five test cases

packages/components/organisms/f-cookie-banner/test/specs/component/f-cookieBanner.component.spec.js
```diff
@@ -0,0 +1,13 @@
+import CookieBannerComponent from '../../../test-utils/component-objects/f-cookieBanner.component';
+
+xdescribe('f-cookieBanner component tests', () => {
```
xander-marjoram: Is it correct that this test is ignored?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
xander-marjoram: I'm assuming this can simply be written as `data-cookie-consent-overlay`

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
...
+        },
+        /**
+         * Check if the cookie banner has been shown to this user
+         */
+        checkCookieBannerCookie () {
+            if (this.legacyBanner) {
+                this.hideBanner = this.$cookies.get('je-banner_cookie') === 2;
+                this.setLegacyCookieBannerCookie();
+            } else {
+                this.hideBanner = this.$cookies.get('je-cookieConsent') === 'full' || this.$cookies.get('je-cookieConsent') === 'necessary';
```
xander-marjoram:
```suggestion
                const cookieConsent = this.$cookies.get('je-cookieConsent');
                this.hideBanner = cookieConsent === 'full' || cookieConsent === 'necessary';
                // or
                this.hideBanner = ['full', 'necessary'].includes(cookieConsent);
```

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,410 @@
...
+            }
+        },
+        /**
+         * Set the cookie for the user's choice
+         */
+        setCookieBannerCookie (cookieValue) {
+            this.$cookies.set('je-cookieConsent', cookieValue, {
```
xander-marjoram: I would pull the cookie names out into a constants file

packages/components/organisms/f-cookie-banner/src/demo/Index.vue
```diff
@@ -0,0 +1,26 @@
+// Demo component for testing `f-cookie-banner` independently via `vue serve --open src/demo/Index.vue`
```
siggerzz: This file can be deleted 👍🏼

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
siggerzz: Can you add `data-test-id="cookieBanner-component"` here please?

packages/components/organisms/f-cookie-banner/test/specs/component/f-cookieBanner.component.spec.js
siggerzz: If you add the following, the component test will pass.

```
describe('f-cookieBanner component tests', () => {
    beforeEach(() => {
        browser.url('?path=/story/components-organisms--cookie-banner-component'); //This URL may be wrong, worth checking in storybook!
        browser.switchToFrame(0);
        CookieBannerComponent .waitForCookieBannerComponent();
    });
    
    
    .....
```


I'll create a PR to address this in the component generator 👍🏼

siggerzz:
Just a few comments from me!

Would you mind adding the following spec file to this component, so we have automated a11y tests? - https://github.com/justeat/fozzie-components/blob/master/packages/components/molecules/f-alert/test/specs/accessibility/axe-accessibility.spec.js

Thanks!

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: `attr=` removed and `data-test-id="cookieBanner-component"` added

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: We do not need both

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: the `id` is a hangover from the GTM version and can be removed

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Will do

packages/components/organisms/f-cookie-banner/test/specs/component/f-cookieBanner.component.spec.js
simonsandersje: Added - and good idea to update the generator!

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Seems a bit of overkill for one constant IMHO, as the legacy cookie `je-banner_cookie` won't be necessary for more than a couple of weeks (fingers crossed)

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: More leftovers from the GTM code - removed

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Removed the method and set `hideBanner` directly on the close button

packages/components/organisms/f-cookie-banner/package.json
```diff
@@ -0,0 +1,57 @@
...
+  ],
+  "scripts": {
+    "prepublishOnly": "yarn lint && yarn test && yarn build",
+    "build": "vue-cli-service build --target lib --name f-cookie-banner ./src/index.js",
+    "demo": "vue serve --port 8080 ./src/demo/Index.vue",
+    "lint": "vue-cli-service lint",
+    "lint:fix": "yarn lint --fix",
+    "lint:style": "vue-cli-service lint:style",
+    "test": "vue-cli-service test:unit",
+    "webdriver:delay:chrome": "yarn test:wait-for-server && yarn webdriver:start:chrome",
```
siggerzz:
```
    "webdriver:delay:chrome": "yarn test:wait-for-server && yarn webdriver:start:chrome",
    "webdriver:start:chrome": "wdio ../../wdio.conf.js"
```

should be removed in favour of

```
    "test-component:chrome": "wdio ../../../../wdio.conf.js --suite component",
    "test-a11y:chrome": "wdio ../../../../wdio.conf.js --suite a11y"
```

This is already part of the generator, but I added this change since this branch was created 👍🏼

simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,393 @@
...
+                    tabindex="0"
+                    role="button"
+                    button-type="ghost"
+                    is-full-width
+                    @click.native="nonAcceptActions">
+                    {{ copy.nonAcceptButtonText }}
+                </button-component>
+            </div>
+        </div>
+        <div v-if="legacyBanner">
```
kevinrodrigues: I think you could use just a `v-else` here?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,393 @@
...
+        return {
+            copy: { ...localeConfig },
+            theme,
+            hideBanner: false,
+            legacyBanner: false
+        };
+    },
+
+    watch: {
+        isHidden (newVal) {
```
kevinrodrigues, line 136: Do you need these two watchers? Looks like they're props that determine the visibility of things when they're passed in on render rather than something happening in the component itself?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,393 @@
+<template>
```
kevinrodrigues, line 1: Is it worth splitting some of the components out into smaller ones? Would mean you'd have smaller chunks of logic in smaller components. I think these are the main components? 

1. `c-cookieBanner-CTA.vue`
2. `legacyBanner.vue`
3. `cookieBanner.vue` (Would just be the main component or wrapper you have already in this file.)

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: As discussed in Slack...

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: I have moved the legacy banner to a component, as this will make it easier to remove in a couple of weeks time. However, I don't think it's worth moving the cookie banner or buttons out to components, as the cookie banner is just a heading and three paragraphs, and the buttons are already imported from f-button - just seems like an unnecessary layer of complexity (KISS).

packages/components/organisms/f-cookie-banner/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+Running below `yarn` commands from the component folder, starts a development
+server displaying a preview example of the component implementation.
+
+```bash
+# cd /packages/f-cookie-banner
+yarn install
+
+# followed by
+yarn demo
```
dandel10n: This is out of date I am afraid. You can copy the right instructions from [here](https://github.com/justeat/fozzie-components/tree/master/packages/components/atoms/f-button#development)

packages/components/organisms/f-cookie-banner/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "lint:style": "vue-cli-service lint:style",
+    "test": "vue-cli-service test:unit",
+    "test-component:chrome": "wdio ../../../../wdio.conf.js --suite component",
+    "test-a11y:chrome": "wdio ../../../../wdio.conf.js --suite a11y"
+  },
+  "browserslist": [
+    "extends @justeat/browserslist-config-fozzie"
+  ],
+  "dependencies": {
+    "@justeat/f-button": "0.6.1",
```
dandel10n: f-button can go to dev dependencies

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+                role="dialog"
+                aria-labelledby="cookieConsentLabel"
+                aria-modal="true">
+                <h2
+                    id="cookieConsentLabel"
+                    :class="$style['c-cookieBanner-title']">
+                    <a
+                        ref="cookieBannerHeadingFocus"
+                        name="cookieConsentTitle"
+                        tabindex="0"
```
dandel10n, line 22: I don't think specifying tabindex here is needed because it is 0 for a link by default

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+                                <span>{{ $t('cookiePolicyLinkText') }}</span>
+                            </a>
+                        </template>
+                    </i18n>
+                </p>
+            </div>
+
+            <div :class="$style['c-cookieBanner-CTA']">
+                <button-component
+                    tabindex="0"
```
dandel10n: No need to specify tabindex for a button because it is 0 by default

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+            :class="[$style['c-cookieBanner-card'], { [$style['c-cookieBanner-ios']]: checkIosSafari() }]">
+            <div
+                :class="$style['c-cookieBanner-content']"
+                role="dialog"
+                aria-labelledby="cookieConsentLabel"
+                aria-modal="true">
+                <h2
+                    id="cookieConsentLabel"
+                    :class="$style['c-cookieBanner-title']">
+                    <a
```
kevinrodrigues: I think this should be a button. `<button type='button'>`

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+                            </a>
+                        </template>
+                    </i18n>
+                </p>
+            </div>
+
+            <div :class="$style['c-cookieBanner-CTA']">
+                <button-component
+                    tabindex="0"
+                    role="button"
```
dandel10n: Role should come from the button component so no need to specify it here

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
+<template>
+    <div
+        ref="cookieBanner"
+        :class="[$style['c-cookieBanner'], $style['c-cookieBanner-overlay'], { [$style['c-cookieBanner--hidden']]: hideBanner }]"
```
kevinrodrigues: Worth breaking these so they sit on new lines.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+        data-cookie-consent-overlay
+        data-test-id="cookieBanner-component"
+        :aria-hidden="hideBanner">
+        <div
+            v-if="!legacyBanner"
+            :class="[$style['c-cookieBanner-card'], { [$style['c-cookieBanner-ios']]: checkIosSafari() }]">
```
kevinrodrigues: Same.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+                    </i18n>
+                </p>
+            </div>
+
+            <div :class="$style['c-cookieBanner-CTA']">
+                <button-component
+                    tabindex="0"
+                    role="button"
+                    is-full-width
+                    @click.native="acceptActions">
```
dandel10n: You don't need `.native` here with the 0.6.1 version of f-button

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+                    {{ $t('acceptButtonText') }}
+                </button-component>
+                <button-component
+                    tabindex="0"
+                    role="button"
+                    button-type="ghost"
+                    is-full-width
+                    @click.native="nonAcceptActions">
```
dandel10n:
```suggestion
                <button-component
                    button-type="ghost"
                    is-full-width
                    @click="nonAcceptActions">
```
As I commented above button should look like this

packages/components/organisms/f-cookie-banner/src/assets/scss/common.scss
```diff
@@ -0,0 +1,16 @@
...
+@import '~@justeat/f-utils/src/scss/mixins/index';
+@import '~include-media/dist/include-media'; // Cleaner media query declarations – http://include-media.com
+@import '~@justeat/fozzie/src/scss/settings/variables';
+@import '~@justeat/fozzie/src/scss/base/reset';
+@import '~@justeat/fozzie-colour-palette/src/scss/';
+
+/**
+ * Component Variables
+ */
+// $cookieBanner-background-color: $brand--orange;
```
SachaZvetelman: Do we need this?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+        this.checkLegacyBannerFlag();
+        this.checkCookieBannerCookie();
+        this.focusOnTitle();
+    },
+
+    methods: {
+        /**
+         * Actions for "Accept all cookies" button
+         */
+        acceptActions () {
```
kevinrodrigues: You could combine this with the method below and use a param to determine the resentEvents & removeUnnecessaryCookies calls & the consent type.  (can be done in another pr)

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+        font-family: $font-family-base;
+    }
+
+    .c-cookieBanner-overlay {
+        position: fixed;
+        top: 0;
+        bottom: 0;
+        left: 0;
+        right: 0;
+        background-color: rgba(0, 0, 0, 0.7);
```
dandel10n, line 284: We should have a suitable background colour in our fozzie-colour-palette

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+            <div
+                :class="$style['c-cookieBanner-content']"
+                role="dialog"
+                aria-labelledby="cookieConsentLabel"
+                aria-modal="true">
+                <h2
+                    id="cookieConsentLabel"
+                    :class="$style['c-cookieBanner-title']">
+                    <a
+                        ref="cookieBannerHeadingFocus"
```
SachaZvetelman: The name is a bit confusing, in my opinion. The `ref` is to reference the element, and "Focus" seems more to do with what you're going to do with the `ref` in a method than what the element is about. Dropping the "focus" on this `ref` makes more sense for me.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+                        name="cookieConsentTitle"
+                        tabindex="0"
+                        data-consent-title>
+                        {{ $t('mainTitle') }}
+                    </a>
+                </h2>
+
+                <p :class="$style['c-cookieBanner-text']">
+                    {{ $t('textLine1') }}
```
SachaZvetelman, line 33: Do we have some more descriptive names for these? `textLine1`, `textLine2` and ``textLine3` don't say a lot.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: @kevinrodrigues The name anchor is only there to allow focus on the heading (`<h2>` is not focusable) for a11y. Can a button act as a name anchor?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+                                :href="$t('cookiePolicyLinkUrl')"
+                                target="_blank">
+                                <span>{{ $t('cookiePolicyLinkText') }}</span>
+                            </a>
+                        </template>
+                    </i18n>
+                </p>
+            </div>
+
+            <div :class="$style['c-cookieBanner-CTA']">
```
SachaZvetelman: Should be `c-cookieBanner-cta`, I think.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+                    button-type="ghost"
+                    is-full-width
+                    @click.native="nonAcceptActions">
+                    {{ $t('nonAcceptButtonText') }}
+                </button-component>
+            </div>
+        </div>
+        <legacy-banner
+            v-else
+            @hide-legacy-banner="hideBanner = true"
```
SachaZvetelman: Could we please extract this into a `method`?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+
+    components: {
+        ButtonComponent,
+        LegacyBanner
+    },
+
+    mixins: [VueGlobalisationMixin],
+
+    props: {
+
```
SachaZvetelman: Super nit: unnecessary empty line :)

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+    data () {
+        const locale = globalisationServices.getLocale(tenantConfigs, this.locale, this.$i18n);
+        const localeConfig = tenantConfigs[locale];
+        const theme = globalisationServices.getTheme(locale);
+
+        return {
+            config: { ...localeConfig },
+            tenantConfigs,
+            theme,
```
SachaZvetelman: You don't need these I believe, if you're using the `VueGlobalisationMixin`

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+
+    methods: {
+        /**
+         * Actions for "Accept all cookies" button
+         */
+        acceptActions () {
+            this.setCookieBannerCookie('full');
+            this.setLegacyCookieBannerCookie();
+            this.dataLayerPush('full');
+            this.hideBanner = true;
```
SachaZvetelman: I personally would rename `hideBanner`. It's a boolean, so it should be called something like `isBannerHidden`, or `shouldDisplayBanner`.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+        /**
+         * Set focus to the cookie consent banner title for accessibility
+         */
+        focusOnTitle () {
+            this.$refs.cookieBannerHeadingFocus.focus();
+        },
+        /**
+         * Check if the cookie banner has been shown to this user
+         */
+        checkLegacyBannerFlag () {
```
SachaZvetelman, line 187: We can probably rename this to something like `hasBannerBeenShownToUser` which would describe its intent better, I believe.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+            this.legacyBanner = this.config.displayLegacy;
+        },
+        /**
+         * Check if the cookie banner has been shown to this user
+         */
+        checkCookieBannerCookie () {
```
SachaZvetelman, line 193: Same here.
Btw, the JSDoc for this seems a copy-paste from the one above.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+            if (this.legacyBanner) {
+                this.hideBanner = this.$cookies.get('je-banner_cookie') === 2;
```
SachaZvetelman: I think we should extract these cookie names into constants.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+            const dataLayer = window.dataLayer || [];
+            dataLayer.push({ event: 'trackConsent', userData: { consent: consentLevel } });
+        },
+        /**
+         * Check for excluded cookies/storage
+         * @param {String} cookieName
+         * @returns {Bool}
+         */
+        isNotExcluded (cookieName) {
+            return this.config.cookieExclusionList.every(arrVal => cookieName.lastIndexOf(arrVal, arrVal.length - 1) === -1);
```
SachaZvetelman: Could we please rename `arrVal` to something more meaningful for the context? Maybe `cookieExclusion` since it's a list of excluded cookies, it seems.

SachaZvetelman: I'm not sure I fully understand what this function does at first glance if I'm honest 😬

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+        },
+        /**
+         * Remove unnecessary cookies
+         */
+        removeUnnecessaryCookies () {
+            const cookies = Object.keys(this.$cookies.getAll());
+            for (let i = 0; i < cookies.length; i++) {
```
SachaZvetelman, line 243: Any particular reason for using a standard `for` loop over something more functional? 🙂

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+                if (this.isNotExcluded(cookies[i])) this.$cookies.remove(cookies[i]);
+            }
+        },
+        /**
+         * Check for iOS browser
+         * @returns {Bool}
+         */
+        checkIosSafari () {
```
SachaZvetelman: Same as above. This to me should be named `isIosBrowser` or similar.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+                    dataLayer[i].eventResent = dataLayer[i].event;
+                    dataLayer[i]['gtm.uniqueEventId'] = Math.floor(100000 + Math.random() * 900000);
+                    dataLayer.push(dataLayer[i]);
+                }
+                if (dataLayer[i].event && dataLayer[i].event === 'trackConsent' && (dataLayer[i].userData.consent === 'full' || dataLayer[i].userData.consent === 'necessary')) {
+                    dataLayer.push({ eventResent: false });
+                    break;
+                }
+            }
+        }
```
SachaZvetelman, line 272: I'm not sure if we got this from somewhere else, but it looks rather complex and hard to understand. Is it something we can simplify?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,364 @@
...
+
+        .c-cookieBanner-CTA {
+            min-width: initial;
+            display: flex;
+            flex-direction: row-reverse;
+            margin: 0;
+        }
+    }
+
+    @media (max-width: 600px) {
```
SachaZvetelman: Do we have breakpoints somewhere we can reuse? 🤔

packages/components/organisms/f-cookie-banner/src/components/LegacyBanner.vue
```diff
@@ -0,0 +1,76 @@
...
+            </div>
+        </div>
+    </div>
+</template>
+
+<script>
+export default {
+    data () {
+        return {
+            hideBanner: false
```
SachaZvetelman, line 44: Same as above.

packages/components/organisms/f-cookie-banner/src/components/LegacyBanner.vue
```diff
@@ -0,0 +1,76 @@
...
+                position: absolute;
+                top: 12px;
+                right: 8px;
+                width: 10px;
+                height: 10px;
+                background: url('//dy3erx8o0a6nh.cloudfront.net/images/icon-close-banner.png') no-repeat 50%;
+                background-size: 10px 10px;
+                border: none;
+                cursor: pointer;
+            }
```
SachaZvetelman, line 91: Can we use our button component so we don't have to do this?

packages/components/organisms/f-cookie-banner/src/components/_tests/CookieBanner.test.js
```diff
@@ -0,0 +1,301 @@
...
+                propsData
+            });
+
+            // Assert
+            expect(wrapper.exists()).toBe(true);
+        });
+    });
+
+    describe('methods', () => {
+        describe('isNotExcluded()', () => {
```
SachaZvetelman: We normally don't put the `()` in the name of the tests here. Could we please remove them?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
kevinrodrigues: Ok. I thought the ref was focusing on mount?

kevinrodrigues (dismissed):

SachaZvetelman (commented):
Sorry for making too many comments, @simonsandersje.
I think we should have split this into two PRs. The first one with the basic structure and the second one with the functionality.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: It is, but it can't focus on the `<h2>`, so I put the name anchor in - happy to look into another approach, I guess

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
dandel10n: Link without href should trigger error for semantic/accessibility check. We shouldn't use link or button on title to make it focusable because it breaks the semantics.

packages/components/organisms/f-cookie-banner/src/components/LegacyBanner.vue
simonsandersje: @SachaZvetelman I'll chat with Maryia about adding the close-cross style to f-button - although I am worried that by the time we get that done, we'll need to delete the legacy banner code anyway (it's a stop-gap to keep feature parity for the UK, till the UK is ready for the new banner), but it'll then be there for someone else in the future

packages/components/organisms/f-cookie-banner/src/components/_tests/CookieBanner.test.js
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: @SachaZvetelman This is from GTM, so I'll ask Pete Crawley/Rob Ellison if they can provide a simplified version

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
kevinrodrigues: Looking at it again do we actually need it to focus on the title? I think one of the dismissable ctas makes more sense.

packages/components/organisms/f-cookie-banner/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+# Changelog
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+
+v0.1.0
+------------------------------
+*January 21, 2021*
```
ashleynolan: 🕐

packages/components/organisms/f-cookie-banner/src/assets/scss/common.scss
```diff
@@ -0,0 +1,16 @@
...
+// n.b. This variables isn't actually used (unless there are theme specific vars from fozzie)
+// Instead, the theme is handled via a data-attribute on the component (via the `@mixin theme()` below)
+$theme: 'jet' !default;
+
+@import '~@justeat/f-utils/src/scss/functions/index';
+@import '~@justeat/f-utils/src/scss/mixins/index';
+@import '~include-media/dist/include-media'; // Cleaner media query declarations – http://include-media.com
+@import '~@justeat/fozzie/src/scss/settings/variables';
+@import '~@justeat/fozzie/src/scss/base/reset';
+@import '~@justeat/fozzie-colour-palette/src/scss/';
```
ashleynolan: You can now replace this with just `@import  '~@justeat/fozzie/src/scss/fozzie';`

(The PR to the generator I'm doing now will have that in).

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,371 @@
+<template>
+    <div
+        ref="cookieBanner"
+        :class="[
+            $style['c-cookieBanner'],
+            $style['c-cookieBanner-overlay'],
```
ashleynolan: If this is a modifier of `c-cookieBanner` it should be double dash – so `c-cookieBanner--overlay`

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,371 @@
...
+            { [$style['c-cookieBanner--hidden']]: hideBanner }
```
ashleynolan: As this is a state (added/removed based on JS) should use `is-hidden`.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,371 @@
...
+        right: 0;
+        background-color: rgba(0, 0, 0, 0.7);
+        z-index: 99999991;
+    }
+
+    .c-cookieBanner--hidden {
+        display: none;
+    }
+
+    .c-cookieBanner-card {
```
ashleynolan, line 292: Hard to imagine, but why do you need `c-cookieBanner-overlay` and `c-cookieBanner-card` to both be fixed (as `c-cookieBanner-card` is nested inside `c-cookieBanner-overlay` which is fixed to the bottom anyway)

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,371 @@
...
+                }
+            }
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+    .c-cookieBanner {
+        font-family: $font-family-base;
```
ashleynolan: This should get set outside the component globally (or is this in there for legacy reasons).

If it's there to support legacy, I'd add a comment to indicate this will be able to be removed at a later date once it's no longer used in XXX platform.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
ashleynolan: I'd think making them more descriptive would tie them a bit too tightly to the content – which if they then changed would mean the name would be redundant? Hard to say without seeing the content, but I imagine cookie banner text will be pretty generic.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,371 @@
...
+        acceptActions () {
+            this.setCookieBannerCookie('full');
+            this.setLegacyCookieBannerCookie();
+            this.dataLayerPush('full');
+            this.hideBanner = true;
+        },
+        /**
+         * Actions for "Accept only required cookies" button
+         */
+        nonAcceptActions () {
```
ashleynolan: Is this meant to be `non` or would it make more sense as `acceptRequiredActions`? (just reads a little funny to me)

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
ashleynolan: > https://github.com/justeat/fozzie/blob/master/src/scss/settings/_variables.scss#L76-L83

```suggestion
    @media ('<narrowMid') {
```

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,371 @@
...
+        }
+    }
+
+    .c-cookieBanner-content,
+    .c-cookieBanner-CTA {
+        margin: 0 auto;
+        padding: spacing(x4);
+    }
+
+    @media (max-width: 768px) {
```
ashleynolan:
```suggestion
    @media ('<mid') {
```

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,371 @@
...
+    .c-cookieBanner-CTA {
+        min-width: 352px;
+    }
+
+    .c-cookieBanner-text {
+        margin: 0;
+        padding: 0;
+    }
+
+    .c-cookieBanner-link {
```
ashleynolan: Is this used in the template? I couldn't find it when searching

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: @SachaZvetelman @kevinrodrigues @dandel10n Legal specifically don't want us to focus on the CTAs. If the company is seen to be pushing users to "Accept all", then that is not legal. The law states that all CTAs must have equal prominence, so if we set focus on the "Accept all", we could be seen to be forcing (disabled) users to accept all cookies. If we set focus to the "Accept only necessary" button, then we run the risk of users just hitting return and our conversion levels dropping off the cliff (as well as being open to pushing users to make a certain decision - again not legal). After discussing with legal, we agreed that the best thing to do was to focus on the title, so that all users (even screen reader users) would have the opportunity to read the cookie policy statement and have the chance to understand what the modal is about.

The only other recommendation I can think of is to have a visually-hidden link that screen readers focus to, similar to the "Skip to content" link, at the top of the cookie banner - this risks screen readers repeating the text "Cookies", as most screen reader users set the reader to read headings, unless someone can suggest some semantic text that would work for a skip link.

Happy to set up a meeting to discuss this.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: @kevinrodrigues IMHO, it is clearer to have distinct methods for the button actions, but happy for it to be done in a separate ticket

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: @dandel10n I'll raise this as a separate ticket to add a modal background colour to the fozzie-color-palette, as we'll need to update the backgrounds for all the components that use modals and this work is rolling on way too long now.

packages/components/organisms/f-cookie-banner/src/assets/scss/common.scss
simonsandersje: Nope - this seems to have been generated by the generator, so I'll remove it.

packages/components/organisms/f-cookie-banner/README.md
simonsandersje: Done

packages/components/organisms/f-cookie-banner/package.json
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/assets/scss/common.scss
simonsandersje: Removed

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: @dandel10n Confusion here. I am using 0.6.1, but I cannot get the buttons to work without `.native` - maybe we can pair on this next week to see what I am doing wrong?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: See above

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: @SachaZvetelman @ashleynolan I have been racking my brain all night on this one. I could rename the vars `cookieBannerTextLine1` etc, but that seems overly complex given that we know this is the cookie banner. I have tried to get some sense from the translations of what is in each line, but each language differs slightly (these are translations not transliterations, so word order is not the same). Maybe `topLineText`, `middleLineText` and 'bottomText`? But these seem to be just a variation on 1,2 and 3, and we might at some point decide on different screen positions. 

We can maybe discuss it in the a11y meeting I'm setting up (to discuss the focus issue).

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Hmmm - tried removing these and it all breaks. :(

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: This is checking the legacy banner flag in the JSON config, so the name seemed to make sense to me - happy to change it to whatever you want

packages/components/organisms/f-cookie-banner/src/components/LegacyBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Removed unused css

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Renamed

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Agree - I'll remove it and if we later need it for legacy apps, I can add it back in

packages/components/organisms/f-cookie-banner/CHANGELOG.md
simonsandersje: I'll update this when the comments have finished flooding in!

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/assets/scss/common.scss
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: @ashleynolan If I remove either `fixed` declaration, the banner doesn't display how I want - happy to look at whether there's a better way to pin the banner to the bottom with you

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
SachaZvetelman: @simonsandersje, it was a minor comment, I didn't want you to rack your brain all night on this thing! If we cannot find an easy, straight-forward name, leave it as it is. It's not a big deal at all. I'm marking this conversation as resolved. 🙂

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/LegacyBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Cos I'm an old man?

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done... and extended the jsdocs comments

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Anchor removed

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: @SachaZvetelman Pete reckons this is as simple as it can be. Besides, he thinks we should keep third party code as is, in case we need to get Pete/Rob involved in debugging, so that they'll recognise and understand the code. As this is GTM code and not our code, I agree with Pete. If you really really think we need to refactor this code, then I think we'll need to carefully document any changes to help the data guys understand what we've done and why.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Just re-reading this, it is the comment that is wrong, and has been updated.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -0,0 +1,359 @@
...
+        },
+        /**
+         * Resend GTM events
+         */
+        resendEvents () {
+            const dataLayer = window.dataLayer || [];
+            const excludeEvents = ['gtm.js', 'gtm.click', 'gtm.linkClick', 'gtm.triggerGroup', 'trackConsent', 'showConsent'];
+            let i;
+            for (i = 0; i < dataLayer.length; i++) {
+                if (dataLayer[i].event && !excludeEvents.includes(dataLayer[i].event)) {
```
DamianMullins, line 262: Can add `dataLayer[i]` to a variable and reference that rather than repeat the lookup throughout the loop.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Swapped for absolute

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: @DamianMullins This is third party code, which I have pasted in as supplied. I'll ask the GTM guys if they can change this, otherwise, we'll have to document what we change and why in case we need them to help troubleshoot at any time.

xander-marjoram (dismissed):

packages/components/organisms/f-cookie-banner/CHANGELOG.md
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: Done

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: KISS is the real reason - didn't feel like it needed anything more complex

kevinrodrigues (approved):

xander-marjoram (approved):

PR #628, opened 2021-01-21
f-tabs@0.1.0 - generated output

matthewhardern:
Output from generator only. Created component

### DIFF
--- a/packages/components/molecules/f-tabs/package.json
+++ b/packages/components/molecules/f-tabs/package.json
@@ -1,0 +1,55 @@
+{
+  "name": "@justeat/f-tabs",
+  "description": "Fozzie Tabs – Switchable slots for content&#34;",
+  "version": "0.1.0",
+  "main": "dist/f-tabs.umd.min.js",
+  "files": [
+    "dist",
+    "test-utils"
+  ],
+  "homepage": "https://github.com/justeat/fozzie-components/tree/master/packages/f-tabs",
+  "contributors": [
+    "Github contributors <https://github.com/justeat/fozzie-components/graphs/contributors>"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:justeat/fozzie-components.git"
+  },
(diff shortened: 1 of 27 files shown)


packages/components/molecules/f-tabs/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@justeat/f-tabs",
+  "description": "Fozzie Tabs – Switchable slots for content&#34;",
```
ashleynolan, line 3: Is the `&#34` meant to be there?

packages/components/molecules/f-tabs/src/components/Tabs.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+    <div
+        :class="$style['c-tabs']"
+        data-test-id='tabs-component'>
```
kevinrodrigues, line 4: Minor, should be double quotes here. (is this generated like this?)

kevinrodrigues (approved):

packages/components/molecules/f-tabs/package.json
matthewhardern: it was generated this way, will change but can we fix this in generator?

packages/components/molecules/f-tabs/src/components/Tabs.vue
matthewhardern: generated this way, I will make the change

packages/components/molecules/f-tabs/package.json
ashleynolan: Yeh, I'll take a look at that 👍

packages/components/molecules/f-tabs/src/components/Tabs.vue
ashleynolan: I've updated this in the generator now too 👍

dandel10n (approved):

PR #631, opened 2021-01-22
#Trivial Correct README.md headers

BanksySan:
Corrected the header and the Fozzie Bear image path.

### DIFF
--- a/packages/components/atoms/f-form-field/README.md
+++ b/packages/components/atoms/f-form-field/README.md
@@ -1,10 +1,11 @@
+<div style="text-align: center">
 
-<div align="center">
-  <h1>f-form-field</h1>
+# f-form-field
 
-  <img width="125" alt="Fozzie Bear" src="../../bear.png" />
+<img width="125" alt="Fozzie Bear" src="../../../../bear.png" />
 
-  <p>Fozzie Form Field Component</p>
+Fozzie Form Field Component.
+
 </div>
 
 ---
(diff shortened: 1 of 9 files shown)


packages/components/atoms/f-form-field/README.md
```diff
@@ -1,10 +1,11 @@
+<div style="text-align: center">
```
DamianMullins, line 1: Looking at the preview, this section is no longer centre-aligned following this change?

ashleynolan: I think this needs to stay as `align="center"` to keep gh-pages markdown happy.

BanksySan: @DamianMullins @ashleynolan 

Fixed.  Could you have another look?

ashleynolan: More out of interest than anything else – what's the benefit to using `style="text-align: center"` over `align="center"`?

I'm a bit averse to inline styles unless they're absolutely necessary.

DamianMullins: The only working examples I've found online use the `align="center"` syntax — https://gist.github.com/DavidWells/7d2e0e1bc78f4ac59a123ddf8b74932d

ashleynolan: I would stick to the old syntax tbh, unless there's a need to change it. It's just holding up the rest of the changes getting merged and I don't think it really adds anything to change it?

simonsandersje (approved):

oliversweb (approved):

PR #634, opened 2021-01-22
f-searchbox@v4.0.0-beta.20 - Fix dropdown flicker & display loading i…

kevinrodrigues:
### Added
- Spinner loading indicator for Loqate.
- Ability to hide dropdown when final address is picked.

### Fixed
- Drop down minor flicker when setTimeout is used.

### Video
https://www.loom.com/share/98bb00ed16b446ac99e7e88a47c06e6e


packages/components/molecules/f-searchbox/src/components/formElements/FormFullAddressContinueWithSuggestion.vue
```diff
@@ -5,6 +5,7 @@
                 $style['c-continueWithSuggestion-btn'],
                 { [$style.selected]: selected }
             ]"
+            data-test-id="continue-with-suggestion"
```
kevinrodrigues, line 8: @noorshahJustEat Locator for continue with suggestion component.

dandel10n (approved):

simonsandersje (approved):

packages/components/molecules/f-searchbox/src/components/formElements/FormSearchField.vue
```diff
@@ -31,6 +31,12 @@
                 @keydown.down="setKeyboardSuggestion(1)">
 
             <span :class="$style['c-search-placeholder']">{{ copy.fieldPlaceholder }}</span>
+
+            <div
```
AlixRichards, line 35: I might be remembering wrong, but I thought you had separated the loading spinner into its own component, could we use that here/did I make that up? 😂 🤔

AlixRichards (approved):
LGTM - just added a question 👍

packages/components/molecules/f-searchbox/src/components/formElements/FormSearchField.vue
kevinrodrigues: I think maybe it was the SCSS styles I moved to fozzie so it was shareable? at the moment this is pulling from there.

AlixRichards: I think you're right - that's what I'm thinking of 👍

PR #635, opened 2021-01-25
f-checkout@0.43.0 - Exporting module and removing its registration from the Checkout component.

SachaZvetelman:
v0.43.0
*January 25, 2021*

### Changed
- The package now also exports its module, besides the component, so the consuming application can handle the registration of the Vuex module.

### Removed
- Registration of `checkout.module` from the component.
- Duplicated tests in `Checkout.test.js`.


tommcclean (dismissed):

tommcclean (dismissed):

FayeCarter (dismissed):

packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -3,6 +3,18 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.43.0
+-------------------------------
+*January 25, 2021*
+
+### Changed
+- The package now also exports its module, besides the component, so the consuming application can handle the registration of the Vuex module.
```
DamianMullins, line 11: Do you need to update anything in the readme to explain this change?

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -259,6 +240,15 @@ export default {
             'updateCustomerDetails'
         ]),
 
+        /**
+         * Loads the necessary data to render a meaningful checkout component.
+         *
+         */
+        async initialise () {
+            this.setAuthToken(this.authToken);
+            await Promise.all([this.loadCheckout(), this.loadAvailableFulfilment()]);
```
kevinrodrigues, line 249: Do you need to handle any possible rejections here from either of the two array items?

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1252,30 +1252,25 @@ describe('Checkout', () => {
...
-            it('should call `selectionChanged` with the first fulfilment time when there are fulfilment times', async () => {
-                // Arrange
-                const setAuthTokenSpy = jest.spyOn(VueCheckout.methods, 'setAuthToken');
+            it('should call `initialise`', async () => {
+            // Arrange
+                const initialiseSpy = jest.spyOn(VueCheckout.methods, 'initialise');
 
-                const wrapper = shallowMount(VueCheckout, {
+                // Act
+                shallowMount(VueCheckout, {
```
xander-marjoram: Is this testing the watcher or the mounted hook?

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -294,58 +275,77 @@ describe('Checkout', () => {
...
-                propsData
+                expect(loadCheckoutSpy).toHaveBeenCalled();
             });
 
-            expect(loadAvailableFulfilmentSpy).toHaveBeenCalled();
-        });
-    });
+            it('should call `loadAvailableFulfilment`', () => {
+                // Arrange & Act
+                const loadAvailableFulfilmentSpy = jest.spyOn(VueCheckout.methods, 'loadAvailableFulfilment');
```
kevinrodrigues, line 336: Minor one but I think it's better to set the spy up after we've mounted the component. I haven't checked but i'm pretty sure this test would fail (& others would too)  if the methods object was removed from the instance or the `loadAvailableFulfilment ` method was ported to a helper file (mixin maybe)? With the wrapper instance it would fail within the test if the method was removed completely from the vm instance and would be a better failure imo.

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: @kevinrodrigues, handling all the different exceptions (and rejections) will be done in the future as part of a separate PR. This line hasn't been introduced now. It's been moved into a function, though :)

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
SachaZvetelman: Good catch. I copy-pasted a test and didn't update it properly. Thanks, @xander-marjoram. I've sorted it out now.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
SachaZvetelman: Hmmm, haven't seen that behaviour, @kevinrodrigues. You may be right, though.
All our tests in checkout are structured this way. For now, are you ok with leaving it as it is?

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/CHANGELOG.md
SachaZvetelman: Done now ✔️

kevinrodrigues (approved):

tommcclean (approved):

packages/components/organisms/f-checkout/README.md
```diff
@@ -27,18 +27,24 @@
     npm install @justeat/f-checkout
     ```
 
-2.  Import the component
+2.  Import the component and its module, ensuring the latter is registered in the Vuex store upon creation.
```
ashleynolan, line 30: Not a change request – just wanted to know for my own knowledge. Why do we need to do this rather than it just being bundled?

Is it so that it becomes part of the global store?

If that's the case, should we look to namespace the store actions etc that are merged into the global store so we don't hit name collisions down the line?

tommcclean: Before it was only registering the story module on mounted, however the component was being hidden with v-if / v-show, which meant it wasn't being mounted on form load. So pulling it out enables CoreWeb to register the state module, regardless of when the component is actually mounted. It doesn't let you register modules after the page is rendered.

PR #637, opened 2021-01-25
f-checkout@0.44.0 When not logged in, retrieve serviceType from basket endpoint

pkedziora:
When the user is not logged in, retrieve serviceType from global basket orchestrator
This is needed because main Checkout GET endpoint requires authentication now

## UI Review Checks

No UI changes

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -246,6 +257,10 @@ export default {
          */
         async initialise () {
             this.setAuthToken(this.authToken);
+            if (!this.isLoggedIn) {
```
SachaZvetelman, line 261: Ultra nit: add an empty line before this one, please.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -327,11 +330,12 @@ describe('Checkout', () => {
                     localVue,
                     propsData
                 });
+                await flushPromises();
```
SachaZvetelman, line 333: Just curious, but why did we need to add this `flushPromises` in these tests? 🙂

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -341,9 +345,27 @@ describe('Checkout', () => {
...
 
                 expect(loadAvailableFulfilmentSpy).toHaveBeenCalled();
             });
+
+            describe('if isLoggedIn set to `false`', () => {
```
SachaZvetelman, line 353: We should have the opposite case tested as well. If it's set to `true`, we should verify there's no call made.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -948,6 +970,44 @@ describe('Checkout', () => {
             });
         });
 
+        describe('loadBasket ::', () => {
+            describe('when `getBasket` request fails', () => {
+                let wrapper;
+
+                beforeEach(async () => {
```
SachaZvetelman: We don't really need the `beforeEach` in here and the case below, given there's only one test. I'd personally remove it, but up to you.

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -139,6 +139,37 @@ export default {
...
+            tenant,
+            language,
+            timeout
+        }) => {
+            // TODO: deal with exceptions.
+            const config = {
+                method: 'get',
+                headers: {
+                    'Content-Type': 'application/json',
+                    'Accept-Tenant': tenant,
```
SachaZvetelman, line 159: We had to remove this from other the Checkout API calls. Do we need it for this API?

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
pkedziora: The test framework seems to return from mount to the test after first await. I now added another call so the test doesn't wait for this call to complete unless the flushPromises is added

packages/components/organisms/f-checkout/src/store/checkout.module.js
pkedziora: Yep. The basket orchestrator calls in production have it

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
pkedziora: Fixed

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
pkedziora: Fixed

packages/components/organisms/f-checkout/src/components/Checkout.vue
pkedziora: Fixed

packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -195,6 +198,33 @@ describe('CheckoutModule', () => {
...
+                    headers: {
+                        'Content-Type': 'application/json',
+                        'Accept-Tenant': payload.tenant,
+                        'Accept-Language': payload.language
+                    },
+                    timeout: payload.timeout
+                };
+
+                axios.get = jest.fn(() => Promise.resolve({ data: basketDelivery }));
+            });
```
SachaZvetelman: Same as before. We could remove this `beforeEach` in this case. Sorry, didn't catch it in my original review.

pkedziora: Fixed

SachaZvetelman (dismissed):

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -214,6 +245,10 @@ export default {
                 ...state.customer,
                 ...customer
             };
+        },
+
+        UPDATE_BASKET_DETAILS (state, { serviceType }) {
```
kevinrodrigues: Shouldn't these follow the same syntax setup as the other mutations? i.e the `:` & `=>` are missing here and in a few other places. I think it still fine without tho so happy to leave as is for now, might be one for making consistent later.

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -139,6 +139,37 @@ export default {
...
+                    'Content-Type': 'application/json',
+                    'Accept-Tenant': tenant,
+                    'Accept-Language': language
+                },
+                timeout
+            };
+
+            const { data } = await axios.get(url, config);
+            const basketDetails = {
+                serviceType: data.ServiceType.toLowerCase()
```
kevinrodrigues, line 167: Do you need to check that data exists before trying to access it? Will it always be available? If not you might get an error at `ServiceType`/

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -948,6 +987,36 @@ describe('Checkout', () => {
...
+                const wrapper = mount(VueCheckout, {
+                    store: createStore(),
+                    i18n,
+                    localVue,
+                    propsData
+                });
+
+                it('should emit success event', async () => {
+                    expect(wrapper.emitted(EventNames.CheckoutBasketGetSuccess).length).toBe(1);
+                    expect(wrapper.emitted(EventNames.CheckoutBasketGetFailure)).toBeUndefined();
```
kevinrodrigues, line 1021: Probably worth having this a separate `IT`:

`it('should not emit a 'CheckoutBasketGetFailure' when basket is successful')`

kevinrodrigues: In fact i dont think you need this one as the one above looks to cover the failure. So this test really only needs the expect on the line above. Then the test above only needs the failure one.

packages/components/organisms/f-checkout/src/store/checkout.module.js
pkedziora: Fixed

packages/components/organisms/f-checkout/src/store/checkout.module.js
pkedziora: It should always be available

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
pkedziora: Fixed

kevinrodrigues (approved):

SachaZvetelman (approved):

PR #642, opened 2021-01-27
f-checkout@0.45.0 - Only call the Checkout `GET` endpoint when the user is logged in

SachaZvetelman:
v0.45.0
*January 27, 2021*

### Changed
- Only call the Checkout `GET` endpoint when the user is logged in.
- Refactored tests.

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -256,17 +256,17 @@ export default {
          *
          */
         async initialise () {
+            const initialLoadPromises = [this.loadAvailableFulfilment()];
+
             this.setAuthToken(this.authToken);
 
-            if (!this.isLoggedIn) {
+            if (this.isLoggedIn) {
+                initialLoadPromises.push(this.loadCheckout());
+            } else {
                 await this.loadBasket();
             }
 
-            const promises = this.isLoggedIn
-                ? [this.loadCheckout(), this.loadAvailableFulfilment()]
-                : [this.loadAvailableFulfilment()];
-
-            await Promise.all(promises);
+            await Promise.all(initialLoadPromises);
         },
 
         /**
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -262,7 +262,11 @@ export default {
                 await this.loadBasket();
             }
 
-            await Promise.all([this.loadCheckout(), this.loadAvailableFulfilment()]);
+            const promises = this.isLoggedIn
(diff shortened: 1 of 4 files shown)


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -256,13 +256,17 @@ export default {
...
-            if (!this.isLoggedIn) {
+            if (this.isLoggedIn) {
+                initialLoadPromises.push(this.loadCheckout());
+            } else {
                 await this.loadBasket();
             }
 
-            await Promise.all([this.loadCheckout(), this.loadAvailableFulfilment()]);
+            await Promise.all(initialLoadPromises);
         },
```
SachaZvetelman: I personally don't like how this looks in the end. Any thoughts?

pkedziora: I prefer Promise.all([this.loadCheckout(), this.loadAvailableFulfilment()]); as it's more readable, does not require to look for the definition of initialLoadPromises

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -365,6 +350,21 @@ describe('Checkout', () => {
...
+                    const loadCheckoutSpy = jest.spyOn(VueCheckout.methods, 'loadCheckout');
+
+                    shallowMount(VueCheckout, {
+                        store: createStore(),
+                        i18n,
+                        localVue,
+                        propsData
+                    });
+                    await flushPromises();
+
```
FayeCarter, line 368: Do you need an `// Assert` here?

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: But we don't always call `loadCheckout`. That's the method we need to call only if we're authenticated.

kevinrodrigues: You could do the below instead? Then you could switch the condition above for the loadBasket() call instead.

```javascript
this.isLoggedIn
? Promise.all([this.loadCheckout(), this.loadAvailableFulfilment()])
: Promise.all([this.loadAvailableFulfilment()])

SachaZvetelman: I had pretty much that in my previous commit (https://github.com/justeat/fozzie-components/pull/642/commits/3abfc4bd41dbdb5c941d53f5eabfd44c5ce92583) but didn't really like it either. But I guess it's still better than what I have now.

FayeCarter (approved):

kevinrodrigues (approved):

PR #646, opened 2021-01-27
f-checkout@0.46.0 - Submit checkout data to CheckoutApi

xander-marjoram:
### Changed
- Reduce nesting within state by moving `address` and `time` out of `fulfilment`
- Update checkout URL
- Fill out checkout post data
- Use latest `f-form-field` with separate text/value
- Update tests

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -6,7 +6,7 @@
                 {{ $t('labels.addressGroup') }}
             </legend>
             <form-field
-                :value="fulfilment.address.line1"
```
xander-marjoram, line 9: @oliversweb and I thought that the nesting here was unnecessary, as the fulfilment address will be the only address on the page.

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -40,36 +40,48 @@ export default {
...
-                ? [this.$t('asapFulfilmentOption')]
-                : [];
-
+            const times = [];
             this.availableFulfilment.times.forEach(time => {
                 const formattedTime = this.$d(new Date(time.from), 'short');
 
-                times.push(formattedTime);
+                times.push({
+                    text: formattedTime,
```
xander-marjoram, line 48: The latest f-form-field has support for different text and value fields, so we can show something like "Wednesday 12:00" in the dropdown but when we submit to the API it will be in ISO format.

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -40,36 +40,48 @@ export default {
...
         /**
         * Update the selected available fulfilment time.
         *
         * @param {string} selectedFulfilmentTime The time emited when dropdown value is changed.
         **/
         selectionChanged (selectedFulfilmentTime) {
             this.selectedAvailableFulfilmentTime = selectedFulfilmentTime;
+            this.updateFulfilmentTime({
+                from: selectedFulfilmentTime,
+                to: selectedFulfilmentTime
```
xander-marjoram, line 83: I understand the APIs will support different from/to times at a later date? We can update this here when that time comes.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -276,19 +277,30 @@ export default {
         async submitCheckout () {
             try {
                 const checkoutData = {
-                    mobileNumber: this.customer.mobileNumber
+                    customer: {
```
xander-marjoram, line 285: This can potentially be pulled out into a mapper to keep this method trimmed down.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -276,19 +277,30 @@ export default {
...
-                }
-
                 if (!this.isLoggedIn) {
                     await this.setupGuestUser();
                 }
 
-                await this.postCheckout({
-                    url: 'myPostUrl',
+                await this.patchCheckout({
+                    url: `${this.tenant}/checkout`,
```
xander-marjoram: Currently missing `/checkoutId` on the end.

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: Yeah, I think we should do that so the `submitCheckout` method is small and simple.

SachaZvetelman (dismissed):

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -40,36 +40,48 @@ export default {
...
+            if (this.availableFulfilment.isAsapAvailable && times.length) {
+                times[0].text = this.$t('asapFulfilmentOption');
+            }
+
             return times;
         }
     },
 
     watch: {
         'fulfilmentTimes' () {
```
kevinrodrigues: does this need to be in a quote?

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -40,36 +40,48 @@ export default {
...
             if (this.fulfilmentTimes.length > 0) {
```
kevinrodrigues: Doesn't have to change here but can just be `this.fulfilmentTimes.length` as the length check will return a boolean.

packages/components/organisms/f-checkout/src/components/Selector.vue
kevinrodrigues: Is it worth adding a comment here about here. It's hard to see that the `selectedFulfilmentTime` differs on glance.

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -40,36 +40,48 @@ export default {
...
-                this.selectionChanged(this.fulfilmentTimes[0]);
+                this.selectionChanged(this.fulfilmentTimes[0].value);
```
kevinrodrigues: You should be able to pull out the `fulfilmentTimes` from the watcher param instead?

packages/components/organisms/f-checkout/src/components/Selector.vue
SachaZvetelman: Not anymore, actually. It had to, given it was a nested object.

packages/components/organisms/f-checkout/src/components/Checkout.vue
xander-marjoram: @SachaZvetelman Do you mind if I do this in a follow-up PR?

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -40,36 +40,50 @@ export default {
...
             return times;
         }
     },
 
     watch: {
-        'fulfilmentTimes' () {
-            if (this.fulfilmentTimes.length > 0) {
-                this.selectionChanged(this.fulfilmentTimes[0]);
+        fulfilmentTimes (newFulfilmentTimes) {
+            if (newFulfilmentTimes.length) {
```
kevinrodrigues: Do you need the length check now? The watcher should only trigger when there's a change to the times now so i assume it will always be populated unless it ever gets cleared to a empty array?

xander-marjoram: That's a good point. I can't think of a situation where we try to push an empty array but if that does happen in future we can fix it then. It's arguably more predictable without the length check because otherwise the value that gets submitted might be out-of-sync with what's in the dropdown.

kevinrodrigues (approved):

zlatinivanov-je (approved):

PR #647, opened 2021-01-27
f-button@v0.7.0 — Update button focus styles.

DamianMullins:
### Changed

- Updated focus styles to provide more control over when focus outlines are displayed. This allows us to only show the focus outlines when using a keyboard. See this article for more info - https://matthiasott.com/notes/focus-visible-is-here.
- Updated package URL to match new structure.

### DIFF
--- a/packages/components/atoms/f-button/src/components/Button.vue
+++ b/packages/components/atoms/f-button/src/components/Button.vue
@@ -144,7 +144,7 @@ $btn-sizeXSmall-lineHeight      : 1;
 
     // Show focus styles on keyboard focus.
     &:focus-visible {
-        outline: 2px solid $btn-default-outline-color;
+        outline: 2px dotted $btn-default-outline-color;
     }
 
     &:hover,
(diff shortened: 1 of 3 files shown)


kevinrodrigues (dismissed):

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -137,14 +137,20 @@ $btn-sizeXSmall-lineHeight      : 1;
...
-    &:focus {
-        outline: 2px dotted $btn-default-outline-color;
+    // Hide focus styles if they're not needed, for example, when an element receives focus via the mouse.
+    &:focus:not(:focus-visible) {
+        outline: 0;
+    }
+
+    // Show focus styles on keyboard focus.
+    &:focus-visible {
+        outline: 2px solid $btn-default-outline-color;
```
dandel10n: I thought that this dotted outline came from new pie design

DamianMullins: Have just checked and, yes. it looks like buttons should have dotted outlines. They should also have a rounded outline but I've not been able to figure out how to do that with `outline` as there is no outline radius property. 

I'll set this back to dotted outlines and will have a think about the rounded styles.

kevinrodrigues (approved):

dandel10n (approved):

PR #649, opened 2021-01-28
f-footer@4.8.0 - showCountrySelector prop

dandel10n:
### Added
- `showCountrySelector` property to turn it off if there is one in the header


ashleynolan (approved):
Will the one in the footer be used in some places? Just wondering if it can be removed at a later date

xander-marjoram (approved):

PR #654, opened 2021-01-28
f-tabs@0.2.0 - First changes, adds styles, layout and components

matthewhardern:
This is the beginning of a tabs component, at a high level it is required to accomplish part of a design in the new stamp cards page. The component consists of two new components `Tab` and `Tabs`. 

The underlying choice of how to handle the way the tabs worked was made based on a few points.

	1. The tabs should be allowed to animate if needed
	2. Only `Tab` ’s should be allowed to be registered as a tab within the `Tabs` slot. 
	3. It must not be complicated to implement.

Based on this I decided not to use`$children` as an option for accessing the slot contents. It’s removed in the next version of vue and it’s also very painful to work with `<transition>` elements.   

I have therefore used`provide/inject` to inject a callback function for the `Tab` component to register itself with the parent `Tabs` component.  This way we have access to more data on the child component / `Tab` as it is able to send tab data when registering, so we can then use the title prop to add the text to the tab button. 

For this PR I am not including unit tests and will add in a later PR to reduce the PR size. 

All style choices are made based on the design given and until we need to modify this e.g we have a PIE design for tabs then I thought best to style tabs buttons according to design. 

Animation can be applied by adding `:animate=“true”` to the `Tabs` component.

## UI Review Checks

Screenshot 2021-01-28 at 14 36 04

Screenshot 2021-01-28 at 14 36 42

Below is snippet from design file.

Screenshot 2021-01-28 at 14 43 56

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/molecules/f-tabs/src/components/Tab.vue
+++ b/packages/components/molecules/f-tabs/src/components/Tab.vue
@@ -1,0 +1,123 @@
+<template>
+    <transition
+        v-if="animateTab"
+        :name="transitionName">
+        <div
+            v-show="isActive"
+            :class="[$style['c-tab']]">
+            <slot />
+        </div>
+    </transition>
+    <div v-else>
+        <div
+            v-show="isActive"
+            :class="[$style['c-tab']]">
+            <slot />
+        </div>
+    </div>
+</template>
+
+<script>
+export default {
+    name: 'Tab',
+    props: {
+        title: {
+            required: true,
+            type: String
+        },
+        name: {
+            required: true,
+            type: String
+        },
+        selected: {
+            required: false,
+            type: Boolean,
+            default: false
+        }
+    },
+    inject: ['register', 'tabsComponent'],
+    computed: {
+        isActive () {
+            return this.tabsComponent.activeTab === this.name;
+        },
+        transitionName () {
+            return this.tabsComponent.animationDirection === 'LEFT' ?
+                this.$style['fade-in-right'] : this.$style['fade-in-left'];
+        },
+        animateTab () {
+            return this.tabsComponent.animate;
+        }
+    },
+    created () {
+        this.register({
+            name: this.name,
+            title: this.title,
+            selected: this.selected
+        });
+    }
+};
+</script>
+
+<style lang="scss" module>
+/* stylelint-disable -- does not understand syntax. */
+.c-tab {
+    padding: spacing() 0;
+    left: 0;
+    top: 0;
+    grid-area: 1/1;
+}
+@keyframes fadeInLeft {
+    from {
+        transform: translate3d(-40px, 0, 0);
+    }
+
+    to {
+        transform: translate3d(0, 0, 0);
+        opacity: 1;
+    }
+}
+
+.fade-in-left {
+    &:global(-enter-to) {
+        opacity: 0;
+        animation-duration: 0.7s;
+        animation-fill-mode: both;
+        animation-name: fadeInLeft;
+    }
+    &:global(-enter) {
+        opacity: 0;
+        transform: translate3d(-40px, 0, 0);
+    }
+    &:global(-leave-to) {
+        opacity: 0;
+        transition: opacity 0.3s;
+    }
+}
+@keyframes fadeInRight {
+    from {
+        transform: translate3d(40px, 0, 0);
+    }
+
+    to {
+        transform: translate3d(0, 0, 0);
+        opacity: 1;
+    }
+}
+
+.fade-in-right {
+    &:global(-enter-to) {
+        opacity: 0;
+        animation-duration: 0.7s;
+        animation-fill-mode: both;
+        animation-name: fadeInRight;
+    }
+    &:global(-enter) {
+        opacity: 0;
+        transform: translate3d(40px, 0, 0);
+    }
+    &:global(-leave-to) {
+        opacity: 0;
+        transition: opacity 0.3s;
+    }
+}
+</style>
--- a/packages/components/molecules/f-tabs/src/components/Tab.vue
+++ b/packages/components/molecules/f-tabs/src/components/Tab.vue
@@ -30,7 +30,6 @@ export default {
             type: String
         },
         selected: {
-            required: false,
             type: Boolean,
             default: false
         }
--- a/packages/components/molecules/f-tabs/src/components/Tab.vue
+++ b/packages/components/molecules/f-tabs/src/components/Tab.vue
@@ -4,20 +4,22 @@
         :name="transitionName">
         <div
             v-show="isActive"
-            :class="[$style['c-tab']]">
+            :class="$style['c-tab']">
             <slot />
         </div>
     </transition>
     <div v-else>
         <div
             v-show="isActive"
-            :class="[$style['c-tab']]">
+            :class="$style['c-tab']">
             <slot />
         </div>
     </div>
 </template>
 
 <script>
+import { DIRECTION } from '../constants';
+
 export default {
     name: 'Tab',
     props: {
@@ -40,7 +42,7 @@ export default {
             return this.tabsComponent.activeTab === this.name;
         },
         transitionName () {
-            return this.tabsComponent.animationDirection === 'LEFT' ?
+            return this.tabsComponent.animationDirection === DIRECTION.LEFT ?
                 this.$style['fade-in-right'] : this.$style['fade-in-left'];
         },
         animateTab () {
@@ -58,16 +60,22 @@ export default {
 </script>
 
 <style lang="scss" module>
-/* stylelint-disable -- does not understand syntax. */
+
+$fade-out-transition-duration: 0.3s;
+$fade-in-transition-duration: 0.7s;
+$fade-in-transition-distance: 40px;
+
+
 .c-tab {
     padding: spacing() 0;
     left: 0;
     top: 0;
     grid-area: 1/1;
 }
+
 @keyframes fadeInLeft {
     from {
-        transform: translate3d(-40px, 0, 0);
+        transform: translate3d(-#{$fade-in-transition-distance}, 0, 0);
     }
 
     to {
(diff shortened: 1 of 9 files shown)


packages/components/molecules/f-tabs/src/components/Tab.vue
```diff
@@ -0,0 +1,123 @@
...
+        title: {
+            required: true,
+            type: String
+        },
+        name: {
+            required: true,
+            type: String
+        },
+        selected: {
+            required: false,
```
dandel10n: You can remove required from here as you have a default

packages/components/molecules/f-tabs/src/components/Tabs.vue
```diff
@@ -1,46 +1,144 @@
...
-    data () {
-        const locale = globalisationServices.getLocale(tenantConfigs, this.locale, this.$i18n);
-        const localeConfig = tenantConfigs[locale];
-
+    data: () => ({
+        direction: null,
+        activeTab: '',
+        tabs: []
+    }),
+    computed: {},
```
dandel10n: Empty `computed` can be removed

dandel10n (commented):
Few minor comments from me.

I never used vue `inject` so would be really cool if you could have a short run through of the code/demo to show how it works.

matthewhardern:
> Few minor comments from me.
> 
> I never used vue `inject` so would be really cool if you could have a short run through of the code/demo to show how it works.

Happy to put in a demo / run through if people want

minrwhite (dismissed):
Happy with the functionality included here - as I believe was discussed elsewhere, testing will be added following delivery of the functional component into the stamp card page due to time constraints; I'd also second the request for a demo to share with more of the web core team how this component operates and how it currently is intended to be used. You can use the opportunity to capture details of any questions that are raised as part of that session that should be recorded in the README for this component, and include that as part of the PR that adds the testing.

packages/components/molecules/f-tabs/src/components/Tab.vue
```diff
@@ -0,0 +1,122 @@
...
+            name: this.name,
+            title: this.title,
+            selected: this.selected
+        });
+    }
+};
+</script>
+
+<style lang="scss" module>
+/* stylelint-disable -- does not understand syntax. */
```
ashleynolan: Out of interest – which syntax doesn't it understand? (is it the keyframes?)

packages/components/molecules/f-tabs/src/components/Tab.vue
```diff
@@ -0,0 +1,122 @@
...
+.c-tab {
+    padding: spacing() 0;
+    left: 0;
+    top: 0;
+    grid-area: 1/1;
+}
+@keyframes fadeInLeft {
```
ashleynolan, line 76: Super NIT – can you add a line break between the `}` and the start of this keyframe block

packages/components/molecules/f-tabs/src/components/Tab.vue
```diff
@@ -0,0 +1,122 @@
...
+    &:global(-enter) {
+        opacity: 0;
+        transform: translate3d(-40px, 0, 0);
+    }
+    &:global(-leave-to) {
+        opacity: 0;
+        transition: opacity 0.3s;
+    }
+}
+@keyframes fadeInRight {
```
ashleynolan, line 106: Super NIT – can you add a line break between the `}` and the start of this keyframe block

packages/components/molecules/f-tabs/src/components/Tab.vue
```diff
@@ -0,0 +1,122 @@
...
+.fade-in-left {
+    &:global(-enter-to) {
+        opacity: 0;
+        animation-duration: 0.7s;
+        animation-fill-mode: both;
+        animation-name: fadeInLeft;
+    }
+    &:global(-enter) {
+        opacity: 0;
+        transform: translate3d(-40px, 0, 0);
```
ashleynolan: I would move out the `40px` as a variable so that it can be used on line 70, here, line 97 and line 115 (prefixing with `-` where necessary)

packages/components/molecules/f-tabs/src/components/Tab.vue
```diff
@@ -0,0 +1,122 @@
...
+    }
+    &:global(-leave-to) {
+        opacity: 0;
+        transition: opacity 0.3s;
```
ashleynolan: Same with the transition duration – I'd pull this out as a variable so it can be changed for both the left and right durations using the same variable.

packages/components/molecules/f-tabs/src/components/Tab.vue
```diff
@@ -0,0 +1,122 @@
...
+    to {
+        transform: translate3d(0, 0, 0);
+        opacity: 1;
+    }
+}
+
+.fade-in-left {
+    &:global(-enter-to) {
+        opacity: 0;
+        animation-duration: 0.7s;
```
ashleynolan: And the same for this duration – would be useful to have it pulled out as a variable so that this and the usage on line 109 stay consistent.

packages/components/molecules/f-tabs/src/components/Tabs.vue
```diff
@@ -1,46 +1,142 @@
...
-        data-test-id='tabs-component'>
-        {{ copy.text }}
+        data-test-id="tabs-component"
+        :class="$style['c-tabs']">
+        <div :class="$style['c-tabs-header']">
+            <div :class="$style['c-tabs-buttons']">
+                <button
+                    v-for="({ name, title }, i) in tabs"
+                    :key="i"
+                    :class="[{ [$style['c-tabs-button--active']]: activeTab === name }, $style['c-tabs-button']]"
```
ashleynolan:
```suggestion
                    :class="[
                        (activeTab === name ? $style['c-tabs-button--active' : ''),
                        $style['c-tabs-button']
                    ]"
```

packages/components/molecules/f-tabs/src/components/Tabs.vue
```diff
@@ -1,46 +1,142 @@
...
+            <slot />
+        </div>
     </div>
 </template>
 
 <script>
-import { globalisationServices } from '@justeat/f-services';
-import tenantConfigs from '../tenants';
+
+const DIRECTION = {
```
ashleynolan: Could pull these out into a constants file – that way you could also reference it in `transitionName` in the `Tab.vue`.

packages/components/molecules/f-tabs/src/components/Tabs.vue
```diff
@@ -1,46 +1,142 @@
...
+    LEFT: 'LEFT',
+    RIGHT: 'RIGHT'
+};
 
 export default {
-    name: 'VueTabs',
-    components: {},
+    name: 'Tabs',
     props: {
         locale: {
```
ashleynolan: Looks like you can probably delete the locale prop now?

packages/components/molecules/f-tabs/src/components/Tabs.vue
```diff
@@ -1,46 +1,142 @@
...
         return {
-            copy: { ...localeConfig }
+            register (tab) {
+                component.addTab(tab);
+            },
+            tabsComponent
         };
+    },
+    methods: {
+        selectTabIndex (name) {
```
ashleynolan, line 71: Could you add some JSDoc comments for these methods?

packages/components/molecules/f-tabs/src/components/Tabs.vue
```diff
@@ -1,46 +1,142 @@
...
+    padding: spacing(x1.5) spacing(x2);
+    cursor: pointer;
+    color: $tabs-link-colour;
+    border: none;
+    background-color: transparent;
+    outline: none;
+
+    @include media('<narrowMid') {
+        /* stylelint-disable */
+        flex: 1 1 0px;// important px for IE
```
ashleynolan, line 132: Why's this important for IE? Does the flex styling not work otherwise?

packages/components/molecules/f-tabs/src/index.js
```diff
@@ -6,13 +6,15 @@
 
 
 // Import vue component
-import VueTabs from '@/components/Tabs.vue';
+import Tabs from '@/components/Tabs.vue';
+import Tab from '@/components/Tab.vue';
```
ashleynolan, line 10: Guessing there's a reason, but what's the reason to export both components, rather than pull in the `Tab` component into the `Tabs.vue`?

packages/components/molecules/f-tabs/src/components/Tab.vue
matthewhardern: It's actually the `:global(-enter-to) {` that causes the issue. For some reason it does not see it as valid syntax. I will reduce the disabled area scope to just those bits though

packages/components/molecules/f-tabs/src/components/Tabs.vue
matthewhardern: Unsure as to the benefit of `(activeTab === name ? $style['c-tabs-button--active' : '')`, would you be able to explain?

packages/components/molecules/f-tabs/src/components/Tabs.vue
matthewhardern: IE has a problem parsing the flex property. For some reason it can not handle the 0 without a unit apparently, something I have been made aware of.

packages/components/molecules/f-tabs/src/components/Tabs.vue
ashleynolan: It's ultimately styliistic preference, but the reasons I prefer it are that it puts the conditional at the front (`if X then Y` rather than `Y if X`) and it is a little cleaner to me with the amount of brackets needed (you don't need to wrap the `$style` in extra `[]` using this way)

packages/components/molecules/f-tabs/src/index.js
matthewhardern: Well the thought process was that technically you cou make your own version of `Tab.vue` if you needed a really custom tab for some reason. As you can register to the `Tabs.vue` by injecting the provided callback

ashleynolan: Ok, cool. As long as we have it documented why we've done that in the README, then I'm happy.

packages/components/molecules/f-tabs/src/components/Tabs.vue
matthewhardern: ATM i will leave the stylistic change until people agree on a format. Will adjust the rest

packages/components/molecules/f-tabs/src/index.js
matthewhardern: I have added a bit into the readme to explain

minrwhite (approved):

packages/components/molecules/f-tabs/src/components/Tabs.vue
```diff
@@ -1,46 +1,147 @@
...
+.c-tabs-button {
+    padding: spacing(x1.5) spacing(x2);
+    cursor: pointer;
+    color: $tabs-link-colour;
+    border: none;
+    background-color: transparent;
+    outline: none;
+
+    @include media('<narrowMid') {
+        /* stylelint-disable */
```
ashleynolan, line 131: Not a big one by any means (can be in a future PR), but would be cleaner to use:

`/* stylelint-disable-next-line whatever-the-stylelint-rule-is */`

ashleynolan (approved):
Looks good – one super small comment, but happy for that to be fixed in a future PR

packages/components/molecules/f-tabs/src/components/Tabs.vue
matthewhardern: @ashleynolan thanks for that, was trying to figure out what the one line version was, will make change in future PR

PR #651, opened 2021-01-28
f-searchbox@v4.0.0 beta.21 popup overlay

kevinrodrigues:
PR adds pop over overlay on small screens for Loqate. Using `f-mega-modal`

### Added
- Readme information on how to render f-searchbox in non-vue instances.
- Tests to cover changes.
- `FormFullAddressSearchModalOverlay.vue` - Allows small screens to display results in a modal (`f-mega-modal`).

### Fixed
- Casing in readme non-vue instructions.

### Video
https://www.loom.com/share/88443eda46bd45249cc0ae8e8dbeaa36

PR #659, opened 2021-01-29
#trivial - Reduce bundlewatch maxsize & target umd js files.

kevinrodrigues:
Noticed my [PR](https://github.com/justeat/fozzie-components/pull/651) is breaking in CI this will target the umd files. Have reduced the maxsize so the umd files do not exceed massively.

### Changes
- Bundle watch maxSize value to `100kb` & target files `umd.min.js`.


DamianMullins (dismissed):

DamianMullins (dismissed):

matthewhardern (dismissed):

xander-marjoram:
Looks like you may need to add some more exclusions to the bundlewatch config file

kevinrodrigues:
> Looks like you may need to add some more exclusions to the bundlewatch config file

> Looks like you may need to add some more exclusions to the bundlewatch config file

Ill pick this up on monday as ill be back in the web team then. Need to look into the patterns we need.

DamianMullins:
Looks like the oidc package is failing now, but as it bundles to UMD you should be able to [update the generated filename to include the `umd` part](https://github.com/justeat/fozzie-components/blob/9e6bfd1c38d26b636be89ca10412f9474b0c9d72/packages/services/f-consumer-oidc/webpack.config.js#L21).

kevinrodrigues:
> Looks like the oidc package is failing now, but as it bundles to UMD you should be able to [update the generated filename to include the `umd` part](https://github.com/justeat/fozzie-components/blob/9e6bfd1c38d26b636be89ca10412f9474b0c9d72/packages/services/f-consumer-oidc/webpack.config.js#L21).

Yup I added this locally earlier & works ok for this package. The issue is that if it's added then other umd files fail the 100kb limit now. 😄

PR #666, opened 2021-02-01
#trivial Update bundle watch max size fix

kevinrodrigues:
### Fixed
- `bundlewatch` config updated in line with new structure.


dandel10n (dismissed):

bundlewatch.config.js
```diff
@@ -36,8 +36,8 @@ packageFolders.forEach(folder => {
 const filteredPackages = packageNames.filter(package => !excludedPackages.includes(package));
 
 const files = filteredPackages.map(package => ({
-    path: `${package}/dist/*.js`,
-    maxSize: '200kB'
+    path: `${package}/dist/*+(.min|.min.umd|es.js).js`,
```
xander-marjoram: Is the potential for `es.js.js` correct?

kevinrodrigues: Good spot. Should be `.es` ill just re-check locally and push. 👍

kevinrodrigues: Actually just realised we don't need this check as vue icons outputs a min file too.

DamianMullins: Might be good to keep it in there as it is a valid bundle target when a build tool supports it via the "modules" field in package.json so it would be good to know if that bundle ever goes over the threshold. We set that field to enable to tree-shaking unused icons.

kevinrodrigues: Good point. Added it back 👍

DamianMullins (approved):

tommcclean (approved):

PR #667, opened 2021-02-01
f-media-element@0.2.0 - layout +  styles + components

matthewhardern:
This is the start of a new media element component. It's designed to provide a component that can take in a Title, text and image and display them in a bunch of orientations and alignments. It also allows changing text size to be one from SM, MD, LG, XL, XXL. 

The component consists of two equal spaced blocks one the content the other the image, this allows for full flexibility.

I will be adding some unit tests in an additional PR to allow this to be a smaller PR. 

## UI Review Checks

Example being used:

Screenshot 2021-02-01 at 15 51 26

Some other examples: 

Screenshot 2021-02-01 at 15 44 36
Screenshot 2021-02-01 at 15 45 11
Screenshot 2021-02-01 at 15 45 32

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/molecules/f-media-element/src/components/MediaElement.vue
+++ b/packages/components/molecules/f-media-element/src/components/MediaElement.vue
@@ -9,20 +9,13 @@
         <div
             :class="[
                 $style['c-mediaElement-content'],
-                contentAlignClass
+                contentAlignClass,
+                fontSizeClass
             ]">
-            <h3
-                :class="[
-                    $style['c-mediaElement-title'],
-                    fontSizeClass
-                ]">
+            <h3 :class="$style['c-mediaElement-title']">
                 {{ title }}
             </h3>
-            <p
-                :class="[
-                    $style['c-mediaElement-text'],
-                    fontSizeClass
-                ]">
+            <p :class="$style['c-mediaElement-text']">
                 {{ text }}
             </p>
         </div>
@@ -34,7 +27,7 @@
             <img
                 :class="[$style['c-mediaElement-img'], 'c-mediaElement-img--override']"
                 :src="imageUrl"
-                alt="Media Element Image">
+                :alt="altText">
         </div>
     </div>
 </template>
@@ -75,6 +68,10 @@ export default {
         textSize: {
             type: String,
             default: 'md'
+        },
+        altText: {
+            type: String,
+            default: 'Media Element Image'
         }
     },
     data () {
@@ -122,17 +119,17 @@ export default {
         fontSizeClass () {
             switch (this.textSize) {
                 case FONT_SIZE.SM:
-                    return this.$style['c-mediaElement-fontSize--sm'];
+                    return this.$style['c-mediaElement--fontSizeSmall'];
                 case FONT_SIZE.MD:
-                    return this.$style['c-mediaElement-fontSize--md'];
+                    return this.$style['c-mediaElement--fontSizeMedium'];
                 case FONT_SIZE.LG:
-                    return this.$style['c-mediaElement-fontSize--lg'];
+                    return this.$style['c-mediaElement--fontSizeLarge'];
                 case FONT_SIZE.XL:
-                    return this.$style['c-mediaElement-fontSize--xl'];
+                    return this.$style['c-mediaElement--fontSizeXLarge'];
                 case FONT_SIZE.XXL:
-                    return this.$style['c-mediaElement-fontSize--xxl'];
+                    return this.$style['c-mediaElement--fontSizeXXLarge'];
                 default:
-                    return this.$style['c-mediaElement-fontSize--md'];
+                    return this.$style['c-mediaElement--fontSizeMedium'];
             }
         }
     }
@@ -141,15 +138,58 @@ export default {
 
 <style lang="scss" module>
 
+$font-sizes: (
+    fontSizeSmall: (
+        title: heading-s,
+        text: body-s,
+    ),
+    fontSizeMedium: (
+        title: heading-m,
+        text: body-l,
+    ),
+    fontSizeLarge: (
+        title: heading-l,
+        text: subheading-s,
+    ),
+    fontSizeXLarge: (
+        title: heading-xl,
+        text: subheading-s,
+    ),
+    fontSizeXXLarge: (
+        title: heading-xxl,
+        text: subheading-l,
+    )
+);
+
 .c-mediaElement {
     display: flex;
     width: 100%;
 }
+
+/**
+ * Modifier – .c-mediaElement--stack
+ *
+ * Applies flex direction column
+ */
 .c-mediaElement--stack {
     flex-direction: column;
-}
-.c-mediaElement--stack.c-mediaElement--reverse {
-    flex-direction: column-reverse;
+
+    /**
+     * Modifier – .c-mediaElement--reverse
+     *
+     * When stacked in flex col applies col-reverse
+     */
+    &.c-mediaElement--reverse {
+        flex-direction: column-reverse;
+
+        & .c-mediaElement-text {
+            margin-bottom: spacing(x3);
+        }
+    }
+
+    & .c-mediaElement-title {
+        margin-top: spacing(x3);
+    }
 }
 
 .c-mediaElement-content {
@@ -158,63 +198,31 @@ export default {
     justify-content: center;
     flex: 1 1 0;
 
-    &--center {
+    /**
+     * Modifier – .c-mediaElement-content--center
+     *
+     * Aligns the content center using flex within the container
+     */
+    &.c-mediaElement-content--center {
         align-items: center;
         text-align: center;
     }
-    &--left {
+    /**
+     * Modifier – .c-mediaElement-content--left
+     *
+     * Aligns the content left using flex within the container
+     */
+    &.c-mediaElement-content--left {
         align-items: flex-start;
     }
-    &--right {
+    /**
+     * Modifier – .c-mediaElement-content--right
+     *
+     * Aligns the content right using flex within the container
+     */
+    &.c-mediaElement-content--right {
         align-items: flex-end;
-    }
-}
-
-.c-mediaElement-title.c-mediaElement-fontSize {
-    &--sm {
-        @include font-size(heading-s);
-    }
-    &--md {
-        @include font-size(heading-m);
-    }
-    &--lg {
-        @include font-size(heading-l);
-        margin-bottom: spacing(x0.5);
-    }
-    &--xl {
-        @include font-size(heading-xl);
-        margin-bottom: spacing(x0.5);
-    }
-    &--xxl {
-        @include font-size(heading-xxl);
-        margin-bottom: spacing(x0.5);
-    }
-}
-
-.c-mediaElement--stack .c-mediaElement-title {
-    margin-top: spacing(x3);
-}
-
-.c-mediaElement-text.c-mediaElement-fontSize {
-    @include font-size(body-l);
-    margin-top: spacing(x0.5);
-    &--sm {
-        @include font-size(body-s);
-    }
-    &--md {
-        @include font-size(body-l);
-    }
-    &--lg {
-        @include font-size(subheading-s);
-        margin-bottom: spacing(x0.5);
-    }
-    &--xl {
-        @include font-size(subheading-s);
-        margin-bottom: spacing(x0.5);
-    }
-    &--xxl {
-        @include font-size(subheading-l);
-        margin-bottom: spacing(x0.5);
+        text-align: right;
     }
 }
 
@@ -223,14 +231,52 @@ export default {
     flex-direction: column;
     flex: 1 1 0;
 
-    &--center {
+    /**
+     * Modifier – .c-mediaElement-imgWrapper--center
+     *
+     * Aligns the image center using flex within the container
+     */
+    &.c-mediaElement-imgWrapper--center {
         align-items: center;
     }
-    &--left {
+    /**
+     * Modifier – .c-mediaElement-imgWrapper--left
+     *
+     * Aligns the image left using flex within the container
+     */
+    &.c-mediaElement-imgWrapper--left {
         align-items: flex-start;
     }
-    &--right {
+    /**
+     * Modifier – .c-mediaElement-imgWrapper--right
+     *
+     * Aligns the image right using flex within the container
+     */
+    &.c-mediaElement-imgWrapper--right {
         align-items: flex-end;
+    }
+}
+
+/**
+ * Modifier – .c-mediaElement--$size
+ *
+ * For each of the above map values under $font-sizes we loop over them applying the value to both title
+ * and text depending on the key using map-get()
+ */
+@each $size, $value in $font-sizes {
+    .c-mediaElement--#{$size} {
+        & .c-mediaElement-title {
+            @include font-size(map-get($value, 'title'));
+        }
+        & .c-mediaElement-text {
+            @include font-size(map-get($value, 'text'));
+            @if $size == fontSizeSmall {
+                margin-top: spacing(x0.5);
+            }
+            @else if $size == fontSizeMedium {
+                margin-top: spacing();
+            }
+        }
     }
 }
 
--- a/packages/components/molecules/f-media-element/src/components/MediaElement.vue
+++ b/packages/components/molecules/f-media-element/src/components/MediaElement.vue
@@ -3,17 +3,17 @@
         data-test-id="mediaElement-component"
         :class="[$style['c-mediaElement'], { [$style['c-mediaElement--stack']]: stacked }, { [$style['c-mediaElement--reverse']]: reverse }]">
         <div :class="[$style['c-mediaElement-content'], contentAlignClass]">
-            <h3 :class="$style['c-mediaElement-title']">
+            <h3 :class="[$style['c-mediaElement-title'], { [$style['c-mediaElement-title--large']]: titleLarge }]">
                 {{ title }}
             </h3>
-            <p :class="$style['c-mediaElement-text']">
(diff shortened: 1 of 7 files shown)


packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -1,18 +1,120 @@
 <template>
     <div
-        :class="$style['c-mediaElement']"
-        data-test-id="mediaElement-component">
-        {{ componentName }}
+        data-test-id="mediaElement-component"
+        :class="[$style['c-mediaElement'], { [$style['c-mediaElement--stack']]: stacked }, { [$style['c-mediaElement--reverse']]: reverse }]">
```
minrwhite: Would be good to split this out to a multi-line construct, for readability. I believe you should also merge the objects here into one:

```suggestion
        :class="[$style['c-mediaElement'], {
            [$style['c-mediaElement--stack']]: stacked,
            [$style['c-mediaElement--reverse']]: reverse
        }]">
```

packages/components/molecules/f-media-element/stories/MediaElement.stories.js
```diff
@@ -10,11 +10,21 @@ export default {
...
+
+MediaElementComponent.args = {
+    title: 'Stampcards',
+    text: 'See the stamps you’ve collected and any discounts you’ve earned.',
+    stacked: false,
+    reverse: true,
+    contentAlign: 'left',
+    imageAlign: 'center',
+    imageUrl: 'https://via.placeholder.com/250',
+    textSize: 'xxl'
```
minrwhite: Can we use the `ALIGN` and `TEXT_SIZE` values from config.js to provide a list of selectable options for the controls?

packages/components/molecules/f-media-element/src/components/MediaElement.vue
ashleynolan:
```suggestion
        :class="[
            $style['c-mediaElement'],
            (stacked ? $style['c-mediaElement--stack'] : ''),
            (reverse ? $style['c-mediaElement--reverse'] : '')
        ]">
```

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -1,18 +1,120 @@
...
+        <div :class="[$style['c-mediaElement-content'], contentAlignClass]">
+            <h3 :class="[$style['c-mediaElement-title'], $style['c-mediaElement-text'], textSizeClass]">
```
ashleynolan: I find these easier to parse if they're on separate lines:

```suggestion
            <h3 :class="[
                $style['c-mediaElement-title'],
                $style['c-mediaElement-text'],
                textSizeClass]">
```

ashleynolan: Also, should this have `c-mediaElement-title` and `c-mediaElement-text` on it? I'd just have `c-mediaElement-title` and then use selectors where it makes sense to share the styles between the two classes.

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -1,18 +1,120 @@
...
+                {{ title }}
+            </h3>
+            <p :class="[$style['c-mediaElement-text'], textSizeClass]">
```
ashleynolan: Similar to above – I'd put these on separate lines (we tend to do this in other components)

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -1,18 +1,120 @@
...
+                {{ text }}
+            </p>
+        </div>
+        <div :class="[$style['c-mediaElement-imgWrapper'], imageAlignClass]">
+            <img
+                :class="$style['c-mediaBlock-img']"
```
ashleynolan:
```suggestion
                :class="$style['c-mediaElement-img']"
```

packages/components/molecules/f-media-element/src/config.js
```diff
@@ -0,0 +1,13 @@
+export const ALIGN = {
```
ashleynolan: Would rename this file `constants.js` to match the naming in other components.

ashleynolan (changes requested):

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: agreed, will follow what ash has suggested for this

packages/components/molecules/f-media-element/stories/MediaElement.stories.js
matthewhardern: Will do, will add select boxes for font size and alignment options

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: will make the change

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: agreed will make the suggested change

matthewhardern: I have now replaced this and have added a font size class that is applied to both

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: agreed

packages/components/molecules/f-media-element/src/config.js
matthewhardern: will rename

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: I have modified thanks, I have also added a class to allow anyone to override the image styles, for example incase they have specific needs like a border. Until I have a greater idea of what people need style's wise on the image then this I thought for now would be the best route.

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: fixed

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: fixed

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: fixed

packages/components/molecules/f-media-element/stories/MediaElement.stories.js
matthewhardern: have added the better controls

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: fixed

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -1,18 +1,140 @@
...
+        </div>
+        <div
+            :class="[
+                $style['c-mediaElement-imgWrapper'],
+                imageAlignClass
+            ]">
+            <img
+                :class="[$style['c-mediaElement-img'], 'c-mediaElement-img--override']"
+                :src="imageUrl"
+                alt="Media Element Image">
```
ashleynolan: Doesn't need to be part of this PR, but would be good to be able to pass in an `alt` for this image.

minrwhite (dismissed):

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -21,13 +143,95 @@ export default {
...
+    }
+    &--xl {
+        @include font-size(heading-xl);
+        margin-bottom: spacing(x0.5);
+    }
+    &--xxl {
+        @include font-size(heading-xxl);
+        margin-bottom: spacing(x0.5);
+    }
+}
```
ashleynolan: We usually don't use shorthand Sass selectors, just because it's harder to do a `find` in text ediors when searching for a classname. So I'd update this to be:

```suggestion
.c-mediaElement-title {
    &.c-mediaElement-fontSize--sm {
        @include font-size(heading-s);
    }
}
```

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -1,18 +1,140 @@
...
+        ]">
+        <div
+            :class="[
+                $style['c-mediaElement-content'],
+                contentAlignClass
+            ]">
+            <h3
+                :class="[
+                    $style['c-mediaElement-title'],
+                    fontSizeClass
```
ashleynolan: Just a preference, as it would mean only adding one class instead of two, but you could put the size class on the `mediaElement` element as a modifier and then just have the title and text react to that outer class.

```
<div class="c-mediaElement c-mediaElement--fontSizeLarge">
    …
    <h3 class="c-mediaElement-title">
    </h3>
</div>
```

```scss
.c-mediaElement--fontSizeLarge {
    & .c-mediaElement--title {
        font-size: xxx;
    }
    & .c-mediaElement--text {
        …
    }
}
```

packages/components/molecules/f-media-element/src/components/MediaElement.vue
ashleynolan: See above commentt on sizing, as I think may make it a little bit cleaner with an outer modifer class.

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -21,13 +143,95 @@ export default {
...
+
+    &--center {
+        align-items: center;
+    }
+    &--left {
+        align-items: flex-start;
+    }
+    &--right {
+        align-items: flex-end;
+    }
```
ashleynolan: Same as above. Would adjust this to be longhand for better codebae searchability.
```suggestion
.c-mediaElement-imgWrapper {
    display: flex;
    flex-direction: column;
    flex: 1 1 0;
}

    .c-mediaElement-imgWrapper--center {
        align-items: center;
    }
    .c-mediaElement-imgWrapper--left {
        align-items: flex-start;
    }
    .c-mediaElement-imgWrapper--right {
        align-items: flex-end;
    }
```

ashleynolan (commented):
Looks good – just a few comments on the textSize classes, otherwise LGTM

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -21,13 +143,95 @@ export default {
...
 .c-mediaElement {
     display: flex;
+    width: 100%;
+}
+.c-mediaElement--stack {
+    flex-direction: column;
+}
+.c-mediaElement--stack.c-mediaElement--reverse {
+    flex-direction: column-reverse;
+}
```
ashleynolan:
```suggestion
.c-mediaElement--stack {
    flex-direction: column;
 
       &.c-mediaElement--reverse {
           flex-direction: column-reverse;
       }
}

```

ashleynolan: I'd also add code comments to the modifiers to describe what they do. E.g. from button:

```scss
/**
 * Modifier – .o-btn--secondary
 *
 * Accompanying secondary style button (knocked back slightly from the primary button)
 */
```

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: I like this idea, makes sense. WIll make the change

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: Yep, I have now added these to all the modifiers

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: I have modified these for alignment to be as described, but I am now handling the font sizing differently

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: font sizing method has now changed, see below for comment, is based on your other suggestion.

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: Have added a prop for this

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: I have gone along with this suggestion, however I have now added a `$font-sizes` map. Using this map I loop over to create the appropriate sizes, reason for this is I think it was starting to bloat out the css with repeated parts, which could be solved with a loop. @ashleynolan could you please take a look again and see if this suits you.

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -1,33 +1,283 @@
...
+            }
+        },
+        /**
+         * Returns the class to modify the font size of the content based on the prop textSize
+         * @returns {*}
+         */
+        fontSizeClass () {
+            switch (this.textSize) {
+                case FONT_SIZE.SM:
+                    return this.$style['c-mediaElement--fontSizeSmall'];
```
ashleynolan: Super NIT – as this is applied on the same level as `c-mediaElement-content`, the modifier should be `c-mediaElement-content--fontSizeSmall` etc.

Could also probably make it `c-mediaElement-content--small` `c-mediaElement-content--medium` and I think it would still make sense as a classname?

packages/components/molecules/f-media-element/src/components/MediaElement.vue
ashleynolan: 👍 Think that looks good

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -1,33 +1,283 @@
 <template>
     <div
-        :class="$style['c-mediaElement']"
-        data-test-id="mediaElement-component">
-        {{ componentName }}
+        data-test-id="mediaElement-component"
+        :class="[
+            $style['c-mediaElement'],
+            (stacked ? $style['c-mediaElement--stack'] : ''),
```
FayeCarter, line 6: Could you do ```{ [$style['c-mediaElement--stack'']]: stacked }```

matthewhardern: What I did originally and ash suggested to change

matthewhardern: Ash suggested that change, I had it like this originally. @ashleynolan could you explain why the suggested change is better?

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: Agree on the first point disagree with the second. As the class purely effects the font size only, I think it's best to name it as such.

packages/components/molecules/f-media-element/src/components/MediaElement.vue
ashleynolan: I think this way is a little cleaner (less nested brackets and the conditional at the front of the statement) – equally, if other people don't like it and prefer the other way happy to discuss (shouldn't stop this PR though)

packages/components/molecules/f-media-element/src/components/MediaElement.vue
ashleynolan: Cool – happy for you to keep it as `fontSizeSmall` etc 👍

packages/components/molecules/f-media-element/src/components/MediaElement.vue
SachaZvetelman: I'd personally favour `[$style['c-mediaElement--stack'']]: stacked` over `(stacked ? $style['c-mediaElement--stack'] : '')` 🙂

matthewhardern: I will leave as is for now, but can we please agree as a team what choice we want to make, my preference is the way @FayeCarter suggests.

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -75,6 +68,10 @@ export default {
         textSize: {
             type: String,
             default: 'md'
```
minrwhite, line 70:
```suggestion
            default: FONT_SIZE.MD
```

minrwhite (commented):
One suggested change, but happy if this goes in a subsequent PR and otherwise looks good to me.

minrwhite (approved):

packages/components/molecules/f-media-element/src/components/MediaElement.vue
ashleynolan: Yep – I'll setup a github discussion so we can agree on an approach going forward 👍

ashleynolan (approved):

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: will do

PR #668, opened 2021-02-01
fozzie-components@3.9.0 - Page Object Spike #globalconfig

LTurns:
Page object spike based on [https://webdriver.io/docs/pageobjects.html](https://webdriver.io/docs/pageobjects.html)

Registration is a class that imports it's parent, Page, from `page.object` from within the `f-wdio-utils` package. This page.object file contains all the main storybook URLs and some general functions that can be used by all components. 

The `f-wdio-utils` package, when published, is imported into the Registration class. When f-registration is updated in CoreWeb, the tests will still pass due to this extension. I have tested this with yalc and all is working. :)

Registration also imports a `selector.js` file which contains the data-test-ids.

### Added
- `f-wdio-utils` package in the services folder - for the `page-object` file and other webdriverio utils. 

- ### Changed
- component feature tests
- component object into a class that imports `page.object` from `f-wdio-utils`


packages/components/organisms/f-registration/test-utils/component-objects/f-registration.component.js
```diff
@@ -1,66 +1,130 @@
...
 
-exports.isEmailEmptyErrorDisplayed = () => emailEmptyError().isDisplayed();
-exports.isEmailInvalidErrorDisplayed = () => emailInvalidError().isDisplayed();
-exports.isEmailExistsErrorDisplayed = () => emailExistsError().isDisplayed();
-exports.isPasswordEmptyErrorDisplayed = () => passwordEmptyError().isDisplayed();
+    isInvalidErrorDisplayed(fieldName) {
+        return this.fields[fieldName].invalidError.isDisplayedInViewport();
+    };
+    
+    displayRequestErrorText(){
```
siggerzz: Perhaps rename this to `getRequestErrorText`?

packages/components/organisms/f-registration/test/specs/component/f-registration.component.spec.js
```diff
@@ -17,22 +29,71 @@ describe('f-registration component tests', () => {
...
         // Assert
-        expect(RegistrationComponent.isFirstNameEmptyErrorDisplayed()).toBe(true);
-        expect(RegistrationComponent.isLastNameEmptyErrorDisplayed()).toBe(true);
-        expect(RegistrationComponent.isEmailEmptyErrorDisplayed()).toBe(true);
-        expect(RegistrationComponent.isPasswordEmptyErrorDisplayed()).toBe(true);
+        expect(registration.isMaxLengthErrorDisplayed(field)).toBe(true);
     });
 
-    it.skip('should show and be able to click the legal documentation', () => {
+    forEach(['firstName', 'lastName', 'email', 'password'])
```
siggerzz: This test can be deleted IMO. We should only test stuff like this in CoreWeb where we're able to mock the underlying API's. Happy to explain on a call if this reasoning is still unclear :)

packages/services/f-wdio-utils/.eslintignore
```diff
@@ -0,0 +1,3 @@
+node_modules
+dist
```
siggerzz, line 2: This can be removed.

packages/services/f-wdio-utils/.eslintignore
```diff
@@ -0,0 +1,3 @@
...
+page.object.js
```
siggerzz, line 2: I don't think the page object should be gitignored?

packages/services/f-wdio-utils/src/page.object.js
```diff
@@ -0,0 +1,24 @@
+class Page {
+    constructor () {
+        this.title = 'Component URLS';
+    }
+
+    openOrganism (path) {
```
siggerzz: This could probably be a single function that utilises a switch statement
```
openComponent(componentType, path) {
    let formattedUrl = ''
    
    switch(componentType.toLowerCase()) {
      case atom:
        formattedUrl = `/iframe.html?id=components-atoms--${path}`
        break;
      case molecule:
        formattedUrl = `/iframe.html?id=components-molecules--${path}`
        break;
      case organism:
        formattedUrl = `/iframe.html?id=components-organisms--${path}`
        break;
      default:
        throw new Error(`${componentType} is not a valid component type. Please use 'atom', 'molecule', or 'organism'`);
    }
    
    browser.url(formattedUrl)
}
```

siggerzz (changes requested):
Just a few minor comments! Looking good though, well done! 🎉

packages/components/organisms/f-registration/test-utils/component-objects/f-registration.component.js
LTurns: yep, makes sense, will do!

packages/components/organisms/f-registration/test/specs/component/f-registration.component.spec.js
LTurns: Yeah I wondered this at the time actually - wasn't sure how to go about showing that no field errors are displayed when the form is filled out correctly. We have some similar tests in f-checkout but can't assert on them until there's a redirect anyway so guessing those could go too? Waiting for the API makes more sense so happy to delete!

packages/services/f-wdio-utils/src/page.object.js
LTurns: ooh yeah could do! I guess this way means having to specify the type too though? Would it be quicker? Is it better to have them all in one? I'm not sure which way I prefer 😄

packages/services/f-wdio-utils/.eslintignore
LTurns: oh yeah so yarn lint doesn't seem to like the page.object - keeps coming back with `UnhandledPromiseRejectionWarning: Unhandled promise rejection.` Help? 😆

siggerzz: As discussed, I mistook this file for .gitignore. Ignore my comment :)

packages/services/f-wdio-utils/src/page.object.js
siggerzz: Whilst this solution is more complex, from the perspective of a user writing a new test in a spec file, you don't want to have to think about what type a component is.

We'll be able to extend this in the future to something like:


```

// f-checkout.component.js

let url = '?path=/story/components-organisms--checkout-component'

withLocale(locale) {
    url = `${url}&knob-locale=${locale}`
    
    return this;
}

withCollection() {
    url = `${url}&knob-Checkout%20Url=%2Fcheckout-collection.json`
    
    return this;
}

openComponent() {
    super.openComponent('organism', url);   
}
```


Then in the spec file we can do something like:

```
f-checkout-collection.spec.js


....

    beforeEach(() => {
         CheckoutComponent.withLocale('en-AU').withCollection();
         CheckoutComponent.openComponent()
         CheckoutComponent.waitForComponent();
     });
```


With this approach, the spec file doesn't have to worry about what function to call :)

Hope that makes sense, but this is my thinking atm as to how we could simply this going forward!

LTurns: yeah definitely agree! It makes sense and will be so much nicer and more organised! 😄

babel.config.js
```diff
@@ -3,7 +3,8 @@ module.exports = api => {
     const isTest = api.env('test');
     const presets = [];
     const plugins = [
-        '@babel/plugin-proposal-optional-chaining'
+        '@babel/plugin-proposal-optional-chaining', 
+        "@babel/plugin-proposal-class-properties"
```
FayeCarter: Should these be single quotes?

LTurns: oh snap, yeah they should be - will fix now!

FayeCarter (approved):

siggerzz (approved):

PR #671, opened 2021-02-03
f-checkout@0.48.0 - Include user note in PATCH request

xander-marjoram:
### Added
- Mapper service for creation of checkout PATCH payload.
- Handling of user note and including it in the PATCH request.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/components/organisms/f-checkout/src/services/tests/mapper.test.js
```diff
@@ -0,0 +1,95 @@
...
+const defaultParams = {
+    address: {},
+    customer: {},
+    isCheckoutMethodDelivery: true,
+    time: {},
+    userNote: ''
+};
+
+describe('checkout mapper', () => {
+    it('should map customer details correctly', () => {
```
FayeCarter, line 12: Just out of curiosity, is it worth this be an `it.each`?

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -281,32 +285,13 @@ export default {
          */
         async submitCheckout () {
             try {
-                const checkoutData = {
```
FayeCarter, line 284: 😲 This is exciting.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -281,32 +285,13 @@ export default {
...
-                };
-
                 if (!this.isLoggedIn) {
                     await this.setupGuestUser();
                 }
 
                 await this.patchCheckout({
                     url: `checkout/${this.tenant}/${this.checkoutId}`,
-                    data: checkoutData,
+                    data: mapCheckoutRequest(this),
```
xander-marjoram: What do we think of this? Would people rather I create a new object like this, rather than passing `this` in in its entirety?
```javascript
data: mapCheckoutRequest({
    address: this.address,
    customer: this.customer,
    time: this.time,
    userNote: this.userNote
})
```

packages/components/organisms/f-checkout/src/services/tests/mapper.test.js
xander-marjoram: So I did try this to see how it worked out, but I don't think it works for two reasons:
1. I'm not a huge fan of putting the property names in magic strings
2. Note the property `mobileNumber` being mapped to `phoneNumber`, so the `it.each` needs another parameter which makes it more untidy.

```javascript
it.each([
    ['firstName', 'firstName', 'Test'],
    ['lastName', 'lastName', 'Tester'],
    ['mobileNumber', 'phoneNumber', '07890123456']
])('should map customer %s correctly', (property, expectedProperty, value) => {
    // Arrange
    const customer = { [property]: value };

    // Act
    const requestBody = mapCheckoutRequest({
        ...defaultParams,
        customer
    });

    // Assert
    expect(requestBody.customer[expectedProperty]).toBe(customer[property]);
});
```

What do you think @FayeCarter?

FayeCarter: Ahh I see, yeah sounds good to me

packages/components/organisms/f-checkout/src/components/Checkout.vue
xander-marjoram: I decided this was probably the better approach even if it's a few more lines.

packages/components/organisms/f-checkout/src/services/mapper.js
```diff
@@ -0,0 +1,40 @@
+const mapCheckoutRequest = ({
```
SachaZvetelman: Should we be more specific in terms of what request to checkout this is about? Given we have various "checkout requests".

packages/components/organisms/f-checkout/src/services/mapper.js
```diff
@@ -0,0 +1,40 @@
...
+                    ],
+                    postalCode: address.postcode || null
+                }
+            } : {}),
+            geolocation: null
+        }
+    },
+    notes: [
+        {
+            type: 'delivery',
```
SachaZvetelman, line 34: Is this correct, or should it sometimes be `'collection'`?

xander-marjoram: I checked with Dave on Slack and he said
> "There is no need to differentiate between collection and delivery. These are delivery notes. For now, we will show them for both delivery and collection to not cause any regressions with what is already there."

packages/components/organisms/f-checkout/src/services/tests/mapper.test.js
```diff
@@ -0,0 +1,95 @@
...
+        expect(requestBody.fulfilment.location.address.lines).toStrictEqual([
+            address.line1,
+            address.line2,
+            '',
+            address.city,
+            ''
+        ]);
+    });
+
+    it('should not map address if checkout method is not delivery', () => {
```
kevinrodrigues: It would be nice to wrap some of these in describes. It will help describe the scenario better. For example:

`describe('when the checkout method is not delivery') => it('should not map the address')`

helps shorten IT descriptions too.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -47,7 +47,9 @@
 
                 <form-selector />
 
-                <user-note data-test-id="user-note" />
+                <user-note
+                    data-test-id="user-note"
+                    @input="updateUserNote($event.target.value)" />
```
dandel10n, line 52: Does this work without `.native`?

packages/components/organisms/f-checkout/src/services/tests/mapper.test.js
```diff
@@ -0,0 +1,95 @@
...
+        };
+
+        // Act
+        const requestBody = mapCheckoutRequest({
+            ...defaultParams,
+            customer
+        });
+
+        // Assert
+        expect(requestBody.customer.firstName).toBe(customer.firstName);
```
kevinrodrigues, line 27: You should be able to pull out the values you need from above and just reference them here? e.g the `firstName, lastName etc`

xander-marjoram: Do you mean just like this?
```suggestion
        expect(requestBody.customer.firstName).toBe('Test');
```

packages/components/organisms/f-checkout/src/components/Checkout.vue
xander-marjoram: Yes, it seems to! 😄

packages/components/organisms/f-checkout/src/services/tests/mapper.test.js
kevinrodrigues: I mean from the requestBody call? Would just help in these expects as you can just reference the const value instead of navigating the three levels down. Not a biggy tho.

kevinrodrigues (approved):

FayeCarter (approved):

packages/components/organisms/f-checkout/src/components/Checkout.vue
dandel10n: Weird because user-note is wrapped in a div so it should silently braik with listening to the div

packages/components/organisms/f-checkout/src/services/tests/mapper.test.js
xander-marjoram: Ah I see what you mean, something like this?:

```javascript
const { customer } = mapCheckoutRequest({
    ...defaultParams,
    customer
});

const {
    firstName, lastName, phoneNumber, dateOfBirth
} = customer;

expect(firstName).toBe(customer.firstName);
expect(lastName).toBe(customer.lastName);
expect(phoneNumber).toBe(customer.mobileNumber);
expect(dateOfBirth).toBe(null);
```

If so, I think I like it as-is as the nested structure is quite clear

kevinrodrigues: The structure, if we care about it, which I think shouldn't matter, could be tested. In this test we only really care about the customer details being mapped properly & not the actual structure of the response, if the nested structure isn't obvious the test should probably change (easiest place is the description). The `customer` object can be described as the `thing` that holds the key bits of information we need so i think that should be enough to cover what the nested response outputs.

PR #672, opened 2021-02-04
f-checkout@0.49.0 - Initial APIs failure user-friendly error page.

SachaZvetelman:
v0.49.0
*February 4, 2021*

### Added
- `Error.vue` component to display a user-friendly error page when we cannot load the necessary information to render the page.

**Images**

**Desktop/Tablet:**
image

**Mobile:**
image


packages/components/organisms/f-checkout/src/components/Error.vue
```diff
@@ -0,0 +1,65 @@
...
+import Card from '@justeat/f-card';
+import '@justeat/f-card/dist/f-card.css';
+
+export default {
+    components: { Card }
+};
+</script>
+
+<style lang="scss" module>
+div.c-checkout-error {
```
SachaZvetelman: Had to add the `div.` bit for more specificity.

packages/components/organisms/f-checkout/stories/Checkout.stories.js
```diff
@@ -44,7 +44,7 @@ export const CheckoutComponent = () => ({
             default: text('Checkout ID', checkoutId)
         },
         checkoutUrl: {
-            default: select('Checkout Url', [deliveryUrl, collectionUrl], deliveryUrl)
+            default: select('Checkout Url', [deliveryUrl, collectionUrl, 'An invalid URL'], deliveryUrl)
```
LTurns, line 47: Is there a reason this isn't called 'invalidUrl' to match the others? teeny but would be nice if they were the same format 😄

SachaZvetelman: This one was a hardcoded text, whereas the other ones are JS variables. I'm happy to extract it into a variable, if people prefer :)

LTurns: ahh okay cool, that makes sense! Probably no need to change it then! :)

packages/components/organisms/f-checkout/src/assets/images/jet-sad-bag.svg
```diff
@@ -0,0 +1,89 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
xander-marjoram: Please can you run this through [SVGOMG](https://jakearchibald.github.io/svgomg/)?
Using `precision = 2` cuts the size by almost 70% without any appreciable drop in quality.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -1,68 +1,73 @@
...
+                    :class="$style['c-checkout-form']"
+                    @submit.prevent="onFormSubmit">
+                    <guest-block v-if="!isLoggedIn" />
+
+                    <form-field
+                        :value="customer.mobileNumber"
+                        name="mobile-number"
+                        :label-text="$t('labels.mobileNumber')"
+                        :has-error="!isMobileNumberValid"
+                        @input="updateCustomerDetails({ ['mobileNumber']: $event })">
```
xander-marjoram: I think the square brackets and quotes can be removed again, assuming this came in through fixing a merge conflict

SachaZvetelman: Ah, it was probably a conflict, yes. Will take a look, thanks.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -1,68 +1,73 @@
...
+                        {{ $t('buttonText') }}
+                    </f-button>
+                </form>
+
+                <checkout-terms-and-conditions
+                    v-if="!isLoggedIn" />
+            </card>
+        </div>
+
+        <error v-else />
```
xander-marjoram: Component names should be multi-word

packages/components/organisms/f-checkout/src/components/Error.vue
FayeCarter: Could you call the class `c-checkout-errorComponent` instead? Or is that too long of a name?

xander-marjoram: What else is overriding this? Is it possible to get around this with another class?

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -154,6 +154,52 @@ describe('Checkout', () => {
                 expect(addressBlock.exists()).toBe(false);
             });
         });
+
+        describe('hasCheckoutLoadedSuccessfully ::', () => {
+            it('should render the checkout form and not the error page when set to `true`', () => {
```
xander-marjoram, line 159: If you can think of how to word the test name, it might be good to combine these two cases using `it.each`

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -923,7 +969,7 @@ describe('Checkout', () => {
             });
 
             describe('when `getCheckout` request fails', () => {
-                it('should emit failure event', async () => {
+                it('should emit failure event and set `hasCheckoutLoadedSuccessfully` to `false`', async () => {
```
xander-marjoram, line 972: "and" in a test name usually indicates it should be multiple test cases

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
```diff
@@ -0,0 +1,60 @@
...
+    it('should show the header', () => {
+        // Arrange & Act
+        const wrapper = shallowMount(Error, {
+            store: createStore(),
+            i18n,
+            localVue,
+            propsData
+        });
+
+        const header = wrapper.find('[data-test-id="checkout-error-page-header"]');
```
xander-marjoram: Do we prefer this to importing the `Header` component and using `wrapper.findComponent(Header)`?

packages/components/organisms/f-checkout/src/tenants/en-GB.js
```diff
@@ -41,7 +41,11 @@ const messages = {
 
     errorMessages: {
         errorHeading: 'Error',
-        genericServerError: 'Something went wrong, please try again later'
+        genericServerError: 'Something went wrong, please try again later',
+        pageLoad: {
+            heading: 'We can\'t show you the page',
```
xander-marjoram: Nit: I think "this page" reads slightly better, while we wait for proper copy

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
SachaZvetelman: I'm not sure I follow. The header in this context refers to the error page's heading (maybe I should rename it to `heading`), not the Checkout Header component.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
SachaZvetelman: I personally don't agree with this. An `and` in this case indicates an additional `expect`, not a different test case.
I think having more than one `expect` statement per test is perfectly fine, although I know some people prefer not to have them.

I'd agree with your comment in a case in which I had something like the following `when A or B, then C`. That `or` probably means there should be two separate tests.

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
xander-marjoram: Renaming it to heading sounds good 👍

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
SachaZvetelman: I think we had a similar conversation in a recent PR (don't remember if it was mine or somebody else's). In this case, when there are two test cases that are opposite of each other, I don't see the value in adding a `.each` and make the name a bit more complicated.

If we had a combination of various inputs in a test, in that case, I'd fully support the `.each`.

What do you think, @xander-marjoram?

packages/components/organisms/f-checkout/src/components/Error.vue
SachaZvetelman: @FayeCarter, the styles for the other components we have in the project don't have the `Component` suffix.

@xander-marjoram, it's the `f-card` styles overriding my styles. Did you mean to add another class that reverts/overrides those styles or add a "dummy" class just so that we have a more specific selector? It's the latter, I'd personally leave it as I have it now, but I'm more than happy to hear suggestions here! (or anywhere, hehe) 🙂

xander-marjoram: Which properties are being overridden? If there is something specific that makes this component visually different to f-card then it probably worth its own class, which you could nest inside this one (which also helps with the extra specificity).

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
xander-marjoram: It comes down to personal preference. I won't let it block the PR though because it's not a particularly bad case of using too many `expect`s.

packages/components/organisms/f-checkout/src/components/Error.vue
```diff
@@ -0,0 +1,65 @@
...
+
+    &.c-checkout-error--verticalPadding {
+        padding: spacing(x8) spacing() spacing(x4);
+
+        @include media('<=narrow') {
+            border: none;
+            padding-top: spacing(x2);
+            padding-bottom: spacing(x2);
+        }
+    }
```
SachaZvetelman, line 52: @ashleynolan, any suggestions/thoughts here? It's to override the padding from `f-card`.

SachaZvetelman:
The `danger` check above will be sorted out in @ashleynolan's PR: https://github.com/justeat/fozzie-components/pull/678, so we can ignore it for now.

LTurns (dismissed):

FayeCarter (dismissed):

dandel10n (approved):

xander-marjoram (approved):

PR #685, opened 2021-02-05
f-user-message@v1.2.0 - Support for isFlush prop

tommcclean:
_Support a new prop `isFlush` which defaults to true. When set to true, top and bottom margins are eliminated. When false top and bottom margins are set._

_This is to account for the fact the component sits flush in some pages but not in others._

---

## UI Review Checks
image

### DIFF
--- a/packages/components/molecules/f-user-message/CHANGELOG.md
+++ b/packages/components/molecules/f-user-message/CHANGELOG.md
@@ -3,11 +3,8 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-
-Latest (add to next release)
-------------------------------
-
 v1.2.0
+------------------------------
 *February 5, 2021*
 
 ### Added
@@ -15,25 +12,19 @@ v1.2.0
 
 
 v1.1.0
+------------------------------
 *January 29, 2021*
 
 ### Added
 - Added the component to storybook
+- Stylelint added to lint styling on build.
 
 ### Changed
 - Update service to use new values as the contract has changed
-
-
-*January 12, 2021*
-
-### Changed
 - 'jet' theme instead of 'je'
 - Updated config for latest `sass-loader`.
 - Switches import in `common.scss` in line with fozzie v5-beta.
 - Update axios version for security advisory
-
-### Added
-- Stylelint added to lint styling on build.
 
 
 v1.0.0-beta.1
--- a/packages/components/molecules/f-user-message/CHANGELOG.md
+++ b/packages/components/molecules/f-user-message/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Latest (add to next release)
 ------------------------------
 
+v1.2.0
+*February 5, 2021*
+
+### Added
+- New Prop 'isFlush' for when the component needs to sit flush with the content around it
+
+
 v1.1.0
 *January 29, 2021*
 
--- a/packages/components/molecules/f-user-message/src/components/UserMessage.vue
+++ b/packages/components/molecules/f-user-message/src/components/UserMessage.vue
@@ -2,7 +2,10 @@
     <div
         v-if="userMessageEnabled"
         :data-theme="theme"
-        :class="$style['c-userMessage']"
+        :class="[
+            $style['c-userMessage'],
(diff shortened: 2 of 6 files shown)


FayeCarter (dismissed):

packages/components/molecules/f-user-message/CHANGELOG.md
```diff
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 Latest (add to next release)
```
dandel10n: Could you please move version to this line instead of "Latest (add to next release)"

tommcclean: Fixed that ta

dandel10n (commented):
I think there shouldn't be surrounding margins in component styles. If margins are needed they should be added to a component call with a separate class

packages/components/molecules/f-user-message/src/components/UserMessage.vue
```diff
@@ -2,7 +2,10 @@
     <div
         v-if="userMessageEnabled"
         :data-theme="theme"
-        :class="$style['c-userMessage']"
+        :class="[
```
kevinrodrigues: I think this works too (saving the use of a ternary):

```javascript
:class="{
      [$style['c-userMessage']]: true,
      [$style['c-userMessage--flush']]: isFlush
]}"

File: packages/components/molecules/f-user-message/src/components/UserMessage.vue [hunk 3]
```diff
@@ -32,6 +35,10 @@ export default {
         locale: {
             type: String,
             default: ''
+        },
+        isFlush: {
+            type: Boolean,
+            default: true
```
kevinrodrigues, line 41: is this mean't to be set to true?

kevinrodrigues (dismissed):

packages/components/molecules/f-user-message/src/components/UserMessage.vue
DamianMullins: Agree, also that's how the [Vue docs](https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax) explain how to add conditional classes, so good to stick with standard syntax I reckon 😺 

You can actually just specify the single class as conditional.

```js
:class="[$style['c-userMessage'], {
    [$style['c-userMessage--flush']]: isFlush
}]"
```

kevinrodrigues (approved):

packages/components/molecules/f-user-message/CHANGELOG.md
```diff
@@ -3,31 +3,29 @@
...
+v1.2.0
 ------------------------------
+*February 5, 2021*
+
+### Added
+- New Prop 'isFlush' for when the component needs to sit flush with the content around it
+
 
 v1.1.0
+------------------------------
```
ashleynolan, line 15: 👏 🥳

packages/components/molecules/f-user-message/src/components/UserMessage.vue
tommcclean: Applied suggestion cheers

packages/components/molecules/f-user-message/src/components/UserMessage.vue
```diff
@@ -65,6 +71,11 @@ export default {
     color: $white;
     background-color: $orange;
     max-width: 100%;
+    margin: spacing(x2) 0;
```
ashleynolan, line 74: Inclined to agree with @dandel10n about outside margins being added outside the component if possible. Is it not possible to add these margins on the consuming side?

```
   <user-message class="c-home-messageBanner" />
```

And then add the styles to that class potentially?

tommcclean: The way the classes are output by $style means I only get a generated class which I can't hook onto; I was originally going to do that as a "quick fix", the only way I could do this on the consuming side is to add elements to the DOM, and then add CSS to HomeWeb which is naff.

The way I see this, there are two valid cases to use this component in, flush and non flush; and it makes sense to me to make the component handle each with a prop so that its entirely self contained.

packages/components/molecules/f-user-message/CHANGELOG.md
tommcclean: Im confused, is something wrong here :)

packages/components/molecules/f-user-message/src/components/UserMessage.vue
tommcclean: Ill give that suggestion a go in Homeweb

PR #699, opened 2021-02-09
f-checkout@0.55.0 - Utilise vue svg loader for displaying SVGs

charliearlie:
### Added
- `vue-svg-loader` to Webpack config
- Utilised `vue-svg-loader` for error page SVG

Not sure if this is overkill when we have the Vue icons package, but using the path to an SVG and using it as an image was causing issues. 

Also my first PR so go easy on me ☺️

---

## UI Review Checks

image
Not much in regards to UI except that this SVG now works when used in Core Web.

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/Error.vue
+++ b/packages/components/organisms/f-checkout/src/components/Error.vue
@@ -8,7 +8,7 @@
         :class="[$style['c-checkout-error'], $style['c-checkout-error--verticalPadding']]"
     >
         // TODO: Load image from CDN in future
-        <sad-bag-icon />
+        <sad-bag-icon-decorator />
 
         <h1
             :class="$style['c-checkout-error-heading']"
@@ -26,13 +26,13 @@
 
 <script>
 import Card from '@justeat/f-card';
-import SadBagIcon from '../assets/images/jet-sad-bag.svg';
+import SadBagIconDecorator from '../assets/images/jet-sad-bag.svg';
 import '@justeat/f-card/dist/f-card.css';
 
 export default {
     components: {
         Card,
-        SadBagIcon
+        SadBagIconDecorator
     }
 };
 </script>
--- a/packages/components/organisms/f-checkout/src/components/Error.vue
+++ b/packages/components/organisms/f-checkout/src/components/Error.vue
@@ -7,9 +7,7 @@
         data-test-id="checkout-error-page-component"
         :class="[$style['c-checkout-error'], $style['c-checkout-error--verticalPadding']]"
     >
-        <img
-            src="../assets/images/jet-sad-bag.svg"
-            alt="">
+        <sad-bag-icon />
 
         <h1
             :class="$style['c-checkout-error-heading']"
@@ -27,10 +25,14 @@
 
 <script>
 import Card from '@justeat/f-card';
+import SadBagIcon from '../assets/images/jet-sad-bag.svg';
 import '@justeat/f-card/dist/f-card.css';
 
 export default {
-    components: { Card }
+    components: {
+        Card,
+        SadBagIcon
+    }
 };
 </script>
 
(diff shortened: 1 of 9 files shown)


packages/components/organisms/f-checkout/src/components/Error.vue
```diff
@@ -7,9 +7,7 @@
         data-test-id="checkout-error-page-component"
         :class="[$style['c-checkout-error'], $style['c-checkout-error--verticalPadding']]"
     >
-        <img
-            src="../assets/images/jet-sad-bag.svg"
-            alt="">
+        <sad-bag-icon />
```
ashleynolan: Does this inline the svg? I think for images like this (long-term) it would be good to load decorative images like this via cdn somehow, but think this is good as a step towards that (as we shouldn't add it to the `f-vue-icons` package).

May be worth adding a todo comment saying that the image should be added to CDN eventually?

charliearlie: Yeah it does inline it. For some strange reason using it as an img was causing inconsistencies.

Will add a todo comment now :)

ashleynolan: Cool – worth checking with @SachaZvetelman if there's a spike ticket for moving this to CDN eventually too (and if not creating one so we do this as part of engineering excellence later)

SachaZvetelman: I haven't created that spike ticket yet, so @charliearlie, go for it if you don't mind, please 😄

kevinrodrigues: Is it worth tagging inline svgs with `decorator`? e.g `sad-bag-icon-decorator` so when we come to switching them all out we can easily find them and do them all at once.

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/src/components/Error.vue
charliearlie: Definitely a good idea in case we have more in the future. I've added that now.

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/src/components/Error.vue
DamianMullins: @ashleynolan Out of curiosity, what is the reason for keeping the SVG out of `f-vue-icons`? Is it because it's a larger image rather than a simple icon? If so, do we have any guidelines on where to draw the line? 

Also, we already have [two bag images in f-vue-icons](https://justeat.github.io/fozzie-components/@justeat/storybook/index.html?path=/story/components-atoms--icons), should they be removed?

ashleynolan: I think the general rule should be based around a couple of factors (imo – happy for other opinions).

- Is the image/icon purely decorative or does it have a functional purpose? if it's functional, I think it should always be inlined and be in f-vue-icons.
- If the image is decorative, then it's likely we should be lazy loading the image rather than embedding it in the page load. I think we can make exceptions to this if the inlined icon/image is tiny, but for decorative images at the top of pages (like on registration/checkout) these are purely decorative and should be lazy-loaded for sure.

ashleynolan: Those bag icons we should definitely remove – they were put in for registration, but they're only decorative so they'll move out of there once we work out how to serve them from cdn instead.

charliearlie:
Closing due to CI not running for me.

FayeCarter (dismissed):

FayeCarter (dismissed):

dandel10n (dismissed):

kevinrodrigues (approved):

FayeCarter (approved):

PR #720, opened 2021-02-12
fozzie-components@v3.13.0 – Contributing and Dependency guides

ashleynolan:
### Added
- `Contributing Guide` to Storybook docs.
- `Dependencies Guide` to Storybook docs.

### Changed
- `Getting Started` guide updated with links to new sections.
- `CONTRIBUTING.md` updated with new info to be in-line with the Storybook docs.

### Fixed
- Image path in `Typography Setup Guide`.
- Storybook helper popup now visible (previously had black text on black bg).
- Link styling in Storybook content updated so the links are clearly visible.

### DIFF
--- a/packages/storybook/config/storybook/manager-head.html
+++ b/packages/storybook/config/storybook/manager-head.html
@@ -88,4 +88,12 @@
         font-size: 14px;
         margin-bottom: 4px;
     }
+
+    /* Theme dropdown overlay in storybook */
+    div.css-16n0kwd {
+        border-bottom-color: #fff;
+    }
+    div.css-wmj67n {
+        background: #fff;
+    }
 </style>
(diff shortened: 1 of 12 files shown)


stories/guides/component-dependencies.stories.mdx
```diff
@@ -0,0 +1,137 @@
+import LinkTo from '@storybook/addon-links/react';
```
tommcclean, line 1: Ill swing around after and put some of the context stuff in here as they can be viewed as dependencies

tommcclean (approved):

siggerzz (approved):

packages/storybook/config/storybook/manager-head.html
```diff
@@ -88,4 +88,12 @@
         font-size: 14px;
         margin-bottom: 4px;
     }
+
+    /* Theme dropdown overlay in storybook */
+    div.css-16n0kwd {
```
ashleynolan, line 93: Unfortunately, this is the only way to hook onto the storybook overlay popup.

I think they're based off storybook generated classnames, so they'll only need to be updated if in a later storybook release, they change the original classname.

PR #721, opened 2021-02-12
f-status-banner@v1.0.0

kevinrodrigues:
### Added
- `PageBanner.vue`, `MainBannerContainer.vue` components.
- Copy for each tenant.
- Test to cover changes.
- `constants.js` file.
- `f-searchbox`, `f-header` & `f-footer` components.

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


simonsandersje (dismissed):

packages/components/organisms/f-status-banner/src/constants.js
```diff
@@ -0,0 +1,2 @@
+/* eslint-disable import/prefer-default-export */
+export const BASE_HERO_URL = 'https://d30v2pzvrfyzpo.cloudfront.net/status-banner/';
```
minrwhite, line 2: Will this CloudFront distribution always be the same for all locales?

kevinrodrigues: Yup the base url will and then the locale is pulled in [here](https://github.com/justeat/fozzie-components/pull/721/files#diff-2a00b3d793dfc8d90b89b326bc905d27bfa36ea2a328f813172b4b7ae59f2a6bR55).

kevinrodrigues:
Please do not merge this PR yet.

bundlewatch.config.js
```diff
@@ -1,20 +1,21 @@
...
+];
 
 const excludedPackages = [
     'packages/services/f-metadata',
     'packages/tools/generator-component',
-    'packages/tools/storybook', 
-    'packages/services/f-wdio-utils'
+    'packages/tools/storybook',
+    'packages/services/f-wdio-utils',
+    'packages/components/organisms/f-status-banner' // Excluding as the footer is pushing this package over the 100kb.
```
kevinrodrigues, line 18: I've had to exclude this one. I've spend a bit of time trying to get the footer file down in size, need to look at it separately so have excluded this file rather than bumping the 100kb file for all packages. (This component sits at 136kb)

minrwhite (approved):

simonsandersje (approved):

bundlewatch.config.js
ashleynolan: Didn't see the PR for the status banner which included the footer, but would it be better to keep the footer out of the status banner? Do we need to include it?

Just thinking for things like registration and checkout we don't include the header as part of those components for instance – but if there's a reason we're doing it then happy for this to go through.

ashleynolan: Not sure if you meant to reolve this?  🙏 😆 

Fine either way, but just interested to know the reasoning

kevinrodrigues: Apologies I didn't see the comment you made yesterday :| So I cleaned this up earlier today and coreWeb now uses a default layout and a stripped back version of f-status-banner page now: https://github.com/justeat/fozzie-components/pull/737

At the time I hadn't realised it was not required.

ashleynolan: 👍  Was just a bit slow on the reviews while helping look after Alice! I've left a small comment on that PR too just to keep you on your toes 😆

PR #726, opened 2021-02-16
f-footer@v4.9.0 - Reduce bundle size in f-footer.

kevinrodrigues:
### Changed

- Marked `f-vue-icons` as external via config.
- Move `f-vue-icons` to peer `peerDependencies`.

### Before

Screen Shot 2021-02-16 at 3 34 43 PM

### After

Screen Shot 2021-02-16 at 3 21 57 PM

## UI Review Checks

## Tested in HomeWeb (offers)

Screen Shot 2021-02-16 at 3 39 07 PM

### DIFF
--- a/packages/components/organisms/f-footer/package.json
+++ b/packages/components/organisms/f-footer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@justeat/f-footer",
-  "version": "4.8.3",
+  "version": "4.9.0",
   "main": "dist/f-footer.umd.min.js",
   "files": [
     "dist",
@@ -40,11 +40,11 @@
   ],
   "dependencies": {
     "@justeat/f-services": "1.3.0",
-    "@justeat/f-vue-icons": "1.10.0",
     "v-click-outside": "3.1.2"
   },
   "peerDependencies": {
-    "@justeat/f-trak": ">=0.6.0"
+    "@justeat/f-trak": ">=0.6.0",
+    "@justeat/f-vue-icons": "1.10.0"
   },
   "devDependencies": {
     "@samhammer/vue-cli-plugin-stylelint": "2.0.1",
--- a/packages/components/organisms/f-footer/CHANGELOG.md
+++ b/packages/components/organisms/f-footer/CHANGELOG.md
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v4.9.0
+------------------------------
+*February 16, 2021*
+
+### Changed
+- Marked `f-vue-icons` as external via config.
+- Move `f-vue-icons` to peer `peerDependencies`.
+
 
 v4.8.3
 ------------------------------
--- a/packages/components/organisms/f-footer/vue.config.js
+++ b/packages/components/organisms/f-footer/vue.config.js
@@ -17,6 +17,11 @@ module.exports = {
                 additionalData: `@import "../assets/scss/common.scss";`
             });
     },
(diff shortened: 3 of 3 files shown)


packages/components/organisms/f-footer/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@justeat/f-footer",
-  "version": "4.8.3",
+  "version": "4.9.0",
```
DamianMullins, line 3: I'd argue this should be a major version bump as users will now need to pass `f-vue-icons` as a peer dep?

DamianMullins (commented):
Probably worth updating the readme steps to explain that `f-vue-icons` is now a peer dep.

dandel10n (approved):
:rocket:

packages/components/organisms/f-footer/package.json
kevinrodrigues: Yeah i wasn't actually sure at first but agree. Will update 👍

kevinrodrigues:
> Probably worth updating the readme steps to explain that `f-vue-icons` is now a peer dep.

Good shout. Will do.

packages/components/organisms/f-footer/package.json
```diff
@@ -40,11 +40,11 @@
   ],
   "dependencies": {
     "@justeat/f-services": "1.3.0",
-    "@justeat/f-vue-icons": "1.10.0",
     "v-click-outside": "3.1.2"
   },
   "peerDependencies": {
-    "@justeat/f-trak": ">=0.6.0"
+    "@justeat/f-trak": ">=0.6.0",
+    "@justeat/f-vue-icons": "1.10.0"
```
DamianMullins, line 47: We don't want to pin a single version as a peer dependency, you can set a range here instead.

```suggestion
    "@justeat/f-vue-icons": ">=1.10.0"
```

DamianMullins:
So I've just pulled this branch into my project and used [Webpack Bundle Analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) to view the resulting bundles and it looks like the footer package ends up being larger than it was before?

## Before

With the icons included in the UMD bundle

Screenshot 2021-02-16 at 19 49 09

## After

With the icons excluded from the UMD bundle but added in as a peer dep

Screenshot 2021-02-16 at 19 49 48

It looks like webpack adds the entire icon package to the footer bundle with no tree-shaking and so the icons are duplicated.

I've also tried this without custom webpack split chunks and the icons are still duplicated.

I _think_ this is because the bundle contains commonjs `require`s rather than esmodule `import`s and so it can't do static analysis and do the tree-shaking.

Would be good to catch up when you have time so we can check this through as there could be something i am missing 😺

kevinrodrigues:
Going to close this.

@DamianMullins I managed to get this to work with menuweb but then the bundle size remains the same in both menu & in fozzie components, so no benefit :\,  I'm going to look at a better way to reduce the size in this component itself as i think the main issue is the imports in the flagIcon.vue file.

ashleynolan:
@kevinrodrigues I think the best solution would be if we could somehow specify for some icons to not be inlined, but included as img includes somehow (so for flag icons, they don't need to be inlined, they could be lazy loaded for instance).

I would guess the best way to do this potentially would be to specify an option to output the icon as a link to the unpkg icon that's published in f-icons (as all the icons are hosted there when that package gets published)

May be worth a catchup to discuss, but off the top of my head that could help with icons we don't _need_ to be inlined?

ashleynolan:
@kevinrodrigues For reference btw, here's the link to one of the icons on unpkg > [alert icon](https://unpkg.com/browse/@justeat/f-icons/dist/icons/alert.svg)

This link will automatically point to the latest version of the package (can also nail down to specific versions if needed)

PR #735, opened 2021-02-22
f-content-cards@v3.3.0: Stamp Card

minrwhite:
## Added
- Stamp Card component, card Type and story

#### Incomplete stamp:

Desktop:
image

Mobile:
image

#### Complete stamp:

Desktop:
image

Mobile:
image

---

## UI Review Checks

- [ ] This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist) - *Accessibility requirements, specifically around stamps, to be addressed in upcoming work*

## Browsers Tested

- [x] Chrome (latest)
- [x] Firefox (latest)
- [ ] Internet Explorer 11


minrwhite:
I've had a query about how the scheme as a whole is supposed to work - please let me know if you have questions about this so I can direct you to details.

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
```diff
@@ -0,0 +1,315 @@
...
+        },
+
+        testId: {
+            type: String,
+            default: null
+        }
+    },
+
+    computed: {
+        title () {
```
dandel10n: I think there is no benefit here in using computed properties for shortening prop names. `this.card.title`, `this.card.description`, `this.card.image`, `this.card.url`, `this.card.discountPercentage`, `this.card.isReadyToClaim` etc should be fine.

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
```diff
@@ -0,0 +1,315 @@
...
+    }
+};
+</script>
+
+<style lang="scss" module>
+
+$stampCard-subStatus-colour: #017a39; /* $color-green in PIE - not in fozzie-colour-palette yet */
+$stampCard-expiryInfo-colour: $grey--dark;
+
+$stampCard-responsive-mobileViewBreakpoint: '<narrowMid';
```
dandel10n, line 154: Why do we need a variable for `<narrowMid` here?

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
```diff
@@ -0,0 +1,315 @@
...
+        expiryDate () {
+            return this.card.expiryDate;
+        },
+        expiryLine () {
+            return this.card.expiryLine;
+        },
+        isReadyToClaim () {
+            return !!this.card.isReadyToClaim;
+        },
+        totalRequiredStamps () {
```
dandel10n: Can we for now move it to data instead of computed prop as it is a fixed value?

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
```diff
@@ -0,0 +1,315 @@
...
+        :class="[
+            $style['c-stampCard1']
+        ]"
+        @click="onClickContentCard"
+    >
+        <div
+            :class="[
+                $style['c-stampCard1-headerDetails']
+            ]"
+        >
```
dandel10n:
```suggestion
        <div :class="[$style['c-stampCard1-headerDetails']]">
```
Here and everywhere where there is only one class

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
```diff
@@ -0,0 +1,315 @@
+<template>
+    <component
+        :is="url ? 'a' : 'div'"
+        :href="url"
+        :data-test-id="testId"
+        :class="[
+            $style['c-stampCard1']
+        ]"
```
dandel10n:
```suggestion
        :class="[$style['c-stampCard1'] ]"
```

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
```diff
@@ -0,0 +1,315 @@
...
+            ]"
+        >
+            <img
+                :class="[
+                    $style['c-stampCard1-icon']
+                ]"
+                :data-test-id="testIdForSection('image')"
+                :src="image"
+                :alt="title"
+            >
```
dandel10n: Could you please move `>` to the last passed attribute here and in other places? It will inline it with other templates style. The same goes for first attribute to be on the same line as the element like 
```
<img :class="smth"
      :src="smth"
      :alt="smth">
```

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
```diff
@@ -0,0 +1,315 @@
...
+            <h3
+                :class="[
+                    $style['c-stampCard1-title']
+                ]"
+                :data-test-id="testIdForSection('title')"
+            >
```
dandel10n:
```suggestion
            <h3 :class="[$style['c-stampCard1-title']]"
                :data-test-id="testIdForSection('title')">
```

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
```diff
@@ -0,0 +1,315 @@
...
+                {{ title }}
+            </h3>
+
+            <p
+                :class="[
+                    $style['c-stampCard1-statusText']
+                ]"
```
dandel10n:
```suggestion
            <p :class="[$style['c-stampCard1-statusText']]"
```

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
```diff
@@ -0,0 +1,315 @@
...
+}
+
+.c-stampCard1-headerDetails {
+    margin-bottom: spacing();
+}
+
+.c-stampCard1-icon {
+    float: left;
+    margin-right: spacing(x2);
+    width: spacing(x7);
```
dandel10n: For width and height it is better to use px or variables than spacing, which is used for paddings and margins

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
minrwhite: The purpose of having these is not only to shorten prop names, but also as a clue (when changing the template, or stories) as to what keys *are* and *should* be available in the incoming card object. I'd prefer to leave it in if possible.

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
minrwhite: A couple reasons - 

1. to avoid putting the 'magic' string everywhere we need it
2. to add some semantics around the meaning of this for the component - e.g. `mobileViewBreakpoint`. 
3. Essentially some insurance around easily changing this if UX decide the breakpoint is happening at the wrong screen width, as we've only been provided portrait and landscape designs.

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCard1.vue
dandel10n: Sorry, I still don't see the benefit. If you want to leave a clue for the object keys than it is better to do in props (via validator function)

dandel10n: Here is [an example](https://stackoverflow.com/questions/43913556/nested-property-types-in-vue-js) of what I mean. In React there is "object of shape" for this purpose but I couldn't find an alternative in Vue closer that validator.

dandel10n: Or another way is to document this in README :)

minrwhite: Thanks for the reasoning - in my opinion the validator function in props is the wrong place to do it in this instance as we need to be fairly widely accepting of data provided by braze - some keys may be specified while others may be not, and we should do our best effort to show the card based on the data we _do_ have available, where a validator function would reject outright.

The README however, I can get behind, I'll make this change.

dandel10n (approved):

AlixRichards (approved):
I really like the `StampCard1.md` 👍  - made it much easier for me to see what was going on quickly

PR #736, opened 2021-02-22
f-card@v1.0.0 – Removing unecessary config and story controls added

ashleynolan:
### Removed
- `tenant` and `theme` config, as not used in Card component.

### Changed
- Storybook config switched over to using `controls` rather than `knobs`.
- Reordered props (Alphabetical order).
- Updated README.
- Updated config for latest `sass-loader`.
- Switches import in `common.scss` in line with fozzie v5-beta.

---

## UI Review Checks

Screen Shot 2021-02-22 at 17 32 24


packages/components/atoms/f-card/CHANGELOG.md
```diff
@@ -4,11 +4,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-Latest (add to next release)
+v1.0.0
 ------------------------------
-*December 30, 2020*
+*February 22, 2020*
```
minrwhite: I know 🤯 but it's true 
```suggestion
*February 22, 2021*
```

ashleynolan: 🤦

packages/components/atoms/f-card/stories/card.stories.js
```diff
@@ -1,41 +1,40 @@
-import {
-    withKnobs, boolean, select, text
-} from '@storybook/addon-knobs';
-import Card from '../src/components/Card.vue';
 import { withA11y } from '@storybook/addon-a11y';
+import Card from '../src/components/Card.vue';
 
 export default {
     title: 'Components/Atoms',
```
minrwhite, line 5: I think this should be scoped to the package, otherwise it injects the derived or given args into other stories with the same `title` here (or has in my experience).

```suggestion
    title: 'Components/Atoms/f-card',
```

ashleynolan: Cool – will make sure I do this for any others I update too (as I'm going to convert a few of the packages over to controls and make the stories a little more useful)

ashleynolan: It looks like if I define this, it does this to the nav – I think it's probably only important when there are multiple stories for the same component?

Screen Shot 2021-02-22 at 17 30 18


Useful to know though, as I'll be splitting the button one up into multiple stories next…

minrwhite: I think if you don't define the 'storyName' for an export, it chooses it as the default, and then doesn't render the story as a sub-option in the menu - don't quote me on it though, as I haven't tried this.

minrwhite (dismissed):

packages/components/atoms/f-card/stories/card.stories.js
```diff
@@ -1,41 +1,40 @@
...
-        isPageContentWrapper: {
-            default: boolean('isPageContentWrapper', false)
-        }
-    },
-    parameters: {
-        notes: 'some documentation here'
-    },
+    props: Object.keys(argTypes),
     template:
         '<card :locale="locale" :cardHeading="cardHeading" :cardHeadingPosition="cardHeadingPosition" :isRounded="isRounded" :hasOutline="hasOutline" :isPageContentWrapper="isPageContentWrapper"><p>Some Card Content</p></card>'
```
dandel10n: :locale should be removed from here

packages/components/atoms/f-card/stories/card.stories.js
```diff
@@ -1,41 +1,40 @@
...
+    decorators: [withA11y],
+    parameters: {
+        backgrounds: {
+            values: [
+                { name: 'Grey', value: '#f5f5f5' },
+                { name: 'Dark', value: '#333' }
+            ],
+            default: 'Grey'
+        },
+        notes: 'some documentation here'
```
dandel10n: Out of interest, why do we have `notes: 'some documentation here'` in some stories?

ashleynolan: It was in there before, but agree it shouldn't be. Will remove it now

packages/components/atoms/f-card/stories/card.stories.js
ashleynolan: Have updated

packages/components/atoms/f-card/stories/card.stories.js
ashleynolan: Ok – I'll have a play around with the conventions and update all the stories at once to reflect the change if this works (otherwise will have some with `storyName` set and others with it not set), if that sounds ok?

minrwhite: Sounds good to me

dandel10n (dismissed):

minrwhite (dismissed):

ashleynolan:
@benwhite-deltas @dandel10n Sorry to be annoying – just pushed the small change Maryia noticed, so can you both re-tick?

minrwhite (approved):

charliearlie (approved):

dandel10n (approved):

PR #737, opened 2021-02-23
f-status-banner@v1.1.0 - Remove header, footer & bundlewatch exclude.

kevinrodrigues:
We want the consuming application to render the header and footer themselves. This PR gives the user more control in their own application. i.e lets them serve a default layout or template that may already contain a header & footer.

### Removed
- `header` & `footer` components so the consuming application can render their own.
- Tests that covered header & footer.
- `bundlewatch` exclude reverted for `f-status-banner`.

## UI Review Checks

- [x] Unit tests have been [updated]

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/components/organisms/f-status-banner/CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v1.0.0
```
charliearlie: Seems the previous change was 1.0.0 and in package.json this has been bumped to 1.1.0  👀

kevinrodrigues: Fixed.

simonsandersje (dismissed):

minrwhite (approved):

charliearlie (approved):

DamianMullins (approved):

packages/components/organisms/f-status-banner/src/components/StatusBanner.vue
```diff
@@ -1,36 +1,19 @@
...
-            data-test-id="main-banner-container"
-            :base-hero-url="baseHeroUrl"
-            :copy="copy"
-            :locale="componentLocale" />
-
-        <footer-component
-            data-test-id="footer-component"
-            :locale="componentLocale" />
-    </div>
+    <main-banner-container
```
ashleynolan, line 2: QQ now this is just calling a banner container component. Could that component just be included here now to avoid the need for two `.vue` files, or is there a reason to route it through the `StatusBanner.vue` file and then into the `MainBannerContainer.vue`?

kevinrodrigues: Do you mean combine the `PageBanner.vue` & `MainBannerContainer.vue` components into this file and delete those? It would also mean moving the logic from the other components into here too. I wanted `MainBannerContainer.vue` to be the thing you pass props through on, so when we have other status pages requirements we can control things inside of the main container rather than keeping the logic in one file. Can refactor tho and move it all here as we don't have that requirement yet?

ashleynolan: I meant just combing what's in `MainBannerContainer.vue` into this component and keeping PageBanner.vue as is, but it could be useful to use this as a shell for different status layouts as you mention to keep the base component cleaner.

Maybe keep it as it is for now and once we get the requirements for other layouts we can always revisit it if needed.

PR #738, opened 2021-02-23
f-checkout@v0.60.0 - Add analytics

FayeCarter:
### Added
- `analytics.module` to store.
- `mapAnalyticsName` and `mapAnalyticsNames` to `mapper` service.
- Analytics tracking to `Checkout` component.
- Tests to cover changes.

TODO in next PR: 
- Add correct error handling.


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -550,16 +562,23 @@ export default {
             }
         },
 
+        /**
+         * Check for is valid - no inline messages
```
ashleynolan: Typo? `for` > `form`

packages/components/organisms/f-checkout/src/services/analytics.js
```diff
@@ -0,0 +1,84 @@
+import Trak from '@justeat/f-trak';
+
+const fieldMapper = {
```
ashleynolan: Would be good to add a comment as to the format of this. I'm guessing it maps names (on the left) to GA name values (on the right), but that's just me assuming so would be good to clarify in a comment

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -319,6 +360,32 @@ export default {
 
         [UPDATE_USER_NOTE]: (state, userNote) => {
             state.userNote = userNote;
+        },
+
+        [UPDATE_FIELD_CHANGES]: (state, field) => {
```
ashleynolan: This name isn't particularly clear what it's linked to. Does it update the list of changed fields in `state.changes` if one of the fields in the form changes? If so, I'd probably call this `UPDATE_CHANGED_FIELDS` or something similar?

packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -106,6 +119,14 @@ describe('CheckoutModule', () => {
         expect(state).toEqual(defaultState);
     });
 
+    describe('getters ::', () => {
+        describe('analyticsData` ::', () => {
+            it('should return `isLoggedIn', () => {
+                expect(analyticsData(state)).toMatchSnapshot();
```
ashleynolan: I think the test description is a little bit misleading for this check as you aren't checking specifically for the presence of `isLoggedIn` (someone could update the test without updating the description and it would still pass) – you're just checking for a valid snapshot.

I'd either change the name of the test to reference the snapshot, or try and check more specifically for `isLoggedIn` being present in `analyticsData(state)`

packages/components/organisms/f-checkout/src/services/tests/analytics.test.js
```diff
@@ -0,0 +1,179 @@
...
+            ['address.city', 'addressCity'],
+            ['city', 'addressCity'],
+            ['address.postcode', 'addressPostcode'],
+            ['postcode', 'addressPostcode'],
+            ['customer.firstName', 'firstName'],
+            ['customer.lastName', 'lastName'],
+            ['customer.mobileNumber', 'phone'],
+            ['mobileNumber', 'phone'],
+            ['customer.email', 'email']
+        ])('should correct %s to %s', (provided, expected) => {
```
ashleynolan:
```suggestion
        ])('should map the fieldname %s to the analytics key %s', (provided, expected) => {
```

packages/components/organisms/f-checkout/src/services/tests/analytics.test.js
```diff
@@ -0,0 +1,179 @@
...
+                'mobilePhone',
+                'address.line1',
+                'customer.firstName',
+                'lastName',
+                'customer.email',
+                'city'
+            ];
+
+            const expected = [
+                'addressCity',
```
ashleynolan: More out of interest – does the function re-order the valued into alphabetical order here then? Or does the test just do a fuzzy match where the ordering doesn't matter when they're compared in the expect statement?

packages/components/organisms/f-checkout/src/services/tests/analytics.test.js
```diff
@@ -0,0 +1,179 @@
...
+            trackFormInteraction(action, analyticsData, error);
+
+            // Assert
+            expect(trackEventSpy).toHaveBeenCalledWith(expectedEvent);
+        });
+
+        it.each([
+            ['checkout', true],
+            ['checkout_guest', false]
+        ])('should set `name` to %s if isLoggedIn is %s', (expectedName, isLoggedInState) => {
```
ashleynolan:
```suggestion
        ])('should set `name` to %s if `isLoggedIn` is %s', (expectedName, isLoggedInState) => {
```

packages/components/organisms/f-checkout/src/services/tests/analytics.test.js
```diff
@@ -0,0 +1,179 @@
...
+
+    afterEach(() => {
+        jest.clearAllMocks();
+    });
+
+    describe('trackInitialLoad :: ', () => {
+        it('should call `event` method of `f-trak`', () => {
+            // Arrange
+            const expectedEvent = {
+                custom: {
```
ashleynolan: Could save declaring this object twice by putting it up on line 20 as a let and then just amending the `pageData` part for the tests on line 47 (as it looks like the same object for both sets of tests otherwise)

packages/components/organisms/f-checkout/src/services/analytics.js
```diff
@@ -0,0 +1,84 @@
...
+            pageData: {
+                name: `Checkout 1 ${pageName}`,
+                group: 'Checkout'
+            }
+        }
+    });
+};
+
+/**
+ * Updates passed fields to match analytics requirements.
```
ashleynolan:
```suggestion
 * Updates passed field names to match the expected analytics name requirements.
```

ashleynolan (commented):
Looks good – few comments but no major issues from me

packages/components/organisms/f-checkout/src/store/checkout.module.js
FayeCarter: updated thanks

packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
FayeCarter: This, I think, was a half-written test, sorry. This has now been removed

packages/components/organisms/f-checkout/src/services/tests/analytics.test.js
FayeCarter: It sorts alphabetically. I've added a test specifically for that now 👍

packages/components/organisms/f-checkout/src/services/tests/analytics.test.js
FayeCarter: Good point, thanks

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1922,32 +1943,21 @@ describe('Checkout', () => {
...
             });
 
             afterEach(() => {
                 jest.clearAllMocks();
             });
 
-            it('should emit `CheckoutValidationError` with validation state if form is invalid', async () => {
+            it('should call `trackFormInteractionSpy` with correct `action`', async () => {
                 // Arrange
-                const mockValidationState = {
```
FayeCarter, line 1937: The test hasn't been deleted, I've moved it into `when the form is Invalid` describe block.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1960,60 +1970,184 @@ describe('Checkout', () => {
...
+                    // Act
+                    await wrapper.vm.onFormSubmit();
 
-                // Assert
-                expect(submitCheckoutSpy).toHaveBeenCalled();
+                    // Assert
+                    expect(wrapper.emitted(EventNames.CheckoutValidationError).length).toBe(1);
+                });
+
+                it('should call `mapAnalyticsFieldNames` with `invalidFields`', async () => {
```
FayeCarter: New tests start here

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1960,60 +1970,184 @@ describe('Checkout', () => {
...
-                const wrapper = mount(VueCheckout, {
-                    store: createStore(),
-                    i18n,
-                    localVue,
-                    propsData,
-                    mocks: {
-                        $v,
-                        $logger
-                    }
+                const mockValidationState = {
```
FayeCarter, line 1987: This is old, it's just been moved

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1960,60 +1970,184 @@ describe('Checkout', () => {
...
+
+                it('should try to call `submitCheckout`', async () => {
+                    // Act
+                    await wrapper.vm.onFormSubmit();
+
+                    // Assert
+                    expect(submitCheckoutSpy).toHaveBeenCalled();
+                });
+
+                it('should call `trackFormInteractionSpy` with correct `action`', async () => {
```
FayeCarter, line 2064: New tests

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1960,60 +1969,142 @@ describe('Checkout', () => {
...
+                    // Act
+                    await wrapper.vm.onFormSubmit();
 
-                // Assert
-                expect(submitCheckoutSpy).toHaveBeenCalled();
+                    // Assert
+                    expect(wrapper.emitted(EventNames.CheckoutValidationError).length).toBe(1);
+                });
+
+                it('should call `trackFormInteraction` with and action `inline_error` and validation errors', async () => {
```
simonsandersje: Nit: should "...with and action..." read "...with an action..."?

FayeCarter: 🤦‍♀️

packages/components/organisms/f-checkout/src/components/_tests/helpers/setup.js
```diff
@@ -50,6 +50,18 @@ const defaultState = {
...
 };
 
+const defaultAnalyticsState = {
+    serviceType: '',
+    restaurantId: '',
+    basket: {
+        id: '',
+        total: 0
+    },
+    isLoggedIn: false,
```
SachaZvetelman: Do these belong to the `analytics` module? I think they belong to the `checkout` module only. You can always reference a module from another module.

packages/components/organisms/f-checkout/src/components/_tests/helpers/setup.js
```diff
@@ -74,13 +93,23 @@ const i18n = {
     }
 };
 
-const createStore = (state = defaultState, actions = defaultActions) => new Vuex.Store({
+const createStore = (
+    state = defaultState,
+    actions = defaultActions,
```
SachaZvetelman: We should probably rename these to `checkoutState = defaultCheckoutState` and `checkoutActions = defaultCheckoutActions`.

packages/components/organisms/f-checkout/src/demo/get-basket-collection.json
```diff
@@ -1,4 +1,5 @@
 {
+  "BasketId": "11111111",
```
SachaZvetelman, line 2: Maybe a silly question, but is this related to the PR or is it something else? 😄

simonsandersje (dismissed):

packages/components/organisms/f-checkout/src/services/mapper.js
```diff
@@ -49,4 +49,37 @@ const mapUpdateCheckoutRequest = ({
...
+/**
+ * Updates passed field names to match the expected analytics name requirements.
+ * If passed fields are an array. Sorts the array alphabetically
+ */
+const mapAnalyticsFieldNames = fields => {
+    if (Array.isArray(fields)) {
+        return fields.map(field => (analyticFieldNameMapper[field] ? analyticFieldNameMapper[field] : field)).sort();
+    }
+    return analyticFieldNameMapper[fields] ? analyticFieldNameMapper[fields] : fields;
+};
```
SachaZvetelman: As we discussed, I'd probably split this into two separate functions. You have two clearly distinct behaviours within one function currently.

packages/components/organisms/f-checkout/src/services/tests/mapper.test.js
```diff
@@ -122,3 +122,72 @@ describe('checkout mapper', () => {
...
+            'firstName',
+            'lastName',
+            'mobilePhone'
+        ];
+
+        // Act & Assert
+        expect(mapAnalyticsFieldNames(provided)).toEqual(expected);
+    });
+
+    it('should sort corrected field names alphabetically', () => {
```
SachaZvetelman: I'm struggling to understand the difference between this test and the one above. Could you please help me? :)

packages/components/organisms/f-checkout/src/store/analytics.module.js
```diff
@@ -0,0 +1,131 @@
...
+    namespaced: true,
+
+    state: () => ({
+        serviceType: '',
+        restaurantId: '',
+        basket: {
+            id: '',
+            total: 0
+        },
+        isLoggedIn: false,
```
SachaZvetelman: Same as above.

packages/components/organisms/f-checkout/src/store/analytics.module.js
```diff
@@ -0,0 +1,131 @@
...
+    }),
+
+    actions: {
+        updateState ({ rootState, commit }, payload) {
+            const checkoutState = {
+                isLoggedIn: rootState.checkout.isLoggedIn,
+                basket: rootState.checkout.basket,
+                restaurantId: rootState.checkout.restaurantId,
+                serviceType: rootState.checkout.serviceType
+            };
```
SachaZvetelman: Related to my comments above, I don't think there's a need for this. We're just duplicating state here rather than referencing the `checkout` module directly.

packages/components/organisms/f-checkout/src/store/analytics.module.js
```diff
@@ -0,0 +1,131 @@
...
+         */
+        updateChangedFields ({ commit }, field) {
+            commit(UPDATE_CHANGED_FIELDS, field);
+        },
+
+        /**
+         * Pushes initial state of checkout to the dataLayer.
+         */
+        trackInitialLoad ({ state, dispatch }) {
+            window.dataLayer = window.dataLayer || [];
```
SachaZvetelman, line 54: I don't think this should go here. This, in my opinion, should be in `f-trak`.
Feel free to do it in a separate PR, but let's not forget about it.

packages/components/organisms/f-checkout/src/store/analytics.module.js
```diff
@@ -0,0 +1,131 @@
...
+
+            Trak.event({
+                event: 'Form',
+                custom: {
+                    form: {
+                        name: formName,
+                        action,
+                        error: mapAnalyticsFieldNames(error) || null,
+                        autofill: JSON.parse(JSON.stringify(state.autofill)),
+                        changes: JSON.parse(JSON.stringify(state.changedFields))
```
SachaZvetelman: Is there any reason to first stringify it and then parsing it?

packages/components/organisms/f-checkout/src/store/analytics.module.js
```diff
@@ -0,0 +1,131 @@
...
+        },
+
+        [UPDATE_AUTOFILL]: (state, payload) => {
+            const { customer, address, serviceType } = payload;
+
+            if (customer.mobileNumber) {
+                state.autofill.push('phone');
+            }
+
+            if (serviceType === 'delivery') {
```
SachaZvetelman: Do we need this conditional at all? We have `if (value)` below, so I assume in the case of collection we won't have a value, and most importantly, probably `address` will be `[]`, so there won't be a `forEach`. I might be wrong.

packages/components/organisms/f-checkout/src/store/tests/analytics.module.test.js
```diff
@@ -0,0 +1,308 @@
...
+            });
+        });
+
+        describe('updateChangedFields ::', () => {
+            it(`should call ${UPDATE_CHANGED_FIELDS} with passed field`, () => {
+                // Arrange
+                const field = 'firstName';
+
+                // Act
+                updateChangedFields({ commit }, field);
```
SachaZvetelman: `updateChangedFields` expects one field. Shouldn't it be `updateChangeField` instead?

packages/components/organisms/f-checkout/src/store/tests/analytics.module.test.js
```diff
@@ -0,0 +1,308 @@
...
+                trackInitialLoad({ state, dispatch });
+
+                // Assert
+                expect(trackEventSpy).toHaveBeenCalledWith(expectedEvent);
+            });
+
+            it.each([
+                ['Checkout 1 Overview', true],
+                ['Checkout 1 Guest', false]
+            ])('should set `pageData.name` to %s if `isLoggedIn` is %s', (expectedName, isLoggedInState) => {
```
SachaZvetelman: Personally, these look more like two different `describe` (`when user is logged in` vs `when user is not logged in` or similar), rather than `it.each`.

packages/components/organisms/f-checkout/src/store/tests/analytics.module.test.js
```diff
@@ -0,0 +1,308 @@
...
+                trackFormInteraction({ state }, payload);
+
+                // Assert
+                expect(trackEventSpy).toHaveBeenCalledWith(expectedEvent);
+            });
+
+            it.each([
+                ['checkout', true],
+                ['checkout_guest', false]
+            ])('should set `name` to %s if `isLoggedIn` is %s', (expectedName, isLoggedInState) => {
```
SachaZvetelman: Same as above.

packages/components/organisms/f-checkout/src/demo/get-basket-collection.json
FayeCarter: I think it's adding in bits from another PR. But if they're not there it's harder to test the analytics stuff

packages/components/organisms/f-checkout/src/components/_tests/helpers/setup.js
FayeCarter: Good point 🙂

packages/components/organisms/f-checkout/src/services/mapper.js
FayeCarter: Ah yeah, sorry. I'll sort that out

packages/components/organisms/f-checkout/src/services/tests/mapper.test.js
FayeCarter: It was from a comment from Ash where he wanted to know if the method sorted or not. I thought it would be worth adding, but with the new methods it will probably not be needed

packages/components/organisms/f-checkout/src/store/analytics.module.js
FayeCarter: Good point, I'll do it after this PR if that's ok 🙂

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -632,33 +642,34 @@ export default {
...
+                this.trackFormInteraction({ action: 'success' });
             } catch (error) {
                 this.handleErrorState(error);
             } finally {
                 this.shouldDisableCheckoutButton = false;
             }
         },
 
+        /**
+         * Check to see if any `Vuelidate` validation errors
```
ashleynolan, line 654: Doesn't need to be part of this PR, but could we update these JSDoc comments to include a `@returns {type}` at the bottom of them? Just helps when debugging down the line.

[Like this sort of thing in searchbox](https://github.com/justeat/fozzie-components/blob/master/packages/components/molecules/f-searchbox/src/components/Form.vue#L201-L206).

packages/components/organisms/f-checkout/src/components/_tests/helpers/setup.js
ashleynolan: Would the same renaming make sense for the analytics variables below too? (`checkoutAnalyticsState`, etc)

I don't have the full context, so more a question for @FayeCarter and @SachaZvetelman

packages/components/organisms/f-checkout/src/store/analytics.module.js
ashleynolan: If you do remove the conditional because of this, I'd probably add a comment just saying that it's only relevant for delivery orders and why (as it's info I wouldn't know without context of the data being passed in)

packages/components/organisms/f-checkout/src/store/tests/analytics.module.test.js
ashleynolan: Not sure if a typo, but did you mean `updateChangedField`?

ashleynolan (commented):
Few small comments

packages/components/organisms/f-checkout/src/services/tests/mapper.test.js
FayeCarter: Just realised I gave them the same name.

packages/components/organisms/f-checkout/src/components/_tests/helpers/setup.js
FayeCarter: I've updated all of these now 👍

packages/components/organisms/f-checkout/src/store/analytics.module.js
FayeCarter: I think I need both, hoping I'm wrong.
I do `JSON.stringify` to remove the observer
Then `JSON.parse` to turn it back to an array

packages/components/organisms/f-checkout/src/store/tests/analytics.module.test.js
FayeCarter: Changed this now

packages/components/organisms/f-checkout/src/components/Checkout.vue
FayeCarter: I'll do this as a separate PR. This one is getting a bit big 😄

packages/components/organisms/f-checkout/src/store/analytics.module.js
FayeCarter: Thanks, I've added a comment to this.
@SachaZvetelman I was hoping the checkout state for address would be `{ address: { line1: '', line2: '' ....}` because that's what the default state allowed?

packages/components/organisms/f-checkout/src/store/analytics.module.js
FayeCarter: Ignore me. I was wrong 🤦‍♀️

simonsandersje (dismissed):

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -80,6 +80,8 @@ export default {
             const { data } = await axios.get(url, config);
 
             commit(UPDATE_STATE, data);
+
+            dispatch('analytics/updateAutofill', state, { root: true });
```
xander-marjoram, line 84: I'm wondering if this should live inside the Vue component (push the analytics after we've called the action) rather than in the checkout store, just to keep things a bit more separated?

FayeCarter: Hey! So the aim was to keep as much out of the checkout component as possible so that it didn't get clogged up. More than happy to move it if you think it would be neater though.

packages/components/organisms/f-checkout/src/store/tests/analytics.module.test.js
```diff
@@ -0,0 +1,391 @@
...
+
+            it('should call `event` method of `f-trak`', () => {
+                // Act
+                trackInitialLoad({ rootState, dispatch });
+
+                // Assert
+                expect(trackEventSpy).toHaveBeenCalledWith(expectedEvent);
+            });
+
+            describe('when user is loggedIn is true', () => {
```
kevinrodrigues: Minor: `isloggedIn` plus a few other areas.

FayeCarter: 🤦

FayeCarter: Fixed

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/src/store/mutation-types.js
```diff
@@ -9,3 +9,5 @@ export const UPDATE_ISSUES = 'UPDATE_ISSUES';
 export const UPDATE_ORDER_PLACED = 'UPDATE_ORDER_PLACED';
 export const UPDATE_STATE = 'UPDATE_STATE';
 export const UPDATE_USER_NOTE = 'UPDATE_USER_NOTE';
+export const UPDATE_CHANGED_FIELD = 'UPDATE_CHANGED_FIELD';
```
xander-marjoram: Nit: alphabetical please 😸

packages/components/organisms/f-checkout/src/store/tests/analytics.module.test.js
```diff
@@ -0,0 +1,391 @@
...
+
+                it(`should call ${UPDATE_AUTOFILL} with correct correct field`, () => {
+                    // Act
+                    updateAutofill({ commit }, checkoutState);
+
+                    // Assert
+                    expect(commit).toHaveBeenCalledWith(UPDATE_AUTOFILL, autofillFields);
+                });
+
+                it('should call `mapAnalyticsNames` with correct correct field', () => {
```
xander-marjoram: `correct correct`

packages/components/organisms/f-checkout/src/store/tests/analytics.module.test.js
```diff
@@ -0,0 +1,391 @@
...
+                expect(trackEventSpy).toHaveBeenCalledWith(expectedEvent);
+            });
+
+            describe('when user `isLoggedIn` is true', () => {
+                it('should set `name` to `Checkout 1 Overview`', () => {
+                    // Arrange
+                    const expectedName = 'Checkout 1 Overview';
+
+                    rootState.checkout.isLoggedIn = true;
+                    expectedEvent.custom.pageData.name = expectedName;
```
xander-marjoram, line 206: I would probably create a new object and override the value on the new object, otherwise you could have a situation where one test affects another.

packages/components/organisms/f-checkout/src/store/tests/analytics.module.test.js
FayeCarter: 🤦

packages/components/organisms/f-checkout/src/store/tests/analytics.module.test.js
xander-marjoram: You could maybe call the default one `baseEvent`, and then create a new variable inside each test called `expectedEvent` which you assert against.

FayeCarter: I've created a `beforeEach` for this so hopefully will prevent any issues

xander-marjoram (approved):

kevinrodrigues (approved):

PR #739, opened 2021-02-23
f-checkout@0.59.0 - Placing order and redirecting to payment page.

SachaZvetelman:
_Note: even though I've tested this in chunks, I haven't been able to do a full test given there are some dependencies that haven't been implemented in the underlying APIs (e.g. the `PATCH` Checkout API). It's possible that we'll have to amend a few things in the near future, but at least this is a step forward and hopefully we can close this ticket for now without being blocked._

v0.59.0
*February 24, 2021*

### Added
- `placeOrder` to call the Order Placement API.
- Redirect to the payment page after the order has been placed.

### Changed
- Restructured component object into page object model
- Refactored component and accessibility tests
- Extracted svgTransform to global test utils for reference by other packages
- How we pass the parameters to the `PATCH` Checkout request.


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -379,6 +405,39 @@ export default {
...
+            const data = {
+                basketId: this.basket.id,
+                applicationId: 7, // Responsive Web
+                customerNotes: {
+                    noteForRestaurant: this.userNote
+                },
+                applicationName: this.applicationName,
+                applicationVersion: '1',
+                referralState: 'None',
+                deviceId: '127.0.0.1', // TODO: TBC
```
SachaZvetelman, line 431: Still waiting for the Order team to see what we need to pass in here, or if we'll have to pass anything at all.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -2050,6 +2104,108 @@ describe('Checkout', () => {
...
+            });
+
+            afterEach(() => {
+                jest.clearAllMocks();
+                jest.clearAllTimers();
+            });
+
+            it('should redirect to the payment page after 1 second', () => {
+                // Arrange
+                const windowLocationSpy = jest.spyOn(window.location, 'assign');
```
FayeCarter: Could this have been done with this?
https://github.com/justeat/fozzie-components/pull/739/files#diff-78d9e6a8935b90cf33c6b12834013ca9bc4ae399aecaae2d684ec50eff9e40a0R81

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -379,6 +405,39 @@ export default {
             }
         },
 
+        /**
+         * Redirect to the payment page.
+         */
+        redirectToPayment () {
+            setTimeout(() => { // TODO: remove this when the order team handles this automatically.
+                window.location.assign(this.paymentPageUrl);
```
pkedziora: paymentPageUrl is in format /pay/[orderId]
It needs to be amended at runtime based on response from place order call.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -154,6 +154,16 @@ export default {
             required: true
         },
 
+        placeOrderUrl: {
```
FayeCarter, line 157: Can you add these to the Readme?

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: Good point, shouldn't have missed that. One of the problems of not being able to test everything end-to-end at the moment.

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -214,6 +222,35 @@ export default {
...
+         *
+         * @param {Object} context - Vuex context object, this is the standard first parameter for actions
+         * @param {Object} payload - Parameter with the different configurations for the request.
+         */
+        placeOrder: async ({ commit, state }, {
+            url, data, timeout
+        }) => {
+            const authHeader = state.authToken && `Bearer ${state.authToken}`;
+
+            const config = {
```
kevinrodrigues, line 236: This config looks to be the same shape or similar throughout? May be worth extracting this out so that the config can accept some options to cater for the other asynchronous calls so it injects a common object which can be extended for other calls and then passed through to axios. That way the other actions can just reuse or extend that without having to repeat it.

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/src/store/checkout.module.js
SachaZvetelman: If you're ok with this, @kevinrodrigues, I think we can do this as a separate PR, as an improvement. There are a few nuances, though. Some APIs expect `application/json`, some others `application/json;v=1`, some expect `x-je-feature`, some `x-je-tenant`, etc.

FayeCarter (dismissed):

packages/components/organisms/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -79,66 +83,68 @@ class Checkout extends Page {
...
-        if ( checkout.isAuthenticated ) {
-            authQueryString = '&knob-Auth%20token=a'; 
+        if (checkout.isAuthenticated) {
+            authQueryString = '&knob-Auth%20token=a';
             createGuestUrl = '';
         }
-        
-        let url = `checkout-component${checkoutUrl}&knob-Available%20Fulfilment%20Url=%2Fcheckout-available-fulfilment.json${authQueryString}${basketUrl}${createGuestUrl}`
+
+        const url = `checkout-component${checkoutUrl}&knob-Available%20Fulfilment%20Url=%2Fcheckout-available-fulfilment.json${authQueryString}${basketUrl}${createGuestUrl}${placeOrderUrl}${paymentPageUrlPrefix}`;
```
SachaZvetelman, line 99: Lines 91, 92 and 99 are the only ones I've changed. The other changes are linting fixes.

packages/components/organisms/f-checkout/src/store/checkout.module.js
kevinrodrigues: Yup i saw that. More of an improvement thing later I was thinking 👍

kevinrodrigues (approved):

FayeCarter (approved):

PR #741, opened 2021-02-23
f-content-cards@4.0.0: State Slots

minrwhite:
### Updated
- `ContentCards` component now surfaces card loading state via several named 'slots' as documented in README.md
- `PostOrderCard` now always opens its link in a new tab

---

## UI Review Checks

- [ ] This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist) - *N/A, only functional code is affected*

## Browsers Tested

- [x] Chrome (latest)
- [x] Firefox (latest)
- [ ] Internet Explorer 11


packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -336,9 +367,20 @@ export default {
             };
         }
     },
-    render () {
-        return this.$scopedSlots.default({
-            cards: this.cards
-        });
+    render (h) {
```
ashleynolan, line 384: Not sure if `h` is a convention, but what does it stand for?

packages/components/organisms/f-content-cards/stories/ContentCards.stories.js
```diff
@@ -162,58 +225,127 @@ export function ContentCardsCustom (args, { argTypes }) {
...
+         */
+        beforeCreate: zeroCardsMockSetup,
+
+        template
+    };
+}
+
+ContentCardsError.storyName = 'Error state';
+
+ContentCardsError.args = {
```
ashleynolan: Not sure what best practice is in storybook, but would be good to split this file up as it's getting pretty big now.

Perhaps having different files imported for each sub-story?

packages/components/organisms/f-content-cards/src/components/ContentCards.js
minrwhite: It's a convention, mentioned in the quote at the bottom of [this section](https://vuejs.org/v2/guide/render-function.html#JSX):

> **Aliasing createElement to h is a common convention you’ll see in the Vue ecosystem** and is actually required for JSX. Starting with version 3.4.0 of the Babel plugin for Vue, we automatically inject const h = this.$createElement in any method and getter (not functions or arrow functions), declared in ES2015 syntax that has JSX, so you can drop the (h) parameter. With prior versions of the plugin, your app would throw an error if h was not available in the scope.

I figured it would be better to have it b `h` for forward compatibility - but happy to rename back to `createElement` if this would be preferred.

minrwhite: Having slept on it, I think the best of both worlds will to keep it as `h`, but add a comment linking to the documentation above, and explaining that it is short for `createElement`, as this is apparently as yet unfamiliar usage within JET.

DamianMullins: We actually use `h` in the menu project as, like you say, it's a convention. You'll see in React/Preact also. 

There is some more [info and history here on CSS Tricks](https://css-tricks.com/what-does-the-h-stand-for-in-vues-render-method/).

minrwhite:
Marked as DNM in deference to #744 for the time being.

packages/components/organisms/f-content-cards/README.md
```diff
@@ -122,108 +122,108 @@ The following rudimentary example can be used as a guide for implementing this c
...
 ```
 
+
+## Slots
+
+The `ContentCards` component makes heavy use of slots in order to surface different states of data retrieval
+
+### `loading`
+
+The `loading` slot can used by the consuming code to surface a loading state to the user before card data has
```
dandel10n: "slot can be used" or "slot is used"

packages/components/organisms/f-content-cards/README.md
```diff
@@ -122,108 +122,108 @@ The following rudimentary example can be used as a guide for implementing this c
...
+been retrieved
+
+### `default`
+
+The `default` slot is used to display the cards when they have been received from the cards source instance, or
+alternatively injected via the `custom-cards` prop.
+
+### `no-cards`
+
+The `no-cards` slot can used to surface a message to the user when no cards have been received from the
```
dandel10n: the same here

packages/components/organisms/f-content-cards/README.md
```diff
@@ -122,108 +122,108 @@ The following rudimentary example can be used as a guide for implementing this c
...
+cards source(s)
+
+### `error`
+
+The `error` slot can used by the consuming code to surface a message to the user when an error has been
```
dandel10n: and here

packages/components/organisms/f-content-cards/src/components/cardTemplates/PostOrderCard.vue
```diff
@@ -79,6 +79,18 @@ export default {
         };
     },
 
+    computed: {
+        cardContainerCard () {
+            return {
+                ...this.card,
+                target: {
+                    attribute: '_blank',
```
dandel10n, line 87: For accessibility reasons, if we open link in a new tab, we need to add visually hidden text "opens in new tab" or something like that for screen readers to announce

packages/components/organisms/f-content-cards/stories/content-cards/all-cards.js
```diff
@@ -0,0 +1,72 @@
...
+
+        methods,
+
+        /**
+         * Ensures that card mocks are set up
+         */
+        beforeCreate: allCardsMockSetup,
+
+        template: `
+<content-cards
```
dandel10n: Indentation here is a bit off

packages/components/organisms/f-content-cards/stories/content-cards/custom-cards.js
```diff
@@ -0,0 +1,84 @@
...
+            }
+        },
+
+        /**
+         * Ensures that card mocks are set up
+         */
+        beforeCreate: zeroCardsMockSetup,
+
+        template: `
+<content-cards
```
dandel10n: And here

packages/components/organisms/f-content-cards/stories/content-cards/error.js
```diff
@@ -0,0 +1,48 @@
...
+
+        methods,
+
+        /**
+         * Ensures that card mocks are set up
+         */
+        beforeCreate: zeroCardsMockSetup,
+
+        template: `
+<content-cards
```
dandel10n: And here :)

packages/components/organisms/f-content-cards/stories/content-cards/loading.js
```diff
@@ -0,0 +1,58 @@
...
+         * Reject the promise set up by the above mock
+         */
+        beforeDestroy () {
+            this.rejectMockPromise({
+                status: 500
+            });
+        },
+
+        template: `
+<content-cards
```
dandel10n: The same here

packages/components/organisms/f-content-cards/stories/content-cards/no-cards.js
```diff
@@ -0,0 +1,49 @@
...
+
+        methods,
+
+        /**
+         * Ensures that card mocks are set up
+         */
+        beforeCreate: zeroCardsMockSetup,
+
+        template: `
+<content-cards
```
dandel10n: And here :)

packages/components/organisms/f-content-cards/README.md
minrwhite: @dandel10n IMO, "can be used", as it's not compulsory to supply content for the loading slot.

packages/components/organisms/f-content-cards/README.md
minrwhite: As above.

packages/components/organisms/f-content-cards/README.md
minrwhite: As above.

packages/components/organisms/f-content-cards/stories/content-cards/all-cards.js
minrwhite: @dandel10n - I was keeping it so the top level element in the template was at column zero for each of these, as IMO it makes it easier for the reader to interpret the template, as no indentation level is prescribed by linting for templates given inline here. What indent rules should I be following here, in your opinion?

dandel10n: I thought like this would look better: https://github.com/justeat/fozzie-components/blob/e0001773a4a12ec2686bb98fd90674e48106777b/packages/components/organisms/f-footer/stories/footer.stories.js#L34-L40

minrwhite: Thanks - I'll action for this and the ones below.

packages/components/organisms/f-content-cards/README.md
dandel10n: At the moment it says "slot can used"

minrwhite: 🤦‍♂️ There goes my brain, inserting words again. Thanks for pointing that out, will fix.

packages/components/organisms/f-content-cards/src/components/cardTemplates/PostOrderCard.vue
minrwhite: As discussed - I've added this to a ticket for actioning once we have translations for copy in our international markets.

packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -163,7 +184,7 @@ export default {
              * Emits voucher code click event with given ongoing url
              **/
             emitVoucherCodeClick (url) {
-                component.$emit('voucherCodeClick', {
+                component.$emit('voucher-code-click', {
```
kevinrodrigues: Be good to keep this event name in a central place. I've always found it helps with working out which events the application is using in a single file.

dandel10n (dismissed):

packages/components/organisms/f-content-cards/README.md
```diff
@@ -122,108 +122,108 @@ The following rudimentary example can be used as a guide for implementing this c
...
+
+> If this is missing or nully, all child components will also be rendered without test id
+> attributes.
 
 ## Events
 
 The following events can be emitted by the component, with the shape given:
 
-### `@voucherCodeClick`
+### `@voucher-code-click`
```
kevinrodrigues: While we're here is it possible to covert this to a table. Like [this](https://github.com/justeat/fozzie-components/tree/master/packages/components/molecules/f-searchbox#available-events)

DamianMullins (approved):

kevinrodrigues (approved):

packages/components/organisms/f-content-cards/src/components/ContentCards.js
ashleynolan: Catching up on this – think that was the best solution @benwhite-deltas. Thanks for the context too – really helpful

PR #743, opened 2021-02-24
f-header@4.8.0 - f-button update

dandel10n:
### Changed
- f-button and f-vue-icons packages version bump
- Country selector buttons updated
- Language specified for french link for Canada in country selector

### DIFF
--- a/packages/components/organisms/f-header/src/components/CountrySelector.vue
+++ b/packages/components/organisms/f-header/src/components/CountrySelector.vue
@@ -6,9 +6,8 @@
         data-test-id="countrySelector-toggle"
         v-on="isBelowMid ? null : { mouseover: openCountrySelector, mouseleave: closeCountrySelector }"
         @keyup.esc="closeCountrySelector">
-        <f-button
+        <button
             :tabindex="isBelowMid && !navIsOpen ? -1 : 0"
-            button-type="icon"
             class="c-nav-list-text c-countrySelector-btn"
             :aria-expanded="countrySelectorIsOpenOnDesktopView ? 'true' : 'false'"
             :aria-haspopup="!isBelowMid"
@@ -23,14 +22,14 @@
             <span class='c-countrySelector-title'>
                 {{ copy.countrySelector.selectYourCountryText }}
             </span>
-        </f-button>
+        </button>
 
         <div :class="['c-nav-panel', { 'is-visible': countrySelectorIsOpen }]">
             <div class="c-nav-popoverList c-nav-popoverList--countrySelector">
                 <header class="c-nav-popoverList-header">
                     <f-button
-                        button-type="icon"
-                        button-size="xsmall"
+                        button-type="ghost"
+                        is-icon
                         class="c-nav-popoverList-header-button"
                         :aria-label="copy.countrySelector.goBackToMainMenu"
                         @click="closeCountrySelector">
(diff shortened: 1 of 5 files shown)


packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -6,9 +6,8 @@
         data-test-id="countrySelector-toggle"
         v-on="isBelowMid ? null : { mouseover: openCountrySelector, mouseleave: closeCountrySelector }"
         @keyup.esc="closeCountrySelector">
-        <f-button
+        <button
```
dandel10n, line 9: Changed back to regular button because f-button icon is now not suitable for flag button as supposed to use just simple icons like arrows, crosses etc

AlixRichards (dismissed):

DamianMullins (dismissed):

packages/components/organisms/f-header/src/components/CountrySelector.vue
ashleynolan: I think one of the pie-designs for f-button includes having a button with an icon alongside it, but we need to amend the styles of f-button to do this.

[Designs are here](https://app.abstract.com/projects/da082ad2-0d14-40e8-a365-8eca471246e9/branches/master/commits/a60ab83f64456eca407dca472587b118d1a176f7/files/7daa54bc-5a0f-4634-ae91-82ebae259f3c/layers/D1D03086-5F69-421E-8591-33B6FD0F9D90?collectionId=7cce2661-005a-464f-9829-d70715caaffb&collectionLayerId=8ceb3ee2-1feb-4c91-b0f6-51783d824245) (you can see the `button with icon` design on these).

Would implementing these in f-button help potentially?

kevinrodrigues, line 9: Should this now specify the button type? e.g `<button type="button">`

dandel10n: @ashleynolan the problem is not with button with icon and text but with just an icon. When I tried to use f-button for it flag icon became blue. The behavior is correct for icons like cross, arrow, plus etc but not for flag icon.

regular button:
Screenshot 2021-02-24 at 12 16 48

f-button with isIcon prop:
Screenshot 2021-02-24 at 12 30 57

ashleynolan: Ah ok – I think this is an error in `f-button` – for the ghost type it shouldn't apply a text colour ([which it does here](https://github.com/justeat/fozzie-components/blob/master/packages/components/atoms/f-button/src/components/Button.vue#L357)). So if we fix this in future you should be able to use that component again then.

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -175,16 +176,12 @@ $countrySelector-flag-height : 16px;
...
-
-.c-nav-popoverList-go-back-icon {
-    transform: rotate(180deg);
-    width: 28px;
-    height: 28px;
-    display: inline-block;
 
-    use {
-        fill: $countrySelector-icon-color;
+    .c-nav-popoverList-go-back-icon {
```
minrwhite, line 180: What's the reason for this now being nested within `.c-nav-popoverList-header-button`?

minrwhite (approved):

packages/components/organisms/f-header/src/components/CountrySelector.vue
dandel10n: I moved it at first to overwrite f-button component styles but later decided not to use f-button for this button. But the class if very specific and won't be used outside of `c-nav-popoverList-header-button` so nesting shouldn't be a problem. But I can move it back if you like for consistency

dandel10n: I did this to overwrite f-button width and height styles. We will fix it in f-button component later because at the moment it is more specific than class in the consuming app.

charliearlie (approved):

PR #748, opened 2021-02-25
f-checkout@0.61.0 - Get geo location for checkout address

oliversweb:
---
Added a call during 'submit' to get the lat lng for the address (from the location services api) so it can be sent with the request to checkoutapi.

v0.61.0
*March 1, 2021*

### Added
- `getGeoLocation` to call the location services api.

### Changed
- Called the api during the Submit.
- Mapped the results into the `PATCH` Checkout request model.
- Refactored/Added unit tests.


siggerzz (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -505,6 +520,35 @@ export default {
             }
         },
 
+        /**
+         * Look up the geo details for the customers address, skip on failure.
+         *
+         */
+        async lookupGeoLocation () {
+            const postData = { addressLines: [] };
+            Object.keys(this.address).forEach(propName => {
```
SachaZvetelman: I'd rename `propName` to something else, given that `prop` in Vue's context refers normally to `props` and not `property` :)

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -505,6 +520,35 @@ export default {
...
+                if (this.address[propName]) {
+                    postData.addressLines.push(this.address[propName]);
+                }
+            });
```
SachaZvetelman: I think we can simplify this by doing:
```
const addressLines = Object.values(this.address).filter(addressLine => !!addressLine);

const postData = {
  addressLines
};

...
...
```

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -505,6 +520,35 @@ export default {
...
+
+            try {
+                if (postData.addressLines.length) {
```
SachaZvetelman: Are we sure we need this validation here? (asking, not suggesting 😄)

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -48,7 +49,8 @@ export default {
             isAsapAvailable: false
         },
         authToken: '',
-        isLoggedIn: false
+        isLoggedIn: false,
+        geolocation: null
```
SachaZvetelman, line 58: Nit: in line 225 we're camel-casing `geoLocation`, so we should camel-case here it as well.

packages/components/organisms/f-checkout/src/components/Checkout.vue
oliversweb: I didn't want to make an api call if for any reason the address fields are empty plus the api fails if <1 address line.

packages/components/organisms/f-checkout/src/store/checkout.module.js
oliversweb: I did have camel casing before but if I use this casing then it matches the casing on the outgoing model in the Submit, so then I can replace like for like in the mapper rather than assigning one to the other.

packages/components/organisms/f-checkout/src/components/Checkout.vue
xander-marjoram: Do we want to log anything if this `if` statement fails?

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -319,6 +340,10 @@ export default {
 
         [UPDATE_USER_NOTE]: (state, userNote) => {
             state.userNote = userNote;
+        },
+
+        [UPDATE_GEO_LOCATION]: (state, [lng, lat]) => {
+            state.geolocation = { latitude: lat, longitude: lng };
```
xander-marjoram: Nit: Can you spread this object out over multiple lines please?

packages/components/organisms/f-checkout/src/store/mutation-types.js
```diff
@@ -8,3 +8,4 @@ export const UPDATE_IS_FULFILLABLE = 'UPDATE_IS_FULFILLABLE';
 export const UPDATE_ISSUES = 'UPDATE_ISSUES';
 export const UPDATE_STATE = 'UPDATE_STATE';
 export const UPDATE_USER_NOTE = 'UPDATE_USER_NOTE';
+export const UPDATE_GEO_LOCATION = 'UPDATE_GEO_LOCATION';
```
xander-marjoram: Nit: alphabetical please

packages/components/organisms/f-checkout/src/components/Checkout.vue
kevinrodrigues: You could move this check to the vuex action instead, where the actual axios call is happening. Just means the async call (Vuex wise) will still happen but the actual axios call won't. Also it would be easier, if it's required later, to commit a different type of state update if we wanted to display a message or $emit a value up to coreWeb if  address lines aren't populated etc. (Also means the fail safe will be in the vuex action, which means no one else will be able to call the api if the criteria  doesn't meet)

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -204,6 +204,16 @@ export default {
         getAddressTimeout: {
```
kevinrodrigues: Tests :)

kevinrodrigues:
@oliversweb Looks really good 👍

packages/components/organisms/f-checkout/src/components/Checkout.vue
oliversweb: Discussed and leaving as is 👍

packages/components/organisms/f-checkout/src/components/Checkout.vue
oliversweb: On it

packages/components/organisms/f-checkout/src/components/Checkout.vue
oliversweb: Not currently

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1671,6 +1672,123 @@ describe('Checkout', () => {
...
+                        propsData,
+                        mocks: {
+                            $logger
+                        }
+                    });
+
+                    await flushPromises();
+                });
+
+                it('then it should not call `logWarn`', async () => {
```
kevinrodrigues: Could you change these so it starts with a `should`. e.g `it('should not call logWarn')`

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1671,6 +1672,123 @@ describe('Checkout', () => {
...
+                    // Assert
+                    expect($logger.logWarn).not.toHaveBeenCalledWith();
+                });
+
+                it('then it should call `getGeoLocation`', async () => {
+                    // Act
+                    await wrapper.vm.lookupGeoLocation();
+
+                    // Assert
+                    expect(getGeoLocationSpy).toHaveBeenCalled();
```
kevinrodrigues: Do we want to check that it's called with the correct values? i.e `url`, `postData` & `timeout`

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1671,6 +1672,123 @@ describe('Checkout', () => {
...
+                });
+
+                it('then it should call `logWarn`', async () => {
+                    // Act
+                    await wrapper.vm.lookupGeoLocation();
+
+                    // Assert
+                    expect($logger.logWarn).toHaveBeenCalledWith(
+                        'Geo location lookup failed',
+                        expect.anything(),
```
kevinrodrigues: Here you'd want to pass in the store. Haven't tested it but `wrapper.vm.store` should work.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1671,6 +1672,123 @@ describe('Checkout', () => {
...
+                        expect.anything()
```
kevinrodrigues: This one should be the error object.

packages/components/organisms/f-checkout/src/services/tests/mapper.test.js
```diff
@@ -121,6 +122,43 @@ describe('checkout mapper', () => {
         expect(notesRequest.length).toBe(1);
         expect(notesRequest[0].note).toBe(userNote);
     });
+
+    describe('when geo location', () => {
```
kevinrodrigues: If you structure it like below then you can `describe` each scenario better.

```javascript
describe('geoLocation')
  describe('when geolocation is present')
    it('should map the location correctly')
  describe('when geolocation is null')
    it('should map the location correctly')
```

We also might only really need one of these tests as all we're really testing is that the mapper works, one test should be fine.

packages/components/organisms/f-checkout/README.md
```diff
@@ -78,11 +78,13 @@ The props that can be defined are as follows:
 | `createGuestUrl` | `String` | - | URL for the API called to load the Create a Guest User. |
 | `getBasketUrl` | `String` | - | URL for the API called to get Basket Details.<br><br>The data returned from this API contains the serviceType, which determines if the Checkout component is created for Collection or Delivery when the user is not authenticated. |
 | `placeOrderUrl` | `String` | - | URL for the API called to place the order.<br><br>The data returned from this API contains the orderId, which is needed to redirect the user to the payment page. |
+| `getGeoLocationUrl` | `String` | - | URL for the API that can return geo location information (Latitude and Longitude) for a given address.<br>The `tenant` must be provided as the last segment of the URL and all calls must be authenticate [(Api Specification Link).](https://pages.github.je-labs.com/Architecture/api_specifications/#tag/Location-Services/paths/~1geocode~1{tenant}/post) |
```
xander-marjoram: Nit: `authenticated`

packages/components/organisms/f-checkout/src/demo/get-geo-location.json
```diff
@@ -0,0 +1,16 @@
+{
+  "type": "Feature",
+  "properties": {
+    "geocodingProvider": "Google",
+    "accuracyScore": 100,
+    "matchScore": 100
+  },
+  "geometry": {
+    "type": "Point",
+    "_comment": "I have verified with the Location Services Team and Longitute will be the first element and Latitude will be the second",
```
xander-marjoram: Nit: `Longitude`

kevinrodrigues (dismissed):

xander-marjoram (dismissed):

kevinrodrigues (approved):

SachaZvetelman (approved):

PR #754, opened 2021-03-01
f-cookie-banner@v0.3.1 - Default banner visiblity to hidden.

DamianMullins:
### Changed
- `shouldHideBanner` data property defaults to `true` in order to avoid flash of cookie banner after it has been dismissed.

### DIFF
--- a/packages/components/organisms/f-cookie-banner/CHANGELOG.md
+++ b/packages/components/organisms/f-cookie-banner/CHANGELOG.md
@@ -3,7 +3,7 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-v0.3.0
+v0.3.1
 ------------------------------
 *March 1, 2021*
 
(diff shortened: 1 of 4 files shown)


packages/components/organisms/f-cookie-banner/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.3.0
```
kevinrodrigues: v0.3.1

kevinrodrigues (dismissed):

packages/components/organisms/f-cookie-banner/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.3.1
+------------------------------
+*March 1, 2021*
+
+### Changed
+- `shouldHideBanner` data property defaults to `true` in order to avoid flash of cookie banner after it has been dismissed.
```
ashleynolan, line 11: Would this be a breaking change potentially rather than a patch? (although it's at pre v1 atm).

DamianMullins: I did consider that when I made the change but I think it is more of a fix (patch) as it fixes a minor UI issue and requires no work from people using the package. I'll update the changelog heading to reflect that.

dandel10n (approved):

kevinrodrigues (approved):

PR #756, opened 2021-03-02
f-header@4.9.0 - popover component

dandel10n:
Instead of having `CountrySelector` component and `Navigation` which both have popover functionality, I've separated  `Popover` component to `@justeat/f-popover` and `CountrySelectorPanel` and `UserNavigationPanel` components to go inside popover component.  

For the end users everything should stay the same.

### Added
- `@justeat/f-popover` component
- `CountrySelectorPanel` component
- `UserNavigationPanel` component
- Main menu title for all the tenants for screen readers

### Updated
- `@justeat/f-button` updated to v1.1.0
- tests to cover the changes

### Removed
- `CountrySelector` component


charliearlie (approved):

packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
```diff
@@ -0,0 +1,145 @@
...
+
+                <h3>{{ copy.countrySelector.selectYourCountryText }}</h3>
+            </header>
+
+            <ul
+                class="c-nav-popoverList c-nav-popoverList--twoColumns"
+                data-test-id="c-countrySelector-list">
+                <li
+                    v-for="(country, i) in countries"
+                    :key="i + '_Country'"
```
DamianMullins, line 22: Might be good to use `country.key` in the key here rather than `index` as it allows Vue to optimise DOM updates if these items ever change. We were but by this on the menu recently where text would update in items but images remained in the wrong place. 

It's probably not much of an issue here as the data is static but it's a good practice to follow.

```suggestion
                    v-for="country in countries"
                    :key="country.key + '_Country'"
```

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -669,15 +684,18 @@ $nav-popover-padding               : spacing(x2);
...
-            }
+    .c-nav-popoverList--twoColumns {
+        @include media('>=mid') {
+            column-count: 2;
+        }
+    }
+
+    // TODO: MAKE THIS NOT USE FLOATS
+    // global modifier for list items horizontally aligned
+    .c-nav-list-item--horisontallyAlignedOnWiderView {
```
DamianMullins, line 696: Could use the `AboveMid` naming convention here?

```suggestion
    .c-nav-list-item--horisontallyAlignedAboveMid {
```

packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
dandel10n: The problem with having `country.key` here is that we have `ca` and `ch` key used multiple times as we show separate links for different languages for Canada and Switzerland. We use `country.key` for showing correct flags

dandel10n: I can specify separate keys for flags and have unique keys for every link

zlatinivanov-je (approved):
~~Just a couple of suggestions on `$emit` events:~~
- ~~Would it make sense to select emit event names including a reference to the component, will be able to easily identify and remain unique to the component~~
- ~~Also could be a good idea to use constants file containing the event names, will avoid possible typo issues and provides IntelliSense.~~ 

Ignore the above I was thinking of those being emitted in a wider scope.

packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
DamianMullins: Right, yeah I see. I'll leave it up to you if you want to make this change or not 😺

dandel10n: Actually, I have a next PR ready to come, so I will include this change to that one

packages/components/organisms/f-header/src/components/Navigation.vue
dandel10n: I will include this change to the next PR

packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
xander-marjoram: I guess we could potentially use the culture as a key? i.e., `en-CA` and `fr-CA`.

packages/components/organisms/f-header/test-utils/component-objects/f-header.selectors.js
```diff
@@ -17,7 +17,7 @@ export const NAVIGATION = {
     }, 
     countrySelector: {
         link: '[data-test-id="country-selector"] button', 
-        currentIcon: '[data-test-id="action-button-component"] svg[class*="current"]',
-        countryList: '[class="c-countrySelector-list"] li', 
+        currentIcon: '[data-test-id="current-flag-icon"]',
+        countryList: '[data-test-id="c-countrySelector-list"] li'
```
xander-marjoram, line 21: Does/should the data-test-id have the `c-` prefix? I thought this was only for styles

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -52,7 +52,7 @@
             <ul class="c-nav-list">
                 <li
                     v-if="showOffersLink"
-                    class="c-nav-list-item">
+                    class="c-nav-list-item--horisontallyAlignedOnWiderView">
```
xander-marjoram, line 55: `horizontally`

packages/components/organisms/f-header/test-utils/component-objects/f-header.selectors.js
dandel10n: Good point, I will add the change to the next PR

PR #763, opened 2021-03-02
f-popover@0.2.0 - component markup

dandel10n:
### Added
- Component markup
- Storybook story

Screenshot 2021-03-03 at 10 10 54


kevinrodrigues (dismissed):

packages/components/atoms/f-popover/stories/Popover.stories.js
```diff
@@ -3,18 +3,38 @@
 //     withKnobs, select, boolean
 // } from '@storybook/addon-knobs';
```
minrwhite: Can we remove these unless we're going to use them?

minrwhite (dismissed):

kevinrodrigues (dismissed):

tommcclean (dismissed):

packages/components/atoms/f-popover/src/components/Popover.vue
```diff
@@ -2,14 +2,14 @@
...
-        I am a VPopover Component (GB)
+        <slot></slot>
     </div>
 </template>
 
 <script>
 
 export default {
-    name: 'VPopover',
+    name: 'Popover',
```
ashleynolan: I think this was added as a [styleguide recommendation](https://vuejs.org/v2/style-guide/#Base-component-names-strongly-recommended). Think we need to put a PR in to amend other components to match this format too

packages/components/atoms/f-popover/src/components/Popover.vue
```diff
@@ -18,15 +18,32 @@ export default {
 
 <style lang="scss" module>
 
+$popover-radius                : 3px;
```
ashleynolan: Should update this to `2px` to match the new PIE guidelines for border radius.

packages/components/atoms/f-popover/src/components/Popover.vue
```diff
@@ -18,15 +18,32 @@ export default {
...
-    display: flex;
-    justify-content: center;
-    min-height: 80vh;
-    width: 80vw;
-    margin: auto;
-    border: 1px solid $red;
-    font-family: $font-family-base;
-    @include font-size(heading-m);
+    @include media('>=mid') {
+        background-color: $white;
```
ashleynolan: Should use `$color-bg--component` here

packages/components/atoms/f-popover/src/components/Popover.vue
```diff
@@ -18,15 +18,32 @@ export default {
...
+        border-top: 0;
+        border-radius: 0 0 $popover-radius $popover-radius;
+        padding: 0 $popover-padding;
+        width: auto;
+
+        // tooltip arrow
+        &:before {
+            bottom: 100%;
+            border: $tooltip-width solid transparent;
+            border-bottom: $tooltip-width solid $white;
```
ashleynolan: and same here – `$color-bg--component`

packages/components/atoms/f-popover/stories/Popover.stories.js
```diff
@@ -1,20 +1,36 @@
...
+        <div style="position: relative; padding: 16px;">
+            Some text
+            <popover 
+                style="position: absolute;
+                        top: 100%;
+                        width: 300px;
+                        opacity: 1;
+                        z-index: 99999;
+                        left: 0;">
+                <card
```
ashleynolan: Do you need the card component in this base story? Would be better if it could just be a header/text passed in so that we don't tie the two components together in the base story if possible (can always have additional stories to show other ways the component can work)

packages/components/atoms/f-popover/src/components/Popover.vue
ashleynolan: (can just use `$border-radius` variable)

dandel10n: Good point, I just found a popover design in abstract, will check and update

ashleynolan: Cool – if it's not 2px, make sure you flag it with design, as that's the only border-radius they've defined in the design-tokens atm

dandel10n: It is 2px so all good, plus I removed border and updated the shadow according to the latest designs.

DamianMullins (approved):

minrwhite (approved):

PR #766, opened 2021-03-03
f-checkout@v0.63.0 -export analytics module

FayeCarter:
v0.63.0
*March 3, 2021*

### Added
- `analytics.module` export.

### DIFF
--- a/packages/components/organisms/f-checkout/CHANGELOG.md
+++ b/packages/components/organisms/f-checkout/CHANGELOG.md
@@ -3,6 +3,15 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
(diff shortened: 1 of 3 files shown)


SachaZvetelman (approved):

LTurns (approved):

JoshuaNg2332 (approved):

ashleynolan (approved):

DamianMullins (approved):

tommcclean:
As discussed - I think you need to rename your state module so that it doesn't interfere with one with the same name in Coreweb already

kevinrodrigues:
> As discussed - I think you need to rename your state module so that it doesn't interfere with one with the same name in Coreweb already

@FayeCarter I've used this naming convention as part of this discussion for f-searchbox recently: https://github.com/justeat/fozzie-components/discussions/730#discussioncomment-397679

You could attach the namespace for f-checkout to that module name.

PR #767, opened 2021-03-03
f-button@1.1.0 - f-button--icon sizes #trivial

### LINKED ISSUE #762: f-button icon button CSS specificity
Hi all, I've noticed that [this rule within `f-button`](https://github.com/justeat/fozzie-components/blob/fb9a39ecea8bf3bac521f453c574d91b3eb149d7/packages/components/atoms/f-button/src/components/Button.vue#L332-L335) is causing issues because it has a higher specificity than most four other rules, you can [see an example of this in our storybook docs in the mega-modal](https://justeat.github.io/fozzie-components/@justeat/storybook/?path=/story/components-molecules--mega-modal-component) where we're using f-button but the icon style overrides the width and height.

I'm not sure how best to fix this as the SVG which the CSS is targeting isn't actually in `f-button` but rather it's provided via a slot. Normally I'd create a new class and apply it to the SVG.

Any help on this would be appreciated.

dandel10n:
Related to https://github.com/justeat/fozzie-components/issues/762

### Added
- Separate story for `f-button--icon`

### Changed
- `f-button--icon` styles to show different button sizes for `--sizeMedium`, `--sizeLarge`, `--sizeSmall` and `--sizeXSmall`
- `f-button--icon` styles to show blue background for large f-button--icon instead of orange

Screenshot 2021-03-03 at 11 47 18

Screenshot 2021-03-03 at 11 47 25

Screenshot 2021-03-03 at 11 47 33

Screenshot 2021-03-03 at 11 47 42

### DIFF
--- a/packages/components/atoms/f-button/src/components/Button.vue
+++ b/packages/components/atoms/f-button/src/components/Button.vue
@@ -192,7 +192,8 @@ $btn-sizeXSmall-lineHeight      : 1;
  * Sets the btn colour to site primary colour
  */
 
-.o-btn--primary {
+.o-btn--primary,
+.o-btn--icon.o-btn--primary.o-btn--sizeLarge {
     background-color: $btn-primary-bgColor;
 
     &,
@@ -328,13 +329,13 @@ $btn-sizeXSmall-lineHeight      : 1;
  */
 
 .o-btn--icon {
+    display: flex;
+    align-items: center;
+    justify-content: center;
     border-radius: 50%;
-    svg {
-        width: 90%;
-        height: 80%;
-    }
 
-    &.o-btn--primary {
+    &.o-btn--primary,
+    &.o-btn--primary.o-btn--sizeLarge {
         path {
             fill: $btn-primary-textColor;
         }
@@ -375,6 +376,47 @@ $btn-sizeXSmall-lineHeight      : 1;
                 fill: $color-link-active;
             }
         }
+    }
+}
+
+.o-btn--icon.o-btn--sizeLarge {
+    width: 56px;
+    height: 56px;
+    padding: 0;
+
+    svg {
+        width: 21px;
+        height: 21px;
+    }
+}
+.o-btn--icon.o-btn--sizeMedium {
+    width: 48px;
+    height: 48px;
+    padding: 0;
+
+    svg {
+        width: 21px;
+        height: 21px;
+    }
+}
+.o-btn--icon.o-btn--sizeSmall {
+    width: 40px;
+    height: 40px;
+    padding: 0;
+
+    svg {
+        width: 18px;
+        height: 18px;
+    }
+}
+.o-btn--icon.o-btn--sizeXSmall {
+    width: 32px;
+    height: 32px;
+    padding: 0;
+
+    svg {
+        width: 18px;
+        height: 18px;
     }
 }
 
--- a/packages/components/atoms/f-button/src/components/Button.vue
+++ b/packages/components/atoms/f-button/src/components/Button.vue
@@ -334,8 +334,7 @@ $btn-sizeXSmall-lineHeight      : 1;
     justify-content: center;
     border-radius: 50%;
 
-    &.o-btn--primary,
-    &.o-btn--primary.o-btn--sizeLarge {
+    &.o-btn--primary {
         path {
             fill: $btn-primary-textColor;
         }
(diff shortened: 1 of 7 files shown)


DamianMullins (dismissed):

DamianMullins (dismissed):

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -192,7 +192,8 @@ $btn-sizeXSmall-lineHeight      : 1;
  * Sets the btn colour to site primary colour
  */
 
-.o-btn--primary {
+.o-btn--primary,
+.o-btn--icon.o-btn--primary.o-btn--sizeLarge {
```
DamianMullins, line 205: I saw you removed this style earlier, did you actually need it in the end?

dandel10n: Oh, looks like I messed up with a rebase

dandel10n: Oh no, this one is correct, I removed one about fill colour and background should be changed

dandel10n: For regular primary button background colour changes to orange if size is large when for icon button it should stay blue

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -378,6 +378,47 @@ $btn-sizeXSmall-lineHeight      : 1;
     }
 }
 
+.o-btn--icon.o-btn--sizeLarge {
+    width: 56px;
```
ashleynolan: I'd probably pull out these value as variables and define them at the top of the file, perhaps:

`$btn--icon--sizeLarge-btnLength`
&
`$btn--icon--sizeLarge-iconSize`

and similar for the other sizes.

Just makes them easier to update in the future if they ever change.

packages/components/atoms/f-button/stories/IconButton.stories.js
```diff
@@ -0,0 +1,65 @@
...
+        },
+        actionType: {
+            default: select('Action Type', ['button', 'submit', 'reset'], 'button')
+        },
+        href: {
+            default: text('href', '')
+        }
+    },
+    template: `
+    <div style="display: flex; width: 300px; justify-content: space-between">
```
ashleynolan: I'd be tempted to pull these out and assign the styles to a class defined in [our storybook config](https://github.com/justeat/fozzie-components/blob/master/packages/storybook/config/storybook/preview-head.html#L4), rather than adding inline styles.

packages/components/atoms/f-button/test-utils/component-objects/f-button.component.js
```diff
@@ -8,7 +8,7 @@ module.exports = class Buttons extends Page {
     open (buttonType = '') {
         const url = buttonType === 'link' ? '&knob-Button%20Type=link&knob-href=link' : '';
 
-        super.openComponent('atom', `button-component${url}`);
+        browser.url(`/iframe.html?id=components-atoms-f-button--button-component${url}`)
```
ashleynolan, line 11: Is there a way of making this work with the class method?

I think  

```js
super.openComponent('atom', `f-button-button-component${url}`);
```
should work?

dandel10n: The changes should be done in page object model to make it work as I made f-button story to be a folder and not just a page, but @LTurns  told that they are refactoring the page object at the moment and recommended to do this  just by calling `browser.url` directly

packages/components/atoms/f-button/stories/IconButton.stories.js
dandel10n: I thought that the styles there are universal for all the stories and here I specified some just for this specific one

packages/components/atoms/f-button/test-utils/component-objects/f-button.component.js
LTurns: yeah exactly, Jonkers is working on refactoring the page object model with the components at the moment, so it saves the hassle of having to make these extra changes when it's going to be restructured 😄

packages/components/atoms/f-button/stories/IconButton.stories.js
ashleynolan: I'd move them out, just as a pattern like this could be used for spacing other things (multiple checkboxes, radio buttons, form-fields etc)

AlixRichards (approved):

ashleynolan (approved):

PR #769, opened 2021-03-03
f-checkout@v0.64.0 - rename analytics module

FayeCarter:
### Changed
- `analytics` to `fCheckout.module`.
- `analytics.module` to `fCheckoutAnalytics.module`.

### Removed
- `f-trak` dependency.


DamianMullins (dismissed):

SachaZvetelman (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -308,7 +308,7 @@ export default {
             'getGeoLocation'
         ]),
 
-        ...mapActions('analytics', [
+        ...mapActions('checkoutAnalytics', [
```
kevinrodrigues: Would be good to centralise this name so we can change it once throughout. E.g [here](https://github.com/justeat/fozzie-components/blob/master/packages/components/molecules/f-searchbox/src/services/constants.js#L13)

packages/components/organisms/f-checkout/src/index.js
```diff
@@ -8,7 +8,7 @@
 // Import vue component
 import VueCheckout from '@/components/Checkout.vue';
 import CheckoutModule from '@/store/checkout.module';
-import AnalyticsModule from '@/store/analytics.module';
+import CheckoutAnalyticsModule from '@/store/checkoutAnalytics.module';
```
kevinrodrigues: Not sure this one needs to change? When you register the module you can attach the name space there which should avoid the clash? (could be wrong but worth seeing if that's the case)

packages/components/organisms/f-checkout/src/components/Checkout.vue
FayeCarter: @kevinrodrigues I've changed `checkout` to `fCheckoutModule` and `analytics` to `fCheckoutAnalyticsModule` Are these good?

packages/components/organisms/f-checkout/src/index.js
FayeCarter: Hey, I can't remember if we talked since this, but it's so the `...mapState('analytics` ....` calls don't get confused with the CoreWeb analytics module.

packages/components/organisms/f-checkout/src/store/checkoutAnalytics.module.js
```diff
@@ -1,5 +1,6 @@
-import Trak from '@justeat/f-trak';
+// import Trak from '@justeat/f-trak';
```
tommcclean: del

FayeCarter: noooooo 🤦

packages/storybook/context/dataLayer.context.js
```diff
@@ -0,0 +1,3 @@
+export default () => {
+    window.dataLayer = window.dataLayer || [];
```
tommcclean, line 2: You should import this in the index to make sure storybook pulls it in

packages/storybook/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.28.0
+------------------------------
+*March 3, 2021*
+
+### Changed
+- Ordering of sub-pages updated with new documentation added.
```
tommcclean: Message is copy of below

packages/components/organisms/f-checkout/stories/Checkout.stories.js
```diff
@@ -7,8 +7,8 @@ import {
 } from '@justeat/storybook/constants/globalisation';
 
 import VueCheckout from '../src/components/Checkout.vue';
-import CheckoutModule from '../src/store/checkout.module';
-import AnalyticsModule from '../src/store/analytics.module';
+import fCheckoutModule from '../src/store/checkout.module';
```
tommcclean, line 10: Id have just called it CheckoutModule, but not important :D

packages/storybook/CHANGELOG.md
FayeCarter: 🤦 it has been a long day

tommcclean (dismissed):

tommcclean (approved):

packages/components/organisms/f-checkout/src/store/checkoutAnalytics.module.js
```diff
@@ -78,20 +78,18 @@ export default {
          * Pushes `form` event to the dataLayer with correct data
          */
         trackFormInteraction ({ state, rootState }, { action, error }) {
-            const formName = rootState.checkout.isLoggedIn ? 'checkout' : 'checkout_guest';
+            const formName = rootState[VUEX_CHECKOUT_MODULE].isLoggedIn ? 'checkout' : 'checkout_guest';
 
             const mappedError = error ? mapAnalyticsNames(error) : null;
 
-            Trak.event({
+            window.dataLayer.push({
```
ashleynolan, line 85: Not sure if I missed the conversation  – why are we switching to dataLayer calls directly in the code?

Would be better to abstract this as otherwise we end up with 3rd party specific API calls (i.e. GA dataLayer calls) throughout our code. Main reason we try to avoid this is because Google has a history of changing the API structure of these calls and that makes them every brittle.

If we abstract these calls, we only have to change the call to dataLayer in one place rather than lots of places.

ashleynolan (changes requested):

packages/components/organisms/f-checkout/src/store/checkoutAnalytics.module.js
FayeCarter: Hey, so `f-trak` doesn't work in CoreWeb, @tommcclean had a look into it yesterday and found that a dependency, way down the line that is doesn't work on isomorphic web apps. 😬

ashleynolan: Ok, makes sense – in that case, we should create some tickets to fix `f-trak` so that it's isomorphic.

Don't mind this going in as a temporary solution, but we should make sure we fix the root problem too otherwise it'll come back to bite us later.

DamianMullins (approved):

ashleynolan (approved):
Approving – will look to switch these calls back to trak calls once that package is updated to be isomorphic (tickets to be created)

PR #770, opened 2021-03-03
f-footer@4.9.1 Fix Mobile Link lists collapse on scroll

zlatinivanov-je:
Fix Mobile Link lists collapse on scroll
----------------

## Fix 

- Links list collapses on page scroll for small screens
  - window resize event filters on screen width to avoid toggle links display, when address bar collapses

## Changed

- Class `is-collapsed` is only applied to below `wide`
- Preserves collapsed state for session

See bug video attached: 
https://user-images.githubusercontent.com/26770126/109804765-88243400-7c1a-11eb-9327-6c715e754941.mp4


packages/components/organisms/f-footer/src/components/LinkList.vue
```diff
@@ -70,7 +70,7 @@ export default {
 
     data () {
         return {
-            panelCollapsed: false,
+            panelMobileCollapsed: true,
```
dandel10n: Can we leave the name `panelCollapsed` as it is used not only on mobiles but on tablets too, and basically can be used on any width if needed

packages/components/organisms/f-footer/src/components/LinkList.vue
```diff
@@ -2,7 +2,7 @@
     <div
         v-if="linkList.links.length"
         :class="[$style['c-footer-panel'], {
-            'is-collapsed': panelCollapsed && isBelowWide
+            'is-collapsed': panelMobileCollapsed
```
dandel10n: Can we leave a check for `isBelowWide` here for not having this class for wider view?

zlatinivanov-je: That was causing the issue, as the value changes on scroll on mobile, although those should only be on resize or when mounted. Additionally it seems to add complexity that is does not seem to be needed.

packages/components/organisms/f-footer/src/components/LinkList.vue
zlatinivanov-je: ok

packages/components/organisms/f-footer/src/components/LinkList.vue
zlatinivanov-je: Also the check based on `panelCollapsed` only triggers when below wide.

packages/components/organisms/f-footer/src/components/LinkList.vue
```diff
@@ -70,7 +70,7 @@ export default {
 
     data () {
         return {
-            panelCollapsed: false,
+            panelCollapsed: true,
```
ashleynolan: Out of interest – what happens if JS never loads? Do the panels stay collapsed, or are they open in that scenario?

packages/components/organisms/f-footer/src/components/LinkList.vue
ashleynolan: So was the bug that `isBelowWide` was changing when the page was scrolled? Just trying to get my head around what was the cause of the issue itself 😄

zlatinivanov-je: For some reason and only on mobile `isBelowWide` gets a value reset on scroll. I can only explain it with page scrollbar appear when the user scrolls... that has an effect on the page width, and triggers the resize, followed by values reset. 

This change removes the resize collapse reset trigger. The display behaviour is based only on CSS media queries and presence of `is-collapsed`. The resize listener only enables/disables the ability to toggle.

ashleynolan: makes sense – and agree, it's probably the scrollbar appearing/disappearing triggering it (I think I've come across a similar thing before now you've mentioned that)

zlatinivanov-je: Correction, It is the collapsing address bar. - https://developers.google.com/web/updates/2016/12/url-bar-resizing 
@ashleynolan ^^

packages/components/organisms/f-footer/src/components/LinkList.vue
zlatinivanov-je: Wouldn't that completely fail to load the Vue content? Not sure how that works for SSR...

ashleynolan: Yeh, it's SSR that I'm thinking about. I think that's why we have the default state set to false atm, because we'd want those links to be visible if the client JS failed to load for some reason (and some search bots may not parse the JS correctly).

zlatinivanov-je: One way to get around that would be to revert back to `false` and set on mount based on screen size. However that would need to re-introduce screen resize listener for responsive... Let me have a look again...

packages/components/organisms/f-footer/src/components/LinkList.vue
```diff
@@ -88,32 +86,35 @@ export default {
...
-        }
+        this.setPanelCollasped();
     },
-
     destroyed () {
         windowServices.removeEvent('resize', this.onResize);
     },
-
     methods: {
+        setPanelCollasped () {
```
ashleynolan: Small thing – can you add JSDoc comments to any new methods added ([like this style](https://github.com/justeat/fozzie-components/blob/master/packages/components/molecules/f-searchbox/src/components/Form.vue#L410-L416)).

We're trying to add these to all methods going forward as well as go back and add them to existing methods (so if you feel like you can add them to the existing methods below too, that would be awesome too 👍 ).

xander-marjoram: Also typo here, should be `collapsed`

packages/components/organisms/f-footer/src/components/LinkList.vue
```diff
@@ -88,32 +86,35 @@ export default {
...
             }
         },
-
         onResize () {
-            this.currentScreenWidth = windowServices.getWindowWidth();
+            const newScreenWidth = windowServices.getWindowWidth();
 
-            if (this.isBelowWide) {
-                this.panelCollapsed = true;
+            // Mobile trigger `resise` on scroll - address bar collapse
```
xander-marjoram: Typo: resize

packages/components/organisms/f-footer/src/components/LinkList.vue
zlatinivanov-je: Oh no! Not again.

packages/components/organisms/f-footer/src/components/LinkList.vue
```diff
@@ -88,32 +86,51 @@ export default {
...
         onPanelClick () {
             if (this.isBelowWide) {
-                this.panelCollapsed = !this.panelCollapsed;
+                this.interactedState = !this.panelCollapsed;
+                this.setPanelCollapsed();
             }
         },
-
+        /**
+         * Hanles `resize` window events.
```
simonsandersje: Nit: Handles `resi.....

simonsandersje (dismissed):

ashleynolan (approved):

packages/components/organisms/f-footer/src/components/LinkList.vue
```diff
@@ -88,32 +86,51 @@ export default {
...
         onResize () {
-            this.currentScreenWidth = windowServices.getWindowWidth();
+            const newScreenWidth = windowServices.getWindowWidth();
 
-            if (this.isBelowWide) {
-                this.panelCollapsed = true;
+            if (this.currentScreenWidth !== newScreenWidth) {
+                this.currentScreenWidth = newScreenWidth;
+
+                this.setPanelCollapsed();
```
xander-marjoram, line 133: Is it worth throttling this event for manual resize on desktop?

zlatinivanov-je: This is already throttled where the `resize` event listener is added, line [91](https://github.com/justeat/fozzie-components/pull/770/files/37b19880de927d8efcf73fb46e8ba738aa565bb1#diff-f86486deb63b13c5ddc77cfb17849fc0fbdb781766f2db38652de9cc135535c8R91) 
The other uses are based on user interaction.

xander-marjoram (approved):

packages/components/organisms/f-footer/src/components/LinkList.vue
```diff
@@ -88,32 +86,51 @@ export default {
...
-        }
+        this.setPanelCollapsed();
     },
-
     destroyed () {
         windowServices.removeEvent('resize', this.onResize);
     },
-
     methods: {
+        /**
```
kevinrodrigues, line 99: Tests?

PR #776, opened 2021-03-05
f-mega-modal@v0.6.0 - Updated npm dependencies.

DamianMullins:
### Changed
- Updated npm dependencies.


dandel10n (dismissed):
I assume all looked good after the update? I found that I had to clean up some styles after updating f-header

DamianMullins:
@dandel10n All looked good when I ran the change in Storybook, yes, thanks for the fix! I think there are likely some CSS changes that could be handled by `f-button` now but I'll leave that for another PR.

matthewhardern (dismissed):

DamianMullins:
@dandel10n My bad, there were some styles that needed to be updated as the icon alignment was slightly off.

kevinrodrigues (approved):

xander-marjoram (approved):

PR #779, opened 2021-03-08
f-button@v1.1.1 - Minor fixes.

DamianMullins:
### Changed
- Formatting in storybook files.

### Fixed
- Pass button attributes to component correctly.
- CSS class name spelling mistake.

### DIFF
--- a/packages/components/atoms/f-button/src/components/Button.vue
+++ b/packages/components/atoms/f-button/src/components/Button.vue
@@ -8,9 +8,9 @@
             $style[`o-btn--size${buttonSizeClassname}`],
             (isFullWidth ? $style['o-btn--fullWidth'] : '')
         ]"
-        :attributes="$attrs"
         :action-type="buttonActionType"
         :data-test-id="`${componentType}-component`"
+        v-bind="$attrs"
         v-on="$listeners">
         <slot />
     </component>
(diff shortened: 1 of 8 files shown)


packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -8,9 +8,9 @@
             $style[`o-btn--size${buttonSizeClassname}`],
             (isFullWidth ? $style['o-btn--fullWidth'] : '')
         ]"
-        :attributes="$attrs"
         :action-type="buttonActionType"
         :data-test-id="`${componentType}-component`"
+        v-bind="$attrs"
```
minrwhite: As the resolved `component` element ends up being the root element of the `Button` component, and we haven't set [`inheritAttrs: false`](https://vuejs.org/v2/guide/components-props.html#Disabling-Attribute-Inheritance) for `Button`, is it necessary to set this?

ashleynolan (dismissed):

packages/components/atoms/f-button/src/components/Button.vue
DamianMullins: Looks like it isn't, I have removed this now, thanks!

minrwhite (approved):

AlixRichards (approved):

zlatinivanov-je (approved):
Should we have some rules in place for code formatting, some of the changes here look like could change based on personal preferences. Or are there any in the project already?

DamianMullins:
I tend to follow the linting rules and maybe some of them are no longer enforced but I still follow them such as object and import properties on new lines, as well as whitespace between object properties - I find the code easier to scan when there is some breathing room between lines.

I can have a dig around to see if rules exist for that type of thing.

PR #782, opened 2021-03-08
f-checkout@v0.67.0 - add analytics error tracking

FayeCarter:
### Added
- Analytics error tracking.
- Tests to cover changes.

### Fixed
- Guest field inputs Vuex mutation error.


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -240,18 +244,20 @@ export default {
     computed: {
         ...mapState(VUEX_CHECKOUT_MODULE, [
             'address',
+            'basket',
             'customer',
+            'errors',
+            'geolocation',
             'id',
-            'isFulfillable',
+            'isFulfilable',
```
DamianMullins: I think this was right before?

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -240,18 +244,20 @@ export default {
...
             'isLoggedIn',
+            'isable',
```
DamianMullins: This one looks like it's missing the start of the word possibly?

packages/components/organisms/f-checkout/src/components/Checkout.vue
FayeCarter: 🤦

packages/components/organisms/f-checkout/src/components/Checkout.vue
FayeCarter: 🤦

JoshuaNg2332 (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -240,18 +244,19 @@ export default {
     computed: {
         ...mapState(VUEX_CHECKOUT_MODULE, [
             'address',
+            'basket',
             'customer',
+            'errors',
+            'geolocation',
```
JoshuaNg2332, line 250: nit/question: Should this be `geoLocation`? Not a massive deal but just noticed other times we use geolocation, we use `GEO_LOCATION` or something of the sort.

FayeCarter: Hey, so this isn't something I've added in, I've just changed the order of the list, so not 100% on why there's `geolocation` and `geoLocation`. Can change it if needed, but I think it would need to be another ticket.

xander-marjoram: I believe `geolocation` is what is returned from the API, so by having it named that way in the state you can just destructure the response

SachaZvetelman: Yes, it should be in my opinion, but not as part of this PR.
I've mentioned this in this PR: https://github.com/justeat/fozzie-components/pull/748#discussion_r582921311
But I still think we should fix it.

kevinrodrigues (approved):

packages/components/organisms/f-checkout/src/components/Checkout.vue
FayeCarter: I'll create a ticket for it

JoshuaNg2332 (approved):

PR #786, opened 2021-03-09
Add Browserstack Support to tests. #trivial

JoshuaNg2332:
f-footer, f-header, f-registration
*March 12, 2021*

## Added
- Browserstack test config in `package.json`

### Changed
- Restructured component tests to support Browserstack


packages/tools/generator-component/generators/app/templates/__package__.json
```diff
@@ -33,8 +33,8 @@
     "lint:fix": "yarn lint --fix"<% if(config.isComponent) { %>,
     "lint:style": "vue-cli-service lint:style"<% } %>,
     "test": "vue-cli-service test:unit"<% if(config.needsComponentTests) { %>,
-    "test-component:chrome": "wdio ../../../../wdio.conf.js --suite component"<% } %>,
-    "test-a11y:chrome": "wdio ../../../../wdio.conf.js --suite a11y"
+    "test-component:chrome": "wdio ../../../../wdio-chrome.conf.js --suite component"<% } %>,
```
siggerzz: `cross-env-shell JE_ENV=local` will need to be added here too

test/utils/browserstack-helper.js
```diff
@@ -0,0 +1,53 @@
...
+    project: 'Fozzie-Components',
+    build: browserstackName,
+    'browserstack.networkLogs': true
+});
+
+exports.browserStackCapabilities = () => {
+    if (CIRCLE_CI) {
+        return exports.circleCICapabilities();
+    }
+    console.log(exports.localCapabilities());
```
siggerzz: This can be removed

wdio-browserstack.conf.js
```diff
@@ -0,0 +1,305 @@
...
+    exclude: [
+        // 'path/to/excluded/files'
+    ],
+
+    // Suites
+    suites: {
+        shared: [
+            './test/specs/component/shared/*.component.spec.js',
+        ],
+        desktop: [
```
siggerzz: This can be removed

packages/components/organisms/f-footer/test/specs/component/desktop/f-footer.component.spec.js
```diff
@@ -1,18 +1,13 @@
-const Footer = require('../../../test-utils/component-objects/f-footer.component');
-const footer = new Footer();
 const forEach = require('mocha-each');
+const Footer = require('../../../../test-utils/component-objects/f-footer.component');
+const footer = new Footer();
 
-describe('f-footer component tests', () => {
+describe('Shared - f-footer component tests', () => {
```
siggerzz: This should be `Desktop`. Happy to fix as part of another PR though

siggerzz (dismissed):

packages/components/organisms/f-footer/test/specs/component/desktop/f-footer.component.spec.js
JoshuaNg2332: I'll fix this now as it's a quick fix.

siggerzz (approved):

package.json
```diff
@@ -114,5 +117,8 @@
     "packages/components/organisms/*",
     "packages/services/*",
     "packages/tools/*"
-  ]
+  ],
+  "dependencies": {
+    "browserstack-local": "1.4.8"
```
dandel10n, line 122: Should this be in dev dependencies as this is in the root package.json of the monorepo?

JoshuaNg2332: This has been put into `dependencies` as this is to mirror what we have on Coreweb.
@siggerzz Do correct me if I'm wrong here.

JoshuaNg2332: @dandel10n So just checked. This needs to be in the `dependencies` as it needs to start at runtime.
@siggerzz was having problems with this in Coreweb when it was in the dev dependencies before.

dandel10n (approved):

PR #790, opened 2021-03-12
f-checkout@0.69.0 - Customer name jwt claims

charliearlie:
### Added
- Ability to extract customer name from JWT Claims if not retrieved from server

If the customer is authenticated but their name isn't retrieved, we can now retrieve it from the token for display. 

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/components/organisms/f-checkout/src/demo/customer-auth.json
```diff
@@ -0,0 +1,3 @@
+{
+    "token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJlbWFpbCI6ImpvZS5ibG9nZ3NAanVzdGVhdHRha2Vhd2F5LmNvbSIsImNyZWF0ZWRfZGF0ZSI6IjIwMjEtMDItMDhUMTA6Mjc6NDkuMTkzMDAwMFoiLCJuYW1lIjoiSm9lIEJsb2dncyIsImdsb2JhbF91c2VyX2lkIjoiVTdOUkFsV0FnNXpPZHNkUmdmN25rVHlvaTkwWEVvPSIsImdpdmVuX25hbWUiOiJKb2UiLCJmYW1pbHlfbmFtZSI6IkJsb2dncyIsImlhdCI6MTYxNTQ2OTUxNn0.VapH6uHnn4lHIkvN_mS9A9IVVWL0YPNE39gDDD-l7SU"
```
charliearlie: Not great for formatting as the line is too long but unfortunately a token with all the right information is that long!

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -458,6 +458,75 @@ describe('Checkout', () => {
                 expect(result).toBe(false);
             });
         });
+
+        describe('shouldLoadCustomerNameFromClaims ::', () => {
+            it('should return `true` for delivery order without customer name when user is logged in', () => {
```
SachaZvetelman:
```suggestion
            it('should return `true` if `isLoggedIn` is `true` and the customer name doesn\'t exist', () => {
```

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -458,6 +458,75 @@ describe('Checkout', () => {
...
+                // Arrange
+                const wrapper = shallowMount(VueCheckout, {
+                    store: createStore({
+                        ...defaultCheckoutState,
+                        isLoggedIn: true,
+                        serviceType: CHECKOUT_METHOD_DELIVERY,
```
SachaZvetelman: I'm quite sure we need to do this for collection order as well, btw.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -283,6 +283,10 @@ export default {
             (!this.address || !this.address.line1);
         },
 
+        shouldLoadCustomerNameFromClaims () {
+            return this.isLoggedIn && (!this.customer.firstName && !this.customer.lastName);
```
SachaZvetelman, line 287: I'm not sure if it could happen in Production, but what would happen if only the `firstName` is set but not the `lastName`, or vice-versa?

packages/components/organisms/f-checkout/src/demo/customer-auth.json
SachaZvetelman: Do we need this file? We normally have these JSON files to mock APIs.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
FayeCarter, line 469: do you need service type and address here?

FayeCarter: Do you need the service type and address here?

packages/components/organisms/f-checkout/package.json
```diff
@@ -45,6 +45,7 @@
     "@justeat/f-globalisation": "1.0.0",
     "@justeat/f-services": "1.7.0",
     "axios": "0.21.1",
+    "jwt-decode": "^3.1.2",
```
FayeCarter: nit - Could you remove the ^ please?

packages/components/organisms/f-checkout/src/components/Checkout.vue
charliearlie: A good point, but should we override what we get from the API in this case? So just change it to `||`

packages/components/organisms/f-checkout/src/demo/customer-auth.json
charliearlie: Yeah it was something I was wondering. I wasn't exactly sure where to keep something like this which is only for Storybook and testing purposes

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: Not sure. Maybe David M knows?

packages/components/organisms/f-checkout/src/demo/customer-auth.json
SachaZvetelman: Maybe in the `setup.js` file for the tests and then `Checkout.stories.js` for the story, I'd say.

simonsandersje (approved):

packages/components/organisms/f-checkout/src/components/_tests/helpers/setup.js
```diff
@@ -117,12 +118,22 @@ const $logger = {
     logError: jest.fn()
 };
 
+// eslint-disable-next-line
+const mockAuthToken = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.'
```
SachaZvetelman, line 122: Super ultra nit: this is not really a "mock", given we're not mocking any behaviour. Maybe we should rename this? Don't worry too much about it, though.

packages/components/organisms/f-checkout/stories/Checkout.stories.js
```diff
@@ -40,6 +40,14 @@ CheckoutMock.setupCheckoutMethod(placeOrderUrl);
 CheckoutMock.setupCheckoutMethod(getGeoLocationUrl);
 CheckoutMock.passThroughAny();
 
+// eslint-disable-next-line
+const mockAuthToken = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.'
```
SachaZvetelman, line 44: Same as above.

SachaZvetelman (approved):

PR #791, opened 2021-03-12
f-checkout@0.70.0 - Implementing the new version of the OPAPI.

SachaZvetelman:
v0.70.0
*March 15, 2021*

### Changed
- Calling OPAPI's v2.

### DIFF
--- a/packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
+++ b/packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
@@ -5,6 +5,7 @@ import basketDelivery from '../../demo/get-basket-delivery.json';
 import checkoutAvailableFulfilment from '../../demo/checkout-available-fulfilment.json';
 import customerAddresses from '../../demo/get-address.json';
 import geoLocationDetails from '../../demo/get-geo-location.json';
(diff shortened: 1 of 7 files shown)


packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -363,6 +367,43 @@ describe('CheckoutModule', () => {
             });
         });
 
+        describe('placeOrder ::', () => {
+            it('should post the order details to the backend.', async () => {
+                // Arrange
+                payload.url = 'http://localhost/opapi/placeorder';
```
DamianMullins, line 393: It looks like this object is mutated in a few places in this test module, do you think it's worth adding a line to a `beforeEach` to reset it for each test?

SachaZvetelman: I've updated this now to reset `payload` within the top `beforeEach`. Let me know if you're ok with that, @DamianMullins

FayeCarter (dismissed):

FayeCarter (approved):

kevinrodrigues (approved):

PR #795, opened 2021-03-15
f-skeleton-loader@v0.1.0 - Add base component structure

AndyMayJE:
---

Initial commit adding the base structure for f-skeleton-loader.

---

## UI Review Checks


kevinrodrigues (dismissed):
Looks like there's some specific implementation changes here? The first commit should ideally just be the generator changes.

siggerzz (dismissed):

siggerzz (dismissed):

dandel10n (dismissed):

packages/components/molecules/f-skeleton-loader/src/components/SkeletonLoader.vue
```diff
@@ -0,0 +1,75 @@
...
+                'div',
+                {
+                    staticClass: `${this.$style[`c-skeleton-loader__${text}`]} ${this.$style['c-skeleton-loader__bone']}`
+                },
+                children
+            );
+        }
+    },
+    render (createElement) {
+        return createElement(
```
DamianMullins: Any reason why you're not using the `template` to generate these elements?

packages/components/molecules/f-skeleton-loader/src/components/SkeletonLoader.vue
```diff
@@ -0,0 +1,75 @@
...
+<style lang="scss" module>
+
+$skeleton-loader-heading-height: 32px;
+$skeleton-loader-text-height: 16px;
+
+.c-skeleton-loader {
+  position: relative;
+  vertical-align: top;
+
+   &__bone {
```
DamianMullins: We should be following [our CSS naming conventions](https://fozzie.just-eat.com/documentation/css/css-naming) here, i.e. children are denoted by a single dash, and modifiers by a double-dash so, in this case, the class would be `.c-skeleton-loader-bone`.

We also advise against using the ampersand shorthand as it makes finding classes more difficult, for example, say you had the "c-skeleton-loader-bone" class on an element and wanted to find it in code you wouldn't find it by searching for "c-skeleton-loader-bone". There are [some docs which mention this here](https://fozzie.just-eat.com/documentation/css/sass).

Finally, using our naming convention, these classes no longer need to be nested within the `.c-skeleton-loader` class.

Hope that all makes sense, give me a shout if you have any questions 😺

packages/components/molecules/f-skeleton-loader/src/components/SkeletonLoader.vue
AndyMayJE: Yes the next PRs will extend the component to dynamically render groups of nested divs. Therefore its using createElement to do this programmatically.

packages/components/molecules/f-skeleton-loader/src/components/SkeletonLoader.vue
AndyMayJE: @DamianMullins thanks I have not written CSS in a long time so this is very useful.

packages/components/molecules/f-skeleton-loader/src/components/SkeletonLoader.vue
SachaZvetelman: I might be missing something here, but I still don't quite understand why just using `template` from the beginning is not an option, sorry! Do you mind trying to explain a bit more please, @AndyMayJE?

packages/components/molecules/f-skeleton-loader/src/components/SkeletonLoader.vue
DamianMullins: No worries, CSS looks perfect now 😺

packages/components/molecules/f-skeleton-loader/src/components/SkeletonLoader.vue
AndyMayJE: The component is never going to render anything other than dynamically generated nested divs based on the skeletonType prop. Expressing this in template logic could be tricky as nesting may be multiple levels deep. The template would be hard to understand and not really provide a visual representation of what may be rendered. 

A simple article will look like:

```
<div class="c-skeleton-loader-article c-skeleton-loader-bone">
    <div class="c-skeleton-loader-heading c-skeleton-loader-bone"></div>
    <div class="c-skeleton-loader-paragraph c-skeleton-loader-bone">
        <div class="c-skeleton-loader-text c-skeleton-loader-bone"></div>
        <div class="c-skeleton-loader-text c-skeleton-loader-bone"></div>
        <div class="c-skeleton-loader-text c-skeleton-loader-bone"></div>
    </div>
</div>
```

Hope that makes sense? Happy to change approach if you have alternative ideas?

SachaZvetelman: What I had in mind was that we'd have various `.vue` components within `f-skeleton-loader` and that we'd load the corresponding one based on the `prop`. 

For instance, `SkeletonLoader.vue` gets passed in `type="article"` and then it would render an `Article.vue` component what would have the template you described above.

But feel free to carry on with your approach and then we can review it as the component grows in complexity and requirements 😄

AndyMayJE: @SachaZvetelman ok that makes sense thanks. Certainly not 100% set on this approach it might be over complicating what we need. I am going to spend an hour or so investigating. I guess the disadvantage might be managing tiny child components. I did consider and look at using JSX in Vue to express the skeletons.

AndyMayJE: @SachaZvetelman @DamianMullins switched back to <template> in favour of a more simple approach which is all we need for now. Next PRs will introduce small skeleton components that will be loaded based on the skeleton prop.

SachaZvetelman (approved):

FayeCarter (approved):

PR #803, opened 2021-03-18
f-checkout@v0.73.0 - Add token checkout refresh specific changes.

kevinrodrigues:
### Changed
- `Checkout.vue` watcher `authToken` to only initialise checkout when the token updates as below:
  - Truthy > Falsey
  - Falsey > Truthy
  - Truthy > Truthy (Checkout will not reload)

### Added
- Tests to cover authToken watcher changes.

### Removed
- Some tests that are not required as they're now covered in these new changes.

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -315,11 +315,11 @@ export default {
          *
          * */
         async authToken (newTokenVal, oldTokenVal) {
+            this.setAuthToken(this.authToken);
+
             if ((!newTokenVal && oldTokenVal) || (!oldTokenVal && newTokenVal)) {
                 await this.initialise();
             }
-
-            this.setAuthToken(this.authToken);
         }
     },
 
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -304,8 +304,19 @@ export default {
     },
 
     watch: {
-        async authToken () {
-            await this.initialise();
+        /**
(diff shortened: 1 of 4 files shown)


charliearlie (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -304,8 +304,19 @@ export default {
...
+         * Truthy > Falsey
+         * Falsey > Truthy
+         *
+         * Do not reload checkout if the token changes from one token to another token
+         * as it will always be valid.
+         * */
+        async authToken (newTokenVal, oldTokenVal) {
+            if (!newTokenVal || (!oldTokenVal && newTokenVal)) {
+                await this.initialise();
+                this.setAuthToken(this.authToken);
```
pkedziora: This should be outside of the if statement, as we want the store to always have the latest token. This would matter if token changes from truthy to truthy.

kevinrodrigues: Yup my bad just realised that too 👍

pkedziora: also wouldn't it be better to write the condition as if ((!newTokenVal && oldTokenVal) || (!oldTokenVal && newTokenVal)) 
That would make sure we do not reload when token changes from false to false.

kevinrodrigues: Fixed.

pkedziora (dismissed):

pkedziora (approved):

FayeCarter (approved):

PR #814, opened 2021-03-22
fozzie-components@v3.25.0 Fix Fozzie docs link in README

adammorr:
Updates to Fozzie docs link in the readme

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "root",
-  "version": "3.24.0",
+  "version": "3.25.0",
   "private": true,
   "scripts": {
     "build": "cross-env-shell \"lerna run $LERNA_ARGS build --stream\"",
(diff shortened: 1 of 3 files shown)


DamianMullins (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "root",
-  "version": "3.24.0",
+  "version": "3.25.0",
```
DamianMullins, line 3: As this is a readme change I'd have said a #trivial PR or patch version would suffice here but happy to approve in any case.

DamianMullins (approved):

package.json
adammorr: @DamianMullins thanks for pointing that out, I didn't realise. Happy to change this, let me know!

DamianMullins: Probably best to stick with the conventions so updating it would be ideal 👍🏻

dandel10n: Hm, I think that updating readme is a #trivial change and should go without version bump as for end users there is no changes, right?

dandel10n:
Closed as the issue has been fixed

package.json
DamianMullins: Agree, remove the versioning and stick a #trivial prefic in the PR title @adammorr

PR #816, opened 2021-03-22
f-skeleton-loader@0.3.0 Restaurant carousel card

AndyMayJE:
### Added
- Skeleton components to build a single restaurant carousel card

image

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
+++ b/packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
@@ -8,7 +8,6 @@
 import ImageBlock from './Image.vue';
 
 export default {
-    name: 'CarouselCardBackground',
     components: { ImageBlock }
 };
 </script>
--- a/packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
+++ b/packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
@@ -1,6 +1,6 @@
 <template>
-    <div class="c-skeleton-loader-carouselCardBackground c-skeleton-loader-bone">
-        <image-block />
+    <div>
+        <image-block :class="$style['c-carouselCardBackground-image']" />
     </div>
 </template>
 
@@ -12,8 +12,10 @@ export default {
 };
 </script>
 
-<style lang="scss" scoped>
-.c-skeleton-loader-bone {
+<style lang="scss" module>
+
+.c-carouselCardBackground-image {
     height: 200px;
 }
+
 </style>
--- a/packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
+++ b/packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
@@ -1,0 +1,20 @@
+<template>
+    <div class="c-skeleton-loader-carouselCardBackground c-skeleton-loader-bone">
+        <image-block />
+    </div>
+</template>
+
+<script>
+import ImageBlock from './Image.vue';
+
+export default {
+    name: 'CarouselCardBackground',
+    components: { ImageBlock }
+};
+</script>
+
+<style lang="scss" scoped>
+.c-skeleton-loader-bone {
+    height: 200px;
+}
+</style>
--- a/packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardRestaurantData.vue
+++ b/packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardRestaurantData.vue
@@ -1,0 +1,39 @@
+<template>
+    <div class="c-skeleton-loader-restaurantData c-skeleton-loader-bone">
+        <sentence />
+        <rating />
+    </div>
+</template>
+
+<script>
+import Sentence from './Sentence.vue';
+import Rating from './Rating.vue';
+
+export default {
+    name: 'CarouselCardRestaurantData',
+    components: { Sentence, Rating }
+};
+</script>
+
+<style lang="scss" scoped>
+.c-skeleton-loader-restaurantData {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    align-items: center;
+    width: 96%;
+    margin: auto;
+    padding: spacing();
+
+    .c-skeleton-loader-sentence {
+        width: 25%;
+    }
+    /deep/ .c-skeleton-loader-sentence .c-skeleton-loader-text {
+        width: 70%;
+        height: spacing(x1.8);
+    }
+    /deep/ .c-skeleton-loader-sentence .c-skeleton-loader-text:nth-child(2) {
+        margin-bottom: 0;
+    }
+}
+</style>
--- a/packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardRestaurantData.vue
+++ b/packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardRestaurantData.vue
@@ -1,21 +1,25 @@
 <template>
-    <div class="c-skeleton-loader-restaurantData c-skeleton-loader-bone">
-        <sentence />
+    <div :class="$style['c-restaurantData']">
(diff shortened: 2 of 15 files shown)


packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
```diff
@@ -0,0 +1,20 @@
...
+    <div class="c-skeleton-loader-carouselCardBackground c-skeleton-loader-bone">
+        <image-block />
+    </div>
+</template>
+
+<script>
+import ImageBlock from './Image.vue';
+
+export default {
+    name: 'CarouselCardBackground',
```
DamianMullins: No need to set `name` in the components as vue-cli will do this automatically.

packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardRestaurantData.vue
```diff
@@ -0,0 +1,39 @@
...
+    justify-content: space-between;
+    align-items: center;
+    width: 96%;
+    margin: auto;
+    padding: spacing();
+
+    .c-skeleton-loader-sentence {
+        width: 25%;
+    }
+    /deep/ .c-skeleton-loader-sentence .c-skeleton-loader-text {
```
DamianMullins: NIT: Can you add some space between the rules so the file is easier to scan?

packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardRestaurantHeader.vue
```diff
@@ -0,0 +1,53 @@
...
+$restaurant-logo-width: 48px;
+
+.c-skeleton-loader-carouselCardRestaurantHeader {
+    text-align: center;
+    background-color: #fff;
+    margin: auto;
+    width: 96%;
+    padding-bottom: spacing(x2);
+}
+.c-skeleton-loader-image {
```
DamianMullins: NIT: Can you add some space between the rules so the file is easier to scan?

packages/components/molecules/f-skeleton-loader/src/components/skeletons/RestaurantCarouselCard.vue
```diff
@@ -0,0 +1,41 @@
...
+<style lang="scss" scoped>
+.c-skeleton-loader-favouriteCarouselCard {
+    max-width: 400px;
+    text-align: center;
+    margin: auto;
+
+    .c-skeleton-loader-carouselCardBackground {
+        margin-bottom: -(spacing(x8));
+    }
+    .c-skeleton-loader-carouselCardRestaurantHeader,
```
DamianMullins: NIT: Can you add some space between the rules so the file is easier to scan?

packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
```diff
@@ -0,0 +1,20 @@
...
+    components: { ImageBlock }
+};
+</script>
+
+<style lang="scss" scoped>
```
DamianMullins: Do you think it's worth using a mix of `scoped` and `module` in these components? i.e. `scoped` when there are styles targeting child elements (I assume that's what the `/deep/` selectors are doing?) and `module` when not?

AndyMayJE: Potentially with some investigation work that could be feasible. However, we don't necessarily know when a parent or new component might need to target and change a child. I think all children or bones will always need to be controlled by a parent. Mixing could make it harder to understand and manage. When I originally investigated combining with & without module, debugging was tricky and frustrating. 

Yes, the /deep/ allows child access more than one level deep.

packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
AndyMayJE: @DamianMullins thanks didn't know that.

packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
DamianMullins: Sure, makes sense - best to go with one and stick with it in the same context 👍🏻

DamianMullins (dismissed):

packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardRestaurantData.vue
```diff
@@ -0,0 +1,38 @@
...
```
1itvinka: same question as before: if we remove `.c-skeleton-loader-sentence` will it break anything? Less nesting - the better.

1itvinka: and it might be more readable and maintanable to use scss nesting here.
```
/deep/ .c-skeleton-loader-text {
    width: 70%;
    height: spacing(x1.8);

    &:nth-child(2) {
        margin-bottom: 0;
    }
}
```

AndyMayJE: Good spot thanks

packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardRestaurantHeader.vue
```diff
@@ -0,0 +1,59 @@
...
+    width: $restaurant-logo-width;
+    height: $restaurant-logo-width;
+    border: 1px solid white;
+    transform: translateY(-50%);
+}
+
+.c-skeleton-loader-heading {
+    margin: auto;
+    height: spacing(x3);
+    margin-bottom: spacing();
```
1itvinka: I would use a shorthand here, or at least kept the rules for the same property as close together as possible.

1itvinka (dismissed):

packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+    <div class="c-skeleton-loader-carouselCardBackground c-skeleton-loader-bone">
```
dandel10n: Can `c-skeleton-loader-bone` be moved to `SkeletonLoader.vue` and added there to `<component :is="skeletonType" />` like `<component :is="skeletonType" class="c-skeleton-loader-bone" />` instead of adding it to every child component?

AndyMayJE: Great point. I have been thinking about ways to improve this. Passing down or a mixin to generate the class that can be shared. I will address this in the next PRs if that's ok?

1itvinka (dismissed):

dandel10n:
I see that scoped works well here but I have some concerns with this approach (not scoped itself but how the classes defined and used at the moment). You add class names in one components than define styles for these classes in another component and it feels that later it will be hard to maintain. I am happy to catch up to see if we can make it easier. And I think we should look into this sooner than later while component is in its early stage.

packages/components/molecules/f-skeleton-loader/src/components/skeletons/CarouselCardBackground.vue
dandel10n: Another thing I've spotted is that for children components you use for example `c-skeleton-loader-carouselCardBackground` class name. As this class will be used specifically for this component the class name should be just `c-carouselCardBackground` instead

minrwhite:
Ash had some concerns with me using `scoped` rather than `module` on [a previous PR](https://github.com/justeat/fozzie-components/pull/236) where I'd also been after the use of the `/deep/` capability - might be worth reviewing his comments there, @AndyMayJE

AndyMayJE:
Re-working styles to use modules

PR #823, opened 2021-03-23
f-content-cards@v4.4.0 🗯 Adds StampCard promotion card

minrwhite:
### Updated
- Adds `StampCardPromotionCard` component, test, stories and docs
- Factors out some common card styling into a shared scss file
- Factors out some common non-visual behaviour into a shared component

### Fixed
- ContentCards stories when rendered as static site

#### Desktop

image

#### Tablet

image

#### Mobile

image

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/organisms/f-content-cards/src/components/cardTemplates/CardContainer.vue
+++ b/packages/components/organisms/f-content-cards/src/components/cardTemplates/CardContainer.vue
@@ -154,18 +154,15 @@ export default {
 </script>
 
 <style lang="scss" module>
+    @import '../../../src/assets/scss/card-styles';
 
     .c-content-card-body {
         flex-grow: 1;
     }
 
     .c-contentCard {
-        width: 100%;
-        text-decoration: initial;
-        text-align: center;
-        display: flex;
-        flex-direction: column;
-        flex-grow: 1;
+        @include card-container;
+
         margin-right: spacing(x2);
         margin-bottom: spacing(x2);
 
@@ -177,22 +174,8 @@ export default {
             margin-right: 0;
         }
 
-        &,
-        &:hover,
-        &:focus {
-            color: currentColor;
-        }
-
         .c-contentCards--wrap & {
-            display: flex;
-            flex-direction: column;
-            flex: 0 0 40%;
-            margin: 0 spacing() spacing(x3) 0;
-            width: 100%;
-
-            @include media('>=narrowMid') {
-                margin: 0 spacing() spacing(x3);
-            }
+            @include card-container-wrapped;
         }
 
         /**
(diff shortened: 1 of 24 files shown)


packages/components/organisms/f-content-cards/src/components/cardTemplates/CardContainer.vue
```diff
@@ -281,10 +244,10 @@ export default {
...
 
     .c-contentCard-thumbnail {
-        border: 1px solid $grey--lighter;
-        margin-top: - (32px + spacing(x2)); // This offsets the thumbnail above the top of the info card
-        width: 48px;
-        min-height: 48px;
+            border: 1px solid $grey--lighter;
+            margin-top: - (32px + spacing(x2)); // This offsets the thumbnail above the top of the info card
+            width: 48px;
+            min-height: 48px;
```
1itvinka: looks like this change was made by mistake?

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCardPromotionCard.vue
```diff
@@ -0,0 +1,192 @@
...
+
+    props: {
+        card: {
+            type: Object,
+            required: true
+        },
+
+        testId: {
+            type: String,
+            default: null
```
kevinrodrigues, line 72: is this null for a particular reason? Instead of just `''` which is falsey too.

kevinrodrigues (dismissed):

1itvinka (dismissed):

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCardPromotionCard.vue
minrwhite: To be compliant to the docs I was going off [this description](https://vuejs.org/v2/guide/syntax.html#Attributes):

> If isButtonDisabled has the value of `null`, `undefined`, or `false`, the disabled attribute will not even be included in the rendered <button> element.

Which mentions `false` but not necessarily falsey values.

kevinrodrigues: I think it's probably safe to use the empty string as strictly it counts as one of Javascripts false values. Happy to leave as is tho. (just seeing that I thought that there was another reason behind why it was used as we normally stick to the falsey values of the type)

kevinrodrigues (approved):

1itvinka (approved):

packages/components/organisms/f-content-cards/src/components/cardTemplates/StampCardPromotionCard.vue
1itvinka: I think that semantically `null` makes more sense here.

SachaZvetelman: I'm inclined in 99% of the cases to go for stricter comparisons (i.e. `true` and `false` rather than `truthy` and `falsey`). I think `null` makes more sense here.

PR #824, opened 2021-03-24
f-checkout@v0.77.1 - Take address from new location object on GET request

charliearlie:
### Changed
- Load `address` into state via new `location` object. 

Do not merge as awaiting backend to be merged/deployed

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


DamianMullins (dismissed):

packages/components/organisms/f-checkout/src/demo/checkout-delivery.json
```diff
@@ -12,15 +12,16 @@
...
-  "address": {
-    "lines": [
-      "1 Bristol Road",
-      "Flat 1",
-      "",
-      "Somerset"
-    ],
-    "locality": "Bristol",
-    "postalCode": "BS1 1AA"
+  "location": {
```
minrwhite, line 15: Should you be updating [this test](https://github.com/justeat/fozzie-components/blob/66b3b03b1ba63290953610215cf65e1d109f0793/packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js#L161-L170) (which refers to `address` instead of `location`) as part of the updates? It worries me that this test is still passing - which might imply it isn't testing what it is intended to be testing.

charliearlie: Everything we store in state works the same as before. I just extract the address from the location object before it reaches the Vuex store. So all component/module tests can remain the same.

charliearlie: I will update this test to be more clear with the new API schema though

charliearlie: Updated :)

minrwhite (dismissed):
That makes it clearer, thanks!

It would be good to update the tests in that file so that the `checkoutDelivery` in the global scope of the file is not mutated as part of the tests, potentially causing bleed from one test to another, but I'm happy for this do be done as part of a later PR.

minrwhite (approved):

DamianMullins (approved):

PR #826, opened 2021-03-24
f-checkout@v0.76.0 - Add new scheduled & asap values & structure to f…

kevinrodrigues:
Handles asap as a property in Checkout API calls. The time object also now follows a different structure according to the API (Screenshot below)

If the time is ASAP we now pick out the first available time and send that in the schedule object. (Previously blank)

### API (Required format):
Screen Shot 2021-03-24 at 2 35 28 PM

### Changes in this PR with asap set to false (When the user selects anything but asap):
Screen Shot 2021-03-24 at 2 28 00 PM

### Changes in this PR with asap set to true (When asap has been selected):
Screen Shot 2021-03-24 at 2 27 34 PM

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.76.0
```
minrwhite, line 16: Needs a double newline above this:
```suggestion

v0.76.0
```

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -469,9 +471,12 @@ export default {
                     address: this.address,
                     customer: this.customer,
                     isCheckoutMethodDelivery: this.isCheckoutMethodDelivery,
-                    time: this.time,
+                    time: this.hasAsapSelected
+                        ? this.availableFulfilment.times[0]
+                        : this.time,
```
SachaZvetelman: Not sure I understand this. Aren't we always sending `this.time`, and the only thing that should change is whether we send `asap: true` or `asap: false`?

minrwhite: Might be worth pushing this logic back into the store, because as the store is aware of the `availableFulfilment` values, it can update `time` as part of the commit when ASAP is set to true. 

Unless theres some use in retaining the previous value of `time`?

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -103,6 +105,7 @@ export default {
         selectionChanged (selectedFulfilmentTime) {
             this.selectedAvailableFulfilmentTime = selectedFulfilmentTime;
             this.updateChangedField('orderTime');
+            this.updateHasAsapSelected(false);
```
SachaZvetelman: Is this correct? What if the user changes from ASAP to a pre-order time, and then switch back to ASAP? Wouldn't `hasAsapSelected` remain as `false`?

packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js
```diff
@@ -198,10 +198,13 @@ describe('Selector', () => {
...
             let wrapper;
             let updateChangedFieldSpy;
+            let updateHasAsapSelectedSpy;
 
             beforeEach(() => {
                 updateChangedFieldSpy = jest.spyOn(Selector.methods, 'updateChangedField');
-
+                updateHasAsapSelectedSpy = jest.spyOn(Selector.methods, 'updateHasAsapSelected');
+    
+
```
SachaZvetelman: Nit: remove one empty white line, please :)

packages/components/organisms/f-checkout/src/components/Checkout.vue
kevinrodrigues: If asap is set to true then we need to send the time too, at the moment without this the time isn't sent. This will send the first time available in the `availableFulfilment` array.

SachaZvetelman: Yeah, that's what I mean, we need to send the `from` and `to` in every case. I'd expect these to be in the store and sent all the time. Happy to take it offline.

kevinrodrigues: Yup understand that part but that's the issue, the `from` & `to` aren't set in the store until the dropdown changes. So if we want that to always be set the `selection.vue` component needs updating so instead of watching for changes we either set the first time as a new value and read from that (similar to `this.availableFulfilment.times[0]`) or we set it this way as it's already been set in the store via `availableFulfilment` which contains the whole list of times (we only need the one).

Here's what is returned if we revert this change:

Screen Shot 2021-03-24 at 3 22 04 PM

SachaZvetelman: I see. As you said in your comment, I think the `from` and `to` should always be set, and that would make the code simpler. What do you think?

kevinrodrigues: Agree. I'll update the selection component so it updates correctly.

kevinrodrigues: Fixed.

kevinrodrigues: Fixed.

packages/components/organisms/f-checkout/src/components/Selector.vue
kevinrodrigues: Fixed.

packages/components/organisms/f-checkout/CHANGELOG.md
kevinrodrigues: Fixed.

packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js
kevinrodrigues: Fixed.

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -105,13 +109,44 @@ export default {
...
+         *
+         * @param selectedFulfilmentTime
+         */
+        setAsapFlag (selectedFulfilmentTime) {
+            if (this.availableFulfilment.isAsapAvailable
+                && this.availableFulfilment.times[0].from === selectedFulfilmentTime) {
+                this.updateHasAsapSelected(true);
+            } else {
+                this.updateHasAsapSelected(false);
+            }
```
minrwhite:
```suggestion
            this.updateHasAsapSelected(this.availableFulfilment.isAsapAvailable
                && this.availableFulfilment.times[0].from === selectedFulfilmentTime);
```

minrwhite (dismissed):

packages/components/organisms/f-checkout/src/components/Selector.vue
kevinrodrigues: Good suggestion 👍 fixed.

minrwhite (approved):

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -104,11 +110,41 @@ export default {
...
+
+        /**
+         * Set asap to `true` if the availableFulfilment time matches the selected fulfilment time,
+         * otherwise set it to false.
+         *
+         * @param selectedFulfilmentTime
+         */
+        setAsapFlag (selectedFulfilmentTime) {
+            const isAsapSelected = this.availableFulfilment.isAsapAvailable
+                && this.availableFulfilment.times[0].from === selectedFulfilmentTime;
```
1itvinka, line 145: do all intermediate properties in `this.availableFulfilment.times[0].from` always exist? For example can there be a situation when `this.availableFulfilment.times.length` is 0?

kevinrodrigues: This method would only be triggered when the user manually changes the time they want their delivery (via the delivery drop down). So `times` would exist. The `fulfilmentTimes` is what populated the drop down values. This gets pulled from the `availableFulfilment` value, if there's no times then the user won't be able to change anything.

1itvinka (approved):

PR #840, opened 2021-03-29
f-braze-adapter@v3.3.0 👩‍🔧 Fixes TypeError occurrences

minrwhite:
### Fixed
- Handled a condition where, if appboy was not initialised when `BrazeDispatcher.configure()` was called,
  and an in-app message was received, an error would be thrown.

---

### DIFF
--- a/packages/services/f-braze-adapter/src/BrazeDispatcher.js
+++ b/packages/services/f-braze-adapter/src/BrazeDispatcher.js
@@ -20,22 +20,29 @@ function interceptInAppMessageClickEventsHandler (message) {
  * click events of the CTA button
  * @param message
  * @this BrazeDispatcher
+ * @return Promise
  */
 function interceptInAppMessagesHandler (message) {
-    if (message instanceof this.appboy.ab.InAppMessage) {
-        /**
-         * Always subscribe click action to second button
-         * as this is always "success" as opposed to "dismiss"
-         * as confirmed with CRM (AS)
-         */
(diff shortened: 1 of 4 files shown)


kevinrodrigues (approved):

packages/services/f-braze-adapter/src/BrazeDispatcher.js
```diff
@@ -20,22 +20,29 @@ function interceptInAppMessageClickEventsHandler (message) {
...
+                if (message.buttons && message.buttons.length >= 2) {
+                    const [, button] = message.buttons;
+                    // Note that the below subscription returns an ID that could later be used to unsubscribe
+                    button.subscribeToClickedEvent(() => interceptInAppMessageClickEventsHandler.bind(this)(message));
+                }
+            }
+            appboy.display.showInAppMessage(message);
+        })
+        .catch(error => {
+            this.logger('error', `Error handling message - ${error}`, { message, error });
```
DamianMullins, line 44: Where does `this.logger` come from? Is it something we need to configure in consuming apps?

minrwhite: From [here](https://github.com/justeat/fozzie-components/blob/e192a3516766f5620c3d79fe051339cb57938c03/packages/services/f-braze-adapter/src/BrazeDispatcher.js#L136) - I'll message you regarding internal resources for this.

DamianMullins (approved):

PR #841, opened 2021-03-29
f-checkout@v0.78.0 - Refactor error handling

SachaZvetelman:
_Note: this PR is huge. We're going to do a walkthrough before reviews, so please, if you're brave enough to read this PR, keep that in mind 🪂_ 

v0.78.0
*March 30, 2021*

### Added
- Custom errors.

### Changed
- The whole handling of exceptions and errors.
- Several functions and tests.

### Removed
- Redundant and unnecessary functions.


packages/components/organisms/f-checkout/README.md
```diff
@@ -100,6 +100,11 @@ The props that can be defined are as follows:
 | `checkout-setup-guest-success` | Emitted when guest user is created successfully. |
 | `checkout-setup-guest-failure` | Emitted when guest user is not created successfully. |
 | `checkout-validation-error` | Emitted validation error occurs. |
+| `checkout-place-order-success` | Emitted when order is successfully placed. |
```
dandel10n, line 103: Not related to your changes but something to check for the next task - not all the events are here in the readme according to `f-checkout/src/event-names.js` so the list here needs to be checked and updated

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -122,11 +122,14 @@ import ErrorDialog from './ErrorDialog.vue';
...
 
 import {
-    ANALYTICS_ERROR_CODE_BASKET_NOT_ORDERABLE,
+    CreateGuestUserError,
+    UpdateCheckoutError,
+    PlaceOrderError
+} from '../exceptions/exceptions';
+
+import {
     ANALYTICS_ERROR_CODE_INVALID_MODEL_STATE,
```
dandel10n, line 131: I think this constant can be removed if `onInvalidCheckoutForm()` function is refactored as well, if I understood correctly. Would be good to add it to the next task not to miss it.

dandel10n (approved):
All looks good, left few comments which can be covered in another task. I got a bit lost in the unit tests to be honest, maybe later we can refactor them to be in a more consistent style

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: I think we still need this one.
I've added the other comment to the refactoring ticket, though.

packages/components/organisms/f-checkout/README.md
SachaZvetelman: Added this to the refactoring ticket I just created.

1itvinka (approved):

PR #843, opened 2021-03-30
f-footer@v4.15.0 - Fix computed class

FayeCarter:
## Fixed
- Confianza logo styles.

## UI Review Checks

image

### DIFF
--- a/packages/components/organisms/f-footer/src/components/FeedbackBlock.vue
+++ b/packages/components/organisms/f-footer/src/components/FeedbackBlock.vue
@@ -1,7 +1,10 @@
 // .is-invisible class on data-gtm-feedback element is dictated by gtm script for Usabilla
 <template>
     <div
-        :class="[$style['c-feedback'], 'is-invisible']"
+        :class="[
+            $style['c-feedback'],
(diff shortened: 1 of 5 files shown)


1itvinka (approved):

AlixRichards (approved):

packages/components/organisms/f-footer/src/components/FeedbackBlock.vue
```diff
@@ -1,7 +1,10 @@
 // .is-invisible class on data-gtm-feedback element is dictated by gtm script for Usabilla
 <template>
     <div
-        :class="[$style['c-feedback'], 'is-invisible']"
+        :class="[
+            $style['c-feedback'],
+            $style['is-invisible']
```
DamianMullins, line 6: [`is-invisible` is a class from Fozzie](https://github.com/justeat/fozzie/blob/99280c21e6c9c8baa54604ae2c670fd7583db4c8/src/scss/trumps/_utilities.scss#L81) and so shouldn't be reference using the object style otherwise it'll have no effect.

PR #844, opened 2021-03-30
f-form-field@1.9.0 - Pass `value` down to FormDropdown component

xander-marjoram:
This makes it possible to update the selected value in FormDropdown by changing the `value` passed into FormField.
When this prop is not passed in then the first value is selected by default, which is the existing behaviour.

There is no readme update because the props of FormField itself have not changed, only FormDropdown.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [x] Internet Explorer 11

### DIFF
--- a/packages/components/atoms/f-form-field/src/components/FormField.vue
+++ b/packages/components/atoms/f-form-field/src/components/FormField.vue
@@ -24,6 +24,7 @@
                 :id="`${uniqueId}`"
                 :attributes="$attrs"
                 :type="normalisedInputType"
+                :value="value"
                 :class="[
                     $style['c-formField-input'],
                     $style['c-formField-dropdownContainer'],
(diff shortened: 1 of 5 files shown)


1itvinka (approved):

dandel10n (approved):

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -24,6 +24,7 @@
                 :id="`${uniqueId}`"
                 :attributes="$attrs"
                 :type="normalisedInputType"
+                :value="value"
```
ashleynolan, line 27: Could this just be set through the `$attrs` object (so no need for the prop)?

xander-marjoram: Potentially, but personally I find that massively reduces readability

PR #845, opened 2021-03-30
f-checkout@v0.79.0

kevinrodrigues:
### Fixed
- Retry modal now closes when errors / issues are returned and the user tries to dismiss the modal.

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
+++ b/packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
@@ -3,7 +3,8 @@
         :is-open="isOpen"
         data-test-id="checkout-issue-modal"
         has-overlay
-        @close="handleClose">
+        v-bind="$attrs"
+        v-on="$listeners">
         <h3
             data-test-id="checkout-issue-modal-title"
             class="u-noSpacing">
@@ -41,21 +42,17 @@ export default {
     props: {
         isOpen: {
             type: Boolean,
-            required: true
+            default: false
         },
         errorCode: {
             type: String,
-            required: true
+            default: ''
         }
     },
 
     methods: {
         handleButtonClick () {
             this.$emit(EventNames.CheckoutErrorDialogButtonClicked);
-        },
-
-        handleClose () {
-            this.$emit(EventNames.CheckoutDialogCloseButtonClicked);
         }
     }
 };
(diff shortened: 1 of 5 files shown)


simonsandersje (dismissed):

simonsandersje (dismissed):

DamianMullins (dismissed):

SachaZvetelman:
You'll need to update the version 🔢

kevinrodrigues:
> You'll need to update the version 🔢

Version is correct I believe (package.json) I've updated the title tho if that's what you mean?

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
```diff
@@ -3,7 +3,8 @@
         :is-open="isOpen"
         data-test-id="checkout-issue-modal"
         has-overlay
-        @close="handleClose">
+        v-bind="$attrs"
+        v-on="$listeners">
```
charliearlie, line 7: Nice. I guess this is where me being new to Vue got me. 🙈

kevinrodrigues: So it was actually the `v-if` that was breaking it mostly. The mega modal wasn't picking up the `hasOpened` value as that `if` [here](https://github.com/justeat/fozzie-components/pull/845/files#diff-3f7bda367e6ae2f5c141559c1aec5722d6606ee4d895f7d44e9837c20f7efb3cL4) stopped it being populated so the close event would never execute properly :)

packages/components/organisms/f-checkout/test/specs/component/f-checkout.component.spec.js
```diff
@@ -157,7 +157,7 @@ describe('f-checkout component tests', () => {
         expect(checkout.isCheckoutErrorMessageDisplayed()).toBe(true);
     });
 
-    it('should close the checkout error when "Retry" is clicked', () => {
+    xit('should close the checkout error when "Retry" is clicked', () => {
```
kevinrodrigues, line 160: Skipped as there's a problem with the way the form submission works, causing the modal to possibly open twice. I've spoke to @Festivejelly about it and he's improving the way this works so should resolve this soon.

simonsandersje (approved):

PR #835, opened 2021-03-25
f-checkout@0.82.0 - Add generic error handling to checkout

charliearlie:
### Added
- Ability to handle issues from API that client has no knowledge of
- Generic error dialog for these issues

- [x] README and/or UI Documentation has been [created|updated]
- [x] Unit tests have been [created|updated]
- [ ] This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist)

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11
- [x] Mobile (Please list device/browser – Ideally one iPhone model and one Android)

### List any other browsers that this PR has been tested in:

- [View the Browser Support Checklist](http://fozzie.just-eat.com/documentation/general/browser-support)

PR #846, opened 2021-03-30
f-checkout@v0.83.0 - Fix top margin for the error messages

1itvinka:
This PR fixes the top margin for the error message on the checkout.

[This style](https://github.com/justeat/fozzie-components/blob/19d13e7f8cc07911e6b364cfefd354b7d66debf5/packages/components/molecules/f-alert/src/components/Alert.vue#L118) was overwritten by `margin: 0 auto`. To fix the `margin` I removed the shorthand.

**Before:** 
image-20210330-072604

**After:**
Screenshot 2021-03-31 at 11 44 39

### DIFF
--- a/packages/components/organisms/f-checkout/CHANGELOG.md
+++ b/packages/components/organisms/f-checkout/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.83.0
+------------------------------
+*April 6, 2021*
+
+### Fixed
+- Error alert top margin.
+
+
 v0.82.0
 ------------------------------
 *April 1, 2021*
--- a/packages/components/organisms/f-checkout/package.json
+++ b/packages/components/organisms/f-checkout/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-checkout",
   "description": "Fozzie Checkout – Fozzie Checkout Component",
-  "version": "0.82.0",
+  "version": "0.83.0",
   "main": "dist/f-checkout.umd.min.js",
   "files": [
     "dist",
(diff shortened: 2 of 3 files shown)


SachaZvetelman:
@1itvinka, could you please upload a screenshot with the fix? 🙂

dandel10n:
Feels like the alert component itself shouldn't have margin and it should be added here instead as in different places spacing around alerts can be different

1itvinka:
@dandel10n great point! So we don't have the standard spacing around alerts then?

1itvinka:
@dandel10n I think in this case (because the alert component has the default spacing and it's the same for this alert), the fix does the trick. We might need to create a separate ticket to refactor the alert component: remove default spacing, move it to a variable, and fix it in all the places where we use the alerts. Does it make sense?

1itvinka:
@SachaZvetelman I've updated the PR description with the screenshot with a fix

dandel10n:
> @dandel10n I think in this case (because the alert component has the default spacing and it's the same for this alert), the fix does the trick. We might need to create a separate ticket to refactor the alert component: remove default spacing, move it to a variable, and fix it in all the places where we use the alerts. Does it make sense?

I agree, we can remove margins from alerts at all as components should contain only their own styles not the styles for surrounding layout, if it make sense.

packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.80.0
```
1itvinka: the version is v0.80.0, not v0.79.0, so that there's no conflict with this or https://github.com/justeat/fozzie-components/pull/835

DamianMullins (approved):

dandel10n (approved):

PR #787, opened 2021-03-10
f-registration@v0.50.0 - Add a11y markup to f-registration

BanksySan:
Add screen reader accessibility markup to f-registration.

## UI Review Checks

- [ ] README and/or UI Documentation has been [created|updated]
- [ ] Unit tests have been [created|updated]
- [ ] This code has been checked with regard to [our accessibility standards](http://fozzie.just-eat.com/documentation/general/accessibility/checklist)

## Browsers Tested

- [ ] Chrome (latest)
- [ ] Internet Explorer 11
- [ ] Mobile (Please list device/browser – Ideally one iPhone model and one Android)

### List any other browsers that this PR has been tested in:

- [View the Browser Support Checklist](http://fozzie.just-eat.com/documentation/general/browser-support)

PR #858, opened 2021-04-01
f-registration@v0.50.0 - Add a11y markup to f-registration

minrwhite:
Add screen reader accessibility markup to f-registration.

Reopened from #787 due to bad branch name

---

## UI Review Checks

## Browsers Tested

- [ ] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/organisms/f-registration/src/components/Registration.vue
+++ b/packages/components/organisms/f-registration/src/components/Registration.vue
@@ -30,108 +30,108 @@
                 @click="formStart"
                 @focus="formStart"
                 @submit.prevent="onFormSubmit">
-                <error-message
-                    v-if="genericErrorMessage"
-                    :class="$style['c-registration-genericError']">
-                    {{ genericErrorMessage }}
-                </error-message>
+                <section
+                    id="error-summary-container"
+                    :class="$style['is-visuallyHidden']"
+                    role="alert"
+                    data-test-id="error-summary-container">
+                    <p
+                        v-if="genericErrorMessage"
+                        :class="$style['o-form-error']">
+                        <warning-icon :class="$style['o-form-error-icon']" />
+                        {{ genericErrorMessage }}
+                    </p>
+                </section>
+
                 <form-field
+                    ref="firstName"
                     v-model="firstName"
+                    aria-required="true"
+                    aria-describedby="error-message-firstname"
+                    :aria-invalid="!!describeFirstnameErrorMessage"
                     name="firstName"
                     :label-text="copy.labels.firstName"
                     input-type="text"
                     @blur="formFieldBlur('firstName')">
-                    <template #error>
-                        <error-message
-                            v-if="shouldShowFirstNameRequiredError"
-                            data-test-id='error-first-name-empty'>
-                            {{ copy.validationMessages.firstName.requiredError }}
-                        </error-message>
-                        <error-message
-                            v-if="shouldShowFirstNameMaxLengthError"
-                            data-test-id='error-first-name-maxlength'>
-                            {{ copy.validationMessages.firstName.maxLengthError }}
-                        </error-message>
-                        <error-message
-                            v-if="shouldShowFirstNameInvalidCharError"
-                            data-test-id='error-first-name-invalid'>
-                            {{ copy.validationMessages.firstName.invalidCharError }}
-                        </error-message>
+                    <template
+                        v-if="describeFirstnameErrorMessage"
+                        #error>
+                        <p
+                            id="error-message-firstname"
+                            :class="$style['o-form-error']"
+                            data-test-id='error-message-firstname'>
+                            <warning-icon :class="$style['o-form-error-icon']" />
+                            {{ describeFirstnameErrorMessage }}
+                        </p>
                     </template>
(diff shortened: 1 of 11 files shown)


packages/components/organisms/f-registration/test-utils/component-objects/f-registration.selectors.js
```diff
@@ -1,24 +1,28 @@
...
 exports.COOKIES_POLICY_LINK = '[data-test-id="cookies-policy-link"]';
 exports.REQUEST_ERROR = '[data-test-id="error-message-component"]';
 
 exports.FIRST_NAME_INPUT = '[data-test-id="formfield-firstName-input"]';
-exports.FIRST_NAME_EMPTY_ERROR = '[data-test-id="error-first-name-empty"]';
-exports.FIRST_NAME_MAX_LENGTH_ERROR = '[data-test-id="error-first-name-maxlength"]';
-exports.FIRST_NAME_INVALID_ERROR = '[data-test-id="error-first-name-invalid"]';
+// exports.FIRST_NAME_EMPTY_ERROR = '[data-test-id="error-first-name-empty"]';
+// exports.FIRST_NAME_MAX_LENGTH_ERROR = '[data-test-id="error-first-name-maxlength"]';
+// exports.FIRST_NAME_INVALID_ERROR = '[data-test-id="error-first-name-invalid"]';
```
1itvinka: can we remove the commented out code?

LTurns (approved):

packages/components/organisms/f-registration/src/components/Registration.vue
```diff
@@ -282,6 +273,60 @@ export default {
...
+        describeFirstnameErrorMessage () {
+            if (this.$v.firstName.$dirty) {
+                if (!this.$v.firstName.required) {
+                    return 'Please include your first name';
+                }
+                if (!this.$v.firstName.maxLength) {
+                    return 'First name exceeds 50 characters';
+                }
+                if (!this.$v.firstName.meetsCharacterValidationRules) {
+                    return 'Your name can only contain letters, hyphens or apostrophes';
```
1itvinka, line 285: I would be careful with such rules. Is it documented somewhere?

packages/components/organisms/f-registration/test-utils/component-objects/f-registration.component.js
```diff
@@ -1,74 +1,85 @@
...
+            },
+            get input () { return $(EMAIL_INPUT); },
+            get errorMessage () { return $(EMAIL_ERROR_MESSAGE).innerText; }
+        },
         password: {
-            get input () { return $(PASSWORD_INPUT) }, 
-            get emptyError () { return $(PASSWORD_EMPTY_ERROR) }
+            errorMessages: {
+                minLength: 'Password is less than four characters',
+                missing: 'Please enter a password'
```
1itvinka, line 60: It concerns me that we're duplicating text. It might make sense to store it somewhere a variable, together with all the other error messages? Also what about internationalisation?

SachaZvetelman:
Do we have translation files in `f-registration`? If yes, we should move all the copies to the translation files, since now they are hardcoded to English.

packages/components/organisms/f-registration/src/components/Registration.vue
BanksySan: For these the code is the documentation, I actually think this is correct as they're not complicated and they're described by the tests.

packages/components/organisms/f-registration/test-utils/component-objects/f-registration.component.js
BanksySan: @1itvinka, @SachaZvetelman, I agree with you both that these need to be in translation files, they weren't before this though and I didn't want to increase the scope to do that.  Are you OK if I create a new ticket to cover it?

packages/components/organisms/f-registration/src/components/Registration.vue
1itvinka: I mean there're probably names out there that contain numbers, etc, for example, X Æ A-12 😂.

1itvinka (approved):

packages/components/organisms/f-registration/src/components/Registration.vue
```diff
@@ -282,6 +273,60 @@ export default {
          * focus on the input field.
          */
 
+        describeFirstnameErrorMessage () {
+            if (this.$v.firstName.$dirty) {
+                if (!this.$v.firstName.required) {
+                    return 'Please include your first name';
```
ashleynolan, line 279: Should these come from a translation file? (or is that planned to move these out later)

packages/components/organisms/f-registration/src/components/Registration.vue
```diff
@@ -428,8 +473,26 @@ export default {
         },
 
         isFormInvalid () {
-            this.$v.$touch();
-            return this.$v.$invalid;
+            const v = this.$v;
```
ashleynolan, line 476: Could you add a JSDoc comment to this, as the function has changed a fair bit from its initial creation.

PR #862, opened 2021-04-06
f-checkout@0.84.0 - Automatic scroll to server-side errors

SachaZvetelman:
v0.84.0
*April 6, 2021*

### Added
- Scrolling to the server-side error alert automatically.


tommcclean (dismissed):

1itvinka (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -724,6 +726,25 @@ export default {
...
+        */
+        scrollToElement (refElement) {
+            const scrollingDurationInMilliseconds = 650;
+
+            const element = this.$refs[refElement]
+                ? this.$refs[refElement].$el
+                : null;
+
+            if (element) {
+                VueScrollTo.scrollTo(element, scrollingDurationInMilliseconds, { offset: -20 });
```
DamianMullins, line 746: You could avoid importing `vue-scrollto` and use `this.$scrollTo` here.

DamianMullins (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: This didn't work for me. We have a bit of a weird way of including things in our components, in my opinion, given we don't really use `Vue.use(...)` anywhere. I had tried adding it to `index.js` but didn't do anything, so I had to import it directly into my component and use `VueScrollTo.xxxxxxxx`.

Any opinions? cc @tommcclean, @DamianMullins.

DamianMullins: To be honest I based the original comment on the fact you were using `Vue.use` in the component. If you're saying that had no effect/didn't work then the import is fine by me (it's included as an example in the `vue-scrollto` readme so it seems perfectly acceptable).

PR #863, opened 2021-04-07
f-status-banner - component tests for error messages #trivial

LTurns:
*April 6, 2021*

### Added
- Component tests for `MainBannerContainer.vue` and `PageBanner.vue`
- `f-statusBanner.selector.js` file for storing test elements
- Disabled `no` and `dk` urls in cookie-banner component tests


siggerzz (approved):

DamianMullins (approved):

DamianMullins:
Looks good to me. Quick question; what is the reasoning behind the tests being nested inside the `/test/specs/` directory? Typically you'd see either `/test` or `/spec` on their own?

LTurns:
> Looks good to me. Quick question; what is the reasoning behind the tests being nested inside the `/test/specs/` directory? Typically you'd see either `/test` or `/spec` on their own?

Thank youu! I'm not sure why we have both actually - @siggerzz do you know why this extra folder was added? I wonder if it could be removed 🤔

siggerzz:
@DamianMullins @LTurns Yeah we can address this at some point as there wasn't any reasoning behind it. Should probably do it as part of a separate PR though so we can do it in all components 👍🏼

PR #870, opened 2021-04-08
f-checkout@0.85.0 Exchange OTAC to JWT

pkedziora:
`otacToAuthExchanger` prop that provides functionality to exchange OTAC to JWT token for guest checkout. The function will be implemented outside of f-checkout and passed as prop as it's not f-checkout responsibility to retrieve authentication details.

When OTAC is exchanged to JWT, the isLoggedIn flag stays `false` to indicate that we are still in 'guest' mode, while authToken gets set to correct value.

---

## UI Review Checks
No UI changes

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -215,6 +215,11 @@ export default {
             default: ''
         },
 
+        otacToAuthExchanger: {
+            type: Function,
+            default: () => ''
```
tommcclean: Suggestion: Could log a not implemented message

tommcclean:
Remember to update the props list in the readme

packages/components/organisms/f-checkout/src/components/Checkout.vue
pkedziora: Logger would not be available here as per: https://vuejs.org/v2/guide/components-props.html
"Note that props are validated before a component instance is created, so instance properties (e.g. data, computed, etc) will not be available inside default or validator functions."

pkedziora:
@tommcclean Readme updated

packages/components/organisms/f-checkout/src/components/Checkout.vue
pkedziora: @tommcclean I replaced it with default implementation that throws not implemented exception.

tommcclean (approved):

simonsandersje (approved):

PR #872, opened 2021-04-08
fozzie-components@v3.27.0 - Adding test tagging mechanism #globalconfig 

siggerzz:


LTurns (dismissed):

packages/tools/generator-component/generators/app/templates/__package__.json
```diff
@@ -33,9 +33,9 @@
     "lint:fix": "yarn lint --fix"<% if(config.isComponent) { %>,
     "lint:style": "vue-cli-service lint:style"<% } %>,
```
LTurns, line 34: is there anything else we need to change in the generator? Is it worth adding the `@browserstack` tag?

siggerzz: I did think this, but would require changing more files. Happy to do this in another PR once we've got a consensus on what tests should be run against BS

ashleynolan:
Looks like this PR should use the [globalconfig](https://justeat.github.io/fozzie-components/@justeat/storybook/index.html?path=/story/documentation-getting-started-contributing--page#globalconfig) flag rather than the trivial flag (so that it runs danger checks on the root of the mon-repo then, but not packages).

siggerzz:
> Looks like this PR should use the [globalconfig](https://justeat.github.io/fozzie-components/@justeat/storybook/index.html?path=/story/documentation-getting-started-contributing--page#globalconfig) flag rather than the trivial flag (so that it runs danger checks on the root of the mon-repo then, but not packages).

My bad! I always forget 🤦🏼

oliversweb (dismissed):

oliversweb (approved):

LTurns (approved):

PR #874, opened 2021-04-12
f-checkout@0.88.0 - Send null for address lines 3 and 4 in PATCH request

charliearlie:
### Changed
- PATCH Checkout now sends null for lines 3 and 4 of address instead of empty string


packages/components/organisms/f-checkout/src/services/mapper.js
```diff
@@ -39,8 +39,8 @@ const mapUpdateCheckoutRequest = ({
                         lines: [
                             address.line1 || '',
                             address.line2 || '',
-                            '',
-                            ''
+                            null,
```
kevinrodrigues, line 42: Should this be checking lines 3 & 4 then setting it to null if it's not populated instead?

charliearlie: I believe we always sent an empty string and now always send null. Because we don't actually capture address lines 3 and 4. City and postcode are sent separately in `locality` and `postcode`.

kevinrodrigues (dismissed):

SachaZvetelman:
Happy to approve, but I think this should be a minor version bump.

packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.88.0
+------------------------------
+*April 12, 2021*
+
+### Changed
+- PATCH Checkout now sends null for lines 3 and 4 of address instead of empty string
+
```
SachaZvetelman, line 14: Nit: we normally add one extra empty line here.

SachaZvetelman (dismissed):

SachaZvetelman (dismissed):

kevinrodrigues (dismissed):

PR #878, opened 2021-04-12
f-header@4.11.2 - Fixed typo in Italian header

1itvinka:
### Fixed
- Fixed typo in Italian header

### DIFF
--- a/packages/components/organisms/f-header/CHANGELOG.md
+++ b/packages/components/organisms/f-header/CHANGELOG.md
@@ -3,14 +3,18 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-
-Latest (add to next release)
+v4.11.2
 ------------------------------
+*March 16, 2021*
+
+### Fixed
+- Fixed typo in Italian header
+
 *April 9, 2021*
 
 ### Changed
 - Refactored large country selector test into two separate tests
-- - Added string placeholder to tenant component tests
+- Added string placeholder to tenant component tests
 
 *March 31, 2021*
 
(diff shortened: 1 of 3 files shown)


packages/components/organisms/f-header/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-header",
   "description": "Fozzie Header – Globalised Header Component",
-  "version": "4.11.1",
+  "version": "4.12.0",
```
1itvinka: is it a correct version upd for a typo fix? or should it be `4.11.2` instead?

packages/components/organisms/f-header/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
...
 
+v4.12.0
+------------------------------
+*March 16, 2021*
+
+### Fixed
+- Fixed typo in Italian header
+
 
 Latest (add to next release)
```
1itvinka: is it okay to add my changelog comment above this one? Or should I edit the previous one ("Latest (add to next release)") somehow?

kevinrodrigues: You should combine the changelog from here with your versioned changes.

packages/components/organisms/f-header/src/tenants/it-IT.js
```diff
@@ -18,7 +18,7 @@ export default {
             gtm: 'click_account_your_orders'
         },
         accountCredit: {
-            text: 'Credito dell',
+            text: "Credito dell'account",
```
kevinrodrigues, line 21: You should be able to escape the `'` with a backslash so you can keep the single quotes.

packages/components/organisms/f-header/package.json
kevinrodrigues: Think `4.11.2` would make sense here.

packages/components/organisms/f-header/src/tenants/it-IT.js
1itvinka: Do you have any policies or guidelines on this? In f-footer we use the format I used here. And as for me, it's more readable anyway.

JoshuaNg2332 (approved):

kevinrodrigues (approved):

PR #879, opened 2021-04-13
f-checkout@0.89.0 - Automatic scrolling to inline errors.

SachaZvetelman:
v0.89.0
*April 13, 2021*

### Added
- Automatic scrolling to the first inline error when trying to proceed to the payment page.


simonsandersje (dismissed):

packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -43,6 +44,7 @@
             <template #error>
                 <error-message
                     v-if="isAddressLocalityEmpty"
+                    data-error-message
```
kevinrodrigues: We should probably hook the `js` name to this name, just to make it clear as to the reason it's here (a hook for javascript functionality)

`data-js-error-message`

SachaZvetelman: I didn't know we had that naming convention, but sounds good to me. Will amend the code 👍

kevinrodrigues (dismissed):

1itvinka (dismissed):

JoshuaNg2332 (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -735,22 +736,18 @@ export default {
...
-                this.scrollToElement('errorAlert');
+                this.scrollToElement(this.$refs.errorAlert.$el);
             });
         },
 
         /**
          * Scroll to a ref element in the screen.
         */
-        scrollToElement (refElement) {
+        scrollToElement (element, options = { offset: -20 }) {
```
ashleynolan, line 746: Small thing – would be good to update the comment for this function to describe the function params. It's not obvious why the default of offset would be -20 for instance, whereas a comment would describe that.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -798,6 +797,17 @@ export default {
...
         },
 
+        /**
+         * Scroll to the first inline error, no matter which one it is.
+         */
+        scrollToFirstInlineError () {
+            this.$nextTick(() => {
+                const firstInlineError = document.querySelector('[data-js-error-message]');
+
+                this.scrollToElement(firstInlineError, { offset: -100 });
```
ashleynolan, line 807: Probably would be cleared up by adding a comment to the `scrollToElement` function, but it's not obvious what -100 refers to (is it -100 pixels? Why 100 specifically – is that the height of an error element?)

PR #880, opened 2021-04-13
f-checkout@0.91.0 - Logger alignment changes

oliversweb:
v0.91.0
*April 14, 2021*

### Added
- Refactored logging calls
- Refactored error handling
- Updated/Aligned the `Event` info in the Readme


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -711,28 +661,25 @@ export default {
...
          */
-        handleErrorState ({
-            error, messageToDisplay, eventToEmit, logMessage
-        }) {
-            const eventData = {
-                ...this.eventData,
-                error
-            };
+        handleErrorState (error) {
+            const message = this.$t(error.messageKey || 'errorMessages.genericServerError');
```
SachaZvetelman: If the `messageKey` has a value but there is no translation for it, this line will return an empty string, I believe.

packages/components/organisms/f-checkout/src/exceptions/exceptions.js
```diff
@@ -1,26 +1,34 @@
 /* eslint-disable max-classes-per-file */
+import EventNames from '../event-names';
+
 class CreateGuestUserError extends Error {
     constructor (message) {
         super(message);
-        this.message = message;
+        this.messageKey = 'errorMessages.guestUserCreationFailure';
```
SachaZvetelman, line 7: Rather than the key, we could be assigning the message directly. What do you think?

packages/components/organisms/f-checkout/src/mixins/logger.mixin.js
```diff
@@ -0,0 +1,18 @@
...
+            exception: error.name,
+            exceptionMessage: error.message,
+            exceptionStackTrace: error.stack
+        })
+    };
+}
+
+export default {
+    methods: {
+        logInvoker (message, data, callback, error) {
```
SachaZvetelman: I think as we discussed (I might be wrong), I'd make the parameter an object instead: `logInvoker ({ message, data, callback, error }) { ... }`

SachaZvetelman: Can you please add some JSDocs to these methods to explain what it does and what the parameters are?

packages/components/organisms/f-checkout/src/exceptions/exceptions.js
oliversweb: Then you would have to inject the translator into the exception class to resolve the translation, didn't really want that responsibility/complexity on the simple class, also unsure what we gain.

packages/components/organisms/f-checkout/src/components/Checkout.vue
oliversweb: I presume that is an issue on all translations, if we have forgotten to add a translation, e.g. if we have forgotten the 'errorMessages.genericServerError' translation then the same thing will happen. Do you want me to protect this line from that scenario and maybe use a hardcoded string if both translations fail to return a string?

packages/components/organisms/f-checkout/src/exceptions/exceptions.js
SachaZvetelman: Yup, fair enough.

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: What I had in mind was doing the following instead:
`const message = this.$t(error.messageKey) ||  this.$t('errorMessages.genericServerError');`

In this case, if the first key doesn't exist, at least we'd have a second chance with the generic one.

packages/components/organisms/f-checkout/src/mixins/logger.mixin.js
```diff
@@ -0,0 +1,29 @@
...
+            exceptionMessage: error.message,
+            exceptionStackTrace: error.stack
+        })
+    };
+}
+
+export default {
+    methods: {
+        /**
+         * @function logInvoker
```
kevinrodrigues, line 14: Nice docs.

packages/components/organisms/f-checkout/src/mixins/logger.mixin.js
```diff
@@ -0,0 +1,29 @@
...
+         * @param  {string} message   - Text that will appear in kibana log.
+         * @param  {object[]} data    - Bag of objects that will attempt to be displayed
```
kevinrodrigues: Should the `[]` be there?

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-checkout",
   "description": "Fozzie Checkout – Fozzie Checkout Component",
-  "version": "0.88.0",
+  "version": "0.90.0",
```
kevinrodrigues: How come this has skipped a few versions? Are there PRs going in before yours? Might be worth adding a DNMY label if so.

oliversweb: I was preempting those reviewed pr's (x2 ticks) before mine to be merged before mine

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -138,10 +129,17 @@ import {
...
+import loggerMixin from '../mixins/logger.mixin';
 import EventNames from '../event-names';
 import tenantConfigs from '../tenants';
 import { mapUpdateCheckoutRequest, mapAnalyticsNames } from '../services/mapper';
 
+const {
+    CreateGuestUserError,
+    UpdateCheckoutError,
+    PlaceOrderError
+} = exceptions;
```
SachaZvetelman, line 141: If you wanted, you could move this to the import: `import { ..., ..., ... } from '...';`

SachaZvetelman (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -161,7 +159,7 @@ export default {
         ErrorDialog
     },
 
-    mixins: [validationMixin, VueGlobalisationMixin, checkoutValidationsMixin],
+    mixins: [validationMixin, VueGlobalisationMixin, checkoutValidationsMixin, loggerMixin],
```
kevinrodrigues: Could we split this on to single lines now? Easier to scan.

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
oliversweb: I did but it complained that it could not find the Error classes so @benwhite-deltas suggested to do it this way, which worked.

kevinrodrigues (approved):

SachaZvetelman (approved):

PR #881, opened 2021-04-13
f-header@v4.12.0 - Use CSS modules (part 1)

FayeCarter:
### Changed
- Add shared `nav.scss` file.
- Convert `CountrySelectorPanel`, `Logo`, `SkipToMain` and `UserNavigationPanel` components to use css modules.
- Unit and Component tests to support changes.

**part 1 of 2**
[https://github.com/justeat/fozzie-components/pull/882](part2)
## UI Review Checks

**Desktop**
image
image
image

**Mobile**
image
image
image

**Mobile - menu open**
image

### DIFF
--- a/packages/components/organisms/f-header/CHANGELOG.md
+++ b/packages/components/organisms/f-header/CHANGELOG.md
@@ -4,8 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-Latest (add to next release)
+v4.12.0
 ------------------------------
+*April 9, 2021*
+
+### Changed
+- Add shared `nav.scss` file.
+- Convert `CountrySelectorPanel`, `Logo`, `SkipToMain` and `UserNavigationPanel` components to use css modules.
+- Unit and Component tests to support changes.
+
 *April 9, 2021*
 
 ### Changed
@@ -17,7 +24,7 @@ Latest (add to next release)
 ### Added
 - Component tests for locales `au, dk, ie, es, no, it, nz`
 - Component test for offers icon in mobile spec
-- Accessibility tests for the above locales 
+- Accessibility tests for the above locales
 
 ### Changed
 - Small refactor to `open` function in component-object
--- a/packages/components/organisms/f-header/CHANGELOG.md
+++ b/packages/components/organisms/f-header/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v4.12.0
 ------------------------------
-*April 9, 2021*
+*April 14, 2021*
 
 ### Changed
 - Add shared `nav.scss` file.
--- a/packages/components/organisms/f-header/src/assets/scss/navigation.scss
+++ b/packages/components/organisms/f-header/src/assets/scss/navigation.scss
@@ -17,7 +17,6 @@ $countrySelector-flag-width : 16px;
 $countrySelector-flag-height : 16px;
 
 @mixin nav-container-visible () {
-    background: yellow;
     overflow-y: auto;
     left: 0;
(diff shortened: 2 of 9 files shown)


packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -0,0 +1,143 @@
+@import './common.scss';
```
FayeCarter, line 1: None of this is new, just moved

packages/components/organisms/f-header/CHANGELOG.md
```diff
@@ -3,13 +3,24 @@
...
+### Changed
+- Add shared `nav.scss` file.
+- Convert `CountrySelectorPanel`, `Logo`, `SkipToMain` and `UserNavigationPanel` components to use css modules.
+- Unit and Component tests to support changes.
+
+
 v4.11.2
 ------------------------------
-*March 16, 2021*
+*March 13, 2021*
```
1itvinka, line 18: thanks! :)

JoshuaNg2332 (approved):

packages/components/organisms/f-header/src/components/Logo.vue
```diff
@@ -46,33 +48,33 @@ export default {
...
-                    return 'c-icon--onTransparentBg';
-                case 'highlight':
-                    return 'c-icon--onHighlightBg';
-                default:
-                    return '';
-            }
+        isAltLogo () {
+            const isHighlight = this.headerBackgroundTheme === 'highlight';
+            const isTransparent = this.headerBackgroundTheme === 'transparent' && !this.isOpen;
+            return isHighlight || isTransparent;
```
1itvinka, line 70: I would probably do:
```
...
const isTransparent = this.headerBackgroundTheme === 'transparent';
return isHighlight || (isTransparent && !this.isOpen);
```
A very minor thing in this case, but if `isHighlight` is `true` we're still calculating `isTransparent`, even though ideally we'd want to skip it.

1itvinka (approved):

packages/components/organisms/f-header/src/components/Logo.vue
FayeCarter: Ooh good point, I'll update it in the part 2. Thanks!

packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
```diff
@@ -79,6 +86,7 @@ export default {
 </script>
 
 <style lang="scss" module>
+@import '../assets/scss/navigation.scss';
```
ashleynolan, line 89: Out of interest – what in this SCSS file is used in this component?

Can see `$countrySelector-flag-width` and `$countrySelector-flag-height` – is that the only parts? (just wondering as could move those out into `common.scss` if that's the case rather than importing all the other styles too).

packages/components/organisms/f-header/src/components/Logo.vue
```diff
@@ -46,33 +48,33 @@ export default {
         headerBackgroundTheme: {
             type: String,
             default: 'white'
+        },
+        isOpen: {
```
ashleynolan, line 52: Super tiny one – I'd call this `isNavOpen` (assuming that's what it refers to?) as looks like there are a few references to different open states across the component.

packages/components/organisms/f-header/src/components/Logo.vue
```diff
@@ -11,9 +11,11 @@
         }`'>
         <component
             :is="iconComponent"
-            :class="['c-logo-img',
-                     iconClassName,
-                     logoColourModifier]"
+            :class="[
+                $style['c-logo-img'],
+                iconClassName,
+                { [$style['c-logo-img--alt']]: isAltLogo }
```
ashleynolan, line 17: Not sure what the answer is to this one, but do the logo states work ok when JS fails to load?

Just as although we don't aim to support no-JS, but the site should still look styled ok, so would be interested to know if the logo appears ok in that situation?

PR #884, opened 2021-04-13
f-form-field@1.10.0 - Add support for number input type

xander-marjoram:
### Added
- Support for `number` input type.
  - Includes two new optional props, `minNumber` and `maxNumber`.

### Fixed
- vuejs-accessibility violation [`form-control-has-label`](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility/blob/1cb838c853af763ac4d77467cdcd790f6896a810/docs/form-control-has-label.md)

---

## UI Review Checks

20210413T150936

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/atoms/f-form-field/src/components/FormField.vue
+++ b/packages/components/atoms/f-form-field/src/components/FormField.vue
@@ -13,7 +13,7 @@
             <form-label
                 v-if="!isInline"
                 :label-style="normalisedLabelStyle"
-                :for="uniqueId"
+                :label-for="uniqueId"
                 :is-inline="isInline"
                 :data-test-id="testId.label">
                 {{ labelText }}
@@ -21,7 +21,7 @@
 
             <form-dropdown
                 v-if="isDropdown"
-                :id="`${uniqueId}`"
+                :id="uniqueId"
                 :attributes="$attrs"
                 :type="normalisedInputType"
                 :value="value"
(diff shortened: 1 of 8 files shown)


1itvinka (approved):

kevinrodrigues (approved):

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -37,9 +37,12 @@
             <input
                 v-else
                 :id="`${uniqueId}`"
+                :aria-labelledby="`label-${uniqueId}`"
                 :value="value"
                 v-bind="$attrs"
                 :type="normalisedInputType"
+                :min="normalisedInputType === 'number' ? minNumber : false"
```
ashleynolan, line 44: Worth checking, but I don't think you'll need the actual minNumber and maxNumber props.

The `v-bind="$attrs"` line just above will pass through any attributes assigned to the outer element directly onto the input – if you need to normalise those numbers (to only apply them when type number is set), you could grab them off the `$attrs` passed in?

packages/components/atoms/f-form-field/src/components/_tests/FormField.test.js
```diff
@@ -70,6 +70,31 @@ describe('FormField', () => {
...
+
+            it('should not include the `min` and `max` attributes for default inputType', () => {
+                // Arrange & Act
+                const wrapper = shallowMount(FormField);
+                const formInput = wrapper.find('input');
+
+                // Assert
+                expect(formInput.attributes('min')).toBeUndefined();
+                expect(formInput.attributes('max')).toBeUndefined();
+            });
```
ashleynolan, line 96: Probably worth adding a test to make sure that the min/max is set to the number you expect it to be as well when provided?

packages/components/atoms/f-form-field/src/components/FormField.vue
ashleynolan: So I think this could just be `:min="normalisedInputType === 'number' ? $attrs['min'] : false"`

But I'd be tempted to change this to take in a computed prop and put the logic in there.

PR #885, opened 2021-04-14
f-form-field@1.10.1 - Fix double-firing of input and update events

xander-marjoram:
Fixing a long-standing (but not nasty) bug where input and update events were triggered twice.

The alternative was to keep the v-on handlers and replace the `listeners` computed object with `this.$listeners`, but that actually breaks the existing implementation because it passes through the whole event object rather than just the updated value.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/atoms/f-form-field/src/components/_tests/FormField.test.js
+++ b/packages/components/atoms/f-form-field/src/components/_tests/FormField.test.js
@@ -259,7 +259,7 @@ describe('FormField', () => {
             });
         });
 
-        describe('@change', () => {
+        describe('@change ::', () => {
             it('should only trigger one event emission', () => {
                 // Arrange
                 const propsData = {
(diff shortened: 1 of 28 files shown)


1itvinka (approved):

kevinrodrigues (approved):

packages/components/atoms/f-form-field/src/components/_tests/FormField.test.js
```diff
@@ -241,4 +241,41 @@ describe('FormField', () => {
...
+        describe('@input ::', () => {
+            it.each(VALID_INPUT_TYPES)('should only trigger one event emission for inputType `%s`', inputType => {
+                // Arrange
+                const propsData = { inputType };
+                const wrapper = shallowMount(FormField, { propsData });
+
+                const formInput = wrapper.find('input');
+
+                // Act
+                formInput.trigger('input');
```
ashleynolan, line 255: Just a thought, but would it be better to actually change the value of the input and check that the input event was triggered?

PR #889, opened 2021-04-15
f-checkout@v0.92.0 - Fallback to use AuthToken customer phones numbers

oliversweb:
v0.92.0
*April 16, 2021*

### Added
- Use AuthToken phone nos. if checkout api fails to return a customer phone no.


packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -341,6 +340,95 @@ describe('CheckoutModule', () => {
                 // Assert
                 expect(dispatch).toHaveBeenCalledWith(`${VUEX_CHECKOUT_ANALYTICS_MODULE}/updateAutofill`, state, { root: true });
             });
+
+            describe('when the `customer` model is not returned from the api', () => {
+                it('should not error when checking phone number', async () => {
+                    // Arrange
+                    checkoutDelivery.customer = null;
```
BanksySan: We're changing a shared fixture here, it might cause errors when running in parrallel.

BanksySan (changes requested):

packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
kevinrodrigues: You could mock the call from axios and return a falsey value or empty object too. Might be easier to read that way.

packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -341,6 +340,95 @@ describe('CheckoutModule', () => {
...
+
+                    // Act
+                    await getCheckout({ commit, state, dispatch }, payload);
+
+                    // Assert
+                    expect(checkoutDelivery.customer).toBe(null);
+                    expect(axios.get).toHaveBeenCalledWith(payload.url, config);
```
kevinrodrigues, line 358: Do we need this expect and the following expect in each test? Think just the one above is fine?

BanksySan (approved):

dandel10n (approved):

kevinrodrigues (approved):

PR #898, opened 2021-04-20
f-checkout@0.95.0 - Handling duplicated orders.

SachaZvetelman:
_Note: I haven't been able to verify this on a consuming application yet, given one of the underlying APIs is failing. I will test it before merging._

_Note 2: as there are a few `TODOs` in this PR, I'll create a ticket to refactor those._

v0.95.0
*April 26, 2021*

### Added
- Handling of duplicate orders.

### Changed
- Refactored handling order issues.


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -477,6 +479,10 @@ export default {
             }
         },
 
+        toggleDialogError () {
```
ashleynolan, line 478: Can you add a JSDoc comment for this method please

packages/components/organisms/f-checkout/src/tenants/en-GB.js
```diff
@@ -89,13 +89,19 @@ const messages = {
...
 
             DEFAULT_CHECKOUT_ISSUE: {
                 title: 'Something went wrong',
                 message: 'Please try again',
                 buttonText: 'Retry'
+            },
+
+            DuplicateOrder: {
+                title: 'This looks like a duplicate order',
+                message: 'So you\'re not charged twice, your order has not been processed',
```
ashleynolan: You should use a proper apostrophe `’` here – that would save you needing to escape the base `dumb` apostrophe.

Base vertical apostrophes are just known as `dumb apostrophes` ([more info here](https://www.fastcompany.com/3018353/be-smart-dont-use-dumb-quote-marks) if you're a type nerd like me) and it looks nicer to use the proper ones where we remember to.

SachaZvetelman: The only problem I have with this is...I need to find out how to type that apostrophe on a Spanish keyboard 🤦

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -708,18 +719,26 @@ export default {
...
             });
+
             this.trackFormInteraction({ action: 'error', error: `error_${error.message}` });
-            this.genericErrorMessage = message;
 
-            this.$nextTick(() => {
-                this.scrollToElement(this.$refs.errorAlert.$el);
-            });
+            // We don't want to show a dialog and an error message.
+            // TODO: refactor `nonFulfillableError` and `genericErrorMessage` so that we try to use only one, and
```
charliearlie, line 733: 100% agree on that!

charliearlie (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
BanksySan, line 733: Can we move TODOs to tickets?

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -291,23 +297,36 @@ export default {
...
+            } catch (error) {
+                if (error.response && error.response.data) {
+                    const { errorCode } = error.response.data;
+
+                    const checkoutIssue = getIssueByCode(errorCode);
+
+                    commit(UPDATE_ERRORS, (checkoutIssue ? [checkoutIssue] : []));
+                }
+
+                throw error; // Handled by the calling function.
```
BanksySan, line 328: No need for this comment, it's implicit that the error will be handled/rethrown upstream.

packages/components/organisms/f-checkout/src/tenants/en-GB.js
```diff
@@ -89,13 +89,19 @@ const messages = {
             LOCATION_UNDELIVERABLE: {
                 title: 'The restaurant doesn\'t deliver to this address',
                 message: 'Search for a different restaurant or change to a collection order.',
-                buttonText: 'OK'
+                buttonText: 'Ok'
```
BanksySan, line 92: Did you mean to change the casing here?

BanksySan (changes requested):

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: Tickets are going to be created for this. But I think it's also good to keep the note there, so that if someone sees the code, they know it's going to change soon. Unless people generally prefer not to have this. Whatever the team prefers 😄

packages/components/organisms/f-checkout/src/store/checkout.module.js
SachaZvetelman: I'm not a huge fan of comments, personally, but I think in this case it's a useful one. It's not implicit that an error will be handled upstream. If it's not handled, it can crash the application, or throw an unhandled exception to the user. With this comment, my intention was to inform other devs that we should be fine, since the calling function is indeed handling it.
Again, I'm more than happy to remove the comment if people think it's not needed, of course.

packages/components/organisms/f-checkout/src/tenants/en-GB.js
SachaZvetelman: Yeah, since I wanted to match what the iOS app shows, but thanks for checking 👍

packages/components/organisms/f-checkout/src/components/Checkout.vue
BanksySan: If we have a ticket, could we reference it in here, otherwise there's the danger f it not getting removed after it's fixed and just being a confusing comment.

packages/components/organisms/f-checkout/src/tenants/en-GB.js
BanksySan: I think we should be consistent on apostrophe usages.

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: I'm sending you a DM with regards to this.

charliearlie (approved):

BanksySan (approved):

PR #899, opened 2021-04-20
f-checkout@v0.93.0 - Refactor Customer Details Fallback

oliversweb:
v0.93.0
*April 20, 2021*

### Added
- Removed alternative method of getting customer details so all access is via the `$State`


packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -374,6 +373,23 @@ describe('CheckoutModule', () => {
                     // Assert
                     expect(checkoutDeliveryCopy.customer.phoneNumber).toBe(expectedPhoneNumber);
                 });
+
+                it('should not use neither of AuthToken customer names', async () => {
+                    // Arrange
+                    const expectedFirstName = 'Jazz';
+                    const expectedLastName = 'Man';
```
minrwhite, line 380: 🎷

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -22,6 +22,26 @@ import {
...
 
+/**
+* @function resolveCustomerDetails
+* If certain customer details are missing from api `data.customer` object then fallback
+* to the decoded `state.AuthToken` details and re-assign back to `data.customer`.
+* @param  {object} data  - Api response object.
+* @param  {object} state - The current `checkout` state.
+*/
+function resolveCustomerDetails (data, state) {
+    const tokenData = jwtDecode(state.authToken);
```
minrwhite: It might be premature optimization, but in the replaced code, the token was only decoded if the necessary keys (previously only `phoneNumber`) weren't set in the given data. Is it worth doing the same here?

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -421,10 +421,6 @@ export default {
             await Promise.all(promises);
             this.resetLoadingState();
 
-            if (this.shouldLoadCustomerNameFromClaims) {
```
minrwhite, line 424: I don't think we need the `shouldLoadCustomerNameFromClaims` computed property any more?

simonsandersje (approved):

minrwhite (approved):

PR #906, opened 2021-04-28
f-checkout@v0.97.0 - Changed main headers to be h1 elements

1itvinka:
### Changed
- Changed main headers to be h1 elements

I've made sure that the text still looks the same

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/Header.vue
+++ b/packages/components/organisms/f-checkout/src/components/Header.vue
@@ -5,9 +5,11 @@
         <div
             v-if="isLoggedIn"
             data-test-id="user-checkoutHeader">
-            <h2 data-test-id="user-title">
+            <h1
+                data-test-id="user-title"
+                :class="$style['c-checkoutHeader-title']">
                 {{ $t('checkoutHeader.user.title', { name }) }}
-            </h2>
+            </h1>
 
             <p :class="$style['c-checkoutHeader-loginLink']">
                 <a
@@ -23,9 +25,9 @@
         <div
             v-else
             data-test-id="guest-checkoutHeader">
-            <h2 :class="$style['c-checkoutHeader-title']">
+            <h1 :class="$style['c-checkoutHeader-guestTitle']">
                 {{ $t('checkoutHeader.guest.loginTitle') }}
-            </h2>
+            </h1>
 
             <f-button
                 button-type="primary"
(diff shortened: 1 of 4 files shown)


packages/components/organisms/f-checkout/src/components/Header.vue
```diff
@@ -108,6 +110,10 @@ export default {
     text-align: center;
 }
 
+.c-checkoutHeader-userTitle {
```
1itvinka: not sure if this class name is good enough

charliearlie: I think if this is the H1 asking the user to confirm their details then it's definitely good enough 🙂

charliearlie (dismissed):

dandel10n (dismissed):

dandel10n (dismissed):

charliearlie (dismissed):

packages/components/organisms/f-checkout/src/components/Header.vue
ashleynolan: Any reason not to just call it `c-checkoutHeader-title`? Just protects it a little if the content of the h1 changes in the future (away from being a user title)

ashleynolan: Just seen that there's already a title called `c-checkoutHeader-title` so guessing that's why. Looking at the two titles, I'd probably make this class the `c-checkoutHeader-title` and change the classname of the other to be named closer to guest experience – `c-checkoutHeader-guestTitle`?

1itvinka: yeah, that was the reason. The styles are a bit different so I needed a new class name. Your suggestion sounds good. Will change the class names.

DamianMullins (dismissed):

dandel10n (dismissed):

packages/components/organisms/f-checkout/src/components/Header.vue
```diff
@@ -23,9 +25,9 @@
         <div
             v-else
             data-test-id="guest-checkoutHeader">
-            <h2 :class="$style['c-checkoutHeader-title']">
+            <h1 :class="$style['c-checkoutHeader-guestTitle']">
```
siggerzz: Could we add a `data-test-id="user-title"` attribute here too?

1itvinka: I'll add it now

dandel10n (approved):

DamianMullins (approved):

PR #908, opened 2021-04-28
f-http@v0.2.0 - Developing POC

tommcclean:
- This PR develops the POC in the following ways - [checkout the readme if you want to feedback on how it works](https://github.com/tommcclean/fozzie-components/blob/f-http%40v0.2.0/packages/services/f-http/README.md) so far
- Its not finished yet, just trying to use checkpoints so I don't end up with a big PR

### Added
- Support for configuration merging
- Support for basic GET and POST requests

### Changed
- Updated Readme documentation

### DIFF
--- a/packages/services/f-http/src/createClient.js
+++ b/packages/services/f-http/src/createClient.js
@@ -1,0 +1,73 @@
+import axios from 'axios';
+import defaultOptions from './defaultOptions';
+
+let _configuration = null;
+let _axiosInstance = null;
+
+/**
+    * Get a resource
+    * @param {string} resource - The resource to post (URL)
+    * @return {object} - Returns data from response
+*/
+const get = async resource => {
+    try {
+        const response = await _axiosInstance.get(resource);
+
+        return response.data;
+    } catch (error) {
+        if (_configuration.errorCallback) {
+            _configuration.errorCallback(error, resource);
+        }
+
+        throw error;
+    }
+};
+
+/**
+    * Post a resource
+    * @param {string} resource - The resource to post (URL)
+    * @param {object} body - The request body, contents of the resource
+    * @param {object} headers - Any additional request headers you want to post
+    * @return {object} - Returns data from response
+*/
+const post = async (resource, body, headers = {}) => {
+    try {
+        const config = {
+            headers
+        };
+
+        const response = await _axiosInstance.post(resource, body, config);
+
+        return response.data;
+    } catch (error) {
+        if (_configuration.errorCallback) {
+            _configuration.errorCallback(error, resource);
+        }
+
+        throw error;
+    }
+};
+
+/**
+    * Create a httpClient
+    * @param {object} options - Any options to override - refer to documentation for options
+    * @return {object} - Returns an object with restful request methods
+*/
+export default options => {
+    // Merge default configuration with overrides
+    _configuration = {
+        ...defaultOptions,
+        ...options
+    };
+
+    _axiosInstance = axios.create({
+        baseURL: _configuration.baseUrl,
+        timeout: _configuration.timeout
+    });
+
+    return {
+        get,
+        post,
+        readConfiguration: () => _configuration
+    };
+};
--- a/packages/services/f-http/src/createClient.js
+++ b/packages/services/f-http/src/createClient.js
@@ -1,25 +1,27 @@
 import axios from 'axios';
 import defaultOptions from './defaultOptions';
+import handleError from './errorHandler';
 
 let _configuration = null;
 let _axiosInstance = null;
 
 /**
     * Get a resource
-    * @param {string} resource - The resource to post (URL)
+    * @param {string} resource - The resource to get (URL)
+    * @param {object} headers - Any additional request headers you want to provide
     * @return {object} - Returns data from response
 */
-const get = async resource => {
+const get = async (resource, headers = {}) => {
     try {
-        const response = await _axiosInstance.get(resource);
+        const config = {
+            headers
+        };
+
+        const response = await _axiosInstance.get(resource, config);
 
         return response.data;
     } catch (error) {
-        if (_configuration.errorCallback) {
-            _configuration.errorCallback(error, resource);
-        }
-
-        throw error;
+        return handleError(error, _configuration.errorCallback);
     }
 };
 
@@ -27,7 +29,7 @@ const get = async resource => {
     * Post a resource
     * @param {string} resource - The resource to post (URL)
     * @param {object} body - The request body, contents of the resource
-    * @param {object} headers - Any additional request headers you want to post
+    * @param {object} headers - Any additional request headers you want to provide
     * @return {object} - Returns data from response
 */
 const post = async (resource, body, headers = {}) => {
@@ -40,11 +42,7 @@ const post = async (resource, body, headers = {}) => {
 
         return response.data;
     } catch (error) {
-        if (_configuration.errorCallback) {
-            _configuration.errorCallback(error, resource);
-        }
-
-        throw error;
+        return handleError(error, _configuration.errorCallback);
     }
 };
 
--- a/packages/services/f-http/src/tests/createClient.test.js
+++ b/packages/services/f-http/src/tests/createClient.test.js
@@ -2,14 +2,13 @@ import createClient from '../createClient';
 
 describe('createClient', () => {
     describe('initialisation', () => {
-        it('httpClient should be defined', async () => {
-        // Arrange & Act
-        // Assert
+        it('should be defined', async () => {
+            // Arrange, Act & Assert
             expect(createClient).toBeDefined();
         });
 
-        it('httpClient should use default options when not overridden', async () => {
-        // Arrange & Act
+        it('should use default options when not overridden', async () => {
+            // Arrange & Act
             const httpClient = createClient();
             const mergedOptions = httpClient.readConfiguration();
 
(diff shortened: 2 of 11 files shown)


packages/services/f-http/src/createClient.js
```diff
@@ -0,0 +1,73 @@
...
+/**
+    * Get a resource
+    * @param {string} resource - The resource to post (URL)
+    * @return {object} - Returns data from response
+*/
+const get = async resource => {
+    try {
+        const response = await _axiosInstance.get(resource);
+
+        return response.data;
```
BanksySan, line 22: Potential null exception if the response isn't a 2xx.

packages/services/f-http/src/createClient.js
```diff
@@ -0,0 +1,73 @@
...
+    } catch (error) {
+        if (_configuration.errorCallback) {
+            _configuration.errorCallback(error, resource);
```
BanksySan: We should wrap this in a `try-catch`.

packages/services/f-http/src/createClient.js
```diff
@@ -0,0 +1,73 @@
...
+        const config = {
+            headers
+        };
+
+        const response = await _axiosInstance.post(resource, body, config);
+
+        return response.data;
+    } catch (error) {
+        if (_configuration.errorCallback) {
+            _configuration.errorCallback(error, resource);
```
BanksySan: Same here, catch exception.

packages/services/f-http/src/createClient.js
```diff
@@ -0,0 +1,73 @@
...
+
+    _axiosInstance = axios.create({
+        baseURL: _configuration.baseUrl,
+        timeout: _configuration.timeout
+    });
+
+    return {
+        get,
+        post,
+        readConfiguration: () => _configuration
```
BanksySan, line 69: Why is this a method rather than a variable?

packages/services/f-http/src/tests/index.test.js
```diff
@@ -1,11 +1,22 @@
...
+
+    it('httpClient should expose expected methods', async () => {
+        // Arrange & Act
+        const httpClient = createClient();
+
+        // Assert
+        expect(httpClient).toBeDefined();
+        expect(httpClient.get).toBeDefined();
+        expect(httpClient.post).toBeDefined();
+        expect(httpClient.readConfiguration).toBeDefined();
```
BanksySan, line 19: Should there be test cases for null being returned for `data` and for the calls to `errorCallback` throwing?

BanksySan (dismissed):

packages/services/f-http/src/createClient.js
tommcclean: Would be handled by the try-catch and throw to the caller who should handle it, any bad response code would fail before this line as well. Additionally the caller can even provide a function to globally handle errors which this would fall into or they could opt to do as most do already and wrap a try catch around their call.

packages/services/f-http/src/createClient.js
tommcclean: Its in a try catch, this is a call-back provided by the caller; so should their own callback fail they are responsible for handling that by handling errors in their own code. They are forced to consider that, as the default is no callback; and thus would simply rethrow to the line where it is called; rather than the installation code. So its kind of an optional way of handling it all in one place.

packages/services/f-http/src/createClient.js
tommcclean: Same as above

packages/services/f-http/src/tests/index.test.js
tommcclean: I intend to put some tests in for the GET, PUT, POST, PATCH, DELETE's when they are all in there as part of the same ticket; just trying to keep the PR's manageable and small. Much of that code will be shared between them and thus the tests too.

packages/services/f-http/src/createClient.js
tommcclean: Just a matter of preference, makes no difference in reality; though it may be desirable to return a new version of the object from the function rather than exposing the reference later.

packages/services/f-http/src/createClient.js
```diff
@@ -0,0 +1,73 @@
+import axios from 'axios';
+import defaultOptions from './defaultOptions';
+
+let _configuration = null;
+let _axiosInstance = null;
+
+/**
+    * Get a resource
+    * @param {string} resource - The resource to post (URL)
```
SachaZvetelman: "The resource to get", right? 😬

packages/services/f-http/src/createClient.js
```diff
@@ -0,0 +1,73 @@
...
+    * @return {object} - Returns data from response
+*/
+const get = async resource => {
```
SachaZvetelman: I think we should have the ability to pass headers in here as well.

packages/services/f-http/src/createClient.js
```diff
@@ -0,0 +1,73 @@
...
+
+        const response = await _axiosInstance.post(resource, body, config);
+
+        return response.data;
+    } catch (error) {
+        if (_configuration.errorCallback) {
+            _configuration.errorCallback(error, resource);
+        }
+
+        throw error;
```
SachaZvetelman: It seems that this snippet will be repeated pretty much across all the different request types, so going forward you might benefit from extracting this into a separate function.

tommcclean: Thats the plan

packages/services/f-http/src/createClient.js
tommcclean: Fixed

packages/services/f-http/src/createClient.js
tommcclean: Fair - Implemented

packages/services/f-http/src/createClient.js
tommcclean: Extracted and tested

packages/services/f-http/src/tests/errorHandler.test.js
```diff
@@ -0,0 +1,40 @@
...
+        const errorCallbackMock = jest.fn();
+
+        // Act
+        handleError(error, errorCallbackMock);
+
+        // Assert
+        expect(errorCallbackMock.mock.calls.length).toBe(1);
+
+        // Check that the first argument is the error
+        expect(errorCallbackMock.mock.calls[0][0]).toBe(error);
```
SachaZvetelman: Would `expect(errorCallbackMock).toHaveBeenCalledWith(error);` work?

tommcclean: Ta, much better

packages/services/f-http/src/tests/errorHandler.test.js
```diff
@@ -0,0 +1,38 @@
...
+            message: 'This is an example error'
+        };
+
+        const errorCallbackMock = jest.fn();
+
+        // Act
+        handleError(error, errorCallbackMock);
+
+        // Assert
+        expect(errorCallbackMock.mock.calls.length).toBe(1);
```
SachaZvetelman: You can get rid of this now.

packages/services/f-http/src/tests/errorHandler.test.js
```diff
@@ -0,0 +1,38 @@
+import handleError from '../errorHandler';
+
+describe('errorHandler', () => {
+    it('errorHandler should be defined', () => {
```
SachaZvetelman: No need to mention `errorHandler` in every test name since their `describe` block already mentions it.

packages/services/f-http/src/tests/errorHandler.test.js
```diff
@@ -0,0 +1,38 @@
...
+    });
+
+    it('errorHandler should re-throw error when callback is not provided', () => {
+        // Arrange
+        const expectedError = new Error('This is an example error');
+
+        try {
+            // Act
+            handleError(expectedError);
+            throw new Error('Test Failed');
```
SachaZvetelman: Doesn't this defeat the purpose of the test? Shouldn't this throw be done by `handleError`?

packages/services/f-http/src/tests/errorHandler.test.js
tommcclean: I guess I still want to make sure it is just called once and its clear from the test that is the case

SachaZvetelman: I might be wrong, but I think the `toHaveBeenCalledWith();` will fail if it's called more than once. If I'm wrong, you can replace your line with `expect(errorCallbackMock).toHaveBeenCalledTimes(1);` which in my opinion looks slightly nicer and more readable.

tommcclean: Ill go with the latter; prefer it to be obvious if not from the way the test functions, but how it reads.

SachaZvetelman (approved):
Looking like a great and promising package! 📦 🚀

packages/services/f-http/src/tests/createClient.test.js
```diff
@@ -0,0 +1,42 @@
...
+            expect(createClient).toBeDefined();
+        });
+
+        it('should use default options when not overridden', async () => {
+            // Arrange & Act
+            const httpClient = createClient();
+            const mergedOptions = httpClient.readConfiguration();
+
+            // Assert
+            expect(mergedOptions).toBeDefined();
```
kevinrodrigues, line 16: You could shorten this to just checking the object itself? That way you will just have one expect?

packages/services/f-http/src/tests/createClient.test.js
```diff
@@ -0,0 +1,42 @@
...
+
+            const httpClient = createClient({
+                baseUrl: expectedBaseUrl,
+                timeout: expectedTimeout,
+                errorCallback: expectedErrorCallback
+            });
+            const mergedOptions = httpClient.readConfiguration();
+
+            // Assert
+            expect(mergedOptions).toBeDefined();
```
kevinrodrigues, line 36: Same.

kevinrodrigues (approved):

packages/services/f-http/src/tests/createClient.test.js
tommcclean: Yeah good idea, ill switch to this in the next version

PR #909, opened 2021-04-28
f-checkout@v0.103.0 - Bump `f-services`/`f-form-field` version to 1.11.0

oliversweb:
v0.103.0
*May 7, 2021*

### Changed
- Bumped `f-services` version to 1.11.0.
- Bumped `f-form-field` version to 1.11.0.
- Set tel. no. field type as `Tel`


DamianMullins (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -47,6 +47,7 @@
                     <form-field
                         :value="customer.mobileNumber"
                         name="mobile-number"
+                        type="tel"
```
xander-marjoram: We should probably extend f-form-field to accept `input-type="tel"` rather than sticking the attribute on directly.

xander-marjoram (changes requested):

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: I agree, although it can be done in a separate ticket, imho.

DamianMullins (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -48,6 +48,7 @@
                     <form-field
                         :value="customer.mobileNumber"
                         name="mobile-number"
+                        :input-type="tel"
```
xander-marjoram:
```suggestion
                        input-type="tel"
```
Should just be a string rather than a v-bind

simonsandersje (approved):

xander-marjoram (approved):

PR #911, opened 2021-04-29
f-checkout@v0.102.0 - refactor error display

FayeCarter:
### Changed
- `genericErrorMessage` and `nonFulfillableError` replaced with `vuex` `message`.
- One of `ErrorDialog` or `f-alert` to be displayed if an error should be displayed to the user.

### Added
- Tests to cover changes in `ErrorDialog`


packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
```diff
@@ -1,58 +1,106 @@
 <template>
```
SachaZvetelman, line 1: This file should probably be named differently, since it's not just a dialog. It could be a "dialog" (modal) or an alert, right?

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
```diff
@@ -1,58 +1,106 @@
...
         FButton
     },
 
-    props: {
-        isOpen: {
-            type: Boolean,
-            default: false
+    computed: {
+        ...mapState(VUEX_CHECKOUT_MODULE, [
+            'displayError',
```
SachaZvetelman: If this is a boolean value, then let's name it `shouldDisplayError`.

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
```diff
@@ -1,58 +1,106 @@
...
-            type: String,
-            default: ''
+
+        shouldShowModal () {
+            return this.displayError ? this.displayError.shouldShowInDialog : false;
+        },
+
+        shouldShowAlert () {
+            return this.displayError ? this.displayError.shouldShowInAlert : false;
+        },
```
SachaZvetelman: Do we need both? Since it's going to be either a dialog or an alert.
We might need some renaming, though.

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
SachaZvetelman: I might be inclined to have two separate components. One for dialogs, one for alerts, and we should one or the other only. Not sure, though.

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -382,6 +386,21 @@ export default {
...
+        updateDisplayError: ({ commit }, payload) => {
+            const alert = {
+                code: payload,
+                shouldShowInAlert: true
+            };
+
+            let displayError = null;
+
+            if (payload) {
+                displayError = payload[0].code ? payload[0] : alert;
```
SachaZvetelman: Is this correct or should be the other way around: `payload[0] ? payload[0].code : alert`?

SachaZvetelman (commented):
Not sure if calling things "errors" is the best idea, since we have errors and issues that we display. Maybe the generic name should be "problems" or similar, if that makes sense.

ashleynolan:
Re @SachaZvetelman 's comment above about generic naming instead of using Errors – would using `Message` be too generic?

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
ashleynolan: Rather than using `shouldShowInDialog` or `shouldShowInAlert`, I would probably just use a variable to reference a type as that's then a lot more scalable (`this.displayError.displayType` === `alert` or `dialog`).

Unless as Sacha mentioned you can just combine these anyway

packages/components/organisms/f-checkout/src/components/_tests/ErrorDialog.test.js
```diff
@@ -13,7 +13,7 @@ const localVue = createLocalVue();
 localVue.use(VueI18n);
 localVue.use(Vuex);
 
-describe('ErrorDialog', () => {
+xdescribe('ErrorDialog', () => {
```
ashleynolan: Guessing this should be removed before it's merged?

packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -130,7 +130,7 @@ const defaultState = {
 
 let state = CheckoutModule.state();
 
-describe('CheckoutModule', () => {
+xdescribe('CheckoutModule', () => {
```
ashleynolan: Same here as above

packages/components/organisms/f-checkout/src/components/_tests/ErrorDialog.test.js
FayeCarter: Hey, yeah it definitely will.  I had to ignore these to commit to github :)

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
FayeCarter: This one was the actual message. I'll come up with a better name so that it makes more sense. Thanks!

packages/components/organisms/f-checkout/src/components/_tests/ErrorDialog.test.js
```diff
@@ -31,26 +42,132 @@ describe('ErrorDialog', () => {
...
+                // Assert
+                expect(wrapper.vm.restaurantMenuPageUrl).toEqual(`restaurant-${restaurant.seoName}/menu`);
+            });
+        });
+    });
+
     describe('methods ::', () => {
-        describe('handleButtonClick ::', () => {
-            it('should emit the `error-button-click` event', () => {
+        describe('closeErrorDialog ::', () => {
```
FayeCarter, line 103: These tests aren't new, just moved out of `checkout.test`

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
FayeCarter: I've removed alert from the component

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
FayeCarter: changed to `message`

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -335,23 +324,29 @@ export default {
...
+
+            const alert = {
+                name: 'alert',
+                props: {
+                    type: 'danger',
+                    class: this.$style['c-checkout-alert'],
+                    heading: this.$t('errorMessages.errorHeading')
+                },
+                content: this.message
+            };
```
SachaZvetelman: I'd probably extract these two into two separate computed properties.

So you'd end up with something like this:

```
messageType () {
  return this.message && !this.message.shouldShowInDialog
    ? this.alertMessage
    : this.dialogMessage;
}
```

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -547,7 +533,6 @@ export default {
                     logMethod: this.$logger.logInfo
                 });
             } catch (e) {
-                this.toggleDialogError();
```
SachaZvetelman, line 551: Have you tested that it works well when you're placing a duplicated order?

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
```diff
@@ -1,27 +1,31 @@
...
-        v-on="$listeners">
-        <h3
-            data-test-id="checkout-issue-modal-title"
-            class="u-noSpacing">
-            {{ $t(`errorMessages.checkoutIssues.${errorCode}.title`, { serviceType: serviceTypeText }) }}
-        </h3>
+        :is-open="isOpen"
+        @close="closeErrorDialog"
+    >
+        <span v-if="isOpen">
```
SachaZvetelman: Why this span? :)

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
```diff
@@ -30,39 +34,52 @@ import MegaModal from '@justeat/f-mega-modal';
...
+
+        errorCode () {
+            return this.message && this.message.code;
         },
-        errorCode: {
-            type: String,
-            default: ''
+
+        isOpen () {
+            return this.message && this.message.shouldShowInDialog;
```
SachaZvetelman: I might be wrong, but this looks a bit weird.
What if you've got a message and it should be shown in a dialog but it hasn't been shown yet?

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -485,121 +458,78 @@ describe('Checkout', () => {
...
                 const wrapper = shallowMount(VueCheckout, {
                     store: createStore({
                         ...defaultCheckoutState,
-                        isLoggedIn: true
+                        message
                     }),
                     i18n,
                     localVue,
                     propsData
                 });
```
SachaZvetelman: The arrange and act sections should be inside the test, or at least in a `beforeEach`.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -485,121 +458,78 @@ describe('Checkout', () => {
...
-                            seoName: restaurantSeoName
-                        }
+                        message: alertMessage
                     }),
                     i18n,
                     localVue,
-                    propsData
+                    propsData,
+                    mocks: {
+                        $style
```
SachaZvetelman: Do we need this?

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -382,11 +387,11 @@ export default {
...
             commit(UPDATE_HAS_ASAP_SELECTED, payload);
-        }
-    },
+        },
 
-    getters: {
-        firstDialogError: state => state.errors.find(error => error.shouldShowInDialog)
+        updateMessage:  ({ commit }, message) => {
+            commit(UPDATE_MESSAGE, message || null);
+        }
```
SachaZvetelman: Nit, but I'd personally do:
```
      updateMessage:  ({ commit }, message = null) => {
            commit(UPDATE_MESSAGE, message);
      }
```

packages/components/organisms/f-checkout/src/components/Checkout.vue
FayeCarter: I've tested it in storybook with the duplicate error knob and it still works. It's set here now https://github.com/justeat/fozzie-components/pull/911/files/ac174149c8afa9d3c7fa66fb089bced18c5b643e#diff-3f7bda367e6ae2f5c141559c1aec5722d6606ee4d895f7d44e9837c20f7efb3cR719

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
FayeCarter: We do. I have a test that when the `alertMessage` is passed a prop with a class in. Without the `$style` mock it comes back as undefined

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
FayeCarter: This has been removed

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -1,11 +1,14 @@
...
-        <error-dialog
-            :is-open="shouldShowErrorDialog"
-            :error-code="hasNonFulfillableErrorCode"
-            :service-type="serviceType"
-            @close="handleErrorDialogClose"
-            @checkout-error-dialog-button-click="handleErrorDialogButtonClick" />
+        <component
+            :is="messageType.name"
+            v-if="message"
+            ref="errorAlert"
```
SachaZvetelman: We should probably change this ref now. It's no longer necessarily an `errorAlert`.

packages/components/organisms/f-checkout/src/exceptions/exceptions.js
```diff
@@ -25,6 +27,7 @@ class PlaceOrderError extends Error {
         this.messageKey = 'errorMessages.genericServerError';
         this.eventToEmit = EventNames.CheckoutPlaceOrderFailure;
         this.logMessage = 'Place Order Failure';
+        this.shouldShowInDialog = true;
```
SachaZvetelman: Is this actually accurate? For duplicate orders, yes, we want a dialog, but for non-duplicate orders, or unknown failures, we don't.

FayeCarter: Damn. Good point

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -547,8 +540,9 @@ export default {
                     logMethod: this.$logger.logInfo
                 });
             } catch (e) {
-                this.toggleDialogError();
-                throw new PlaceOrderError(e.message);
+                const { errorCode } = e.response.data;
+
+                throw new PlaceOrderError(e.message, errorCode === 'DuplicateOrder');
```
SachaZvetelman: `errorCode === 'DuplicateOrder'` is not very extensible in my opinion.
You already have this information in `checkout-issues.js`, so it would be better to drive this decision from that file, which should be our single source of truth in terms of handling problems (issues or errors).
```
  const { shouldShowInDialog } = checkoutIssues[errorCode];

  throw new PlaceOrderError(e.message, shouldShowInDialog);
```

packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -636,6 +645,27 @@ describe('CheckoutModule', () => {
...
+                                }
+                            }
+                        }));
+
+                        try {
+                            // Act
+                            await placeOrder({ commit, state, dispatch }, payload);
+                        } catch {
+                            // Assert
+                            expect(dispatch).toHaveBeenCalledWith('updateMessage', message);
```
minrwhite, line 666: Given that this might not be called if the placeOrder function erroneously returns without throwing, it would be a good idea to have a [`expect.hasAssertions()`](https://jestjs.io/docs/expect#expecthasassertions) statement earlier in the test, to ensure that we are actually making this assertion.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -547,8 +541,10 @@ export default {
                     logMethod: this.$logger.logInfo
                 });
             } catch (e) {
-                this.toggleDialogError();
-                throw new PlaceOrderError(e.message);
+                const { errorCode } = e.response.data;
+                const { shouldShowInDialog } = checkoutIssues[errorCode];
+
+                throw new PlaceOrderError(e.message, shouldShowInDialog);
```
minrwhite: To future-proof this slightly, and prevent having to add more parameters into the constructor signature as more requirements emerge, I'd recommend passing in the `errorCode` rather than the `shouldShowInDialog` value, and resolve to properties within the constructor by importing `checkout-issues.js` within the exceptions definition file.

minrwhite (dismissed):
Good working through of this problem - I have a couple suggestions, but happy if those go in a later PR if you'd prefer.

packages/components/organisms/f-checkout/src/exceptions/exceptions.js
```diff
@@ -22,12 +23,12 @@ class UpdateCheckoutError extends Error {
...
 
 class PlaceOrderError extends Error {
-    constructor (message, shouldShowInDialog = false) {
+    constructor (message, errorCode) {
         super(message);
         this.messageKey = 'errorMessages.genericServerError';
         this.eventToEmit = EventNames.CheckoutPlaceOrderFailure;
         this.logMessage = 'Place Order Failure';
-        this.shouldShowInDialog = shouldShowInDialog;
+        this.shouldShowInDialog = checkoutIssues[errorCode];
```
minrwhite: If I read the code correctly, this would make `this.shouldShowInDialog` an object, but we want the value for the `shouldShowInDialog`  key from that object in here:
```suggestion
        const issue = checkoutIssues[errorCode] || {}; // in case the errorCode is unrecognised E.g. from a later API version
        this.shouldShowInDialog = issue.shouldShowInDialog || false; // or `true` if that makes more sense as a default value here
```

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: I'm not sure I agree with this, because now the custom errors have a direct dependency on `checkout-issues.js`.

minrwhite: I see what you mean @SachaZvetelman - but we're adding the dependency in here instead.

In that case, given we're already depending on `checkout-issues.js` within the checkout vuex module, would it not be better generally to have try-catches within the vuex actions that result in these errors, and have the store module instantiate and throw the exceptions directly? That keeps the dependency in the store only and slims up the component as no try-catch-rethrow events need to be performed there.

SachaZvetelman (dismissed):

SachaZvetelman (approved):

charliearlie (approved):

PR #913, opened 2021-04-29
f-checkout@0.98.0 - Update checkout issue messages to reflect new copy

charliearlie:
### Changed
- Updated Checkout error messages to reflect latest copy

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
+++ b/packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
@@ -12,7 +12,7 @@
         </h3>
 
         <p data-test-id="checkout-issue-modal-message">
-            {{ $t(`errorMessages.checkoutIssues.${errorCode}.message`) }}
+            {{ $t(`errorMessages.checkoutIssues.${errorCode}.message`, { serviceType }) }}
         </p>
 
         <f-button
--- a/packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
+++ b/packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
@@ -8,11 +8,11 @@
         <h3
             data-test-id="checkout-issue-modal-title"
             class="u-noSpacing">
-            {{ $t(`errorMessages.checkoutIssues.${errorCode}.title`, { serviceType }) }}
+            {{ $t(`errorMessages.checkoutIssues.${errorCode}.title`, { serviceType: serviceTypeText }) }}
         </h3>
 
         <p data-test-id="checkout-issue-modal-message">
-            {{ $t(`errorMessages.checkoutIssues.${errorCode}.message`, { serviceType }) }}
+            {{ $t(`errorMessages.checkoutIssues.${errorCode}.message`, { serviceType: serviceTypeText }) }}
         </p>
 
         <f-button
@@ -52,6 +52,12 @@ export default {
             type: String,
             default: ''
         }
+    },
+
+    data () {
+        return {
+            serviceTypeText: this.$t(`serviceTypes.${this.serviceType}`)
+        };
     },
 
     methods: {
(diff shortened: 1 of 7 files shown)


packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
```diff
@@ -8,7 +8,7 @@
         <h3
             data-test-id="checkout-issue-modal-title"
             class="u-noSpacing">
-            {{ $t(`errorMessages.checkoutIssues.${errorCode}.title`) }}
+            {{ $t(`errorMessages.checkoutIssues.${errorCode}.title`, { serviceType }) }}
```
minrwhite: If I'm reading the code correctly, this is passed in by `Checkout.vue` direct from the store, where it can be one of the `CHECKOUT_METHOD_COLLECTION` or `CHECKOUT_METHOD_DELIVERY` constants.

I think the Error dialog should be interpolating this via a translation string, which will be more important when we roll out to AU, where 'collection' is known as 'pick-up'.

minrwhite (commented):
One observation on this

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
charliearlie: Very good point. I had actually thought of that then forgot during the implementation so I'll change that now! Thanks

packages/components/organisms/f-checkout/src/tenants/en-GB.js
```diff
@@ -51,51 +51,51 @@ const messages = {
...
+                message: 'We\'re sorry, but the minimum spend for your order has not been reached. Please amend your order.',
+                buttonText: 'Back to order'
             },
 
             RESTAURANT_NOT_TAKING_ORDERS: {
-                title: 'We couldn\'t send your order',
-                message: 'The restaurant went offline, so we couldn\'t send your order to them, sorry. Hopefully it\'s only temporary, please try again.',
-                buttonText: 'Retry'
+                title: 'Restaurant not taking orders',
+                message: 'We\'re sorry, but the restaurant is not currently taking orders. Please choose another restaurant to order from.',
```
ashleynolan: You can use `‘` here (and in your other translations) rather than the standard `dumb` apostrophe. Saves escaping it and it's also better typographically when it gets displayed on screen.

See these links if you're interested:
- https://www.presentationzen.com/presentationzen/2009/08/apostrophes-and-quotation-marks.html#:~:text=To%20get%20a%20proper%20apostrophe,by%20the%20best%20of%20them.
- https://www.fastcompany.com/3018353/be-smart-dont-use-dumb-quote-marks

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
zlatinivanov-je: The change is also not reflected in the Changelog

packages/components/organisms/f-checkout/src/tenants/en-GB.js
charliearlie: Yep, I remember you saying. Will update now. Thanks

packages/components/organisms/f-checkout/src/tenants/en-GB.js
```diff
@@ -43,59 +43,59 @@ const messages = {
...
+                message: 'We’re sorry, but the restaurant is not currently taking orders. Please choose another restaurant to order from.',
+                buttonText: 'Ok'
             },
 
             SERVICE_TYPE_UNAVAILABLE: {
-                title: 'We couldn\'t send your order',
-                message: 'It appears the message for this error is not yet added to the spreadsheet',
-                buttonText: 'Retry'
+                title: '{serviceType} unavailable',
+                message: 'We\'re sorry, but the restaurant is not currently taking {serviceType} orders.',
```
minrwhite: If you're changing the others, this should be changed also:
```suggestion
                message: 'We’re sorry, but the restaurant is not currently taking {serviceType} orders.',
```

charliearlie: Changed all of them 🙂

ashleynolan (approved):

minrwhite (approved):

PR #916, opened 2021-04-30
fozzie-components@v3.30.1 - Add error handling and removed extra `spec` folder  #globalconfig

LTurns:
Sorry for the long PR - lots of repetitive changes 😄 

## Added
- Error handling added for `forEach` component tests  - https://www.npmjs.com/package/mocha-each#string-title
- General linting changes and test-cleanup

## Removed
- Unneeded extra intermediate `spec` folder in `test/spec/component`


JoshuaNg2332 (approved):

DamianMullins (approved):

packages/components/organisms/f-header/test/component/f-header.component.mobile.spec.js
```diff
@@ -40,6 +42,7 @@ describe('Mobile - f-header component tests - @browserstack', () => {
             // Act
             header.open(headerData);
             header.waitForComponent();
+            console.log(link);
```
minrwhite, line 45: Any particular reason for this log?

packages/components/organisms/f-header/test/component/f-header.component.mobile.spec.js
```diff
@@ -82,6 +85,7 @@ describe('Mobile - f-header component tests - @browserstack', () => {
             header.open(headerData);
             header.openMobileNavigation();
             header.waitForComponent();
+            console.log(link);
```
minrwhite, line 88: As above - any particular reason for this log?

minrwhite (commented):
Really good that you've taken the opportunity here to reconcile the import order for the spec files, and add printf formatting placeholders for the test cases here.

I've a couple observations below, and a couple of requests:

1. Move the spec files in the tools/generator-component package, so that we don't get any more components with this file structure after you having done this work.

2. I would also like if you could remove all the formerly named `specs/component/README.md` files (barring the one in the generator), as they seem to be just a guide for someone creating a new component, rather than any useful information about an existing one.

LTurns:
> Really good that you've taken the opportunity here to reconcile the import order for the spec files, and add printf formatting placeholders for the test cases here.
> 
> I've a couple observations below, and a couple of requests:
> 
> 1. Move the spec files in the tools/generator-component package, so that we don't get any more components with this file structure after you having done this work.
> 2. I would also like if you could remove all the formerly named `specs/component/README.md` files (barring the one in the generator), as they seem to be just a guide for someone creating a new component, rather than any useful information about an existing one.

Thank you, yeah hopefully all the placeholders will be handy! 

Ah yeah good point on the generator and definitely agree that the README's should be removed in the component folders. I'll do these changes next in a separate PR because this one is already so lengthy! 😄

packages/components/organisms/f-header/test/component/f-header.component.mobile.spec.js
LTurns: my thinking here was that we need to have some sort of error handling for the navigation links - we're not using the `it` statement in this forEach so I thought a console.log might be the easiest way to do this for now? Let me know what you think though

minrwhite (approved):

PR #918, opened 2021-04-30
f-http@v0.3.0 - Support Testing (Unit + Integration types)

tommcclean:
> Learn more about the [new testing setup by referring to the readme](https://github.com/tommcclean/fozzie-components/blob/f-http%40v0.3.0/packages/services/f-http/README.md#unit-testing-guidance)

**F-Http** 0.3.0 makes the module easier to use for testing purposes by...
1. Providing guidance on how to mock the context in unit tests and integration tests
2. Providing the ability to mock HTTP responses using the new mock factory.
3. Doesn't expose axios at all to underlying clients, so if we want to change this in future, we can change it in one place (F-Http)

I have been working with existing components and replacing the testing mechanisms with these; so it is a seamless transition in all cases I can see thus far.

Also added support for a default content type; as most components were applying this per HTTP request.

Coming up in the next PR
- Will update F-Registration to use F-Http instead of Axios and AxiosMockAdapter.
- Will add the support for PATCH, PUT, DELETE methods


packages/services/f-http/src/mocks/mockClient.js
```diff
@@ -0,0 +1,38 @@
...
+/**
+    * Create a httpClient with mocked responses
+    * @param {object} url - The HTTP method for the mocked call (IE: GET)
+    * @param {object} url - The URL the test will mimic
+    * @param {object} requestdata - The request payload if applicable
+    * @param {object} statusCode - The status code to return
+    * @param {object} data - The data to return
+    * @return {object} - Returns an object with mocked restful request methods
+*/
+export default (method, url, requestData, statusCode, data = {}) => {
```
tommcclean, line 16: Improve this sig

packages/services/f-http/src/mocks/mockClient.js
```diff
@@ -0,0 +1,38 @@
...
+
+        case httpVerbs.METHOD_POST:
+            axiosMock.onPost(url, requestData).reply(statusCode, data);
+            break;
+
+        default:
+            throw new Error('Method not provided');
+    }
+
+    const httpClient = createClient();
```
tommcclean, line 32: Should still accept opts

packages/services/f-http/README.md
```diff
@@ -104,11 +104,43 @@ export default {
...
+
+### Integration Testing Guidance
+F-Http exposes a way to create a mockClient, this enables you to mock the underlying HTTPProvider with responses.
+
+**Todo** Support multiple responses from a single test => next update
+
+```js
+import fHttp from '@justeat/f-http';
+
+const $http = fHttp.createMockClient(METHOD_POST, propsData.createAccountUrl, CONSUMERS_REQUEST_DATA, 201);
```
tommcclean, line 128: Cleanup

packages/services/f-http/README.md
```diff
@@ -104,11 +104,43 @@ export default {
...
+      get: jest.fn()
+    }
+  }
+});
+```
+
+### Integration Testing Guidance
+F-Http exposes a way to create a mockClient, this enables you to mock the underlying HTTPProvider with responses.
+
+**Todo** Support multiple responses from a single test => next update
```
tommcclean, line 123: Do in this version

packages/services/f-http/src/httpVerbs.js
```diff
@@ -0,0 +1,7 @@
+const METHOD_GET = 'GET';
```
SachaZvetelman: Should we just call them `GET` and `POST` instead of `METHOD_XXXXX`?

SachaZvetelman (dismissed):
One minor comment.

packages/services/f-http/src/httpVerbs.js
tommcclean: I wanted to do this originally; but DELETE is a keyword in Javascript :D

SachaZvetelman: I thought `delete` was, but didn't know `DELETE` was! 🤯

tommcclean: Hmm, good point; let me check

SachaZvetelman (approved):

FayeCarter (approved):

PR #926, opened 2021-05-05
f-services@v1.11.0 - Revert UK Tel. Regex back to original 1.9.0 version

oliversweb:
v1.11.0
*May 5, 2021*
### Changed

- Reverted Tel. no. regex for en-GB back to 1.9.0 version


packages/services/f-services/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v1.11.0
+------------------------------
+*April 27, 2021*
```
xander-marjoram: 📆

packages/services/f-services/src/validations.js
```diff
@@ -46,7 +46,7 @@ const isValidPostcode = (postcode, locale) => POSTCODE_REGEX[locale].test(postco
 const meetsCharacterValidationRules = value => /^[\u0060\u00C0-\u00F6\u00F8-\u017Fa-zA-Z-' ]*$/.test(value);
 
 const PHONE_REGEX = {
-    'en-GB': /^(\+(44))?[0-9]{10}|(\+(44))?[0-9]{11}$/,
+    'en-GB': /^(\+(44))?[0-9]{10}|^(\+(44))?[0-9]{11}/,
```
xander-marjoram: Does this require any new test cases?

oliversweb: No

xander-marjoram (dismissed):

packages/services/f-services/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v1.11.0
+------------------------------
+*May 5, 2021*
+### Changed
+- Extended Tel. no. regex for en-GB.
```
ashleynolan: Super NIT:

```suggestion
*May 5, 2021*

### Changed
- Extended Tel. no. regex for en-GB.
```

packages/services/f-services/src/validations.js
ashleynolan: What's the difference between the 2 regex's, if we don't need a test to cover what was wrong with the initial pattern?

xander-marjoram (dismissed):

simonsandersje (dismissed):

packages/services/f-services/src/validations.js
minrwhite: The replacement doesn't look right to me either - if I'm reading it correctly it will expect any string as long as it's prefixed by something resembling a valid phone number, where before it would accept either a string *beginning* with a number of 10 digits or a string *ending* with a number of 11 digits. I'd personally go for something like:
```suggestion
    'en-GB': /^(\+(44))?[0-9]{10,11}$/,
```
to avoid the repetition.

JoshuaNg2332 (dismissed):

minrwhite (dismissed):

SachaZvetelman (dismissed):

minrwhite (approved):

SachaZvetelman (approved):

PR #932, opened 2021-05-06
f-braze-adapter@v4.0.0-beta.1 - Support multiple consumers to enable brands to be used again

matthewhardern:
In order to allow brands filtering within content cards component we have had to make adjustments to the braze adapter. The previous version of the adapter applied filtering on brands and enabled card types at a global level, that meant any consumer of braze adapter ( or consumer of the f-content-cards component as it consumes the braze adapter) could potentially have its cards filtered by other consumers of the braze adapter within the same JS environment.

To give an example:

Component 1 consumes the braze adapter and supplies a brands list. Its cards are then returned with only those brands included.

Component 2 then loads and consumes the braze adapter, supplying no brands list. However it still would receive a list of cards with brands filtered out by component 1’s brands in its current form.

SOLUTION

In order to solve this issue this PR changes the way braze adapter works. Firstly when consuming the braze adapter via `BrazeAdapter.initialise` static method, it now returns a new instance of  `BrazeAdapter` each time. On the first request to consume the adapter by any code within the JS environment it creates a `BrazeConsumerRegistry` class and registers the instantiation of the BrazeAdapter as a consumer by storing the config as a BrazeConsumer within the BrazeConsumerRegistry.

On creating the registry it also instantiates the `BrazeDispatcher` class to communicate with the Braze SDK and calls the `BrazeDispatcher` `configure` method. On any subsequent consumption of the `BrazeAdapter`  within the same JS environment it only registers this instantiation in the registry of consumers within `BrazeConsumerRegistry`. 

On configuration of the `BrazeDispatcher` it makes a request to the Braze SDK to retrieve cards. Those cards are then published via `DispatcherEventStream`. The corresponding subscription to the content cards publish event is then fired within the `BrazeConsumerRegistry` (this subscription is created during the instantiation of the `BrazeConsumerRegistry` class) calling the content cards callback of any consumer within the `BrazeConsumerRegistry` via the `applyContentCardCallbacks` method.

*BrazeAdapter* - Class used to instantiate the braze adapter

	* Initialise method is called in order to return a BrazeAdapter object
	* Unsubscribe method is used when consumer no longer requires the braze adapter.

*BrazeConsumerRegistry* - Class used to store a registry of code consuming the adapter.

	* subscribes to the `DispatcherEventStream` to retrieve content cards, in app messages and logs.
	* register method registers a consumer of `BrazeAdapter`
	* unregister removes a consumer from the registry

*BrazeDispatcher* - Class used to communicate with the Braze SDK.

	* configure method called to configure the Braze SDK

*BrazeConsumer* - Class used to store the configuration object passed to `BrazeAdapter`.

	* filters cards returned from the Braze SDK as part of the cards setter.

*DispatcherEventStream* - simple pub/sub class to handle events and pass data between classes.

contentCard service has now been removed in favour of separate functions that are called within the `BrazeConsumer` and are now curried via a `pipe` function as part of the filtering process before passing cards back to the consumer as part of the content cards callback.  This solves the problem of immutability as prior to this the original cards array was modified as the result of each class method when chained together.

Unit Tests have been updated and added where needed for the new files.


packages/services/f-braze-adapter/src/services/utils/removeDuplicateContentCards.js
```diff
@@ -0,0 +1,11 @@
+import orderBy from 'lodash.orderby';
```
minrwhite, line 1: I think these duplication rules wanted to be changed for stampcards - might be worth revising these on a later pass to prevent growing this PR inordinately.

packages/services/f-braze-adapter/src/BrazeAdapter.js
```diff
@@ -0,0 +1,55 @@
...
+    get dispatcher () {
+        return this._consumer;
+    }
+
+    constructor () {
+        // create the consumer registry
+        this._consumerRegistry = GetConsumerRegistry();
+        // create the dispatcher
+        this._dispatcher = GetDispatcher(sessionTimeoutInSeconds);
+    }
```
minrwhite: ~IMO, for testability and maintainability the fewer singletons we're using, the better. As a relatively lightweight class, and the entry point to this module, I think this BrazeAdapter class would be better as the singleton, and that it should instantiate a non-singleton registry and dispatcher when it is constructed.~

Actually, strike this, I see what you're doing here. I still think it's worth keeping the number of singletons down if possible, but worth me having another think about how to do that.

matthewhardern: Yes my thought process here was to have a BrazeAdapter class than can be instantiated to store the record of the consumer. The unsubscribe could be called on the instantiated BrazeAdapter object when no longer needed, it prevents the consumer of BrazeAdapter having to worry about the concept of consumers or registry. However I agree if we can figure out a way of having a less singletons then it would be better.

matthewhardern: One option would be to store a global variable for the consumers array on the `window` object, which would remove the need for the BrazeConsumerRegistry to be a singleton.

matthewhardern: but not a great fan of opening the consumers array on a global level

matthewhardern: Would another option would be to do `BrazeConsumerRegistry.prototype.consumers = [];` then consumers would be shared across instances?

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
```diff
@@ -0,0 +1,114 @@
...
+     * the returned cards. Once the filters have been applied then it calls the callbacks.
+     * @returns {(function(*): void)[]}
+     */
+    getContentCardCallbacks () {
+        const _this = this;
+        // eslint-disable-next-line func-names
+        return Object.keys(this.callbacks).map(key => function (cards) {
+            _this.cards = cards;
+            _this.callbacks[key](_this._cards);
+        });
```
minrwhite: You can switch this around with the arrow function inside the keyword function, you can use the [`thisArg` map parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#parameters) to avoid having to reassign `this` to `_this`. I'd also say that naming functions would be better for traversing stack traces for debugging:
```suggestion
        return Object.keys(this.callbacks).map(function mapContentCardCallbacks(key) {
            return (cards) => {
                this.cards = cards;
                this.callbacks[key](this._cards);
            }
        }, this);
```

matthewhardern: Thanks good suggestion

packages/services/f-braze-adapter/src/BrazeAdapter.js
matthewhardern: We could actually make all the methods static on the BrazeConsumerRegistry leaving us with one copy of `consumers` rather than a new one per instance?

matthewhardern:
```
class BrazeConsumerRegistry {
    /**
     * Static constructor to store one instance of BrazeConsumerRegistry per js env
     * @return {BrazeConsumerRegistry}
     * @constructor
     */
    static consumers = [];

    /**
     * For each consumers content card callbacks fire the callback with supplied cards
     * @param cards
     */
    static applyContentCardCallbacks (cards) {
        this.consumers.reduce((acc, consumer) => [...acc, consumer.getContentCardCallbacks()])
            .forEach(callback => {
                callback(cards);
            });
    }

    /**
     * For each consumers in app message callbacks fire the callback with the supplied message
     * @param message
     */
    static applyInAppMessageCallbacks (message) {
        this.consumers.reduce((acc, consumer) => [...acc, consumer.getInAppMessagesCallbacks()])
            .forEach(callback => {
                callback(message);
            });
    }

    /**
     * For each consumers in app message click callbacks fire the callback with the supplied message
     * @param message
     */
    static applyInAppMessageClickEventsCallbacks (message) {
        this.consumers.reduce((acc, consumer) => [...acc, consumer.getInAppMessagesCallbacks()])
            .forEach(callback => {
                callback(message);
            });
    }

    /**
     * Register a consumer instance in the registry
     * @param consumerOptions
     * @returns {BrazeConsumer}
     */
    static register (consumerOptions) {
        const consumer = new BrazeConsumer(consumerOptions);
        this.consumers = [...this.consumers, consumer];
        return consumer;
    }

    /**
     * Remove the passed consumer from the registry
     * @param consumer {BrazeConsumer}
     */
    static unregister (consumer) {
        this.consumers = [...this.consumers.filter(c => c !== consumer)];
    }
}

export default BrazeConsumerRegistry;
```

packages/services/f-braze-adapter/src/services/BrazeConsumerRegistry.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * Register a consumer instance in the registry
+     * @param consumerOptions
+     * @returns {BrazeConsumer}
+     */
+    register (consumerOptions) {
+        const consumer = new BrazeConsumer(consumerOptions);
+        this.consumers = [...this.consumers, consumer];
```
minrwhite: At the risk of premature optimization, reassignment after a destructure seems overkill here, and below within `unregister()`. FWIW, I'd use `.push()` and `.splice()`.

packages/services/f-braze-adapter/src/BrazeAdapter.js
matthewhardern:
```
import BrazeConsumer from './BrazeConsumer';

class BrazeConsumerRegistry {

    /**
     * For each consumers content card callbacks fire the callback with supplied cards
     * @param cards
     */
    static applyContentCardCallbacks (cards) {
        this.consumers.reduce((acc, consumer) => [...acc, consumer.getContentCardCallbacks()])
            .forEach(callback => {
                callback(cards);
            });
    }

    /**
     * For each consumers in app message callbacks fire the callback with the supplied message
     * @param message
     */
    static applyInAppMessageCallbacks (message) {
        this.consumers.reduce((acc, consumer) => [...acc, consumer.getInAppMessagesCallbacks()])
            .forEach(callback => {
                callback(message);
            });
    }

    /**
     * For each consumers in app message click callbacks fire the callback with the supplied message
     * @param message
     */
    static applyInAppMessageClickEventsCallbacks (message) {
        this.consumers.reduce((acc, consumer) => [...acc, consumer.getInAppMessagesCallbacks()])
            .forEach(callback => {
                callback(message);
            });
    }

    /**
     * Register a consumer instance in the registry
     * @param consumerOptions
     * @returns {BrazeConsumer}
     */
    static register (consumerOptions) {
        const consumer = new BrazeConsumer(consumerOptions);
        this.consumers.push(consumer);
        return consumer;
    }

    /**
     * Remove the passed consumer from the registry
     * @param consumer {BrazeConsumer}
     */
    static unregister (consumer) {
        this.consumers.splice(this.consumers.findIndex(c => c === consumer), 1);
    }
}

BrazeConsumerRegistry.consumers = [];

export default BrazeConsumerRegistry;

```

Above might be a better approach?

packages/services/f-braze-adapter/src/services/_tests/BrazeDispatcher.test.js
```diff
@@ -434,23 +332,23 @@ describe('BrazeDispatcher operation', () => {
...
-
-                it('should log the error to the provided callback', async () => {
-                    // Act
-                    await interceptInAppMessagesHandler(message);
-
-                    // Assert
-                    expect(loggingCallback).toHaveBeenCalledWith(LOG_ERROR, `Error handling message - ${error}`, { message, error });
-                });
-            });
+            // describe('when showInAppMessage throws an error', () => {
```
minrwhite: [`xdescribe()`/`describe.skip()`](https://jestjs.io/docs/api#describeskipname-fn) is your friend when temporarily disabling tests 😄

packages/services/f-braze-adapter/src/services/BrazeConsumerRegistry.js
```diff
@@ -0,0 +1,113 @@
...
+        this.eventStream = new DispatcherEventStream();
+
+        this._dispatcher = new BrazeDispatcher(sessionTimeout, this.eventStream);
+
+        // TODO change this so we call all configure bits inside the dispatcher constructor
+        this._dispatcher.configure({
+            apiKey,
+            userId,
+            enableLogging
+        });
```
minrwhite: Note configure is `async` at this point, so we either need to `await` it (and catch any errors/rejections) - or attach a `.catch()` to the resulting promise, to avoid having the potential for an unhandled promise rejection.

matthewhardern: resolved now by using the new BrazeAdapter initialize static function, which returns the braze adapter instance.

packages/services/f-braze-adapter/src/services/utils/removeDuplicateContentCards.js
matthewhardern: will do on another pass, worth raising another ticket

packages/services/f-braze-adapter/src/services/utils/filterByEnabledCardTypes.js
```diff
@@ -1,3 +1,8 @@
 const filterByEnabledCardTypes = (cards, enabledCardTypes) => cards
-    .filter(({ type }) => (type ? enabledCardTypes.includes(type) : false));
+    .filter(({ type }) => {
+        if (enabledCardTypes.length > 0) {
+            return !type || enabledCardTypes.includes(type);
```
minrwhite: If I'm reading this correctly, this will pass any card object that doesn't have a `type` field? Based on the previous condition I would have expected:
```suggestion
            return type && enabledCardTypes.includes(type);
```

matthewhardern: yes correct sorry, was doing this after brands. I will make the change

packages/services/f-braze-adapter/src/services/utils/filterByBrandName.js
```diff
@@ -1,2 +1,7 @@
-const filterByBrands = (cards, brands) => cards.filter(({ brand }) => !brand || brands.includes(brand));
+const filterByBrands = (cards, brands) => cards.filter(({ brand }) => {
+    if (brands.length > 0) {
+        return !brand || brands.includes(brand);
```
minrwhite, line 8: Ah, the condition of this form makes more sense from a brand perspective - if a card isn't relevant to a brand then it should pass the filter, which is slightly different business logic than that for: `enabledCardTypes` 👍🏽

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
```diff
@@ -0,0 +1,147 @@
...
+        this._cards = [];
+
+        this.brands = [];
+
+        // Here we just resolve a promise on brand to allow passing in a promise that resolves to a array of strings
+        // and also accepts a array strings
+        Promise.resolve(brands).then(brandsList => {
+            this.brands = uniq([...this.brands, ...brandsList]);
+            // check to see if cards has already been set if so re call all the content card callbacks
+            if (this._cards.length > 0) {
```
kevinrodrigues: Below should be enough?

`if (this._cards.length)`

packages/services/f-braze-adapter/src/services/_tests/DispatcherEventStream.test.js
```diff
@@ -0,0 +1,60 @@
...
+        dispatcherEventStream.subscribe(mockEventName2, mockCallback);
+
+        // Act
+        dispatcherEventStream.publish(mockEventName, mockData);
+
+        // Assert
+        expect(mockCallback).not.toHaveBeenCalled();
+    });
+
+    it('should throw and error if the channel being published to is no in the list of subscriptions', () => {
```
kevinrodrigues: `Not`

packages/services/f-braze-adapter/src/services/utils/_tests/sortByCardOrder.test.js
```diff
@@ -0,0 +1,56 @@
...
+        // Arrange
+        expect(orderedCards).toEqual(expected);
+    });
+
+    it('should put cards with NO order key at the back', () => {
+        // Act
+        const orderedCards = sortByCardOrder(mockCards);
+
+        // Arrange
+        expect(orderedCards[3]).toEqual({ id: 4 });
```
kevinrodrigues: Minor, but `orderedCards.slice(-1).pop()` may work better incase the `mockCards` changed in length in future.

packages/services/f-braze-adapter/src/tests/BrazeAdapter.test.js
```diff
@@ -0,0 +1,163 @@
...
+            // Act
+            await brazeAdapter.logCardImpressions([rawCards[0].id, rawCards[1].id]);
+
+            // Assert
+            expect(brazeAdapter.dispatcher.logCardImpressions).toHaveBeenCalledWith([rawCards[0].id, rawCards[1].id]);
+        });
+
+        describe('unsubscribe', () => {
+            it('should unregister the consumer that this braze adapter instance created', async () => {
+
```
kevinrodrigues, line 164: Empty?

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
matthewhardern: yep thanks

packages/services/f-braze-adapter/src/services/_tests/DispatcherEventStream.test.js
matthewhardern: thanks

packages/services/f-braze-adapter/src/services/utils/_tests/sortByCardOrder.test.js
matthewhardern: agreed makes sense

packages/services/f-braze-adapter/src/tests/BrazeAdapter.test.js
matthewhardern: yep my bad thanks

packages/services/f-braze-adapter/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-braze-adapter",
   "description": "Fozzie Metadata Service",
-  "version": "4.0.0-beta-1",
+  "version": "4.0.0-beta.1",
```
BanksySan, line 4: Does changing this have any functional difference or is it just personal preference?

packages/services/f-braze-adapter/src/BrazeAdapter.js
```diff
@@ -0,0 +1,110 @@
+import GetConsumerRegistry from './services/BrazeConsumerRegistry';
+
+export default class BrazeAdapter {
+    /**
+     * Initializes the braze adapter and returns instance of the braze adapter class
```
BanksySan, line 5: Should this read _... and returns *the* single instance of the braze adapter class_?

packages/services/f-braze-adapter/src/BrazeAdapter.js
```diff
@@ -0,0 +1,110 @@
...
+        await this.dispatcher.logCardClick(cardId);
+    }
+
+    /**
+     * Reports card impressions interaction to Dispatcher
+     * TODO use the dispatcher event stream to publish events
+     * @param cardIds
+     */
+    async logCardImpressions (cardIds) {
+        await this.dispatcher.logCardImpressions(cardIds);
```
BanksySan: Do we want to await here?  can we return immediately after this is called?

BanksySan: How do we handle errors here?

packages/services/f-braze-adapter/src/services/BrazeConsumerRegistry.js
```diff
@@ -0,0 +1,132 @@
...
+
+    /**
+     * Remove the passed consumer from the registry
+     * @param consumer {BrazeConsumer}
+     */
+    unregister (consumer) {
+        if (this.consumers.findIndex(c => c === consumer) !== -1) {
+            this.consumers.splice(this.consumers.findIndex(c => c === consumer), 1);
+        } else {
+            throw new Error("Can't find consumer in registry");
```
BanksySan: Could the message be a bit more descriptive?  e.g. _Failed to unregister consumer, consumer not in registry_.

packages/services/f-braze-adapter/src/index.js
```diff
@@ -1,22 +1,4 @@
-import GetDispatcher from './BrazeDispatcher';
+import BrazeAdapter from './BrazeAdapter';
 
-/**
- * sessionTimeoutInSeconds
- * Set session timeout to 0 in order to avoid caching issues with Braze
- * @type {number}
- */
-export const sessionTimeoutInSeconds = 0;
+export default BrazeAdapter.initialise;
```
BanksySan, line 3: Should we await here?  Is there a risk that we could return an as-yet uninitialised adapter?

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
```diff
@@ -0,0 +1,147 @@
...
+     * @param cards
+     */
+    set cards (cards) {
+        this._cards = this.applyFilterMethods(cards);
+    }
+
+    constructor ({
+        enabledCardTypes,
+        brands = [],
+        callbacks,
```
BanksySan: Callbacks should default to an empty array.

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
```diff
@@ -0,0 +1,147 @@
...
+     * Setter that is called when cards are set on the class, applies any filter methods on those cards and
+     * sets the cards variable
+     * @param cards
+     */
+    set cards (cards) {
+        this._cards = this.applyFilterMethods(cards);
+    }
+
+    constructor ({
+        enabledCardTypes,
```
BanksySan: Should default to an empty array.

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
```diff
@@ -0,0 +1,147 @@
...
+        brands = [],
+        callbacks,
+        interceptInAppMessages = {},
+        interceptInAppMessageClickEvents = {},
+        loggerCallbacks = {},
+        customFilters = []
+    }) {
+        this.defaultEnabledCardTypes = [
+            'Home_Promotion_Card_1',
```
BanksySan: Can these be shifted to some shared, global place?  Maybe the `types` folder?

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
```diff
@@ -0,0 +1,147 @@
...
+
+        this.optionsChecks.forEach(check => {
+            Object.keys(check).forEach(key => {
+                if (!check[key]) {
+                    throw new InvalidConsumerConfigError(key);
+                }
+            });
+        });
+
+        this.enabledCardTypes = enabledCardTypes || this.defaultEnabledCardTypes;
```
BanksySan, line 64: Can these be written more succinctly with rest operators?

packages/services/f-braze-adapter/src/services/BrazeDispatcher.js
```diff
@@ -42,10 +42,9 @@ function interceptInAppMessagesHandler (message) {
             appboy.display.showInAppMessage(message);
         })
         .catch(error => {
-            this.logger('error', `Error handling message - ${error}`, { message, error });
+            dispatcherEventStream.publish(LOGGER, ['error', `Error handling message - ${error}`, { message, error }]);
```
BanksySan, line 45: Has `'Error handling message - '` been left in by mistake?

packages/services/f-braze-adapter/src/services/BrazeDispatcher.js
```diff
@@ -54,67 +53,30 @@ function interceptInAppMessagesHandler (message) {
...
-        }
-        return dispatcherInstance;
-    }
-
     /**
      * Sets defaults, checks environment, checks for reinstantiation
      * @param {Number} sessionTimeoutInSeconds
      */
     constructor (sessionTimeoutInSeconds) {
         if (typeof window === 'undefined') throw new Error('window is not defined');
```
BanksySan, line 74: When would this happen?

packages/services/f-braze-adapter/src/services/BrazeDispatcher.js
```diff
@@ -138,45 +102,20 @@ class BrazeDispatcher {
...
-                userId,
-                enabledCardTypes
+                userId
             };
-        } else if (!(apiKey === this.dispatcherOptions.apiKey
-            && userId === this.dispatcherOptions.userId
-            && JSON.stringify(enabledCardTypes.slice().sort())
-                === JSON.stringify(this.dispatcherOptions.enabledCardTypes.slice().sort()))) {
+        } else if (!(apiKey === this.dispatcherOptions.apiKey && userId === this.dispatcherOptions.userId)) {
             throw new Error('attempt to reinitialise appboy with different parameters');
```
BanksySan: Capitol `a`.

packages/services/f-braze-adapter/src/services/DispatcherEventStream.js
```diff
@@ -0,0 +1,26 @@
+
+class DispatcherEventStream {
+    constructor () {
+        this.subscribers = {};
+    }
+
+    publish (eventName, data) {
+        if (!Array.isArray(this.subscribers[eventName])) {
+            throw new Error(`No event subscribers exist for the following event: ${eventName}`);
```
BanksySan: This shouldn't be an error as the client shoudn't know/care whether there are any subscribers.

packages/services/f-braze-adapter/src/services/_tests/BrazeConsumer.test.js
```diff
@@ -0,0 +1,133 @@
...
+    brands: [
+        'a',
+        'b',
+        'c',
+        'd'
+    ]
+};
+
+describe('BrazeConsumer', () => {
+    it.each`
```
BanksySan, line 62: Is this a typo or a template literal?

packages/services/f-braze-adapter/src/services/_tests/BrazeConsumer.test.js
```diff
@@ -0,0 +1,133 @@
...
+
+    it('should return an array of callbacks from getInAppMessagesCallbacks method', () => {
+        // Arrange
+        const consumer = new BrazeConsumer(mockConsumerOptions);
+
+        // Act
+        const callbacks = consumer.getInAppMessagesCallbacks();
+
+        // Assert
+        expect(callbacks).toEqual(expect.arrayContaining([expect.any(Function)]));
```
BanksySan: Can we check that the correct callbacks are returned?

packages/services/f-braze-adapter/src/services/_tests/BrazeConsumer.test.js
```diff
@@ -0,0 +1,133 @@
...
+
+    it('should return an array of callbacks from getLoggerCallbacks method', () => {
+        // Arrange
+        const consumer = new BrazeConsumer(mockConsumerOptions);
+
+        // Act
+        const callbacks = consumer.getLoggerCallbacks();
+
+        // Assert
+        expect(callbacks).toEqual(expect.arrayContaining([expect.any(LogService)]));
```
BanksySan, line 131: Can we check that the correct callbacks are returned?

BanksySan (changes requested):

packages/services/f-braze-adapter/src/BrazeAdapter.js
BanksySan: Not actually a singleton.  Ignore me

packages/services/f-braze-adapter/src/index.js
BanksySan: This exports the method, not the returned value.

packages/services/f-braze-adapter/src/services/_tests/BrazeConsumer.test.js
BanksySan: It's a template literal, it's already a pattern in use though so we'll leave it and have a discussion on their use later.

packages/services/f-braze-adapter/src/BrazeAdapter.js
matthewhardern: this has now been removed

packages/services/f-braze-adapter/src/services/BrazeConsumerRegistry.js
matthewhardern: replaced with your good suggestion

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
matthewhardern: default empty object

packages/services/f-braze-adapter/src/services/BrazeDispatcher.js
matthewhardern: changed

packages/services/f-braze-adapter/src/services/BrazeDispatcher.js
matthewhardern: no its part of the message

packages/services/f-braze-adapter/src/services/BrazeDispatcher.js
matthewhardern: during SSR

packages/services/f-braze-adapter/src/services/DispatcherEventStream.js
matthewhardern: removed as suggested

packages/services/f-braze-adapter/src/services/_tests/BrazeConsumer.test.js
matthewhardern: changed to correct

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
matthewhardern: I not sure, if your thinking of spread operator then no as it needs to be one or the other.

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
matthewhardern: done

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
matthewhardern: agreed

packages/services/f-braze-adapter/src/services/_tests/BrazeConsumer.test.js
matthewhardern: A callback is not returned an object containing callbacks is returned and the type is passed inside of Braze Consumer Registry to fire the correct callback

packages/services/f-braze-adapter/src/BrazeAdapter.js
matthewhardern: it returns a boolean true false from braze sdk so the caller of this function would check for this returned value as it does not throw errors.

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
BanksySan: I was thinking of that.

BanksySan (approved):

minrwhite (approved):

PR #935, opened 2021-05-06
f-http@v0.4.0 - Support all HTTP methods

tommcclean:
I think this is the last PR I need to enable F-Registration to consume this package
Should also put us in a good place to start collecting dependency stats

v0.4.0
*May 6, 2021*

### Added
- Expose HTTPVerbs for PUT, PATCH, DELETE
- Expose methods for PUT, PATCH, DELETE on client
- Expose a method for setting authorisation token for all requests from a client
- Ability to name a client - useful when we can push stats and logs by API or grouping name

### DIFF
--- a/packages/services/f-http/README.md
+++ b/packages/services/f-http/README.md
@@ -106,7 +106,7 @@ export default {
 ```
 
 ### Alternative Implementation
-If you would rather create the HTTPClient when you use it, thats fine too; its just means it can't be reused as easily and you will need to filter the configuration options down to the component.
+If you would rather create the HTTPClient when you use it, that's fine too; it just means it can't be reused as easily and you will need to filter the configuration options down to the component.
 
 ```js
 export default {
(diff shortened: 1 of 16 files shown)


packages/services/f-http/src/createClient.js
```diff
@@ -32,13 +30,59 @@ const get = async (resource, headers = {}) => {
  * @param {object} headers - Any additional request headers you want to provide
  * @return {object} - Returns data from response
  */
-const post = async (resource, body, headers = {}) => {
+const postResource = async (resource, body, headers = {}) => {
+    try {
+        const response = await _axiosInstance.post(resource, body, configBuilder(headers));
```
tommcclean, line 35: Just gonna try and reduce duplication here

packages/services/f-http/README.md
```diff
@@ -144,3 +181,19 @@ baseUrl | Ensure all requests from this client use a relative url | string | ''
...
+## Client Methods
+These are all of the methods exposed by the httpClient
+
+Method | Description | Parameters
+------------- | ------------- | -------------
+get | Perform a GET request for a resource | resource URL _[string]_, headers _[array]_
+post | POST a resource | resource URL _[string]_, body _[object]_, headers _[array]_
+put | Perform a PUT on a resource | resource URL _[string]_, body _[object]_, headers _[array]_
+patch | Perform a PATCH on a resource | resource URL _[string]_, body _[object]_, headers _[array]_
+get | DELETE a resource | resource URL _[string]_, headers _[array]_
```
dandel10n:
```suggestion
delete | DELETE a resource | resource URL _[string]_, headers _[array]_
```

packages/services/f-http/src/requestDispatcher.js
```diff
@@ -0,0 +1,49 @@
...
+    } catch (error) {
+        return handleError(error, _configuration.errorCallback);
+    }
+};
+
+/**
+ * Send a request with a request body
+ * @param {string} method - The HTTP Method of the request (post, put, patch)
+ * @param {string} resource - The resource to get (URL)
+ * @param {object} headers - Any additional request headers you want to provide
```
dandel10n, line 29: Should here be body parameter as well?

packages/services/f-http/README.md
```diff
@@ -102,6 +105,39 @@ export default {
 }
 ```
 
+### Alternative Implementation
+If you would rather create the HTTPClient when you use it, thats fine too; its just means it can't be reused as easily and you will need to filter the configuration options down to the component.
```
dandel10n:
```suggestion
If you would rather create the HTTPClient when you use it, that's fine too; it's just means it can't be reused as easily and you will need to filter the configuration options down to the component.
```

dandel10n (commented):
Few tiny comments but all looks good 👍

simonsandersje (approved):

dandel10n (approved):

PR #940, opened 2021-05-11
f-checkout@v0.109.0 - (a) - Accessibility added for error messages   

LTurns:
Error messages were previously not being read in Checkout. Ticket was to make error handling more accessible, using a similar approach to that in [f-registration](https://github.com/justeat/fozzie-components/blob/master/packages/components/organisms/f-registration/src/components/Registration.vue).

### Added

- `aria-describedby` and `aria-invalid` for all input fields - to read error messages
- `error-summary-section` in checkout to alert of errors after user submits - resembles f-registration


packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -13,10 +13,13 @@
                 label-style="inline"
                 is-grouped
                 :has-error="isAddressLine1Empty"
+                aria-describedby="line1-error"
```
SachaZvetelman, line 16: I'm not an expert in accessibility, unfortunately, but is this correct? If my understanding is correct, we're basically saying that the error message is what describes the form field, and that doesn't sounds correct to me.
cc @ross-gratton

packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -40,10 +43,13 @@
             name="address-locality"
             :label-text="$t('labels.locality')"
             :has-error="isAddressLocalityEmpty"
+            aria-describedby="locality-error"
+            @blur="isFieldValid('locality')"
```
SachaZvetelman: @oliversweb is working on a similar ticket, to see when we trigger the validation exactly. Can the two of you please check you're aligned on this, to ensure we're not doing things twice or one PR will override what the other did? Thanks!

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -36,22 +36,38 @@
                     :class="$style['c-checkout-form']"
                     @submit.prevent="onFormSubmit"
                 >
-                    <guest-block v-if="!isLoggedIn" />
+                    <section
+                        id="error-summary-container"
+                        :class="$style['is-visuallyHidden']"
```
SachaZvetelman: In this case, shouldn't we be using `class="is-visuallyHidden"` instead?

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -791,6 +809,10 @@ export default {
                 data: { ...this.eventData, validationState },
                 logMethod: this.$logger.logWarn
             });
+
+            this.genericFormErrorMessage = errorCount === 1 ?
+                'There is 1 error in the form.' :
+                `There are ${errorCount} errors in the form.`;
```
SachaZvetelman: This needs to be coming from translations. And I'd probably extract this into a computed or similar.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -926,4 +953,9 @@ export default {
 .c-checkout-submitButton {
     margin: spacing(x4) 0 spacing(x0.5);
 }
+
+.c-checkout-genericError {
+        margin-top: 0;
+        margin-bottom: spacing(x2);
+}
```
SachaZvetelman: Nit: indentation looks odd.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -809,9 +831,14 @@ export default {
          */
         isFormValid () {
             this.$v.$touch();
+            this.genericFormErrorMessage = '';
             return !this.$v.$invalid;
         },
 
+        isFieldValid () {
+            return this.$v.customer.mobileNumber.$touch();
```
SachaZvetelman: `isFieldValid` sounds very generic but then in the implementation we're only actioning on `mobileNumber`. Something doesn't look right 😬

SachaZvetelman (commented):
It looks like we're doing a lot of "manual" steps to handle the validations, blur, touch, etc, and I'm not entirely sure we need to do that. I might be completely wrong, but I just want to double-check this.

packages/components/organisms/f-checkout/src/components/Address.vue
xander-marjoram: It sounds reasonable to me, assuming the error message only exists on the DOM when there is an error.

packages/components/organisms/f-checkout/src/mixins/validations.mixin.js
```diff
@@ -10,6 +10,9 @@ export default {
         */
         isFieldEmpty (type, field) {
             return this.$v[type][field].$dirty && !this.$v[type][field].required;
+        },
+        focusField (type, field) {
```
xander-marjoram: This name doesn't quite match what it does, I would call it `touchField` (calls the touch method) or `dirtyField` (makes the field dirty).

packages/components/organisms/f-checkout/src/components/Guest.vue
```diff
@@ -103,7 +113,11 @@ export default {
     methods: {
         ...mapActions(VUEX_CHECKOUT_MODULE, [
             'updateCustomerDetails'
-        ])
+        ]),
+
+        isFieldValid (field) {
```
xander-marjoram: I would expect a method with this name to return a boolean, but it doesn't look like it does that. Is this method needed?

packages/components/organisms/f-checkout/src/mixins/validations.mixin.js
xander-marjoram: I think in f-registration there is a method called `onFieldBlur` which might be a good way to replace this method and also `isFieldValid` above

packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -56,16 +62,20 @@
...
                 <error-message
                     v-if="isAddressPostcodeEmpty"
+                    id="postcode-error"
                     data-js-error-message
                     data-test-id="error-address-postcode-empty">
                     {{ $t('validationMessages.postcode.requiredError') }}
                 </error-message>
                 <error-message
                     v-else-if="!isAddressPostcodeValid"
+                    id="postcode-error"
```
xander-marjoram, line 78: Question rather than suggestion: would it work if you moved the shared attributes like `id` and `data-js-error-message` to the `template` tag?

packages/components/organisms/f-checkout/src/components/Address.vue
LTurns: from what I've read, `aria-describedby` is used for error messages quite frequently in accessibility - it's a way of connecting the input field with the corresponding id in the error message  - so that if the error message shows, it also reads out that information. Some forms also add `aria-invalid=true` to the input field so I could do that too. In most cases, from what I've seen, either `aria-describedby` or `aria-labelledby` is used though

packages/components/organisms/f-checkout/src/mixins/validations.mixin.js
LTurns: ah the `formFieldBlur` function? I actually had it like that originally but it ended up being a bit messier than the current solution - I might take another look this morn though actually, now I'm understanding things a bit more 😄

packages/components/organisms/f-checkout/src/components/Checkout.vue
LTurns: yep definitely agree - looking into this now! 😄

packages/components/organisms/f-checkout/src/components/Address.vue
LTurns: just tried and it doesn't work sadly - looks like they both need to be after the two conditions

packages/components/organisms/f-checkout/src/components/Address.vue
oliversweb: We have spoken and as this will hopefully be merged before mine this code rules and I will take mine out

ashleynolan:
@LTurns Super small thing – but is it possible to embed the video referenced rather than including a link to JIRA? InfoSec don't like us linking to internal tooling on public repos.

LTurns:
> @LTurns Super small thing – but is it possible to embed the video referenced rather than including a link to JIRA? InfoSec don't like us linking to internal tooling on public repos.

ahh yeah thank you, will upload now! :)

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -36,22 +36,39 @@
                     :class="$style['c-checkout-form']"
                     @submit.prevent="onFormSubmit"
                 >
-                    <guest-block v-if="!isLoggedIn" />
+                    <section
+                        id="error-summary-container"
+                        :class="$style['is-visuallyHidden']"
```
ashleynolan: Not 100% sure on this, but as these styles come from fozzie and are shared, should this be: (/cc @dandel10n )

```suggestion
                        :class="is-visuallyHidden"
```

simonsandersje (dismissed):

JoshuaNg2332 (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -36,22 +36,39 @@
                     :class="$style['c-checkout-form']"
                     @submit.prevent="onFormSubmit"
                 >
-                    <guest-block v-if="!isLoggedIn" />
+                    <section
+                        id="error-summary-container"
```
SachaZvetelman: Do we need this `id`?

packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -129,7 +142,10 @@ export default {
     methods: {
         ...mapActions(VUEX_CHECKOUT_MODULE, [
             'updateAddressDetails'
-        ])
+        ]),
+        onAddressFieldBlur (field) {
```
SachaZvetelman: Super nit: empty line between methods, please :)

packages/components/organisms/f-checkout/src/components/Guest.vue
```diff
@@ -103,7 +116,11 @@ export default {
     methods: {
         ...mapActions(VUEX_CHECKOUT_MODULE, [
             'updateCustomerDetails'
-        ])
+        ]),
+
+        onGuestFieldBlur (field) {
+            this.onFieldBlur(VALIDATIONS.guest, field);
+        }
```
SachaZvetelman: Do we need this? Can't we just call `onFieldBlur` directly similar to the other parts of the code?

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -791,6 +810,10 @@ export default {
                 data: { ...this.eventData, validationState },
                 logMethod: this.$logger.logWarn
             });
+
+            this.invalidFieldsCount = invalidFieldsLength === 1 ?
+                this.$t('errorMessages.singleFieldError') :
+                this.$t('errorMessages.multipleFieldErrors', { invalidFieldsLength });
```
SachaZvetelman: This can be extracted into a computed property. And we should probably use a more descriptive name, since it's more a title than just a count, in my humble opinion.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -812,6 +835,10 @@ export default {
             return !this.$v.$invalid;
         },
 
+        onMobileFieldBlur () {
```
SachaZvetelman: Do we need this one?

SachaZvetelman (changes requested):

packages/components/organisms/f-checkout/src/components/Guest.vue
xander-marjoram: The problem is that you can't call `VALIDATIONS` inside the template unless you wrap it with a computed property

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -355,6 +367,19 @@ export default {
                 },
                 content: this.message
             };
+        },
+
+        invalidFieldsSummary () {
+            const invalidFieldCount = this.$v.$dirty
+                && validations.getFormValidationState(this.$v).invalidFields.length;
+
+            if (!invalidFieldCount) {
```
kevinrodrigues: You can return in a single line here:

`if (!invalidFieldCount) return null;`

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -336,6 +336,159 @@ describe('Checkout', () => {
             });
         });
 
+        describe('invalidFieldsSummary ::', () => {
```
kevinrodrigues, line 339: Maybe not for this PR but i think we should consider using `describes` a little more, it would help structure the test and make it easier to follow. Example:

``` javascript
describe('invalidFieldsSummary ::', () => {...
  describe('when invalidFieldCount is falsey', () => {
     it('should return null')
  })
describe('when invalidFieldCount is truthy', () => {
     describe('and error state is equal to 1')
       it('should match the single error snapshot')
     describe('and error state is not equal to 1')
       it('should match the multi error snapshot')
  })
```

LTurns: ah yeah thank you, will definitely keep this in mind for the future - We've been using more `describes` in our  f-header component tests for checking all the different tenants and it's been really helpful as an extra separation between tests - super useful for error handling 😄

kevinrodrigues (dismissed):

kevinrodrigues:
@LTurns Looks like there's a conflict.

Ah thank you - will solve now!

kevinrodrigues (approved):

FayeCarter (approved):

SachaZvetelman (approved):

PR #947, opened 2021-05-12
f-http@v0.4.1 - Return status code

tommcclean:
- Some systems are using `PACT` testing and directly asserting on status codes. I wasn't exposing this, so I am now exposing it.
- I didn't return the full response object as I wanted to avoid the lock in with Axios, so if we need more bits later; we can add the bits we need - for now this resolves the use case.

v0.4.1
*May 11, 2021*

### Changed
- Return status code to support PACT tests further down the stack


tommcclean:
Ignore percy - there aren't any UI changes; known issue

JoshuaNg2332 (approved):

siggerzz (approved):

PR #949, opened 2021-05-12
f-footer@4.21.0 - Adds styles that can be targeted by consumers

juan-amor:
---

Added 3 classes so they can be targeted by css in the consumer application

---

## UI Review Checks

## Browsers Tested

- [X] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/organisms/f-footer/CHANGELOG.md
+++ b/packages/components/organisms/f-footer/CHANGELOG.md
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v4.21.0
+------------------------------
+*May 12, 2021*
+
+### Added
+- Added 3 classes so they can be targeted by css in the consumer application
 
 v4.20.0
 ------------------------------
@@ -18,7 +24,7 @@ v4.19.0
 *April 15, 2021*
 
 ### Changed
-- Expander buttons are present only for narrow screens 
+- Expander buttons are present only for narrow screens
 
 ### Added
 - Tests to cover this change
--- a/packages/components/organisms/f-footer/README.md
+++ b/packages/components/organisms/f-footer/README.md
@@ -91,6 +91,18 @@ The props that can be defined are as follows (if any):
 | `showCountrySelector` | `Bool`   | `true`  | Controls whether to show country selector in footer.                                                          |
 | `euro2020Enabled`     | `Bool`   | `false` | Controls whether to show euro2020 related links in footer. |
 
+### CSS Classes
+
+The footer has its own styles which are scoped to the component using CSS modules to prevent conflicts with existing styles on the page.
+
+In addition to this, the footer exposes some classes which you can target in your application.
+
+| Class | Description |
+| ----- | ----------- |
+| `c-footer-row-linkLists` | Can be used to target the footer with the list of links |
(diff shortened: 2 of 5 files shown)


simonsandersje (approved):

DamianMullins (approved):

packages/components/organisms/f-footer/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v4.21.0
+------------------------------
+*May 12, 2021*
+
+### Added
+- Added 3 classes so they can be targeted by css in the consumer application
```
ashleynolan, line 11: Sorry for delayed comment – I missed this one at the time and only just noticed it

Interested to know what you wanted to do styling-wise with these classes in Menu, as ideally we should make changes through modifiers inside the component (to encapsulate the styles properly).

juan-amor: Hey @ashleynolan , the 3 classes are overridden in Menu Web so they can be hidden for dine-in. 
So in other words:

```
.c-footer--dineIn  {    
    .c-footer-row-linkLists,    
    .c-footer-row-partnerLinks,    
    .c-footer-row-socialLinks {        
       display: none;    
    }
}
```

ashleynolan: Makes sense – I'd change this so that a prop controls this within the footer component itself – via adding a class to that list of links (we do that on quite a few of the features on the header for instance).

Let me know if that's something you may not have capacity for this though and we can create a ticket in our backlog to update this. The less we can expose the styles in components the better as just means they're less likely to get out of sync across consuming applications.

DamianMullins: I'm not sure I understand why we'd only add the classes when a prop has been set? We do take this same approach in the megamodal component where we allow users to target "plain" classes so that they are able to customise the style if required. 

I understand that this could mean that the look of a component could then not match designs, but it also means they not very flexible - plus if someone wants to changes the styles without these classes being present they can still do it and it'll likely be via nested selectors. Best to allow the option to extend the styles to offer more flexibility rather than making components that are black boxes with regards to styling IMO.

Also, I could have got the wrong end of the stick here, let me know if so!

ashleynolan: I guess there are a couple of arguments to encapsulating the styles in the component:

- If Menu needs to hide these links, then other pages may need to in future too. So it makes sense to make these options of the component so we don't then have to manually add the styles into the consuming applications – they would instead get controlled by props which add the classes and styles to do this. This also means the tests can live in the component, rather than having to potentially duplicate tests to check that parts of the footer are being hidden in each consuming application.
- Allowing hook classes like this opens the door for consumers to change the look of the component away from how it may be intended (which is the aim of style encapsulation, to obfuscate the classes so this is much harder). So exposing these classes means in the future, other teams could use the classes to change the spacing/font-size etc of the component, which isn't then consistent with other uses of the footer.

Ideally, we should be doing the same in the modal as well where we try and pull as much as possible into the component as we can, but I think (from my memory of the modal) this was to help with hooking onto the content inside the modal as well? With the footer, all the content is known up front, so we shouldn't hit this problem.

I agree with the flexibility argument for some components – modal I think does need to be a little bit more flexible because of how it uses slots. For footer though, it should be able to look after it's own look and style (imo)

PR #960, opened 2021-05-17
f-cookie-banner@v0.10.0 - UK experimentation

simonsandersje:
showLegacyBanner can be passed in as a prop to allow experimentation in UK rollout

### DIFF
--- a/packages/components/organisms/f-cookie-banner/README.md
+++ b/packages/components/organisms/f-cookie-banner/README.md
@@ -76,7 +76,7 @@ The props that can be defined are as follows:
 | ----- | ----- | ------- | ----------- |
 | `locale` | `String` | `''` | Set the tenant for localisation ['da-DK', 'en-AU', 'en-GB', 'en-IE', 'en-NZ', 'es-ES', 'it-IT', 'nb-NO']. |
 | `isHidden` | `Boolean` | `false` | show/hide the cookie consent banner. |
-| `showLegacyBanner` | `Boolean` | `false` | Use the legacy "passive" banner markup (UK only). |
+| `shouldShowLegacyBanner` | `Boolean` | `false` | Use the legacy "passive" banner markup (UK only). |
 | `cookieExpiry` | `Number` | `7776000` | Expiry time of cookies written to the browser. |
 
 ### CSS Classes
--- a/packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
+++ b/packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
@@ -146,6 +146,7 @@ export default {
     computed: {
         /**
          * Check if the legacy cookie banner should be used
+         * @returns {Bool}
          */
         legacyBanner () {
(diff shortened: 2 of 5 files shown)


packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -115,7 +115,7 @@ export default {
 
         showLegacyBanner: {
             type: Boolean,
-            default: false
+            default: null
         },
```
AlixRichards, line 119: For my understanding - why is the default `null` rather than `false`

simonsandersje: So I can check if the prop has been passed in `return this.showLegacyBanner === null ?`

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -115,7 +115,7 @@ export default {
 
         showLegacyBanner: {
```
SachaZvetelman: I know this wasn't introduced in this PR, but given it's a `boolean`, the naming of the prop should indicate it's a `boolean`. For instance, `shouldShowLegacyBanner`.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
AlixRichards: and it wouldn't work if this was `false` instead? I'm happy to approve I just want to make sure I understand

simonsandersje: If I set it to `false` and then pass in `false` from the consuming app, Vue doesn't know that there has been a change and will always use the prop value rather than check the tenant config. This would mean that every consuming app would need to pass in the correct value for each tenant every time.

AlixRichards: Ahhh cool, I understand now thank you 👍

AlixRichards (approved):

DamianMullins (approved):

packages/components/organisms/f-cookie-banner/README.md
```diff
@@ -76,7 +76,7 @@ The props that can be defined are as follows:
 | ----- | ----- | ------- | ----------- |
 | `locale` | `String` | `''` | Set the tenant for localisation ['da-DK', 'en-AU', 'en-GB', 'en-IE', 'en-NZ', 'es-ES', 'it-IT', 'nb-NO']. |
 | `isHidden` | `Boolean` | `false` | show/hide the cookie consent banner. |
-| `showLegacyBanner` | `Boolean` | `false` | Use the legacy "passive" banner markup (UK only). |
+| `shouldShowLegacyBanner` | `Boolean` | `false` | Use the legacy "passive" banner markup (UK only). |
```
DamianMullins, line 79: The default value here should now be `null`.

PR #961, opened 2021-05-17
f-form-field@1.12.0 - textarea

dandel10n:
### Added
- Textarea element to the formField component

### DIFF
--- a/packages/components/atoms/f-form-field/src/components/FormField.vue
+++ b/packages/components/atoms/f-form-field/src/components/FormField.vue
@@ -156,26 +156,6 @@ export default {
             default: 100
         },
 
-        textareaCols: {
-            type: Number,
-            default: 30
-        },
-
-        textareaRows: {
-            type: Number,
-            default: 8
-        },
-
-        textareaMaxlength: {
-            type: Number,
-            default: 200
-        },
-
-        textareaName: {
-            type: String,
-            default: ''
-        },
-
         hasExtraLabelText: {
             type: Boolean,
             default: false
--- a/packages/components/atoms/f-form-field/src/components/FormField.vue
+++ b/packages/components/atoms/f-form-field/src/components/FormField.vue
@@ -63,7 +63,7 @@
                 :class="[
                     $style['c-formField-field'],
                     $style['c-formField-field--defaultHeight'],
-                    (isSelectionControl ? $style['c-formField-field--focus'] : '')
+                    { [$style['c-formField-field--focus']]: isSelectionControl }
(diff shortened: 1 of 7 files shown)


FayeCarter (dismissed):

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -45,8 +61,9 @@
                 placeholder=" "
                 :data-test-id="testId.input"
                 :class="[
-                    $style['c-formField-input'],
-                    (isSelectionControl ? $style['c-formField-input--focus'] : '')
+                    $style['c-formField-field'],
+                    $style['c-formField-field--defaultHeight'],
+                    (isSelectionControl ? $style['c-formField-field--focus'] : '')
```
DamianMullins: Could use the Vue conditional syntax here? i.e. `{ [$style['c-formField-field--focus']]: isSelectionControl }`.

DamianMullins (approved):

FayeCarter (approved):

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -19,20 +19,36 @@
                 {{ labelText }}
             </form-label>
 
+            <slot v-if="hasInputDescription" />
```
ashleynolan, line 22: Just looking at the storybook entry, this slot content is inside a div – should it be better described in HTML tags for screen readers?

dandel10n: At the moment it is in use for user note textarea in f-checkout and the text is basically the same as the one in textarea placeholder which is read by screen readers anyway. But if needed we can always add/change html tags in consuming app/component where we pass the element to the slot.

ashleynolan: I would at least change it here (and in checkout) so that we use a `p` tag for the text. We shouldn't use block elements on their own like `div`'s to wrap text content as it carries no semantic meaning (plus I'd be interested to know if the placeholder is still read out on screenreaders when some text has been entered in the textarea?).

Unless we think it really limits flexibility, I would probably just take in the text as the slot content (rather than the div and text), and wrap the slot in a `p` tag inside the form field component. That way we aren't reliant on consumers for the accessibility of the component as more people may use this down the line.

dandel10n: Makes sense, I will add it to 1.12.1 as well

PR #964, opened 2021-05-18
f-form-field@1.12.1 - added back missed css class to the textarea element

dandel10n:
### Added
- Added back missed css class to the textarea element
- Wrapped input description slot into a p html element

### DIFF
--- a/packages/components/atoms/f-form-field/src/components/FormField.vue
+++ b/packages/components/atoms/f-form-field/src/components/FormField.vue
@@ -44,7 +44,8 @@
                 v-bind="$attrs"
                 :class="[
                     $style['c-formField-field'],
-                    $style['c-formField-field--focus']
+                    $style['c-formField-field--focus'],
+                    $style['c-formField-field--textarea']
                 ]"
                 data-test-id="formfield-textarea"
                 v-on="listeners" />
@@ -335,7 +336,7 @@ $form-input-focus--boxShadow              : 0 0 0 2px $form-input-focus;
         }
     }
 
-    .c-formField-textarea {
+    .c-formField-field--textarea {
         background-clip: padding-box;
         padding: spacing(x2);
         resize: none;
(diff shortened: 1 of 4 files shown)


packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -44,7 +44,8 @@
                 v-bind="$attrs"
                 :class="[
                     $style['c-formField-field'],
-                    $style['c-formField-field--focus']
+                    $style['c-formField-field--focus'],
+                    $style['c-formField-field--textarea']
```
ashleynolan, line 52: Not related to this PR (will still approve) but do we need the `c-formField-field--focus` class?

Seems like those styles should just be an extension of `c-formField-field` potentially (as it implies a state that doesn't exist)

ashleynolan (dismissed):

packages/components/atoms/f-form-field/src/components/FormField.vue
dandel10n: There is a condition on input element to add focus class only if `isSelectionControl=true` so focus styles were separated into their own class

ashleynolan: Just looking over the old PR (which is actually mine), I think it's an error that I put in a while ago.

I'll update this and get a PR in, but I think the naming is wrong as I may have missed the `!` on [line 234](https://github.com/justeat/fozzie-components/pull/964/files#diff-0c855ee5f708709123fe3e0cd37f1060278c6ead61d3a62a053e5c23b45a03b3R234) – which is why this is confusing me!

DamianMullins (dismissed):

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -19,7 +19,11 @@
                 {{ labelText }}
             </form-label>
 
-            <slot v-if="hasInputDescription" />
+            <p
+                v-if="hasInputDescription"
+                class="u-spacingTop u-spacingBottom--large">
+                <slot />
+            </p>
```
dandel10n, line 26: @ashleynolan here is a change we discussed in #961

ashleynolan (approved):

DamianMullins (approved):

PR #965, opened 2021-05-18
Update documentation for generator component + fix component object #trivial

siggerzz:
- Removes old documentation
- Fixes bug in the generated component object that was causing component / a11y tests to fail


ashleynolan (approved):

ashleynolan:
@siggerzz You shouldn't use trivial for this bump – it should have the generator name for the PR title instead

DamianMullins (approved):

PR #968, opened 2021-05-18
f-checkout@0.115.0 - textarea element to come from f-form-field

dandel10n:
### Removed
- `UserNote.vue` as now it is just another form field from `f-form-field` component

### Updated
- `f-form-field` to version 1.12.1 to include textarea input-type

Before:
Screenshot 2021-05-18 at 14 12 51

After:
Screenshot 2021-05-18 at 14 12 08


packages/components/organisms/f-checkout/src/components/UserNote.vue
```diff
@@ -1,30 +1,31 @@
...
+            input-type="textarea"
             :placeholder="$t('userNote.placeholder')"
-            @input="updateUserNote($event.target.value)" />
+            :cols="30"
+            :rows="7"
+            :maxlength="200"
+            :show-extra-label-text="true"
+            :name="notePlaceholder"
+            :has-input-description="true"
+            @input="updateUserNote($event)">
```
DamianMullins: Do you need to explicitly pass `$event` here?

packages/components/organisms/f-checkout/src/components/UserNote.vue
```diff
@@ -1,30 +1,31 @@
...
-            :class="$style['c-userNote-textArea']"
+        <form-field
+            :label-text="$t('userNote.title')"
+            input-type="textarea"
             :placeholder="$t('userNote.placeholder')"
-            @input="updateUserNote($event.target.value)" />
+            :cols="30"
+            :rows="7"
+            :maxlength="200"
+            :show-extra-label-text="true"
```
DamianMullins: You could omit the "true" part here and the same for `has-input-description` below

```suggestion
            show-extra-label-text
```

packages/components/organisms/f-checkout/src/components/UserNote.vue
dandel10n: f-form-field component emits input event and passes only value so $event in this case is the value: https://github.com/justeat/fozzie-components/blob/e33577a7442ddf150bfa35a69c24ef2f5d9ec1f2/packages/components/atoms/f-form-field/src/components/FormField.vue#L261

JoshuaNg2332 (dismissed):

DamianMullins (dismissed):

tommcclean (dismissed):

siggerzz:
CircleCI is failing with the following error:

```
@justeat/f-checkout: [0-0] RUNNING in chrome - /test/component/f-checkout-delivery.component.desktop.spec.js
@justeat/f-checkout: [0-0] Error in "f-checkout "delivery" component tests should enable a user to submit a postcode with correct characters"
@justeat/f-checkout: Can't call setValue on element with selector "[data-test-id="user-note"] textarea" because element wasn't found
```


You'll want to either modify this selector, or get it to match the selector that's already there 👍🏼 

https://github.com/justeat/fozzie-components/blob/master/packages/components/organisms/f-checkout/test-utils/component-objects/f-checkout-selectors.js#L55

tommcclean (dismissed):

DamianMullins (dismissed):

DamianMullins (approved):

tommcclean (approved):

PR #971, opened 2021-05-19
f-link@v0.2.0 - Add props and styles

FayeCarter:
### Added
- Props `isExternalLink` and `opensInNewLocation` to set `target`, `rel`, and `ariaDescription`.
- Props `isBold`, `isFullWidth`, `noLineBreak` and `hasTextDecoration` to toggle link style.
- Tests to cover changes.

> There will be another PR after this to update the readme

### DIFF
--- a/packages/components/atoms/f-link/CHANGELOG.md
+++ b/packages/components/atoms/f-link/CHANGELOG.md
@@ -6,11 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.2.0
 ------------------------------
-*May 19, 2021*
+*May 20, 2021*
 
 ### Added
-- Props `isExternalLink` and `opensInNewLocation` to set `target`, `rel`, and `arialLabel`.
-- Props `isBold`, `isFullWidth` and `hadTextDecoration` to toggle link style.
+- Props `isExternalLink` and `opensInNewLocation` to set `target`, `rel`, and `ariaLabel`.
+- Props `isBold`, `isFullWidth` and `hasTextDecoration` to toggle link style.
 - Tests to cover changes.
 
 
--- a/packages/components/atoms/f-link/CHANGELOG.md
+++ b/packages/components/atoms/f-link/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.2.0
 ------------------------------
-*May 20, 2021*
+*May 21, 2021*
 
 ### Added
 - Props `isExternalLink` and `opensInNewLocation` to set `target`, `rel`, and `ariaLabel`.
--- a/packages/components/atoms/f-link/CHANGELOG.md
+++ b/packages/components/atoms/f-link/CHANGELOG.md
@@ -6,11 +6,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v0.2.0
 ------------------------------
-*May 21, 2021*
+*May 24, 2021*
 
 ### Added
-- Props `isExternalLink` and `opensInNewLocation` to set `target`, `rel`, and `ariaLabel`.
-- Props `isBold`, `isFullWidth` and `hasTextDecoration` to toggle link style.
+- Props `isExternalLink` and `opensInNewLocation` to set `target`, `rel`, and `ariaDescription`.
+- Props `isBold`, `isFullWidth`, `noLineBreak` and `hasTextDecoration` to toggle link style.
 - Tests to cover changes.
 
 
--- a/packages/components/atoms/f-link/CHANGELOG.md
+++ b/packages/components/atoms/f-link/CHANGELOG.md
@@ -9,7 +9,6 @@ v0.2.0
 *May 19, 2021*
 
 ### Added
-- Props to set `linkText` and `url`.
 - Props `isExternalLink` and `opensInNewLocation` to set `target`, `rel`, and `arialLabel`.
 - Props `isBold`, `isFullWidth` and `hadTextDecoration` to toggle link style.
 - Tests to cover changes.
--- a/packages/components/atoms/f-link/CHANGELOG.md
+++ b/packages/components/atoms/f-link/CHANGELOG.md
@@ -4,6 +4,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.2.0
+------------------------------
+*May 19, 2021*
+
+### Added
+- Props to set `linkText` and `url`.
+- Props `isExternalLink` and `opensInNewLocation` to set `target`, `rel`, and `arialLabel`.
+- Props `isBold`, `isFullWidth` and `hadTextDecoration` to toggle link style.
+- Tests to cover changes.
+
+
 v0.1.0
 ------------------------------
 *May 14, 2021*
--- a/packages/components/atoms/f-link/src/components/Link.vue
+++ b/packages/components/atoms/f-link/src/components/Link.vue
@@ -6,7 +6,7 @@
             { [$style['o-link--noDecoration']]: !hasTextDecoration },
             { [$style['o-link--full']]: isFullWidth }
         ]"
-        :data-test-id="`${linkText} link`"
+        :data-test-id="dataTestId"
         :href="url"
         :target="target"
         :aria-label="ariaLabel"
@@ -27,6 +27,11 @@ export default {
         locale: {
             type: String,
             default: 'en-GB'
+        },
+
+        dataTestId: {
+            type: String,
+            required: true
         },
 
         linkText: {
--- a/packages/components/atoms/f-link/src/components/Link.vue
+++ b/packages/components/atoms/f-link/src/components/Link.vue
@@ -17,9 +17,9 @@
             <slot />
         </a>
         <span
+            v-if="ariaDescription"
             id=description
             class="is-visuallyHidden"
-            v-if="ariaDescription"
         >
             {{ ariaDescription }}
         </span>
--- a/packages/components/atoms/f-link/src/components/Link.vue
+++ b/packages/components/atoms/f-link/src/components/Link.vue
@@ -84,7 +84,7 @@ export default {
             const locationType = this.isExternalLink ? 'opensExternalSiteInNew' : 'openInNew';
             const type = this.target === DEFAULT_LINK_TARGET ? 'opensExternal' : locationType;
 
-            return this.linkText + " - " + this.$t(`ariaDescribedBy['${type}']`) || null;
+            return `${this.linkText} - ${this.$t(`ariaDescribedBy['${type}']`)}` || null;
         }
     }
 };
--- a/packages/components/atoms/f-link/src/components/Link.vue
+++ b/packages/components/atoms/f-link/src/components/Link.vue
@@ -7,23 +7,16 @@
             { [$style['o-link--full']]: isFullWidth }
         ]"
         data-test-id="link"
-        :href="linkHref"
+        :href="url"
         :target="target"
-        rel="noopener"
         :aria-label="ariaLabel"
-    >
-        {{ linkText }}
-    </a>
+        :rel="rel"
+    >{{ linkText }}</a>
 </template>
 
 <script>
-// import { globalisationServices } from '@justeat/f-services';
 import { VueGlobalisationMixin } from '@justeat/f-globalisation';
 import tenantConfigs from '../tenants';
-import {
-    DEFAULT_LINK_TARGET,
-    VALID_LINK_TARGETS
-} from '../constants';
 
 export default {
     name: 'VLink',
@@ -41,9 +34,19 @@ export default {
             required: true
         },
 
-        linkHref: {
+        url: {
             type: String,
             required: true
+        },
+
+        isExternalLink: {
+            type: Boolean,
+            default: false
+        },
+
+        opensInNew: {
+            type: Boolean,
+            default: false
         },
 
         isBold: {
@@ -59,17 +62,6 @@ export default {
         isFullWidth: {
             type: Boolean,
             default: false
-        },
-
-        target: {
-            type: String,
-            default: DEFAULT_LINK_TARGET,
-            validator: value => (VALID_LINK_TARGETS.indexOf(value) !== -1) // The prop value must match one of the valid link types
-        },
-
-        isExternalLink: {
-            type: Boolean,
-            default: false
         }
     },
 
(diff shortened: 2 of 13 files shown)


packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,138 @@
...
-    </div>
+    <a
+        :class="[
+            $style['o-link'],
+            { [$style['o-link--bold']]: isBold },
+            { [$style['o-link--noDecoration']]: !hasTextDecoration },
+            { [$style['o-link--full']]: isFullWidth }
+        ]"
+        :data-test-id="`${linkText} link`"
+        :href="url"
```
dandel10n: You can use `v-bind="$attrs"` so there will be no need for passing url prop but just standard `href=""` on a component call

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,138 @@
...
-        {{ copy.text }}
-    </div>
+    <a
+        :class="[
+            $style['o-link'],
+            { [$style['o-link--bold']]: isBold },
+            { [$style['o-link--noDecoration']]: !hasTextDecoration },
+            { [$style['o-link--full']]: isFullWidth }
+        ]"
+        :data-test-id="`${linkText} link`"
```
dandel10n: This should be in `some-dashed-text` format

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,138 @@
...
 import tenantConfigs from '../tenants';
 
 export default {
     name: 'VLink',
-    components: {},
+
+    mixins: [VueGlobalisationMixin],
+
     props: {
         locale: {
```
dandel10n: Do we need this prop?

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,138 @@
...
+
+    &:hover,
+    &:focus {
+        color: $color-link-hover;
+    }
+
+    &:active {
+        color: $color-link-active;
+    }
+    // have left in in case we want to use eventually
```
dandel10n: Did you transfer this from fozzie? I don't think that we need it as it's been there commented for a long time

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,138 @@
...
+        ariaLabel () {
+            let message;
+
+            if (this.isExternal) {
+                message = this.$t('ariaLabel.externalSite');
+            } else if (this.opensInNewLocation) {
+                message = this.$t('ariaLabel.newLocation');
+            }
+
+            return message ? this.linkText + message : this.linkText;
```
dandel10n: I think we need to add ariaLabel only if we need to add info about link opening in a new location and not duplicate just link text in case if it opens in the same page

packages/components/atoms/f-link/stories/Link.stories.js
```diff
@@ -1,20 +1,77 @@
...
+            default: boolean('hasTextDecoration', true)
+        },
+
+        isFullWidth: {
+            default: boolean('isFullWidth', false)
+        }
+
     },
-    template: `<v-link />`
+    template: `<v-link ` +
```
dandel10n: You don't need `+` on every line here

packages/components/atoms/f-link/stories/Link.stories.js
```diff
@@ -1,20 +1,77 @@
...
+                ':locale="locale" ' +
+                ':dataTestId="dataTestId" ' +
+                ':linkText="linkText" ' +
+                ':url="url" ' +
+                ':isExternal="isExternal" ' +
+                ':opensInNewLocation="opensInNewLocation" ' +
+                ':isBold="isBold" ' +
+                ':hasTextDecoration="hasTextDecoration"  ' +
+                ':isFullWidth="isFullWidth"  ' +
+                '/>'
```
dandel10n: Closing ' should be `

packages/components/atoms/f-link/src/components/Link.vue
FayeCarter: Thanks! this now comes in $attrs

packages/components/atoms/f-link/src/components/Link.vue
FayeCarter: Yeah, it's copied directly. I'll remove it

packages/components/atoms/f-link/CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.2.0
+------------------------------
+*May 19, 2021*
+
+### Added
+- Props to set `linkText` and `url`.
+- Props `isExternalLink` and `opensInNewLocation` to set `target`, `rel`, and `arialLabel`.
```
minrwhite: NIT:
```suggestion
- Props `isExternalLink` and `opensInNewLocation` to set `target`, `rel`, and `ariaLabel`.
```

packages/components/atoms/f-link/src/components/Link.vue
FayeCarter: We don't :D

minrwhite (commented):
One observation, but overall looks good to me 👍🏽

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,127 @@
...
-    >
-        {{ copy.text }}
-    </div>
+    <a
+        :class="[
+            $style['o-link'],
+            { [$style['o-link--bold']]: isBold },
+            { [$style['o-link--noDecoration']]: !hasTextDecoration },
+            { [$style['o-link--full']]: isFullWidth }
+        ]"
```
DamianMullins: You can add the conditional classes in a single object

```suggestion
            $style['o-link'], {
                [$style['o-link--bold']]: isBold,
                [$style['o-link--noDecoration']]: !hasTextDecoration,
                [$style['o-link--full']]: isFullWidth
            }]"
```

packages/components/atoms/f-link/CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.2.0
+------------------------------
+*May 19, 2021*
+
+### Added
+- Props `isExternalLink` and `opensInNewLocation` to set `target`, `rel`, and `arialLabel`.
+- Props `isBold`, `isFullWidth` and `hadTextDecoration` to toggle link style.
```
ashleynolan:
```suggestion
- Props `isBold`, `isFullWidth` and `hasTextDecoration` to toggle link style.
```

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,127 @@
...
 
 export default {
     name: 'VLink',
-    components: {},
+
+    mixins: [VueGlobalisationMixin],
+
     props: {
-        locale: {
+        linkText: {
```
ashleynolan: Could this not just use the slot content (rather than a prop) when we use the component? Would also be able to pass in icons as part of this slot content then if needed down the line.

```
<my-custom-link>Some example link text</my-custom-link>

Comment (minrwhite) [reply packages/components/atoms/f-link/src/components/Link.vue]: I think the tricky part is then extracting that text out from the slot in order to also as part of the `aria-label`, if I'm not mistaken, unless you had something in mind for that, @ashleynolan?

File: packages/components/atoms/f-link/src/components/Link.vue [hunk 12]
```diff
@@ -1,45 +1,127 @@
...
+            type: String,
+            required: true
+        },
+
+        url: {
```
dandel10n: Url is not in use anymore so this can be deleted

FayeCarter: Thought I'd got rid of this already. Thanks 🙂

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,127 @@
...
+        ariaLabel () {
+            let message;
+
+            if (this.isExternal) {
+                message = this.$t('ariaLabel.externalSite');
+            } else if (this.opensInNewLocation) {
+                message = this.$t('ariaLabel.newLocation');
+            }
+
+            return message ? this.linkText + message : null;
```
dandel10n: We don't need to add linkText here as screen reader should read the link content and than aria label if I am not mistaken. Definitely worth testing

packages/components/atoms/f-link/test-utils/component-objects/f-link-selectors.js
```diff
@@ -1 +1 @@
-export const COMPONENT = '[data-test-id="link"]';
+export const COMPONENT = '[data-test-id="link-component"]';
```
dandel10n, line 1: I think CI build is failing because this data-test-id hasn't been added to the component

packages/components/atoms/f-link/src/components/Link.vue
FayeCarter: ~~I think I've found the solution.~~
~~I think I should have been using `aria-description` instead of `aria-label`. This read the text in the slot and then read the description 'opens in new ...'~~
Turns out `aria-description` creates a vue-cli-service linting issue

packages/components/atoms/f-link/src/components/Link.vue
FayeCarter: Annoyingly it didn't with `aria-label`. I've added `aria-description` which seems to be doing the right thing.

packages/components/atoms/f-link/src/components/Link.vue
ashleynolan: You should be able to get the content of a slot using `this.$slots.default` (if you only have one slot like this), so you could either assign that directly, or sanitise it in a computed prop (to make sure it's just a string of text) and then use the computed in the aria attribute

FayeCarter: Brilliant! Thanks so much @ashleynolan

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,123 @@
...
-        {{ copy.text }}
-    </div>
+    <a
+        :class="[
+            $style['o-link'], {
+                [$style['o-link--bold']]: isBold,
+                [$style['o-link--noDecoration']]: !hasTextDecoration,
+                [$style['o-link--full']]: isFullWidth
+            }]"
+        :data-test-id="'link-component'"
```
minrwhite: If this is a literal, I'd just make it a standard attribute, unless this has some weird interaction with where you're setting the `data-test-id` attribute in storybook?
```suggestion
        data-test-id="link-component"
```

packages/components/atoms/f-link/src/components/_tests/Link.test.js
```diff
@@ -1,10 +1,163 @@
-import { shallowMount } from '@vue/test-utils';
+import { shallowMount, createLocalVue } from '@vue/test-utils';
```
ashleynolan, line 1: Not sure if it's possible with CSS Modules (I know we've had issues with Jest + CSS Modules rendering out classnames previously), but would be good to have some tests for the props (so that it checks that classes are added when props are true (such as the bold/full-width ones).

Doesn't need to be part of this PR but would be good to see if this is testable or not in Jest with CSS Modules now.

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,123 @@
...
         return {
-            copy: { ...localeConfig }
+            tenantConfigs
         };
+    },
+
+    computed: {
+        linkText () {
+            return this.$slots.default[0].children[0].text.trim();
+        },
```
minrwhite: I'd worry here that any text from subsequent children after the first would be lost.

I'd be tempted to add the aria-label content using a [locally-registered directive](https://vuejs.org/v2/guide/custom-directive.html#Hook-Functions) hooking into the `componentUpdated` event to ensure that all text from child elements would be available and using the element's [`innerText`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText) property. Happy to walk through this in a pairing session if this would be useful. @ashleynolan - any thoughts on this strategy?

FayeCarter: Thanks for this Ben 🙂

ashleynolan: Nope, sounds good to me – I'd make sure we add some comments to the resulting code just explaining what it's doing for anyone new to this section.

dandel10n (commented):
I would like to suggest one more prop for this component: `no-line-break`. I found that for some links we add `white-space: nowrap;` to avoid link splitting into different lines so would be useful to have a prop to pass here which will add this style via `o-link--noBreak` for example

dandel10n (dismissed):

minrwhite (dismissed):

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,136 @@
...
+            aria-describedby="description"
+            :target="target"
+            :rel="rel"
+            v-bind="$attrs"
+        >
+            <slot />
+        </a>
+        <span
+            v-if="ariaDescription"
+            id=description
```
DamianMullins:
```suggestion
            id="description"
```

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -1,45 +1,136 @@
...
+        <a
+            :class="[
+                $style['o-link'], {
+                    [$style['o-link--bold']]: isBold,
+                    [$style['o-link--noDecoration']]: !hasTextDecoration,
+                    [$style['o-link--full']]: isFullWidth,
+                    [$style['o-link--noBreak']]: noLineBreak
+                }]"
+            data-test-id="link-component"
+            aria-describedby="description"
```
DamianMullins: Would this point to a non-existent element if an aria description is not present?

packages/components/atoms/f-link/src/components/Link.vue
FayeCarter: 🤦

minrwhite: Actually on this - an `id` attribute must be unique. So assuming there may be more than one link per page, this should probably be a system-generated identifier (maybe a slugified version of the link text?) rather than a fixed string.

packages/components/atoms/f-link/src/components/Link.vue
FayeCarter: Fixed, Thank you!

packages/components/atoms/f-link/src/components/Link.vue
FayeCarter: Updated this now, Thanks :)

minrwhite (approved):

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -68,8 +70,11 @@ export default {
     },
 
     data () {
+        uid += 1;
```
DamianMullins, line 73: Does this correctly increment when adding multiple components to the page?

FayeCarter: Hey @DamianMullins!  It does yeah. I've tested it with the three links in registration and each one has a different ID

DamianMullins: 🎉

DamianMullins (approved):

PR #972, opened 2021-05-19
f-mega-modal@v0.9.0 - added the ability to pass `title` and `titleHtmlTag` values

1itvinka:
### Added
- `title` and `titleHtmlTag` props to allow passing the title value, so that we can control the default display of the title inside the component
- Unit tests

### DIFF
--- a/packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
+++ b/packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
@@ -47,6 +47,15 @@
                     </f-button>
                 </slot>
 
+                <component
+                    :is="titleHtmlTag"
+                    v-if="title"
+                    :class="$style['c-megaModal-title']"
+                    data-test-id="mega-modal-title"
+                >
+                    {{ title }}
+                </component>
+
                 <slot />
             </div>
         </div>
@@ -128,6 +137,17 @@ export default {
         closeButtonCopy: {
             type: String,
             default: 'Close modal'
+        },
+
+        title: {
+            type: String,
+            default: ''
+        },
+
+        titleHtmlTag: {
+            type: String,
+            default: 'h3',
+            validator: value => ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'].includes(value)
         }
     },
 
@@ -384,5 +404,9 @@ export default {
     .c-megaModal-closeIcon * {
         fill: $color-link-default;
     }
+}
+
+.c-megaModal-title {
+    margin: 0 spacing(x3);
 }
 </style>
--- a/packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
+++ b/packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
@@ -50,9 +50,8 @@
                 <component
                     :is="titleHtmlTag"
                     v-if="title"
-                    :class="$style['c-megaModal-title']"
-                    data-test-id="mega-modal-title"
-                >
+                    :class="['c-megaModal-title', $style['c-megaModal-title']]"
+                    data-test-id="mega-modal-title">
                     {{ title }}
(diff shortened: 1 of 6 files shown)


packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -47,6 +47,15 @@
                     </f-button>
                 </slot>
 
+                <component
+                    :is="titleHtmlTag"
+                    v-if="title"
+                    :class="$style['c-megaModal-title']"
+                    data-test-id="mega-modal-title"
+                >
+                    {{ title }}
```
DamianMullins, line 55: On menu we use a `slot`, here (`<slot name="heading" />`) rather than a prop, as we pass some additional markup to allow for things like sticky headers etc.

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -47,6 +47,15 @@
                     </f-button>
                 </slot>
 
+                <component
+                    :is="titleHtmlTag"
+                    v-if="title"
+                    :class="$style['c-megaModal-title']"
+                    data-test-id="mega-modal-title"
+                >
```
DamianMullins: NIT 😸 

```suggestion
                    data-test-id="mega-modal-title">
```

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -47,6 +47,15 @@
                     </f-button>
                 </slot>
 
+                <component
+                    :is="titleHtmlTag"
+                    v-if="title"
+                    :class="$style['c-megaModal-title']"
```
DamianMullins: Could you also add the "plain" CSS class here so that applications can target it? e.g. `['c-megaModal-title', $style['c-megaModal-title']`.

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
1itvinka: I'll look into it now, thanks!

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -128,6 +137,17 @@ export default {
         closeButtonCopy: {
             type: String,
             default: 'Close modal'
+        },
+
+        title: {
```
ashleynolan, line 141: Can you make sure you add these new props to the component `README` so that they're documented.

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -128,6 +137,17 @@ export default {
...
+            type: String,
+            default: ''
+        },
+
+        titleHtmlTag: {
+            type: String,
+            default: 'h3',
+            validator: value => ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'].includes(value)
```
ashleynolan: I'd consider removing a couple of these  – I'm not sure we'd ever need a modal to have a heading of `h5`/`h6`?

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
1itvinka: @DamianMullins you can still define your own heading without using the prop. It's not a breaking change, your code will still work. This is for situations when you have a standard heading.  I think that the way it works now covers the situation on menu where you need additional markup and custom styles.

I think that ideally, we should have the sticky headers feature in `f-mega-modal`. 

What do you think?

ashleynolan (approved):

DamianMullins (approved):

PR #974, opened 2021-05-20
f-checkout@0.117.0 - Add mobile number aria label

charliearlie:
### Added
- Aria label with structured mobile number for screen readers

## Browsers Tested

- [ ] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -399,12 +399,7 @@ export default {
         },
 
         formattedMobileNumberForScreenReader () {
-            return Array.from(this.customer.mobileNumber, (digit, index) => {
-                if (index > 0 && (index + 1) % 3 === 0) {
-                    return `${digit}.`;
-                }
-                return digit;
-            }).join(' ');
+            return [...this.customer.mobileNumber].join(' ');
         }
     },
 
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -399,10 +399,8 @@ export default {
         },
 
         formattedMobileNumberForScreenReader () {
-            const usesAreaCode = this.customer.mobileNumber.startsWith('+');
-            const pauseIndices = usesAreaCode ? [6, 9] : [4, 7]; // TODO: Split the number at different indices depending on country
             return Array.from(this.customer.mobileNumber, (digit, index) => {
-                if (pauseIndices.includes(index)) {
+                if (index > 0 && (index + 1) % 3 === 0) {
                     return `${digit}.`;
                 }
                 return digit;
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -57,6 +57,7 @@
                         :has-error="isMobileNumberEmpty || isMobileNumberInvalid"
                         aria-describedby="mobile-number-error"
                         :aria-invalid="!isMobileNumberValid"
+                        :aria-label="formattedMobileNumberForScreenReader"
                         @input="updateCustomerDetails({ mobileNumber: $event })">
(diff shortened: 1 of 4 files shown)


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -395,6 +396,17 @@ export default {
...
+
+        formattedMobileNumberForScreenReader () {
+            const usesAreaCode = this.customer.mobileNumber.startsWith('+');
+            const pauseIndices = usesAreaCode ? [6, 9] : [4, 7]; // TODO: Split the number at different indices depending on country
+            return Array.from(this.customer.mobileNumber, (digit, index) => {
+                if (pauseIndices.includes(index)) {
+                    return `${digit}.`;
+                }
+                return digit;
+            }).join(' ');
```
charliearlie: I feel like this is a bit mental and could maybe just pause after every third digit, but felt like I should structure it how we tend to in the UK 🙃

SachaZvetelman: I personally think we should simplify this and do what you suggested. A pause after 3 digits or so, or even just read a digit at a time and see how fast or how slow that's read by the screenreader. 

We need to support several countries with different phone numbers lengths and ways to structure them.
Even within one country, people can structure it or read it differently.

I might be completely wrong here, but I'd prefer the number to be read at a decent speed, one at a time, rather than the website trying to be too smart and make things more confusing.

cc @ross-gratton.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -395,6 +396,17 @@ export default {
             return invalidFieldCount === 1 ?
                 this.$t('errorMessages.singleFieldError') :
                 this.$t('errorMessages.multipleFieldErrors', { errorCount: invalidFieldCount });
+        },
+
+        formattedMobileNumberForScreenReader () {
+            const usesAreaCode = this.customer.mobileNumber.startsWith('+');
```
SachaZvetelman: `hasAreaCode`

packages/components/organisms/f-checkout/src/components/Checkout.vue
charliearlie: I am more than happy to do that. Less code is always better. I think having no periods at all works but I was just following a suggestion given in the ticket.

charliearlie: I will follow up with Ross whether I should include periods at all or just keep the spacing.

packages/components/organisms/f-checkout/src/components/Checkout.vue
charliearlie: Following the above comment this code `hasBeenDeleted` 😅

packages/components/organisms/f-checkout/src/components/Checkout.vue
charliearlie: We have decided to remove full stops all together as the issue was the number being read out as a whole number and now it isn't. And as you mentioned, this will need to cover other territories so it's safer this way.

siggerzz (dismissed):

minrwhite (dismissed):

dandel10n (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -57,6 +57,7 @@
                         :has-error="isMobileNumberEmpty || isMobileNumberInvalid"
                         aria-describedby="mobile-number-error"
                         :aria-invalid="!isMobileNumberValid"
```
dandel10n: Do you have a console error that `isMobileNumberValid is not defined on the instance but referenced during render`. Maybe while you are here you can fix this one? I checked that the value should be `isMobileNumberInvalid` and not `!isMobileNumberValid`

dandel10n (approved):

1itvinka (approved):

PR #975, opened 2021-05-20
f-card@1.2.0 - add hasFullWidthFooter prop

dandel10n:
### Added
- `hasFullWidthFooter` prop
- Named slot `cardFooter` to render full width content at the bottom of the card without card paddings when `hasFullWidthFooter` set to true
- `.c-card-innerSpacing` css class for card paddings to be able to separate card content paddings from the card itself

### Removed
- `.c-card--separated`, `.c-card--center`, `.c-card--right` css classes as unused

Screenshot 2021-05-20 at 14 25 17

### DIFF
--- a/packages/components/atoms/f-card/src/components/Card.vue
+++ b/packages/components/atoms/f-card/src/components/Card.vue
@@ -2,24 +2,31 @@
     <div
         data-test-id="card-component"
         :class="[
-            $style['c-card'],
-            (isRounded ? $style['c-card--rounded'] : ''),
-            (hasOutline ? $style['c-card--outline'] : ''),
-            (isPageContentWrapper ? $style['c-card--pageContentWrapper'] : '')
-        ]"
-    >
-        <component
-            :is="cardHeadingTag"
-            v-if="cardHeading"
-            :class="[
-                $style['c-card-heading'],
-                (cardHeadingPosition !== 'left' ? $style[`c-card--${cardHeadingPosition}`] : '')
-            ]"
-            data-test-id="card-heading"
-        >
-            {{ cardHeading }}
-        </component>
-        <slot />
+            $style['c-card'], {
+                [$style['c-card--rounded']]: isRounded,
+                [$style['c-card--outline']]: hasOutline,
+                [$style['c-card--pageContentWrapper']]: isPageContentWrapper
+            }]">
+        <div :class="[$style['c-card--innerSpacing']]">
+            <component
+                :is="cardHeadingTag"
+                v-if="cardHeading"
+                :class="[
+                    $style['c-card-heading'],
+                    (cardHeadingPosition !== 'left' ? $style[`c-card--${cardHeadingPosition}`] : '')
+                ]"
+                data-test-id="card-heading">
+                {{ cardHeading }}
+            </component>
+            <slot />
+        </div>
+
+        <div
+            v-if="hasFullWidthBottomElement"
+            data-test-id="card-component-fullWidthBottomElement">
+            <slot
+                name="full-width-bottom-element" />
+        </div>
     </div>
 </template>
 
@@ -53,6 +60,10 @@ export default {
         isRounded: {
             type: Boolean,
             default: false
+        },
+        hasFullWidthBottomElement: {
+            type: Boolean,
+            default: false
         }
     }
 };
@@ -70,11 +81,9 @@ $card--pageContentWrapper-width           : 472px; // so that it falls on our 8p
 .c-card {
     background-color: $card-bgColor;
     display: block;
-    padding: $card-padding;
 }
-
-    .c-card--separated {
-        margin-bottom: spacing();
+    .c-card--innerSpacing {
+        padding: $card-padding;
     }
 
     .c-card--rounded {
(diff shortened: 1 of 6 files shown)


packages/components/atoms/f-card/src/components/Card.vue
```diff
@@ -92,32 +101,27 @@ $card--pageContentWrapper-width           : 472px; // so that it falls on our 8p
...
             margin: spacing(x5) auto;
-            padding-left: spacing(x10);
-            padding-right: spacing(x10);
         }
-    }
 
-    .c-card-heading {
-        margin-bottom: spacing(x2);
-    }
+        .c-card--innerSpacing {
```
1itvinka: can we remove the css nesting here? less nesting - the better.

dandel10n: This one is nested for a reason. `.c-card--pageContentWrapper` modifies `.c-card` and has different inner spacing.

packages/components/atoms/f-card/src/components/_tests/Card.test.js
```diff
@@ -210,4 +210,24 @@ describe('Card', () => {
...
+        // Act
+        const wrapper = shallowMount(Card, { propsData });
+
+        // Act
+        const testedElement = wrapper.find('[data-test-id="card-component-fullWidthBottomElement"]');
+
+        it(`should ${wrapper.vm.hasFullWidthBottomElement ? '' : 'not '}display a bottom positioned full width element`, () => {
+            expect(testedElement.exists()).toBe(expected);
+        });
+    });
```
1itvinka, line 232: I think it might be more readable to have the following format. But I don't have a strong opinion on this. So ignore this comment if you think otherwise.

```
   describe('hasFullWidthBottomElement', () => {
       it.each([
           [true, true],
           [false, false]
       ])('`should return "%s" when `hasFullWidthBottomElement` is "%s"', (passedValue, expected) => {
           ....
       }
   });
```

1itvinka (dismissed):

packages/components/atoms/f-card/src/components/_tests/Card.test.js
dandel10n: The test checks that the specific element is on the page and in your suggested description the test checks that `hasFullWidthBottomElement` is true or false

dandel10n: The outcome at the moment says: "Card › `hasFullWidthBottomElement` is true › should display a bottom positioned full width element"

I will change to "Card › if `hasFullWidthBottomElement` prop is true › should display a bottom positioned full width element"

packages/components/atoms/f-card/src/components/Card.vue
```diff
@@ -53,6 +60,10 @@ export default {
         isRounded: {
             type: Boolean,
             default: false
+        },
+        hasFullWidthBottomElement: {
```
ashleynolan: I'd be tempted to call this a footer, as that's what it is really?

So then you could call this `hasFullWidthFooter` and then the slot itself could just be called `cardFooter`

packages/components/atoms/f-card/src/components/Card.vue
```diff
@@ -2,24 +2,31 @@
...
-        >
-            {{ cardHeading }}
-        </component>
-        <slot />
+            $style['c-card'], {
+                [$style['c-card--rounded']]: isRounded,
+                [$style['c-card--outline']]: hasOutline,
+                [$style['c-card--pageContentWrapper']]: isPageContentWrapper
+            }]">
+        <div :class="[$style['c-card--innerSpacing']]">
```
ashleynolan: This class would just be a child of `c-card` (not a modifier). So would just be called `c-card-innerSpacing`.

We've tended to use `wrapper` or `contentWrapper` in other components I think, so could also call it `c-card-contentWrapper` or something like that?

packages/components/atoms/f-card/stories/card.stories.js
```diff
@@ -19,7 +19,21 @@ export const CardComponent = (args, { argTypes }) => ({
...
+            :card-heading="cardHeading"
+            :card-heading-position="cardHeadingPosition"
+            :card-heading-tag="cardHeadingTag"
+            :is-rounded="isRounded"
+            :has-outline="hasOutline"
+            :is-page-content-wrapper="isPageContentWrapper"
+            :has-full-width-bottom-element="hasFullWidthBottomElement">
+            <p>Some Card Content</p>
+            <template v-slot:full-width-bottom-element>
+                <div>
```
ashleynolan: Would be better to use `p` tags as this gets wrapped in a `div` already when it gets inserted into the slot?

packages/components/atoms/f-card/src/components/Card.vue
dandel10n: I agree with `c-card-innerSpacing`. In terms of `c-card-contentWrapper` there is already `c-card--pageContentWrapper` class and it feels a bit confusing to understand the difference between them, as with `c-card-innerSpacing` it is more understandable straight away imo

DamianMullins (approved):

ashleynolan (approved):

PR #979, opened 2021-05-21
f-stat-client@v0.1.0 - Created a service to handle the publishing of stat data

oliversweb:
v0.1.0
*May 20, 2021*

### Added
- Created package


packages/services/f-stat-client/README.md
```diff
@@ -0,0 +1,109 @@
...
+```js
+import StatClient from '@justeat/f-stat-client';
+
+const client = new StatClient('http://localhost', 9200, 'uk', 'checkoutWeb');
+
+```
+### *How to use*
+```js
+
+await client.publish('GET', '/search', 200, 611);
```
tommcclean: I think this is good for now to work for our prototype - but I suspect we may want to change this later to support accepting an object with parameters - as that will ensure we cover other use cases like search, and any other stats we want to push - though it may be easier to plan forward now so we don't need to change the interface.

packages/services/f-stat-client/src/statClient.js
```diff
@@ -0,0 +1,55 @@
...
+            };
+        }
+
+        if (mock) {
+            conn.Connection = mock.getConnection();
+        }
+
+        this.client = new Client(conn);
+        this.tenant = tenant || 'ns';
+        this.feature = featureName || 'NotSpecified';
```
tommcclean: I would probably just default this at the top as `featureName = 'Generic Front End`

Same applies with tenant, makes sense to declare the defaults all in one place

packages/services/f-stat-client/src/statClient.js
```diff
@@ -0,0 +1,55 @@
...
+        this.indexName = indexName;
+    }
+
+    async publish ({
+        verb, segment, status, timing
+    }) {
+        const payload = {
+            index: this.indexName,
+            body: {
```
tommcclean, line 53: I think this can be a combination of fixed params and dynamic ones. For ex below, opens it up so different people can provide properties relevant to their use case

```
body: {
  tenant: this.tenant
  feature: this.feature,
  ...payload
}
```

tommcclean (dismissed):

packages/services/f-stat-client/README.md
```diff
@@ -0,0 +1,113 @@
...
+## Benefits (Now)
+- Simple publish method that hides away the complexities of making posts requests to ElasticSearch
+- Sensible defaults, with the ability to override all
+- If you run ElasticSearch locally in a Docker Container and use the host `localhost` then you can see you results instantly (see https://www.elastic.co/guide/en/elasticsearch/reference/7.0/docker.html & https://elasticvue.com/)
+- Built-in mock ability by providing a mock on the constructor (see https://github.com/elastic/elasticsearch-js-mock and an example below)
+
+## Benefits (Soon)
+- _encapsualted batch publishing_
+- _encapsualted sample publishing_
+- _encapsualted authentication_
```
tommcclean:
```suggestion
- _encapsulated batch publishing_
- _encapsulated sample publishing_
- _encapsulated authentication_
```

packages/services/f-stat-client/README.md
```diff
@@ -0,0 +1,113 @@
...
+```js
+import StatClient from '@justeat/f-stat-client';
+
+const client = new StatClient('http://localhost:9200', 'uk', 'checkoutWeb');
+
+```
+### *How to use*
+```js
+
+await client.publish({ verb: 'GET', segment: '/search', status: 200, timimg: 611 });
```
tommcclean:
```suggestion
await client.publish({ verb: 'GET', segment: '/search', status: 200, timing: 611 });
```

tommcclean (commented):
Just a couple of typos

tommcclean (dismissed):

packages/services/f-stat-client/src/statClient.js
```diff
@@ -0,0 +1,50 @@
+const { Client } = require('@elastic/elasticsearch');
+
+export default class StatClient {
+    constructor (
```
BanksySan: Can we make this constructor cleaner having it accept an object instead of so many parameters.  Seven parameter is a code smell to me, it lends itself to confusing code that calls it.

BanksySan (changes requested):

tommcclean (approved):

BanksySan (approved):

PR #980, opened 2021-05-24
f-checkout@0.122.0 and f-registration@0.56.0

dandel10n:
### Changed
- f-card package version updated to 1.2.1 to include card-footer slot option for t&c
- Some styles were cleaned up in connection with new f-card version

There is a PR for f-registration already in the queue so version will become correct after merging the latest master later


DamianMullins (dismissed):

BanksySan (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -23,6 +23,7 @@
         >
             <card
                 is-rounded
+                has-full-width-footer
```
FayeCarter, line 26: Should this change if checkout is guest or not?

dandel10n: No, as this prop just allows to add cardFooter slot, and this slot has a condition on it to be shown only for logout users

FayeCarter (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -132,7 +136,11 @@
                     </f-button>
                 </form>
 
-                <checkout-terms-and-conditions v-if="!isLoggedIn" />
+                <template
+                    v-if="!isLoggedIn"
+                    v-slot:cardFooter>
```
DamianMullins: Don't have to do this but you could use the slot shorthand `#cardFooter`

DamianMullins (dismissed):

BanksySan (dismissed):

BanksySan (approved):

PR #994, opened 2021-05-27
fozzie-components@v3.35.0 - Force Percy to run in parallel

siggerzz:
Due to lerna, Percy creates multiple Percy builds if a `test:visual` script exists in multiple packages.

To fix this, I've temporarily stopped using lerna for the visual regression tests, and instead, we're executing the Percy command directly from the root package.json.


DamianMullins (approved):

DamianMullins (changes requested):
You have some PR check failures to address

DamianMullins (dismissed):

dandel10n (approved):

DamianMullins (approved):

PR #995, opened 2021-05-27
f-checkout@v0.124.0 - Fix modal breaking page scrolling.

kevinrodrigues:
Have tried quite a few things to get this to work and this was the only solution that works correctly. Passing a `true` / `false` prop to mega modal never triggers the watch on `isOpen`.

### Fixed
- Scrolling broken when clicking `OK` within `megaModal` component.

## Browsers Tested

- [x] Chrome (latest)


simonsandersje (dismissed):

dandel10n (commented):
I am not sure why it doesn't show conflicts but the latest checkout version in master is 0.121.0 and there is a PR for 0.122.0 in the list already so your should be 0.123.0

DamianMullins (dismissed):

FayeCarter (dismissed):

simonsandersje (dismissed):

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
```diff
@@ -78,6 +82,11 @@ export default {
             }
 
             this.updateMessage();
+
+            const { errorModal } = this.$refs;
+            const { megaModal } = errorModal.$refs;
+
+            if (megaModal) errorModal.close();
```
SachaZvetelman: Could we extract this into a function/method, since we have the same code here and on `mounted`?

kevinrodrigues: Done.

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
```diff
@@ -64,20 +65,35 @@ export default {
...
     methods: {
         ...mapActions(VUEX_CHECKOUT_MODULE, [
             'updateMessage'
         ]),
 
+        getModalContext () {
+            const { errorModal } = this.$refs;
+            const { megaModal } = errorModal.$refs;
+
+            if (megaModal) return errorModal;
```
SachaZvetelman, line 82: Is this correct? It looks weird to me that we're checking if there is an `errorModal.$refs.megaModal` and if there is one, we return `errorModal`, but I don't know the context of this. Just to double check :)

kevinrodrigues: The megaModal reference is to make sure we have the element on the page, if we don't check that then we'd be attempting to close/open the modal without knowing if the modal is actually open or not. The errorModal which is returned is the main part that allows the method selection (open/close) Let me know if that makes sense.

SachaZvetelman: It kinda does. I'll trust you on this one 😄

tommcclean (dismissed):

oliversweb (approved):

packages/components/organisms/f-checkout/src/components/ErrorDialog.vue
ashleynolan: Would be useful to add a JSDoc to this method, as think this explanation helps give some context (as agree it's not obvious from the code, but makes sense with your explanation).

kevinrodrigues: Will update.

PR #996, opened 2021-05-27
f-checkout@0.126.0 - Restore selected fulfilment time (if valid and still available)

oliversweb:
### Changed
- Restored/Set the selected fulfilment time if available/valid on `mount`


packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -4,9 +4,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.123.0
+=======
```
SachaZvetelman: Nit: this isn't the format on the rest of the file :)

packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js
```diff
@@ -413,29 +343,181 @@ describe('Selector', () => {
...
+                        localVue,
+                        propsData
+                    });
+
+                    updateFulfilmentTimeSpy = jest.spyOn(Selector.methods, 'updateFulfilmentTime');
+                });
+
+                afterEach(() => {
+                    jest.clearAllMocks();
+                });
```
SachaZvetelman: You can move this to the parent `describe` block so you don't have to copy-paste this in every `describe`.

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -119,13 +120,17 @@ export default {
...
+         * currently present in the Fulfilment times array, otherwise initialise the with first value in the
+         * Fulfilment times array.
+         * Note; we always need to pre-select a time so the user can always proceeds with a selected time.
          *
-         * @param times
+         * @param {array} times - An array of available fulfilment times
          */
         initFulfilmentTime (times) {
-            if (times.length && times[0].value) {
+            if (this.time && this.time.from && times.length && times.find(i => i.value === this.time.from)) {
```
kevinrodrigues: Do you need the `times.length` check here? Also is it worth moving this check into a method as it's quite long.

packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js
```diff
@@ -413,29 +343,181 @@ describe('Selector', () => {
...
+
+                it('should exist', () => {
```
kevinrodrigues: I think one makes sense to move outside of the current describe as it's usually the first thing you'd want to fail.

packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js
```diff
@@ -413,29 +343,181 @@ describe('Selector', () => {
...
+                    expect(wrapper.vm.initFulfilmentTime).toBeDefined();
+                });
+
+                it('should make a call to `updateFulfilmentTime`', () => {
```
kevinrodrigues: `should make a call to updateFulfilmentTime ... with from & to times passed in`?

packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js
```diff
@@ -413,29 +343,181 @@ describe('Selector', () => {
...
+                });
+
+                it('should update `selectedAvailableFulfilmentTime` with the first available fulfilment `times`', () => {
+                    // Assert
+                    expect(wrapper.vm.selectedAvailableFulfilmentTime).toBe('2020-01-01T01:00:00.000Z');
+                });
+            });
+
+            describe('AND when there are fulfilment `times` available and a pre-selected fulfilment time', () => {
+                describe('AND the pre-selected fulfilment time is still available', () => {
```
kevinrodrigues, line 286: Do you need this describe as i think the outer one just above covers the test fine?

packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js
```diff
@@ -413,29 +343,181 @@ describe('Selector', () => {
...
+                        updateFulfilmentTimeSpy = jest.spyOn(Selector.methods, 'updateFulfilmentTime');
+                    });
+
+                    afterEach(() => {
+                        jest.clearAllMocks();
+                    });
+
+                    it('should not make a call to `updateFulfilmentTime`', () => {
+                        // Assert
+                        expect(updateFulfilmentTimeSpy).not.toHaveBeenCalledWith();
```
kevinrodrigues: `.not.toHaveBeenCalled()`

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -119,13 +120,17 @@ export default {
...
+                this.selectedAvailableFulfilmentTime = this.time.from;
+            } else if (times.length && times[0].value) {
```
SachaZvetelman: This seems a bit more complicated than what it could be, I think. Is it possible to simplify this?
I haven't checked if the code below works, but maybe it does, and if not, maybe it can give some ideas as in how to potentially make it cleaner:

```
initFulfilmentTime (times = []) {
  const selectedTime = times.find(time => time.value === this.time.from) || 
                                          times.length && times[0].value;

  this.selectedAvailableFulfilmentTime = selectedTime;

  this.updateFulfilmentTime({
      from: selectedTime,
      to: selectedTime
  });
}

File: packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js (hunk 7 above)
Comment (SachaZvetelman) [inline packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js]: `not.toHaveBeenCalled`, since you're not asserting on the parameters being (or not being) passed.

File: packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js [hunk 9]
```diff
@@ -413,29 +343,181 @@ describe('Selector', () => {
     });
 
     describe('mounted ::', () => {
-        it('should make a call to `initFulfilmentTime` with the available `fulfilmentTimes`', () => {
+        describe('invokes `initFulfilmentTime`', () => {
```
SachaZvetelman: I think all these tests are in the wrong section.

On mounted, all we want to check is that `initFullfilmentTime` is invoked with the right parameters.

These tests are actually unit tests of `initFulfilmentTime`.

packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js
```diff
@@ -413,29 +343,181 @@ describe('Selector', () => {
...
+                    updateFulfilmentTimeSpy = jest.spyOn(Selector.methods, 'updateFulfilmentTime');
+                });
+
+                afterEach(() => {
+                    jest.clearAllMocks();
+                });
+
+                it('should exist', () => {
+                    expect(wrapper.vm.initFulfilmentTime).toBeDefined();
+                });
```
SachaZvetelman: I personally don't see the value of this test. I know some people like to have this test, but if we don't have the method/computed defined, all the other tests will anyway fail with a very clear reason.

DamianMullins: Agree, we have started removing them in the menu project for that same reason.

kevinrodrigues (dismissed):

DamianMullins (dismissed):

DamianMullins (dismissed):

DamianMullins (dismissed):

kevinrodrigues (approved):

DamianMullins (approved):

PR #998, opened 2021-05-28
f-cookie-banner@v0.13.0 - add method to hide legacy cookie banner(s).

simonsandersje:
## Browsers Tested

- [x] Chrome (latest)


packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -206,6 +208,16 @@ export default {
             this.shouldHideBanner = true;
         },
 
+        /**
+         * Hide the legacy banners
+         */
+        hideAllBanners () {
+            const allBanners = document.getElementsByClassName('cpOverlay');
```
kevinrodrigues, line 215: what's cpOverlay?

kevinrodrigues (approved):

DamianMullins (approved):

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: @kevinrodrigues It's the container for the cookie banner script pushed by GTM. We have a race condition whereby both banners are being loaded and trapping the user on the page until they refresh. I have asked them to not load the script on pages with f-cookie-banner, but they insist on loading it. I have been showering constantly since writing this method!

PR #999, opened 2021-05-28
f-checkout@0.125.0 - Read address from local storage if it exists

charliearlie:
## Read user's address from local storage if it exists.

When a customer selects their full address in search box, their address is stored in local storage to be used later. It's a much better experience for the customer to have the Checkout pre-populate their address with what they have searched rather than an address from their address book as they could be different.

This PR will check to see if that address exists in local storage, if it does, then it will populate the Checkout form with said address.


packages/components/organisms/f-checkout/src/services/addressService.js
```diff
@@ -70,6 +72,25 @@ function getAddressClosestToPostcode (postcode, addresses) {
     return address;
 }
 
+const isAddressInLocalStorage = () => window.localStorage.getItem(storedLocationKey) !== null;
```
tommcclean: You should check that window and window.localstorage exist first - may be worth pulling the check out as it may be needed in a few places

- Window won't exist if the code happens to be run on the server - which it won't atm but good practice
- I think incognito windows in some browsers disable localstorage, so worth checking it exists

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
```diff
@@ -174,4 +174,75 @@ describe('addressService', () => {
...
+                const expectedAddress = {
+                    lines: ['Fleet Place House', 'Farringdon', 'City of London'],
+                    locality: 'London',
+                    postalCode: 'EC4M 7RF'
+                };
+                expect(addressService.getAddressFromLocalStorage()).toEqual(expectedAddress);
+            });
+        });
+    });
 });
```
tommcclean, line 275: Worth adding a test that disables local storage based on my previous comment - to make sure the code still runs through

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -421,8 +421,15 @@ export default {
 
             state.time = fulfilment.time;
 
-            if (fulfilment.location && fulfilment.location.address && fulfilment.location.address.lines) {
-                const { address } = fulfilment.location;
+            let address = null;
+            if (addressService.isAddressInLocalStorage()) {
+                address = addressService.getAddressFromLocalStorage();
+            } else if (fulfilment.location && fulfilment.location.address && fulfilment.location.address.lines) {
+                /* eslint-disable prefer-destructuring */
```
tommcclean, line 428: Is this bit still needed? I imagine it won't ask you to destructure it now its declared on a different line; could be wrong though

packages/components/organisms/f-checkout/src/services/addressService.js
charliearlie: Done 🙂

charliearlie (commented):
Now tests the case in which local storage doesn't exist too.

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
```diff
@@ -174,4 +174,124 @@ describe('addressService', () => {
...
+
+            it('should return false', () => {
+                expect(addressService.isAddressInLocalStorage()).toBe(false);
+            });
+        });
+    });
+
+    describe('getAddressFromLocalStorage ::', () => {
+        beforeEach(() => {
+            const localStorageMock = (function localStorageMock () {
```
tommcclean: Probably worth pulling this bit out to a const at the top so it isn't duplicated

charliearlie: Moved it into a test util file so it can be reused elsewhere. I know I have another task coming up which relies on local storage.

tommcclean (dismissed):

packages/components/organisms/f-checkout/src/services/addressService.js
```diff
@@ -70,6 +72,32 @@ function getAddressClosestToPostcode (postcode, addresses) {
     return address;
 }
 
+const isAddressInLocalStorage = () => {
+    if (window.localStorage) {
+        const address = window.localStorage.getItem(storedLocationKey) ? JSON.parse(window.localStorage.getItem(storedLocationKey)) : null;
+        return !!(address && (address.PostalCode || address.ZipCode));
```
1itvinka, line 82:
```suggestion
        return !!(address?.PostalCode || address?.ZipCode);
```

charliearlie: Unfortunately we can't use optional chaining with Storybook right now. I believe we're either waiting for the next version of Storybook or the next version of Vue.

1itvinka: I would probably even leave it like this: 
```
return address?.PostalCode || address?.ZipCode;
```
as for me, it's much more readable

1itvinka: oh I see!

1itvinka (dismissed):

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
```diff
@@ -174,4 +175,90 @@ describe('addressService', () => {
             });
         });
     });
+
+    describe('isAddressInLocalStorage ::', () => {
```
kevinrodrigues, line 179: Could you add arrange act to these please?

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
```diff
@@ -174,4 +175,90 @@ describe('addressService', () => {
...
+
+            describe('when the address does exist in local storage', () => {
+                it('should return true if there is a post code', async () => {
+                    window.localStorage.setItem('je-full-address-details', JSON.stringify(london3LinesDefault));
+                    expect(addressService.isAddressInLocalStorage()).toBe(true);
+                });
+
+                it('should return false if the customer only searched a postcode', async () => {
+                    window.localStorage.setItem('je-full-address-details', JSON.stringify({
+                        City: null,
```
kevinrodrigues, line 208: I think there should be a lat/long here too from what I remember? (Or was that mapped to field 1 & field 2)?

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
```diff
@@ -174,4 +175,90 @@ describe('addressService', () => {
             });
         });
     });
+
+    describe('isAddressInLocalStorage ::', () => {
+        describe('if localStorage exists', () => {
```
kevinrodrigues, line 180: If we're checking when it exists do you want to check when it doesn't exist?

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
```diff
@@ -174,4 +175,90 @@ describe('addressService', () => {
...
+            });
+
+            describe('when the address does NOT exist in local storage', () => {
+                it('should return false', () => {
+                    expect(addressService.isAddressInLocalStorage()).toBe(false);
+                });
+            });
+
+            describe('when the address does exist in local storage', () => {
+                it('should return true if there is a post code', async () => {
```
kevinrodrigues: `postcode`

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
charliearlie: Line 220 😉

packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.125.0
+------------------------------
+*May 24, 2021*
+
+### Added
+- Reading address from local storage if it exists
```
kevinrodrigues: Is there another pr to populate the address fields coming later?

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
charliearlie: I believe that may be separate as i have another ticket for exactly that. I copied this straight from local storage 😄

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
charliearlie: Done

packages/components/organisms/f-checkout/CHANGELOG.md
charliearlie: Nope, that's already done so updated the changelog to reflect that

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
kevinrodrigues: I actually think it's mapped to field 1 & 2 as you can't pass custom values but worth double checking.

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
```diff
@@ -174,4 +175,101 @@ describe('addressService', () => {
...
+                    expect(addressService.isAddressInLocalStorage()).toBe(false);
+                });
+            });
+
+            describe('when the address does exist in local storage', () => {
+                it('should return true if there is a postcode', async () => {
+                    // Arrange & Act
+                    window.localStorage.setItem('je-full-address-details', JSON.stringify(london3LinesDefault));
+
+                    // Assert
```
kevinrodrigues: Going to be picky :P the Act is actually this part. So here it should be `// Act & Assert`

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
```diff
@@ -174,4 +175,101 @@ describe('addressService', () => {
...
+                        Line2: null,
+                        Line3: null,
+                        Line4: null,
+                        Line5: null,
+                        PostalCode: null,
+                        searchBoxAddress: 'BS3 4RL'
+                    }));
+
+                    // Assert
+                    expect(addressService.isAddressInLocalStorage()).toBe(false);
```
kevinrodrigues, line 221: Same.

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
```diff
@@ -174,4 +175,101 @@ describe('addressService', () => {
...
+            it('should return the address mapped correctly', () => {
+                // Arrange & Act
+                const expectedAddress = {
+                    lines: ['Fleet Place House', 'Farringdon', 'City of London'],
+                    locality: 'London',
+                    postalCode: 'EC4M 7RF'
+                };
+                window.localStorage.setItem('je-full-address-details', JSON.stringify(london3LinesDefault));
+
+                // Assert
```
kevinrodrigues: Same.

packages/components/organisms/f-checkout/src/services/addressService.js
```diff
@@ -70,6 +72,32 @@ function getAddressClosestToPostcode (postcode, addresses) {
...
+    }
+
+    return false;
+};
+
+function mapLocalStorageAddress (address) {
```
kevinrodrigues: Can this be renamed to `getMappedLocalStorageAddress`? Since it's returning something.

packages/components/organisms/f-checkout/src/services/addressService.js
```diff
@@ -70,6 +72,32 @@ function getAddressClosestToPostcode (postcode, addresses) {
     return address;
 }
 
+const isAddressInLocalStorage = () => {
```
kevinrodrigues, line 79: If possible would be good to have JSdocs added to these methods.

kevinrodrigues (approved):

DamianMullins (approved):

PR #1001, opened 2021-05-28
Lerna versioning - DNM

siggerzz:


xander-marjoram:
@siggerzz Is this PR still required?

ashleynolan:
@xander-marjoram it will be eventually. Probably worth keeping it open for now, as we'll want to do this in Q4 I think (it's just a bigger task that we wanted to wait to release Icing before we looked at rolling out

siggerzz:
Closing this. Will look to see if this is worth picking up in the future.

PR #1002, opened 2021-06-01
f-checkout@0.127.0 - Ensure fulfilmentTime and table is sent for dinein

charliearlie:
### Changed
- Moved logic to show fulfilment time dropdown inside selector component to still set the time under the hood
- Passed correct params to mapper to ensure request is made correctly


packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -77,6 +78,10 @@ export default {
          */
         shouldShowPreOrderWarning () {
             return !this.availableFulfilment.isAsapAvailable;
+        },
+
+        shouldShowSelector () {
+            return this.serviceType !== 'dinein' || this.availableFulfilment.times.length > 1;
```
DamianMullins: Is it guaranteed that there will be no `availableFulfilment` times if the service type is dine-in?

If not it looks like this could return true if there are more than one `availableFulfilment` times and the service type is dine-in. This might actually be desired behaviour but thought I'd check 😸

charliearlie: We had a discussion in the Checkout team that we would like the behaviour of this selector to be server driven, instead of me having to alter code in the unlikely scenario that Dine In ever takes pre-orders. So we come to the agreement that if the service type is not dine in, we will always show the selector. If it is dine in, we will only show it if more than one fulfilment time is sent. Currently we are only expecting one, but now it's open to change in the future 🙂

DamianMullins: Sounds good, as long as it has been discussed I'm happy.

packages/components/organisms/f-checkout/src/components/_tests/Selector.test.js
```diff
@@ -189,6 +189,80 @@ describe('Selector', () => {
...
+                        }),
+                        i18n,
+                        localVue,
+                        propsData
+                    });
+                });
+
+                it('should return `false`', () => {
+                    // Assert
+                    expect(wrapper.vm.shouldShowSelector).toBeFalsy();
```
DamianMullins: `toBeFalse`? Looks like `shouldShowSelector` should return a boolean value rather than a truth or falsy value?

packages/components/organisms/f-checkout/src/components/Selector.vue
```diff
@@ -78,6 +79,10 @@ export default {
...
```
FayeCarter: Could you use `CHECKOUT_METHOD_DINEIN` here?

FayeCarter (approved):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -580,10 +576,12 @@ export default {
                     address: this.address,
                     customer: this.customer,
                     isCheckoutMethodDelivery: this.isCheckoutMethodDelivery,
+                    isCheckoutMethodDineIn: this.isCheckoutMethodDineIn,
                     time: this.time,
                     userNote: this.userNote,
                     geolocation: this.geolocation,
-                    asap: this.hasAsapSelected
+                    asap: this.hasAsapSelected,
```
1itvinka, line 583:
```suggestion
                    isAsap: this.hasAsapSelected,
```
just a nitpick. ignore if you think that `asap` is better

charliearlie: I think `asap` is the property name sent to the server. And this is where not having trailing commas is bad, this line wasn't added as part of my work. It's an existing line.

DamianMullins (approved):

PR #1007, opened 2021-06-02
f-checkout@v0.129.0 - save userNote field to session storage

1itvinka:
v0.129.0
*June 2, 2021*

### Added
- Populate userNote field from session storage if it exists
- Save userNote field to session storage when checkout form is submitted

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -112,6 +112,7 @@
                         :label-text="$t(`userNote.${serviceType}.title`)"
                         input-type="textarea"
                         :placeholder="$t(`userNote.${serviceType}.placeholder`)"
+                        :value="userNote"
                         cols="30"
                         rows="7"
                         maxlength="200"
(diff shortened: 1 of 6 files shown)


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -528,6 +531,8 @@ export default {
          */
         async submitCheckout () {
             try {
+                this.saveUserNote();
```
kevinrodrigues, line 532: Is it worth checking that the notes field has a value before calling this? Just to avoid calling it when there's no change.

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
1itvinka: There can be a scenario when there was a user note in session storage, but a customer removed the note and hit "submit". In this case, we should update the session storage so it contains an empty string as a value. We can check if the value was changed, but, as for me, it will only make the code more complex without hardly any benefit.

DamianMullins (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
kevinrodrigues: Sorry i meant if the value existed (not changed, my bad).

DamianMullins (dismissed):

kevinrodrigues (approved):

dandel10n (approved):

siggerzz (approved):

PR #1008, opened 2021-06-02
f-tabs@0.6.0 - Change button type from submit to button

xander-marjoram:
### Changed
- Button type from `submit` (the default) to `button` to prevent submission when tabs are used inside a form.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/molecules/f-tabs/src/components/Tabs.vue
+++ b/packages/components/molecules/f-tabs/src/components/Tabs.vue
@@ -7,13 +7,13 @@
                 <button
                     v-for="({ name, title }, i) in tabs"
                     :key="i"
+                    type="button"
                     :class="[
                         { [$style['c-tabs-button--active']]: activeTab === name },
                         $style['c-tabs-button']
                     ]"
                     :data-test-id="`tab-button-${name}`"
-                    @click="selectTabIndex(name)"
-                >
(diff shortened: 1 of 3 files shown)


DamianMullins (approved):

packages/components/molecules/f-tabs/src/components/Tabs.vue
```diff
@@ -7,13 +7,13 @@
                 <button
```
kevinrodrigues, line 7: Should we be using f-button now seen as we're in this component.

minrwhite: f-button had specifc styling associated that we wanted to avoid here, I believe.

minrwhite (approved):

packages/components/molecules/f-tabs/src/components/Tabs.vue
xander-marjoram: Happy to address this in a later PR if we think it's worth updating.
It feels like the right solution is to have a button variant with the styling we want, rather than avoiding using f-button because of styling.

kevinrodrigues (approved):

PR #1012, opened 2021-06-03
fozzie-components@v3.38.0 - Optional chaining in Storybook

charliearlie:
## Added
- Optional Chaining which works inside Storybook by adding acorn resolution to package.json


1itvinka (dismissed):

charliearlie:
Okay so this works locally but the build is failing in the CI. Will look into this then reopen.

PR #1013, opened 2021-06-04
f-offers@v0.2.0 - Vuex module for offers header

matthewhardern:
This is a PR for the setup of the vuex store serving the search box in the offers header. I have added unit tests for the store module and the helper functions I have added. Helpers are from HomeWeb, I have moved them as part of the migration process for offers.

### DIFF
--- a/packages/components/organisms/f-offers/src/store/types/mutations.js
+++ b/packages/components/organisms/f-offers/src/store/types/mutations.js
@@ -1,0 +1,3 @@
+const EMIT_SEARCH_START = 'searchbox-focus';
+
+export default EMIT_SEARCH_START;
--- a/packages/components/organisms/f-offers/src/store/types/mutations.js
+++ b/packages/components/organisms/f-offers/src/store/types/mutations.js
@@ -1,3 +1,0 @@
-const EMIT_SEARCH_START = 'searchbox-focus';
-
-export default EMIT_SEARCH_START;
--- a/packages/components/organisms/f-offers/src/utils/helpers.js
+++ b/packages/components/organisms/f-offers/src/utils/helpers.js
@@ -2,22 +2,6 @@ import Cookies from 'js-cookie';
 import Trak from '@justeat/f-trak';
 
 /**
- * Function that sets the cookies using js-cookie
- * @param name
- * @param value
- * @param minutes
- */
-const setCookie = (name, value, minutes = 1) => {
-    let expires = '';
-    const millisecondsPerMinute = 60 * 1000;
-    if (minutes) {
-        const date = new Date();
-        expires = date.setTime(date.getTime() + (minutes * millisecondsPerMinute));
-    }
-    Cookies.set(name, (value || ''), { expires, path: '/' });
-};
-
-/**
  * Pushes to the GTM datalayer using the f-trak library
  * @param category
  * @param action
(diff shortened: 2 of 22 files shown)


packages/components/organisms/f-offers/src/utils/helpers.js
```diff
@@ -0,0 +1,24 @@
...
+    window.trak.event({
+        event: 'trackEvent',
+        category,
+        action,
+        label
+    });
+};
+
+export const setGtmEventCookie = (category, action, label) => {
+    setCookie('je-gtm-event', `${category}|${action}|${label}`);
```
minrwhite: I'd look at using `js-cookie` here. We're already using it as a peer dependency in `f-cookie-banner`, `f-braze-adapter` and `f-consumer-oidc`.

packages/components/organisms/f-offers/src/store/types/mutations.js
```diff
@@ -0,0 +1,3 @@
+const EMIT_SEARCH_START = 'searchbox-focus';
+
+export default EMIT_SEARCH_START;
```
minrwhite: It seems strange to me to be exporting as default a single value from a file that seems designed to export multiple as per `actions.js` above, even if for the moment it only contains one - is this to avoid a linting warning?

packages/components/organisms/f-offers/src/utils/helpers.js
```diff
@@ -0,0 +1,24 @@
...
+    if (minutes) {
+        const date = new Date();
+        date.setTime(date.getTime() + (minutes * millisecondsPerMinute));
+        expires = `; expires=${date.toUTCString()}`;
+    }
+    document.cookie = `${name}=${(value || '') + expires}; path=/`;
+};
+
+export const pushToDataLayer = (category, action, label) => {
+    window.trak.event({
```
minrwhite: We probably also want `f-trak` added as a peer dependency if we're going to use this.

packages/components/organisms/f-offers/src/utils/helpers.js
matthewhardern: thanks for the tip

packages/components/organisms/f-offers/src/store/types/mutations.js
matthewhardern: yes this is to avoid a linting warning, I could add a lint disable if that would be ok for this situation.

packages/components/organisms/f-offers/src/utils/helpers.js
matthewhardern: yes, I found out what this is now, I am looking into the package.

packages/components/organisms/f-offers/src/store/types/mutations.js
minrwhite: If you could, and please export the const as well as exprting default, if linting accepts that, to avoid having to change `import EMIT_SEARCH_START from ...` to `import { EMIT_SEARCH_START } from ...` when more mutations types are added.

Alternatively, it might read better to put all store-related constants into a single `constants.js` file and namespace them via `MUTATION_`, `ACTION_`, etc. style prefices.

DamianMullins: We had a similar issue on menu where we simply added an eslint disable comment until more exports were added that allowed us to keep these types of files consistent

```js
/* eslint-disable import/prefer-default-export */

export const EMIT_SEARCH_START = 'searchbox-focus';
```

matthewhardern: I have added a types file and namespaced as I liked that idea @benwhite-deltas

packages/components/organisms/f-offers/src/utils/helpers.js
matthewhardern: I have now included the f-trak library under peer dependency.

packages/components/organisms/f-offers/src/utils/helpers.js
matthewhardern: updated to now use `js-cookie`

minrwhite (approved):

zlatinivanov-je (approved):
LGTM

PR #1014, opened 2021-06-04
f-tabs@0.8.0 - Reverse animation direction, and don't change direction if the tab isn't changing

xander-marjoram:
For click (manually triggered) events, the method is being called twice, so on the second call it was seeing `newIndex` and `previousIndex` as equal, so the direction was set to `DIRECTION.LEFT` every time.

I couldn't see a nice way of making sure the method only gets called once (because the mechanism for programmatically triggered events is different), so instead I fixed it so that all of the logic only happens if the tab is actually being changed.

I also swapped the animation direction because it makes more visual sense (at least to me it does).

### Changed
- Reverse animation direction (when choosing a tab that is further left, its contents should slide in from the left and vice-versa).

### Fixed
- Don't *change* direction if tab selection isn't changing.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/molecules/f-tabs/src/components/Tabs.vue
+++ b/packages/components/molecules/f-tabs/src/components/Tabs.vue
@@ -89,9 +89,9 @@ export default {
             const newIndex = this.tabs.findIndex(t => t.name === name);
 
             if (newIndex > previousIndex) {
-                this.direction = DIRECTION.RIGHT;
-            } else {
                 this.direction = DIRECTION.LEFT;
+            } else {
+                this.direction = DIRECTION.RIGHT;
             }
 
             this.$emit('change', {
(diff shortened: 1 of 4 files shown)


packages/components/molecules/f-tabs/src/components/Tabs.vue
```diff
@@ -83,23 +83,23 @@ export default {
...
+
             const previousIndex = this.tabs.findIndex(t => t.name === this.activeTab);
             const newIndex = this.tabs.findIndex(t => t.name === name);
 
             if (newIndex > previousIndex) {
-                this.direction = DIRECTION.RIGHT;
-            } else {
                 this.direction = DIRECTION.LEFT;
+            } else {
+                this.direction = DIRECTION.RIGHT;
```
1itvinka: minor suggestion:
```
this.direction = newIndex > previousIndex ? DIRECTION.LEFT : DIRECTION.RIGHT;
```

xander-marjoram: I did try this (who doesn't like more concise code? 😄) but thought I'd keep it the same because it's _slightly_ easier to read

xander-marjoram: I used the need to run another build as an excuse to update this 👍🏻

dandel10n (approved):

matthewhardern (approved):

PR #1020, opened 2021-06-08
f-checkout@0.133.0 - separate page for 403 error

dandel10n:
This is a first PR for adding separate 403 error page. Showing the button  will be covered in the next PR.


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -145,7 +145,11 @@
             </card>
         </div>
 
-        <error-page v-else-if="shouldShowErrorPage" />
+        <error-page
+            v-else-if="shouldShowErrorPage"
+            :header="$t(`errorMessages.${getCheckoutError}.heading`)"
```
DamianMullins: What do you think about passing instead passing the checkout error type as a prop and building up the translation in the error page component instead?

dandel10n: Good point, I added the change :+1:

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -308,7 +311,8 @@ export default {
             tenantConfigs,
             hasCheckoutLoadedSuccessfully: true,
             shouldShowSpinner: false,
-            isLoading: false
+            isLoading: false,
+            hasAccessForbiddenError: false
```
SachaZvetelman: I might be wrong, but I think I preferred the approach you showed me when we were pairing of using the custom `AccessForbiddenError`. 
The problem I see with the current approach is that we can have mutually exclusive scenarios (e.g. `hasCheckoutLoadedSuccessfully && hasAccessForbiddenError`, and at the same time, if we want to handle another type of error in the future, this will grow significantly in complexity.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -443,6 +447,10 @@ export default {
 
         formattedMobileNumberForScreenReader () {
             return this.customer.mobileNumber ? [...this.customer.mobileNumber].join(' ') : '';
+        },
+
+        getCheckoutError () {
```
SachaZvetelman: It should probably be called `getCheckoutErrorType` given you're using it as a type in the template.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -685,6 +693,9 @@ export default {
 
                 this.$emit(EventNames.CheckoutGetSuccess);
             } catch (error) {
+                if (error.response.status === 403) {
```
SachaZvetelman: Do we always get an `error.response`? It's worth checking, just in case.

SachaZvetelman (commented):
Have you added tests for the new functionality (403 error code handling)? I might have missed it.

packages/components/organisms/f-checkout/src/components/Checkout.vue
dandel10n: Sure, I've updated the code :white_check_mark:

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -712,15 +711,12 @@ export default {
...
-                    message: 'Get Checkout Basket Failure',
-                    data: this.eventData,
-                    logMethod: this.$logger.logError,
-                    error
-                });
+                if (error.response && error.response.status === 403) {
+                    this.handleErrorState(new AccessForbiddenError(error.message, CHECKOUT_LOADED_TYPE.basket, error.response.status));
+                } else {
+                    this.handleErrorState(new DefaultGetCheckoutError(error.message, CHECKOUT_LOADED_TYPE.basket, error.response.status));
+                }
```
SachaZvetelman: Do we need this here? (a genuine question, because I'm not sure).

packages/components/organisms/f-checkout/src/exceptions/exceptions.js
```diff
@@ -35,8 +36,31 @@ class PlaceOrderError extends Error {
     }
 }
 
+class DefaultGetCheckoutError extends Error {
+    constructor (message, loadedComponent, errorCode) {
+        super(message);
+        this.messageKey = 'errorMessages.pageLoad.description';
+        this.eventToEmit = loadedComponent === CHECKOUT_LOADED_TYPE.checkout ? EventNames.CheckoutGetFailure : EventNames.CheckoutBasketGetFailure;
```
SachaZvetelman: I'm not sure I understand this. The custom error is for `GetCheckout`, but then it's used for handling basket failures as well?
This is linked to my question above re the need to handle 403s in the get basket call.

dandel10n: So, for logged in users loadCheckout() function fires to load the checkout. For logged out users loadBasket() function fires to load the checkout instead. That's why I had to add it in both functions to make it work for both logged in and guest checkouts

tommcclean (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
dandel10n: Changed as discussed privately

packages/components/organisms/f-checkout/src/exceptions/exceptions.js
```diff
@@ -35,8 +35,57 @@ class PlaceOrderError extends Error {
...
+        this.messageKey = 'errorMessages.pageLoad.description';
+        this.eventToEmit = EventNames.CheckoutAvailableFulfilmentGetFailure;
+        this.logMessage = 'Get Checkout Available Fulfilment Times Failure';
+        this.errorCode = errorCode;
+        this.shouldShowInDialog = false;
+        this.errorFormType = 'pageLoad';
+    }
+}
+
+class GetCheckoutLoadBasketError extends Error {
```
SachaZvetelman: I'd rename this to `GetBasketError`, since technically it doesn't have anything to do with the checkout bits :)

packages/components/organisms/f-checkout/src/exceptions/exceptions.js
```diff
@@ -35,8 +35,57 @@ class PlaceOrderError extends Error {
     }
 }
 
+class DefaultGetCheckoutError extends Error {
```
SachaZvetelman: Should we rename this to `GetCheckoutError`?

packages/components/organisms/f-checkout/src/exceptions/exceptions.js
```diff
@@ -35,8 +35,57 @@ class PlaceOrderError extends Error {
...
+        this.errorFormType = 'pageLoad';
+    }
+}
+
+class GetCheckoutLoadBasketError extends Error {
+    constructor (message, errorCode) {
+        super(message);
+        this.messageKey = 'errorMessages.pageLoad.description';
+        this.eventToEmit = EventNames.CheckoutBasketGetFailure;
+        this.logMessage = 'Get Checkout Basket Failure';
```
SachaZvetelman: Same here: `Get Basket Failure`.

packages/components/organisms/f-checkout/stories/Checkout.stories.js
```diff
@@ -49,21 +51,34 @@ CheckoutMock.setupCheckoutMethod(updateCheckoutServerErrorUrl);
...
 
 const checkoutIssues = 'Checkout Issues (Response from server but order not fulfillable)';
 const checkoutServerError = 'Checkout Error (Response from server is an error)';
 const placeOrderError = 'Place Order Duplicate Error (Response from server is an error)';
+const accessForbiddenError = 'Access Forbidden Get Checkout Error (Response from server is an error)';
+const defaultGetCheckoutError = 'Any other Get Checkout Error (Response from server is an error)';
 const ISSUES = 'ISSUES';
 const SERVER = 'SERVER';
+const accessForbiddenErrorCode = '403';
+const defaultGetCheckoutErrorCode = 'default';
```
SachaZvetelman: This looks a bit weird, since it says `ErrorCode` but then it says `default`.

packages/components/organisms/f-checkout/src/exceptions/exceptions.js
```diff
@@ -35,8 +35,57 @@ class PlaceOrderError extends Error {
...
 
+class DefaultGetCheckoutError extends Error {
+    constructor (message, errorCode) {
+        super(message);
+        this.messageKey = 'errorMessages.pageLoad.description';
+        this.eventToEmit = EventNames.CheckoutGetFailure;
+        this.logMessage = 'Get Checkout Failure';
+        this.errorCode = errorCode;
+        this.shouldShowInDialog = false;
+        this.errorFormType = 'pageLoad';
```
SachaZvetelman: Should we extract this `pageLoad` to the constants file? Since we're using it in many places.

packages/components/organisms/f-checkout/stories/Checkout.stories.js
dandel10n: It is used for all errors other than 403 so I came up with 'default'. Can you suggest something more logical for it?

SachaZvetelman: A standard `500` error should be "default enough" in my opinion, hehe.

packages/components/organisms/f-checkout/src/demo/checkout-500-get-error.json
```diff
@@ -0,0 +1,9 @@
+{
+"traceId": "0HLOCKDKQPKIU",
+"faultId": "72d7036d-990a-4f84-9efa-ef5f40f6044b",
+"errors": [
+    {
+        "description": "Get Checkout Error response"
+    }
+]
```
SachaZvetelman: Nit: this should be indented.

packages/components/organisms/f-checkout/src/demo/checkoutMock.js
```diff
@@ -15,6 +15,8 @@ import checkoutServerError from './checkout-server-error.json';
 import getAddress from './get-address.json';
 import placeOrder from './place-order.json';
 import placeOrderDuplicate from './place-order-duplicate.json';
+import accessForbiddenError from './checkout-403-get-error.json';
+import GetCheckoutError from './checkout-500-get-error.json';
```
SachaZvetelman: `getCheckoutError` :)

packages/components/organisms/f-checkout/src/demo/checkoutMock.js
```diff
@@ -67,6 +69,12 @@ export default {
             case '/place-order-duplicate.json':
                 mock.onPost(path).reply(400, placeOrderDuplicate);
                 break;
+            case '/checkout-403-get-error.json':
+                mock.onGet(path).reply(403, accessForbiddenError);
+                break;
+            case '/checkout-500-get-error.json':
+                mock.onGet(path).reply(400, GetCheckoutError);
```
SachaZvetelman: We should return a `500` if the file is a `500` simulation.

packages/components/organisms/f-checkout/stories/Checkout.stories.js
```diff
@@ -49,21 +51,34 @@ CheckoutMock.setupCheckoutMethod(updateCheckoutServerErrorUrl);
...
+CheckoutMock.setupCheckoutMethod(accessForbiddenErrorUrl);
+CheckoutMock.setupCheckoutMethod(getCheckoutErrorUrl);
 CheckoutMock.setupCheckoutMethod(getGeoLocationUrl);
 CheckoutMock.passThroughAny();
 
 const checkoutIssues = 'Checkout Issues (Response from server but order not fulfillable)';
 const checkoutServerError = 'Checkout Error (Response from server is an error)';
 const placeOrderError = 'Place Order Duplicate Error (Response from server is an error)';
+const accessForbiddenError = 'Access Forbidden Get Checkout Error (Response from server is an error)';
+const GetCheckoutError = 'Any other Get Checkout Error (Response from server is an error)';
```
SachaZvetelman: `getCheckoutError`

packages/components/organisms/f-checkout/stories/Checkout.stories.js
```diff
@@ -49,21 +51,34 @@ CheckoutMock.setupCheckoutMethod(updateCheckoutServerErrorUrl);
...
 const ISSUES = 'ISSUES';
 const SERVER = 'SERVER';
+const accessForbiddenErrorCode = '403';
+const GetCheckoutErrorCode = '500';
```
SachaZvetelman: `getCheckoutErrorCode`

SachaZvetelman (approved):

charliearlie (approved):

PR #1021, opened 2021-06-09
f-alert@v0.6.1 - Fix styling issues while rendering the alert header

tommcclean:
v0.6.1
*June 09, 2021*

### Fixed
- Alert icon no longer resizes as content around it grows
- Warning icon no longer receives greater margin than the other alert types
- Items in the header are no longer vertically aligned as the content may exceed one line
- Added some spacing below the title so that the component doesn't feel so mushed

_Before_
image
image

_After_
image
image

### DIFF
--- a/packages/components/molecules/f-alert/src/components/Alert.vue
+++ b/packages/components/molecules/f-alert/src/components/Alert.vue
@@ -49,7 +49,11 @@ import {
     SuccessIcon,
     WarningIcon
 } from '@justeat/f-vue-icons';
-import { globalisationServices } from '@justeat/f-services';
+
+import {
+    globalisationServices
+} from '@justeat/f-services';
+
 import FButton from '@justeat/f-button';
 import tenantConfigs from '../tenants';
 import '@justeat/f-button/dist/f-button.css';
@@ -146,6 +150,7 @@ $alert-borderRadius: $border-radius;
     .c-alert-heading {
         @include font-size(subheading-s);
         margin-top: -1px;
+        margin-bottom: spacing(x0.5);
     }
 
     .c-alert-content {
--- a/packages/components/molecules/f-alert/src/components/Alert.vue
+++ b/packages/components/molecules/f-alert/src/components/Alert.vue
@@ -120,7 +120,6 @@ $alert-borderRadius: $border-radius;
 }
     .c-alert-headingContainer {
         display: flex;
-        align-items: center;
         padding: spacing(x0.5) 0 0;
     }
 
@@ -155,13 +154,11 @@ $alert-borderRadius: $border-radius;
     }
 
     .c-alert-icon {
+        min-width: 20px;
+        width: 20px;
         height: 20px;
-        margin: 0 spacing(x1.5) 0 spacing();
+        margin: 5px spacing(x1.5) 0 spacing();
     }
-
-        .c-alert-icon--warning {
-            margin-right: 10px;
-        }
 
     .c-alert-dismiss {
         text-indent: 0;
--- a/packages/components/molecules/f-alert/CHANGELOG.md
+++ b/packages/components/molecules/f-alert/CHANGELOG.md
@@ -4,6 +4,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.6.1
+------------------------------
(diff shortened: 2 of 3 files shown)


packages/components/molecules/f-alert/src/components/Alert.vue
```diff
@@ -155,14 +154,12 @@ $alert-borderRadius: $border-radius;
     }
 
     .c-alert-icon {
+        min-width: 20px;
```
matthewhardern, line 162: would setting min-width the same as width make it redundant?

tommcclean: No, it seemed to be required to stop the icon from shrinking, I think the flex was essentially overriding the main width when the text was larger than one line. By setting a min width it never shrinks below that size; and vice versa.

matthewhardern: ok makes sense

matthewhardern (dismissed):

1itvinka:
It doesn't feel right that the `line-height` of both the title and the text is bigger than the spacing between these blocks. As far as I see in the design, the spacing between the elements is slightly bigger. https://www.figma.com/file/6FhFpY7PXuxE5q9X1gKu4K/PIE-Global---JE-Light?node-id=782%3A1302. And I would reduce the line-height of the title a bit. I can't find the design for this though, so we can probably use our own judgement :)

tommcclean:
> It doesn't feel right that the `line-height` of both the title and the text is bigger than the spacing between these blocks. As far as I see in the design, the spacing between the elements is slightly bigger. https://www.figma.com/file/6FhFpY7PXuxE5q9X1gKu4K/PIE-Global---JE-Light?node-id=782%3A1302. And I would reduce the line-height of the title a bit. I can't find the design for this though, so we can probably use our own judgement :)

The line heights themselves are pulled from fozzie - not specified in the component so hesistant to override them, though they do match up with the design in pixel amounts. I agree they look a little too close - it doesn't translate well from the design; it seems adding some `margin-bottom: spacing(x0.5);` to the title looks subtly nicer in both long form and short forms.

_Without margin_
image

_With margin_
image

1itvinka:
> > It doesn't feel right that the `line-height` of both the title and the text is bigger than the spacing between these blocks. As far as I see in the design, the spacing between the elements is slightly bigger. https://www.figma.com/file/6FhFpY7PXuxE5q9X1gKu4K/PIE-Global---JE-Light?node-id=782%3A1302. And I would reduce the line-height of the title a bit. I can't find the design for this though, so we can probably use our own judgement :)
> 
> The line heights themselves are pulled from fozzie - not specified in the component so hesistant to override them, though they do match up with the design in pixel amounts. I agree they look a little too close - it doesn't translate well from the design; it seems adding some `margin-bottom: spacing(x0.5);` to the title looks subtly nicer in both long form and short forms.
> 
> _Without margin_
> image
> 
> _With margin_
> image

yeah, with margin it looks much better. 👍

1itvinka (approved):

SachaZvetelman:
Are the designs correct? The title seems to be bigger than the content, and according to the designs, for the size of these notifications, I don't think it should be the case, look: 
image

tommcclean:
> Are the designs correct? The title seems to be bigger than the content, and according to the designs, for the size of these notifications, I don't think it should be the case, look:
> image

The font size for the heading is larger in the designs by 4 pixels - equates to `subheading-s` (20 pixels) compared to `body-l` (16 pixels) - it must just look different when you render it on the page and also take account of any deviance in our global typography from Fozzie.

[The typography styles match the design identically as far as I can see](https://github.com/justeat/fozzie/blob/99280c21e6c9c8baa54604ae2c670fd7583db4c8/src/scss/settings/_variables.scss)

<hr>

Hmm on reflection I can see what you are saying, the design actually has different font sizes for each one going horizontally. We can't use media queries to detect that; as its the container that is getting shorter not the viewport. We also can't allow the developer to choose the size as it wouldn't be right for every display type. 🤔 This is currently matching every variant except the smallest one because checkout is not very wide (and fozzie typography doesn't scale down either currently with the viewport)

tommcclean:
As discussed ill spin up a ticket to work on the typography stuff seperately; looks like they need to scale down more than they do at mobile breakpoints. Though that would only affect the fonts when on a XS device. They would look the same on desktop as they do now.

matthewhardern (approved):

PR #1023, opened 2021-06-09
f-checkout@0.134.0 - Add error code to analytics error message

charliearlie:
### Added
- Error code to analytics when errors occur for a more descriptive error

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -793,7 +793,7 @@ export default {
             const message = this.$t(error.messageKey) || this.$t('errorMessages.genericServerError');
             const eventToEmit = error.eventToEmit || EventNames.CheckoutFailure;
             const logMessage = error.logMessage || 'Consumer Checkout Failure';
-            const errorName = error.errorCode || '';
+            const errorName = error.errorCode ? `${error.errorCode}-` : ''; // This appends the hyphen so it doesn't appear in the logs when the error name does not exist
 
             this.$emit(eventToEmit, { ...this.eventData, error });
 
@@ -804,7 +804,8 @@ export default {
                 error
             });
 
-            this.trackFormInteraction({ action: 'error', error: `error_${errorName}-${error.message}` });
+
+            this.trackFormInteraction({ action: 'error', error: `error_${errorName}${error.message}` });
 
             if (!error.shouldShowInDialog && !error.errorFormType) {
                 this.updateMessage(message);
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -811,6 +811,7 @@ export default {
             const message = this.$t(error.messageKey) || this.$t('errorMessages.genericServerError');
             const eventToEmit = error.eventToEmit || EventNames.CheckoutFailure;
             const logMessage = error.logMessage || 'Consumer Checkout Failure';
+            const errorName = error.errorCode || '';
 
             this.$emit(eventToEmit, { ...this.eventData, error });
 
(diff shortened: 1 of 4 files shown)


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -811,6 +811,7 @@ export default {
             const message = this.$t(error.messageKey) || this.$t('errorMessages.genericServerError');
             const eventToEmit = error.eventToEmit || EventNames.CheckoutFailure;
             const logMessage = error.logMessage || 'Consumer Checkout Failure';
+            const errorName = error.errorCode || '';
```
SachaZvetelman: Isn't this an `errorCode` rather than an `errorName`?

charliearlie: Going by the name inside the error object, I'd say yes.. But the actual value isn't a code. The value is the name of the error, genuinely.

For a `PlaceOrderError` the value will be `PlaceOrderError` 😄

charliearlie: But happy to change to remove the confusion from this line.

SachaZvetelman: Sad days 😢, haha. I'll leave it up to you 😄

simonsandersje (dismissed):

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -2407,12 +2407,23 @@ describe('Checkout', () => {
...
+                error.errorCode = 'OrderError';
+
+                // Act
+                wrapper.vm.handleErrorState(error);
+
+                // Assert
+                expect(trackFormInteractionSpy).toHaveBeenCalledWith({ action: 'error', error: `error_${error.errorCode}-${error.message}` });
+            });
+
             it('should call `trackFormInteraction` with the error information', () => {
```
SachaZvetelman: Could we refactor this to make the test name for explicit? 
```should call `trackFormInteraction` without an `errorCode` when it doesn't exist```

And then I'd do an `// Arrange` with `error.errorCode = null` or similar.

File: packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js [hunk 3]
```diff
@@ -2407,12 +2407,23 @@ describe('Checkout', () => {
...
                 // Act
                 wrapper.vm.handleErrorState(error);
 
                 // Assert
-                expect(trackFormInteractionSpy).toHaveBeenCalledWith({ action: 'error', error: `error_${error.message}` });
+                expect(trackFormInteractionSpy).toHaveBeenCalledWith({ action: 'error', error: `error_-${error.message}` });
```
SachaZvetelman: Nit: this looks a bit weird, since it will output something like `error_-MyMessage`.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -793,6 +793,7 @@ export default {
             const message = this.$t(error.messageKey) || this.$t('errorMessages.genericServerError');
             const eventToEmit = error.eventToEmit || EventNames.CheckoutFailure;
             const logMessage = error.logMessage || 'Consumer Checkout Failure';
+            const errorName = error.errorCode ? `${error.errorCode}-` : ''; // This appends the hyphen so it doesn't appear in the logs when the error name does not exist
```
charliearlie, line 796: I wasn't a big fan of adding the hyphen here but the errorName is only used in one place and I'd probably have to write something a bit over the top to do it another way.

SachaZvetelman (approved):

matthewhardern (approved):

PR #1027, opened 2021-06-09
f-button@1.7.0 Add Vue router link support

zlatinivanov-je:
_Following a regression, page reload, when replacing existing Vue `router-link` project instances with Fozzie `f-button` required ability to extend to support those._

### Added

 - Support of Vue `router-link` when `to` prop has value
 - Test coverage 
 - Story book section for Router Link example

Deployed to QA11 for preview - https://ie-publicweb-haproxy-qa11.bapps.je-labs.com/help 

---

## UI Review Checks


packages/components/atoms/f-button/package.json
```diff
@@ -45,15 +45,19 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "peerDependencies": {
-    "@justeat/browserslist-config-fozzie": ">=1.1.1"
+    "@justeat/browserslist-config-fozzie": ">=1.1.1",
+    "vue": "2.x",
+    "vue-router": "^3.4.8"
```
DamianMullins: Slightly different syntax require here.

```suggestion
    "vue-router": ">=3.4.8"
```

packages/components/atoms/f-button/src/components/RouterLinkButton.vue
```diff
@@ -0,0 +1,18 @@
...
+        <slot />
+    </router-link>
+</template>
+
+<script>
+export default {
+    name: 'RouterLinkButton',
+    props: {
+        to: {
+            type: String,
```
DamianMullins: The router link `to` attribute [can be either a string or a location object](https://router.vuejs.org/api/#to).

packages/components/atoms/f-button/package.json
```diff
@@ -46,15 +46,19 @@
...
+    "@justeat/f-wdio-utils": "0.2.0",
     "@samhammer/vue-cli-plugin-stylelint": "2.0.1",
     "@vue/cli-plugin-babel": "4.4.6",
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.4.6",
     "@vue/test-utils": "1.0.3",
-    "@justeat/f-wdio-utils": "0.2.0",
-    "node-sass-magic-importer": "5.3.2"
+    "node-sass-magic-importer": "5.3.2",
+    "vue": "2.x",
```
dandel10n: As f-button is a part of monorepo vue should be available from the root yarn.lock so there shouldn't be a need to add it here.

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -203,6 +205,38 @@ describe('Button', () => {
...
+
+                it('should render `Router Link` component', () => {
+                    // Act
+                    const wrapper = mount(FButton, { propsData });
+
+                    // Assert
+                    expect(wrapper.find('[data-test-id="router-link-button-component"]').exists()).toBeTruthy();
+                });
+
+                it('should apply `href` attribute', () => {
```
dandel10n:
```suggestion
                it('should apply `to` attribute', () => {
```

dandel10n (commented):
Can you `yalc` local version of f-button to the consuming app to confirm it is working as expected instead of publishing beta version?

packages/components/atoms/f-button/src/components/RouterLinkButton.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+    <router-link
+        :to="to">
```
FayeCarter: Would this be applied automatically with this? 
https://github.com/justeat/fozzie-components/blob/ebf1eac7099bde14a983807c532adf17bcd1f89f/packages/components/atoms/f-button/src/components/Button.vue#L14

packages/components/atoms/f-button/package.json
zlatinivanov-je: Added as this is required by the Storybook as an import... followed example from `f-searchbox`  https://github.com/justeat/fozzie-components/blob/8715e261e368b397b7b57b3998fe163444d67478/packages/components/molecules/f-searchbox/package.json#L74 

Let me see the implications when removed.

packages/components/atoms/f-button/src/components/RouterLinkButton.vue
zlatinivanov-je: Does not seem to work, it breaks the storybook. I assume it is because `to` is not a standard attribute.

packages/components/atoms/f-button/package.json
zlatinivanov-je: done

packages/components/atoms/f-button/src/components/RouterLinkButton.vue
zlatinivanov-je: Screenshot 2021-06-15 at 16 27 52
Hope that helps

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -149,10 +150,11 @@ describe('Button', () => {
         describe('componentType :: ', () => {
             let propsData;
 
-            describe('when `href` prop is empty :: ', () => {
+            describe('when `href` and `to` prop is empty :: ', () => {
```
ashleynolan:
```suggestion
            describe('when `href` and `to` props are empty :: ', () => {
```

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -237,6 +280,45 @@ describe('Button', () => {
                     expect(wrapper.attributes('type')).toBeUndefined();
                 });
             });
+
+            // RouterLink type -> incl. `to` attribute
+            describe('when `to` prop is empty :: ', () => {
+                it('should add `type` attribute', () => {
```
ashleynolan: More a Q than a comment – is this `type="button"` that gets added?

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -237,6 +280,45 @@ describe('Button', () => {
...
+                    // Arrange
+                    const propsData = {
+                        to: null,
+                        actionType
+                    };
+
+                    // Act
+                    const wrapper = mount(FButton, {
+                        propsData,
+                        stubs: ['router-link']
```
ashleynolan: You likely don't need the stub in this test I think.

packages/components/atoms/f-button/stories/RouterButton.stories.js
```diff
@@ -0,0 +1,79 @@
...
+    select,
+    boolean,
+    text
+} from '@storybook/addon-knobs';
+import { withA11y } from '@storybook/addon-a11y';
+import FButton from '../src/components/Button.vue';
+
+Vue.use(Router);
+
+const Foo = { template: '<div>foo</div>' };
```
ashleynolan: Does this just load in some example routes? If so, I'd be inclined to use `Example Route 1` & `Example Route 2` rather than `Foo` and `Bar` for the content.

packages/components/atoms/f-button/src/components/_tests/Button.test.js
zlatinivanov-je: Yes, in this case of `to` prop empty, it renders the default `<button>` element with type attribute. 
Made me think about that and had to go an check, to realise what is the case. :)

packages/components/atoms/f-button/src/components/_tests/Button.test.js
zlatinivanov-je: I do get an error in the tests. `[Vue warn]: Unknown custom element: <router-link> ... ` 
https://vue-test-utils.vuejs.org/guides/using-with-vue-router.html#using-stubs

packages/components/atoms/f-button/stories/RouterButton.stories.js
zlatinivanov-je: Updated accordingly

packages/components/atoms/f-button/src/components/_tests/Button.test.js
zlatinivanov-je: Found an issue when I dug up deeper, an attribute of `action-type` was added. Now fixed. 
Screenshot 2021-06-15 at 17 08 58

dandel10n: Should this test be the same as one for `href` checking for both `type` and `action-type`? I think you can use [describe.each()](https://jestjs.io/docs/api#describeeachtablename-fn-timeout) in this case to remove duplication. Like: 
```
describe.each([
  ["to"],
  ["href"]
])('when %s prop is not an empty string, (attr) => {...})

File: packages/components/atoms/f-button/src/components/Button.vue [hunk 10]
```diff
@@ -79,16 +79,25 @@ export default {
...
+            }
+
+            if (this.$attrs.to) {
+                return 'router-link';
+            }
+
+            return 'action-button';
         },
         /**
          * Passes `actionType` prop to action button if no `href` attribute is applied to the component
```
juan-amor: Update comment with `no "href" or "to" attributes` to catch the logic below

packages/components/atoms/f-button/stories/RouterButton.stories.js
```diff
@@ -0,0 +1,82 @@
+import Vue from 'vue';
+import Router from 'vue-router';
+
+import {
+    withKnobs,
+    select,
+    boolean,
+    text
```
dandel10n: text can be removed as not in use anymore

juan-amor (commented):
Just a very minor change, but LGTM

packages/components/atoms/f-button/stories/RouterButton.stories.js
zlatinivanov-je: done

packages/components/atoms/f-button/src/components/Button.vue
zlatinivanov-je: done

ashleynolan (dismissed):

dandel10n (approved):

ashleynolan (approved):

PR #1034, opened 2021-06-11
f-checkout@0.136.0 - redirectToMenu button for error page

dandel10n:
Allows navigation back to the menu page when errors occur. `Go back to order` will take you back to the menu you were previously on. We also delete the menu cookie `je-mw-basket-xxx` if it exists before navigating back.

### Added
- Button for error pages.
- Tests to cover button changes.
- Added log warn when error component is displayed.

### Screenshots

Screenshot 2021-06-14 at 14 44 42

Screenshot 2021-06-15 at 10 27 26

#### Navigation back via the button:
Screenshot 2021-06-14 at 14 46 45


packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -955,46 +964,24 @@ describe('Checkout', () => {
             });
         });
 
-        describe('dialogMessage ::', () => {
```
dandel10n, line 958: Removed `dialogMessage` and `alertMessage` tests as they are duplicated in other tests on lines 902 and 931

packages/components/organisms/f-checkout/src/components/_tests/ErrorDialog.test.js
```diff
@@ -76,24 +81,6 @@ describe('ErrorDialog', () => {
             });
         });
 
-        describe('restaurantMenuPageUrl ::', () => {
```
dandel10n, line 79: Moved the test to Checkout.tests.js as now the computed prop lives there and not in the ErrorDialog.vue

packages/components/organisms/f-checkout/src/components/Error.vue
```diff
@@ -22,23 +21,74 @@
...
+    },
+
+    computed: {
+        ...mapState(VUEX_CHECKOUT_MODULE, [
+            'basket'
+        ])
+    },
+
+    mounted () {
+        if (this.$logger) {
```
SachaZvetelman: I think it's safe to assume we have the `$logger` here. At least we're assuming that in the rest of `f-checkout`.

packages/components/organisms/f-checkout/src/components/Error.vue
```diff
@@ -22,23 +21,74 @@
...
+                message: 'Consumer Checkout Error Page',
+                data: {},
+                logMethod: this.$logger.logWarn
+            });
+        }
+    },
+
+    methods: {
+        redirectToMenu () {
+            const basketCookie = this.$cookies.get(`je-mw-basket-${this.basket.id}`);
```
SachaZvetelman: Nit: since we're using it twice, we could maybe extract the cookie name into a `const`?

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
```diff
@@ -1,26 +1,56 @@
...
+import { CHEKOUT_ERROR_FORM_TYPE, VUEX_CHECKOUT_MODULE } from '../../constants';
 
 const localVue = createLocalVue();
 
 localVue.use(VueI18n);
+localVue.use(Vuex);
+
+const mockState = {
+    basket: {
+        id: 'menu-sentinel-6-freilich',
```
SachaZvetelman: I know this is a test, but could we please try to use a similar-to-prod example?

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
```diff
@@ -42,4 +72,70 @@ describe('Error', () => {
...
+                            configurable: true,
+                            value: jest.fn()
+                        }
+                    }
+                );
+            });
+
+            afterAll(() => {
+                window.location = oldWindowLocation;
+            });
```
SachaZvetelman, line 93: This seems quite complicated to configure the spy for `window.location.assign`.
We have a few examples in `Checkout.test` already, but in a nutshell, this should do it:

```
const windowLocationSpy = jest.spyOn(window.location, 'assign').mockImplementation();

// Do something

expect(windowLocationSpy).toHaveBeenCalledWith('...');

File: packages/components/organisms/f-checkout/src/components/_tests/Error.test.js [hunk 7]
```diff
@@ -42,4 +72,70 @@ describe('Error', () => {
...
+
+            it('should exist', () => {
+                expect(wrapper.vm.redirectToMenu).toBeDefined();
+            });
```
SachaZvetelman: I think we discussed this with some of the team members and we decided not to add these tests anymore, since we don't think it adds any value. If a method/computed is not defined, all the associated tests will fail.

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
```diff
@@ -42,4 +72,70 @@ describe('Error', () => {
...
+
+            describe('when invoked', () => {
+                it('should assign the window location to the restaurant menu page url so the user can navigate back from the error', () => {
```
SachaZvetelman: Maybe `'should redirect the customer back to the menu'`?

packages/components/organisms/f-checkout/src/components/Error.vue
kevinrodrigues: Sure.

packages/components/organisms/f-checkout/src/components/Error.vue
kevinrodrigues: For some reason the tests were throwing errors so added this in. Do we use the logger in the mounted hook in checkout elsewhere?

simonsandersje (dismissed):

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
kevinrodrigues: This won't work, originally did this and it failed. It's because jsdom tries to assign a location and it then produces errors in the test.

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
kevinrodrigues: Disagree but can remove. I think having the tests fail as early as possible has the advantage of being able to see why a test fails in a more simpler format (it doesn't exist) rather than a more complex failure further down the chain.

packages/components/organisms/f-checkout/src/components/Error.vue
SachaZvetelman: Not sure if we do, but it should be available on `mounted`, so I'd say it might be a missing mock or similar in one of the tests.

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
kevinrodrigues: 👍

packages/components/organisms/f-checkout/src/components/Error.vue
kevinrodrigues: Did add a mock for $logger [here](https://github.com/justeat/fozzie-components/pull/1034/files#diff-77d6d5e41e52cc7565e75f71e027e67a362fae9f1f6e0c8954c6b95ea57f8d50R45)

packages/components/organisms/f-checkout/src/components/Error.vue
kevinrodrigues: Done.

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
kevinrodrigues: Done.

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
kevinrodrigues: Done

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
kevinrodrigues: Done.

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
```diff
@@ -1,26 +1,56 @@
...
+localVue.use(Vuex);
+
+const mockState = {
+    basket: {
+        id: 'MGFkMzgwZjgtYjY3Yi00Nz-v1',
+        total: 0
+    }
+};
+
+const createStore = (state = mockState) => new Vuex.Store({
```
FayeCarter: Could you use the helper file for this? https://github.com/justeat/fozzie-components/blob/master/packages/components/organisms/f-checkout/src/components/_tests/helpers/setup.js#L104

packages/components/organisms/f-checkout/stories/Checkout.stories.js
```diff
@@ -26,7 +26,8 @@ const getBasketDeliveryUrl = '/get-basket-delivery.json';
 const getBasketCollectionUrl = '/get-basket-collection.json';
 const getBasketDineInUrl = '/get-basket-dinein.json';
 const updateCheckoutUrl = '/update-checkout.json';
-const updateCheckoutErrorsUrl = '/update-checkout-errors.json';
+const updateCheckoutRestaurantNotTakingOrdersUrl = '/update-checkout-restaurant-not-taking-orders.json';
+const updateCheckoutAdditionalItemsRequaredUrl = '/update-checkout-additional-items-requared.json';
```
FayeCarter: super nit
`/update-checkout-additional-items-required.json`?

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
kevinrodrigues: Fixed.

packages/components/organisms/f-checkout/stories/Checkout.stories.js
kevinrodrigues: Fixed.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -445,6 +436,10 @@ export default {
 
         formattedMobileNumberForScreenReader () {
             return this.customer.mobileNumber ? [...this.customer.mobileNumber].join(' ') : '';
+        },
+
+        restaurantMenuPageUrl () {
+            return `restaurants-${this.restaurant.seoName}/menu`;
         }
```
SachaZvetelman: Is there any specific benefit to having the `restaurantMenuPageUrl` here and injecting it via a `prop`? `Error.vue` could be the one doing this, since the information needed is in the Vuex store.

packages/components/organisms/f-checkout/src/components/Error.vue
```diff
@@ -22,23 +21,75 @@
...
+            this.logInvoker({
+                message: 'Consumer Checkout Error Page',
+                data: {},
+                logMethod: this.$logger.logWarn
+            });
+        }
+    },
+
+    methods: {
+        redirectToMenu () {
```
SachaZvetelman, line 85: I see we're always deleting the basket cookie here if it exists, but we should only delete the basket cookie when we get a 403 error, not in other cases.

simonsandersje (dismissed):

SachaZvetelman (commented):
The screenshots are showing the error pages have different titles. As far as I know, both titles should be "Something went wrong".

As a side note, when testing (by ad-hoc'ing to CoreWeb), can you please verify that if a user has multiple basket cookies, everything works as expected? We have to delete only the current basket, but the user might have other cookies (I've seen that in the past), so I want to be sure it will still work as expected 😄

packages/components/organisms/f-checkout/src/components/Checkout.vue
kevinrodrigues: Ill have a check as this was already written before i took over the pr.

packages/components/organisms/f-checkout/src/components/Error.vue
kevinrodrigues: Will check but according to Maryia this was working as expect, only the tests and logging was left to do so I assumed this was fine. Ill re-check tho.

packages/components/organisms/f-checkout/src/components/Checkout.vue
dandel10n: @SachaZvetelman @kevinrodrigues I moved it from ErrorDialog.vue as now it is in use in two places (Error.vue and ErrorDialog.vue)

kevinrodrigues: Thanks!

packages/components/organisms/f-checkout/src/components/Error.vue
dandel10n: @SachaZvetelman @kevinrodrigues if user gets the error page and we advice them to return to the menu page, we want it to be a fresh menu page without saved basket, so I added a check that if there is a basket id in cookies than remove it, if not just redirect to menu page.

kevinrodrigues: I re-checked the ACs and it does say delete the cookie when there's a 403? Any other error would remove the basket when going back to menu. Let me know if we should specifically check for the 403 here first.

dandel10n:
> The screenshots are showing the error pages have different titles. As far as I know, both titles should be "Something went wrong".
As a side note, when testing (by ad-hoc'ing to CoreWeb), can you please verify that if a user has multiple basket cookies, everything works as expected? We have to delete only the current basket, but the user might have other cookies (I've seen that in the past), so I want to be sure it will still work as expected 😄

@SachaZvetelman about the title: you are right, I didn't change the default error page copy just added the new one for 403 page, but checking the screenshot from the ticket to add a button, it should be "Something went wrong" now.

About multiple basket cookies, do you mean user can have `je-mw-basket-11111`, `je-mw-basket-2222` etc at the same time? At the moment we remove `je-mw-basket-${this.basket.id}` so the basket id we have from vuex store

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -955,46 +964,24 @@ describe('Checkout', () => {
...
+                const wrapper = shallowMount(VueCheckout, {
+                    store: createStore({
+                        ...defaultCheckoutState,
+                        restaurant
+                    }),
+                    i18n,
+                    localVue,
+                    propsData,
+                    mocks: {
+                        $style
```
FayeCarter: Do you need this mock here?

packages/components/organisms/f-checkout/src/demo/checkoutMock.js
```diff
@@ -10,7 +10,8 @@ import getBasketDelivery from './get-basket-delivery.json';
 import getBasketCollection from './get-basket-collection.json';
 import getBasketDineIn from './get-basket-dinein.json';
 import updateCheckout from './update-checkout.json';
-import updateCheckoutErrors from './update-checkout-errors.json';
+import updateCheckoutRestaurantNotTakingOrders from './update-checkout-restaurant-not-taking-orders.json';
+import updateCheckoutAdditionalItemsRequared from './update-checkout-additional-items-required.json';
```
FayeCarter: Requared again 😄

dandel10n: It was me :face_palm:

packages/components/organisms/f-checkout/src/components/Error.vue
dandel10n: I amended the condition with a check that errorFormType is 403 page

packages/components/organisms/f-checkout/src/components/Error.vue
SachaZvetelman: Any updates on this? Because I don't think we should be adding these validations because of the tests. We should see why the tests are failing/throwing console errors.

packages/components/organisms/f-checkout/src/components/Error.vue
```diff
@@ -22,23 +21,78 @@
...
+            const cookieName = `je-mw-basket-${this.basket.id}`;
+            const basketCookie = this.$cookies.get(cookieName);
+
+            if (basketCookie && this.errorFormType === CHEKOUT_ERROR_FORM_TYPE.accessForbidden) {
```
SachaZvetelman: Nit: this validation should be at the top of the method, in my opinion. If the issue is no `accessForbidden` we don't even care about getting the cookie.

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
```diff
@@ -42,4 +68,84 @@ describe('Error', () => {
...
+                        // Act
+                        wrapper.vm.redirectToMenu();
+
+                        // Assert
+                        expect(cookieRemoveSpy).not.toHaveBeenCalled();
+                    });
+                });
+
+                describe('AND a menu basket cookie already exists', () => {
+                    describe('AND errorFormType is accessForbiddenError', () => {
```
SachaZvetelman: Nit: same as above, this should be probably moved up.

packages/components/organisms/f-checkout/test/visual/f-checkout-authenticated.visual.shared.spec.js
```diff
@@ -67,7 +67,62 @@ describe('f-checkout - Collection - Authenticated - Visual Tests', () => {
         checkout.goToPayment();
 
         // Assert
-        browser.percyScreenshot('f-checkout - Collection - Authenticated - "Restaurant not taking orders" Error Modal', 'shared');
+        browser.percyScreenshot('f-checkout - Delivery - Authenticated - "Restaurant not taking orders" Error Modal', 'shared');
+    });
+
+    it('should display the "Additional Items Requared" modal', () => {
```
SachaZvetelman: "Required".

packages/components/organisms/f-checkout/test/visual/f-checkout-authenticated.visual.shared.spec.js
```diff
@@ -67,7 +67,62 @@ describe('f-checkout - Collection - Authenticated - Visual Tests', () => {
...
+        // Arrange
+        checkout = new Checkout('organism', 'checkout-component');
+        checkout.withQuery('&knob-Service Type', 'delivery')
+                .withQuery('&knob-Is User Logged In', true)
+                .withQuery('&knob-Is ASAP available', true)
+                .withQuery('&knob-Get Checkout Errors', 'additional-items-requared')
```
SachaZvetelman: "required"

packages/components/organisms/f-checkout/test/visual/f-checkout-authenticated.visual.shared.spec.js
```diff
@@ -67,7 +67,62 @@ describe('f-checkout - Collection - Authenticated - Visual Tests', () => {
...
+                .withQuery('&knob-Place Order Errors', 'SERVER');
+
+        const pageUrl = buildUrl(checkout.componentType, checkout.componentName, checkout.path);
+        // Act
+        checkout.open(pageUrl);
+        checkout.waitForComponent();
+        checkout.goToPayment();
+
+        // Assert
+        browser.percyScreenshot('f-checkout - Delivery - Authenticated - "Additional Items Requared" Error Modal', 'shared');
```
SachaZvetelman: "Required".

packages/components/organisms/f-checkout/src/components/Error.vue
```diff
@@ -22,23 +21,78 @@
...
     },
+    mixins: [
+        loggerMixin
+    ],
     props: {
         errorFormType: {
             type: String,
             required: true
+        },
+        restaurantMenuPageUrl: {
```
ashleynolan: I'm always a bit wary of naming things related to pages, as they can change down the line.

Would a name like `redirectUrl` be appropriate (as it looks like that is its actual function?)

packages/components/organisms/f-checkout/src/components/Error.vue
```diff
@@ -69,5 +123,9 @@ export default {
         @include font-size(body-l);
         margin-top: spacing();
     }
+
+    .c-checkout-error-button {
```
ashleynolan: Small one, but do `.c-checkout-error-heading`, `.c-checkout-error-description` and `.c-checkout-error-button` need to be nested?

If not, I'd un-nest them to keep the specificity of the CSS lower.

packages/components/organisms/f-checkout/src/components/Error.vue
kevinrodrigues: Fixed now. Didn't realise the failure was coming from checkout.vue instead of error.vue. 👍

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
```diff
@@ -42,4 +68,71 @@ describe('Error', () => {
...
+
+    describe('methods', () => {
+        describe('`redirectToMenu` ::', () => {
+            const oldWindowLocation = window.location;
+
+            beforeAll(() => {
+                delete window.location;
+
+                window.location = Object.defineProperties(
+                    {},
```
FayeCarter, line 80: Is this right?

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
```diff
@@ -42,4 +68,71 @@ describe('Error', () => {
...
+                            // Assert
+                            expect(cookieRemoveSpy).not.toHaveBeenCalled();
+                        });
+                    });
+
+                    describe('AND a menu basket cookie already exists', () => {
+                        it('should delete the cookie', () => {
+                            // Arrange
+                            jest.spyOn(wrapper.vm.$cookies, 'get').mockReturnValue(() => 'je-mw-basket');
+                            const cookieRemoveSpy = jest.spyOn(wrapper.vm.$cookies, 'remove');
```
FayeCarter, line 125: can this go in the beforeEach for this and the test above?

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -416,7 +404,10 @@ export default {
 
         dialogMessage () {
             return {
-                name: 'error-dialog'
+                name: 'error-dialog',
+                props: {
+                    'redirect-url': this.redirectUrl
```
SachaZvetelman, line 409: Wouldn't `redirectUrl` work instead of `redirect-url`? I'm not sure, but it's worth asking.

SachaZvetelman (approved):

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
kevinrodrigues: Yes, needs two arguments here. First is the object second is the properties.

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
kevinrodrigues: Yes that would work. @dandel10n would you be able to do this in this pr or another? If not I can fix in another PR myself.

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
FayeCarter: Thanks 🙂

kevinrodrigues (approved):

packages/components/organisms/f-checkout/src/components/_tests/Error.test.js
dandel10n: @FayeCarter  @kevinrodrigues  Can we apply it in another PR, as adding the change here would trigger all the builds again and it takes ages?

packages/components/organisms/f-checkout/src/components/Checkout.vue
dandel10n: @SachaZvetelman I tested it now and it works with `redirectUrl` as well.

PR #1039, opened 2021-06-14
f-checkout@0.138.0 - Referral State fix.

SachaZvetelman:
v0.138.0
*June 15, 2021*

### Changed
- Fixed the hardcoded `referralState` data we're passing to the Place Order endpoint.

_Note: updating the version and CHANGELOG once the previous PRs are merged._


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -922,6 +922,15 @@ export default {
...
             }, this.spinnerTimeout);
+        },
+
+        getReferralState () {
+            const cookieName = `je-rw-menu-referral-state-${this.restaurant.id}`;
+            const referralCookie = this.$cookies.get(cookieName);
+
+            return !referralCookie || !referralCookie.menuReferralState
+                ? 'None'
+                : 'ReferredByWeb';
```
1itvinka:
```suggestion
            return referralCookie && referralCookie.menuReferralState
                ? 'ReferredByWeb'
                : 'None';
```
as for me is more readable

tommcclean (dismissed):

tommcclean (dismissed):

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -3210,6 +3237,142 @@ describe('Checkout', () => {
...
+
+                        // Act
+                        const result = wrapper.vm.getReferralState();
+
+                        // Assert
+                        expect(result).toBe('ReferredByWeb');
+                    });
+                });
+
+                describe('AND the `menuReferralState` is not `1`', () => {
```
1itvinka, line 3280:
```suggestion
                describe('and the `menuReferralState` is not `1`', () => {
```
AND looks like an acronym to me. 😬

SachaZvetelman: Whereas I kinda agree, several of our tests follow that convention already, so I'm inclined to leave it as it is, personally.

FayeCarter (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -922,6 +922,15 @@ export default {
...
                 }
             }, this.spinnerTimeout);
+        },
+
+        getReferralState () {
+            const cookieName = `je-rw-menu-referral-state-${this.restaurant.id}`;
+            const referralCookie = this.$cookies.get(cookieName);
+
+            return referralCookie && referralCookie.menuReferralState
+                ? 'ReferredByWeb'
```
pkedziora, line 927: Should this be hardcoded to 'ReferredByWeb' or should it rather take whatever value is in referralCookie.menuReferralState

SachaZvetelman: `referralCookie.menuReferralState` stores either `1` or `0`, so yeah, we need to hardcode it here 👍

pkedziora (dismissed):

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -3210,6 +3237,142 @@ describe('Checkout', () => {
...
+
+                const wrapper = shallowMount(VueCheckout, {
+                    store: createStore({ ...defaultCheckoutState, restaurant: { id: '99999' } }),
+                    i18n,
+                    localVue,
+                    propsData,
+                    mocks: {
+                        $logger,
+                        $cookies: {
+                            get: getSpy
```
kevinrodrigues: I dont think you need to point the internal `get` to the spy here? You should be able to just create the spy above from the wrapper instance and then use that in the assert. The `get` will already be tied to the `jest.fn()` via the setup file anyway so here you can just have `$cookies` i believe.
```javascript
const getSpy = jest.spyOn(wrapper.vm.$cookies, 'get').mockReturnValue(() => { menuReferralState: 1 });

mocks: {
  $cookies
}

expect(getSpy).toHaveBeenCalledWith('je-rw-menu-referral-state-99999');
```

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -3210,6 +3237,142 @@ describe('Checkout', () => {
...
+
+                        const wrapper = shallowMount(VueCheckout, {
+                            store: createStore(),
+                            i18n,
+                            localVue,
+                            propsData,
+                            mocks: {
+                                $logger,
+                                $cookies: {
+                                    get: getSpy
```
kevinrodrigues: Same as above.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -3210,6 +3237,142 @@ describe('Checkout', () => {
...
```
kevinrodrigues: Same

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -3210,6 +3237,142 @@ describe('Checkout', () => {
...
+
+                    const wrapper = shallowMount(VueCheckout, {
+                        store: createStore(),
+                        i18n,
+                        localVue,
+                        propsData,
+                        mocks: {
+                            $logger,
+                            $cookies: {
+                                get: getSpy
```
kevinrodrigues: Same.

packages/components/organisms/f-checkout/src/components/_tests/helpers/setup.js
```diff
@@ -128,6 +128,11 @@ const $logger = {
     logError: jest.fn()
 };
 
+const $cookies = {
```
kevinrodrigues: If you pull master you should get this change too.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
SachaZvetelman: Nope, I just tried this and the test failed. I don't see why it would work, though. But maybe I'm missing something?
I mean, if we don't pass the `getSpy` mock, it won't be used, hence it won't be called.

kevinrodrigues: What was the failure out of interest? I did the same sort of thing earlier [here](https://github.com/justeat/fozzie-components/pull/1034/files#diff-77d6d5e41e52cc7565e75f71e027e67a362fae9f1f6e0c8954c6b95ea57f8d50R124)

kevinrodrigues: It should work because you're passing the $cookie instance as a mock to the wrapper so when you spy on it you're spying on the mocked jest instance. Happy to approve if for for some reason it's not working for you.

SachaZvetelman: Ahhh, wait, I understand what you did now. Ok, I'll implement this 👍

FayeCarter (approved):

kevinrodrigues (approved):

PR #1040, opened 2021-06-14
f-checkout@0.139.0 - Loading spinner on Checkout submission

charliearlie:
### Changed
- Use loading spinner on submit button


yarn.lock
```diff
@@ -2204,11 +2204,6 @@
   resolved "https://registry.yarnpkg.com/@justeat/f-button/-/f-button-1.1.1.tgz#065471a0aa00391ca70f4a108f5904b85b477446"
   integrity sha512-0l43zDtVP5oIOYUqJtZjX4CPEwl9NIu3UgENt6EairOZpfFeGZg/2YJ6vVIWySycQo9v6iHfeaAtDj4JyTWebg==
 
-"@justeat/f-button@1.2.0":
```
charliearlie, line 2207: I'm not sure what's going on here but upgrading f-button has removed this from the yarn lock but didn't add it for the new version?

packages/components/organisms/f-checkout/package.json
```diff
@@ -61,7 +61,6 @@
   "devDependencies": {
     "@babel/plugin-proposal-class-properties": "7.12.13",
     "@justeat/f-alert": "0.5.0",
-    "@justeat/f-button": "1.2.0",
```
1itvinka, line 64: you've removed the `f-button` completely?

charliearlie: Haha, I think I removed it to test why the yarn.lock wasn't adding the new version. Back now

FayeCarter (dismissed):

1itvinka (dismissed):

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -366,6 +366,56 @@ describe('Checkout', () => {
...
+
+                // Act
+                const submitButton = wrapper.find('[data-test-id="confirm-payment-submit-button"]');
+                const spinner = submitButton.find('[data-test-id="action-button-spinner"]');
+
+                // Assert
+                expect(submitButton.attributes('aria-busy')).toBeFalsy();
+                expect(spinner.exists()).toBe(false);
+            });
+        });
```
SachaZvetelman: I personally don't think these unit tests should test what they are testing. `f-button`'s unit tests will cover that the spinner is displayed when it's loading. The same for the `aria-busy`.

I think we should be testing that `isFormSubmitting` is set to `true` and `false` within `onFormSubmit`, and that `isFormSubmitting` is used on the button correctly.

Then, we'd probably want to cover that the button in the `f-checkout` form is looking as expected with a Visual Regression Test.

charliearlie: Yeah, it does make more sense and I wasn't really sure what level to test. So I will go with what you suggested.

On the visual regression test side of things, I was struggling to get the test to stay in a loading state because it seems the Percy configuration is controlled by props, like Storybook. In fact, I think it uses the knobs and URL parameters to control it. So unsure how I'll test a loading state.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -2660,6 +2662,28 @@ describe('Checkout', () => {
                     expect(updateMessageSpy).toHaveBeenCalled();
                 });
 
+                it('should set `isFormSubmitting` to true via `setSubmittingState`', async () => {
```
SachaZvetelman:
```suggestion
                it('should call `setSubmittingState` first with `true` and then with `false`', async () => {
```

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -919,6 +924,10 @@ export default {
             }, this.spinnerTimeout);
         },
 
+        setSubmittingState (isFormSubmitting) {
+            this.isFormSubmitting = isFormSubmitting;
+        },
```
SachaZvetelman, line 933: Nit: probably worth adding a JSDoc here to explain why we're doing this, since I'm quite sure in 2 months someone will come and try to remove this method.

PR #1044, opened 2021-06-15
f-header@v4.17.2 - fix nav links colour when profile dropdown is open

1itvinka:
v4.17.2
*June 15, 2021*

### Fixed
- fix nav links colour when profile dropdown is open


simonsandersje (dismissed):

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -446,7 +446,7 @@ export default {
         },
 
         isAltColour () {
-            return this.headerBackgroundTheme === 'highlight' || (this.headerBackgroundTheme === 'transparent' && !this.navIsOpen);
+            return this.headerBackgroundTheme === 'highlight' || this.headerBackgroundTheme === 'transparent';
```
ashleynolan: Worth testing this on mobile and tablet as I think this was in there for a use-case on one of those devices. I think it should apply the alt colour on mobile when the nav is opened.

dandel10n: Yes, as Ash said it tied up to mobile and tablet device styles. `&& this.isBelowMid` prop check can be added instead of removing `&& !this.navIsOpen`

dandel10n (approved):

ashleynolan (approved):

PR #1045, opened 2021-06-16
f-checkout@0.139.0 - Loading spinner on Checkout submission

charliearlie:
### Changed
- Use loading spinner on submit button


SachaZvetelman (approved):

1itvinka (approved):

ashleynolan (approved):

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -321,7 +321,8 @@ describe('Checkout', () => {
                     localVue,
                     propsData,
                     mocks: {
-                        $logger
+                        $logger,
```
DamianMullins, line 324: Probably not one for this PR but we could reduce a lot of the duplication in this module by creating a helper function to set up mounting of the checkout component. 

In menu we do it like this

```js
const createComponent = ({
    store = createStore(),
    propsData,
	data
} = {}) => mount(VueCheckout, {
    i18n,
    store,
    localVue,
    propsData,
    mocks: {
        $logger,
        $cookies
    },
    data
});
```

And then call that in the test like so

```js
const wrapper = createComponent({
    propsData,
    data () {
        return {
            isLoading: false,
            errorFormType: CHEKOUT_ERROR_FORM_TYPE.default
        };
    }
});
```

DamianMullins (approved):

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
SachaZvetelman: I agree. MenuWeb style :)

DamianMullins: 🤟

PR #1047, opened 2021-06-16
f-checkout@0.141.0 - Fixed an issue in which the chosen delivery/collection time wasn't remembered.

SachaZvetelman:
v0.141.0
*June 17, 2021*

### Changed
- Fixed an issue in which the chosen delivery/collection time wasn't remembered.


pkedziora (dismissed):

packages/components/organisms/f-checkout/src/demo/checkout-dinein.json
```diff
@@ -10,8 +10,8 @@
   },
   "fulfilment": {
     "time": {
-      "from": "2020-01-01T00:00+00:00",
-      "to": "2020-01-01T00:00+00:00"
+      "asap": false,
+      "scheduled": null
```
charliearlie: Is there a specific reason as to why the dine in mock was changed differently to the other two? I am happy to rectify it after, but this is the only thing that stuck out to me in this PR.

charliearlie (dismissed):

packages/components/organisms/f-checkout/src/demo/checkout-dinein.json
SachaZvetelman: Because it's also a valid value (what most users will see when they land on Checkout). I wanted to have different valid values in the files, but there's no specific reason to have this in the `dinein` file. Are you ok to leave it as it is or would you like me to change it to something else?

charliearlie: On dine in there is no selector so in the background the API always sets `asap` to true and sends us one `to` and `from` time slot. But as mentioned, I am happy to patch that after. 🙂

packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+### Changed
+- Fixed an issue in which the chosen delivery/collection time wasn't remembered.
+
+
+v0.140.0
+------------------------------
+*June 16, 2021*
+
+### Changed
+- Fixed cookie removal on 403 error dialog by now using restaurant ID instead of basket ID
```
dandel10n, line 20:
```suggestion
- Fixed cookie removal on 403 error page by now using restaurant ID instead of basket ID
```

dandel10n:
One of Percy tests is failing now with "Cannot read property 'from' of null" error on "f-checkout - Dine in - Authenticated - Base State" so I think we need to revert dine in changes as Charlie mentioned:
Screenshot 2021-06-16 at 16 33 13
Need to update

simonsandersje (dismissed):

simonsandersje (approved):

packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.141.0
+------------------------------
+*June 17, 2021*
+
+### Changed
```
kevinrodrigues, line 11: Fixed?

kevinrodrigues (approved):

packages/components/organisms/f-checkout/CHANGELOG.md
SachaZvetelman: Will fix this later, since we need this merged ASAP and changing this will kick off the whole build progress again.

PR #1048, opened 2021-06-16
f-performance@0.1.0 RUM Vue plugin

zlatinivanov-je:
Implements a Real User Monitoring metrics as Vue plugin
--------

_Currently filed in as JERFC-62 and in active discussion._ 

 - Uses [PerfumeJS](https://github.com/zizzamia/perfume.js) library to provide sensible defaults

See README for more details

Screenshot 2021-06-17 at 12 19 05

Screenshot 2021-06-17 at 12 22 39

Example with custom `logger` Function `(metricName, data) => { console.log(metricName, data) }`

Screenshot 2021-06-17 at 15 26 52

---

## UI Review Checks


simonsandersje (dismissed):

siggerzz (dismissed):

packages/services/f-perf/src/index.js
```diff
@@ -0,0 +1,129 @@
...
+    analyticsTracker: options => {
+        const {
+            metricName,
+            data,
+            eventProperties,
+            navigatorInformation,
+            vitalsScore
+        } = options;
+
+        switch (metricName) {
```
matthewhardern: Not a fan of the giant switch statement, could we instead store an object who's keys represent the correct metric name? a look up table of some kind. Example

```
const lookUpTable = {
    'navigationTiming': 'navigationTiming',
    'firstContentfulPaint': 'fcp'
};

logPerformance(Object.keys(lookUpTable).find(key => lookUpTable[key] === metricName), data)
```

matthewhardern:
Also the component name, any reason why `perf` instead of `perfume`. Also the component name is very specific, if we change library choice for the analytics we would be tied to that name.

packages/services/f-perf/src/index.js
zlatinivanov-je: There is certainly scope for that. I was considering similar approach but enabling `options` to include specific metrics tracked. There has been concern on how much logging will be sent to GTM if every vertical or component starts logging.

zlatinivanov-je:
> Also the component name, any reason why `perf` instead of `perfume`. Also the component name is very specific, if we change library choice for the analytics we would be tied to that name.

I was referring to `perf` as in Performance rather than trying to bind it to Perfume, the library. Maybe `performance` is a better name or something completely different all together. Open to suggestions.

matthewhardern:
> > Also the component name, any reason why `perf` instead of `perfume`. Also the component name is very specific, if we change library choice for the analytics we would be tied to that name.
> 
> I was referring to `perf` as in Performance rather than trying to bind it to Perfume, the library. Maybe `performance` is a better name or something completely different all together. Open to suggestions.

ahh I see makes more sense, yea think `f-performance` or maybe `f-metrics` are good alternatives

packages/services/f-perf/src/index.js
matthewhardern: You may know better how to optimise this switch statement, I was just giving a suggestion, happy with another approach if it reduces the repetition of code. My only thoughts around a lookup table was that would enable us to keep an eye on what event types were being submitted to the logging server, as they would be controlled with PR's being raised to add additional types. As for the use of GTM as our logging platform, I have to agree I don't think that is a best use case for GTM, would we not be better off using something like sentry? @tommcclean I was under this impression we were trialling sentry atm?

zlatinivanov-je: Heard about sentry being mentioned. GTM could provide a different dimension as the data is larger and could potentially be combined to provide a better picture.

matthewhardern: A couple of concerns for me with GTM use. 

1) Metrics tell us nothing meaningful about the user, they either report errors and stack traces or they provide information that is relevant to performance on a global level. GTM seems like a user specific event tracking system.
2) Events use up monthly quotas so we would want to be sure that letting people lose with this would not dramatically cost us anything as a result of the potential significant increase in events being fired. (don't know what we currently get)

I may be alone in this thought though so I think it would be better discussed at a wider group level?

zlatinivanov-je: Absolutely, this is a wider group discussion. I wanted to gauge earlier version views and get the sense of what others may have in mind. 

Thank you so much in contributing @matthewhardern. I am happy to have that open conversation going on here, but will follow up with a Guild meeting before we can go ahead with these changes.

matthewhardern: I have added a slot in the guild to discuss it tomorrow.

xander-marjoram:
@zlatinivanov-je Is this PR still required?

zlatinivanov-je:
Closing due to it being obsolete.

PR #1055, opened 2021-06-17
f-button@v1.9.0 – Adding ghostTertiary iconButton styles and better style validation for types

ashleynolan:
### Changed
- `buttonType` has been tidied up a bit. Have added type validation for when the component is either a standard `button` or an `iconButton` (when `isIcon` is set to `true`). If an invalid type is now set for these variations, the component will throw an error when created.
- New `buttonType` of `ghostTertiary` has been added for use with the `iconButton` component.

---

## UI Review Checks

Screen Shot 2021-06-17 at 17 02 05

## Browsers Tested

- [x] Chrome (latest)


zlatinivanov-je (dismissed):

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -105,72 +107,105 @@ export default {
...
+    },
+    methods: {
+        /**
+         * Throws an error if the `buttonType` is invalid for the type of button being used (component is either a button or iconButton).
+         * This is to ensure visual styles are only used with the intended component type.
+         */
+        validateProps () {
+            let componentType = 'button';
+
+            if (this.isIcon === true) {
```
1itvinka:
```suggestion
            if (this.isIcon) {
```

 or even

```
let componentType = this.isIcon ? 'iconButton' : 'button';
```

simonsandersje (dismissed):

1itvinka:
Looks good to me. But it's hard for me to understand what was changed in `Button.vue` scss variable declaration block.

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -5,11 +5,91 @@ describe('Button', () => {
...
+                'ghost',
+                'link'
+            ])('should not throw an error when the component is a standard button (isIcon="false") and buttonType is set to %s', buttonType => {
+                // Arrange
+                const propsData = {
+                    buttonType,
+                    isIcon: false
+                };
+
+                // Act
```
kevinrodrigues: Can just do `// Act & Assert` here & below instead of splitting it.

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -5,11 +5,91 @@ describe('Button', () => {
...
+                expect(() => {
+                    shallowMount(FButton, { propsData });
+                })
+                    // Assert
+                    .toThrowError('button is set to have buttonType="invalid_value"');
+
+                spy.mockRestore();
```
kevinrodrigues: You could move this into a `beforeEach`?

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -5,11 +5,91 @@ describe('Button', () => {
...
+                // Act
+                expect(() => {
+                    shallowMount(FButton, { propsData });
+                })
+                    // Assert
+                    .not.toThrowError();
+            });
+            it('should throw an error when the component is a standard button (isIcon="false") and buttonType is set to an invalid type', () => {
+                // Arrange
+                const spy = jest.spyOn(global.console, 'error').mockImplementation(() => { });
```
kevinrodrigues: `BeforeEach`

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -5,11 +5,91 @@ describe('Button', () => {
...
+                expect(() => {
+                    shallowMount(FButton, { propsData });
+                })
+                    // Assert
+                    .not.toThrowError();
+            });
+
+            it('should throw an error when the component is an iconButton (isIcon="true") and the buttonType is set to an invalid type', () => {
+                // Arrange
+                const spy = jest.spyOn(global.console, 'error').mockImplementation(() => { });
```
kevinrodrigues: `BeforeEach`

packages/components/atoms/f-button/src/components/_tests/Button.test.js
dandel10n: And //Assert right before expect() function

siggerzz:
Should this have made any changes to the visual regression checks? Looks like there's no visual diffs in the f-button Percy project? 🤔

ashleynolan:
@siggerzz Probably not (as the old buttton styles should be the same), but I may need to add a new test for the new buttonType that's been added.

Happy to pick this up with you as a separate PR though?

ashleynolan:
@1itvinka Its mostly whitespace changes - if you click on the settings cog at the top of the PR you can toggle to hide whitespace changes and that should make that section more readable

packages/components/atoms/f-button/src/components/_tests/Button.test.js
ashleynolan: Done

packages/components/atoms/f-button/src/components/_tests/Button.test.js
ashleynolan: Done

packages/components/atoms/f-button/src/components/_tests/Button.test.js
ashleynolan: Done

packages/components/atoms/f-button/src/components/_tests/Button.test.js
ashleynolan: Changed 👍

packages/components/atoms/f-button/src/components/Button.vue
ashleynolan: Have updated 👍

simonsandersje (dismissed):

kevinrodrigues (dismissed):

1itvinka:
> @1itvinka Its mostly whitespace changes - if you click on the settings cog at the top of the PR you can toggle to hide whitespace changes and that should make that section more readable

cool! I didn't know about this feature. thanks!

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -105,72 +107,101 @@ export default {
...
             return this.isLoading ? 'polite' : 'off';
         }
+    },
+    watch: {
+        $props: {
+            immediate: true,
+            handler () {
+                this.validateProps();
+            }
+        }
```
minrwhite, line 120: It's frustrating that this is the way mutually dependent prop validation has to be done when prop validators are supposed to be a thing, but [this](https://stackoverflow.com/questions/56904327/access-props-value-in-another-props-validator) would seem to indicate it is the accepted way. 

As a slightly cleaner solution, could I suggest that the list of valid button types for each component type not intersect, and therefore the 'isIcon' prop would become unnecessary, as the component type could be inferred from the button type? E.g.
```
export const VALID_BUTTON_TYPES = {
        iconButton: ['primaryIcon', 'secondaryIcon', 'ghostIcon', 'ghostTertiaryIcon'],
        button: ['primary', 'secondary', 'outline', 'ghost', 'link']
};
```
Or would that complicate things elsewhere?

ashleynolan: That could potentially work too, for sure. I'll give it a try when I get chance and see if it's an easy thing to switch to.

It would be a major bump (as the contract with the component would change for consumers potentially), but other than that I can't see any issues with it.

ashleynolan: I Have discussed this with @benwhite-deltas – after some investigation, it's not possible to get rid of the `isIcon` prop without also losing the prop validation, so will keep this like it is for now.

DamianMullins (approved):

kevinrodrigues (approved):

PR #1061, opened 2021-06-21
f-checkout@0.144.0 - Check if all storage and form address values match

charliearlie:
### Added
- More thorough check on whether form address and storage address match

On ConsumerWeb, the address in local storage is compared with the form data by using a deep equals of the objects. Because our fields are named differently, I had to just check if each field matches individually.

### DIFF
--- a/packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
+++ b/packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
@@ -272,4 +272,81 @@ describe('addressService', () => {
             });
         });
     });
+
+    describe('doesAddressInStorageAndFormMatch ::', () => {
+        beforeEach(() => {
+            Object.defineProperty(window, 'localStorage', { value: localStorageMock });
+        });
+
+        afterEach(() => {
+            window.localStorage.clear();
+            jest.resetAllMocks();
+        });
+
+        describe('when the address fields are the same', () => {
+            it('should return true', () => {
+                // Arrange
+                const storageAddress = {
+                    Line1: 'Fleet Place House',
+                    Line2: 'Farringdon',
+                    City: 'London',
+                    PostalCode: 'EC4M 7RF'
+                };
+
+                const formAddress = {
+                    line1: 'Fleet Place House',
+                    line2: 'Farringdon',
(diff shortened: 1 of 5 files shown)


packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -355,7 +355,7 @@ export default {
 
             if (isAddressInLocalStorage) {
                 const storedAddress = addressService.getAddressFromLocalStorage(false);
-                if (storedAddress.Line1 === state.address.line1 && storedAddress.PostalCode === state.address.postcode) {
+                if (addressService.doesAddressInStorageAndFormMatch(storedAddress, state.address)) {
```
charliearlie, line 358: I used this with both addresses passed in as we don't want to be reading from local storage repeatedly. But I could check if they match as part of the check to see if the address is in local storage?

packages/components/organisms/f-checkout/src/services/tests/addressService.test.js
```diff
@@ -272,4 +272,72 @@ describe('addressService', () => {
...
+        });
+
+        describe('when the postcode is different', () => {
+            it('should return false', () => {
+                // Arrange
+                const storageAddress = {
+                    Line1: 'Fleet Place House',
+                    Line2: 'Farringdon',
+                    City: 'London',
+                    PostalCode: 'EC4M 7RF'
```
matthewhardern, line 328: What would happen if we removed the postcode space? But they were the same characters

charliearlie: It would see them as different, but as mentioned in the description, this wouldn't be a regression because in ConsumerWeb they just deep equals the objects. I could trim the fields and lowercase everything, but I don't want to make the function extremely convoluted.  And I think we're safer calling the API for the coordinates for a slight change rather than not at all. But happy to be corrected on that one.

matthewhardern: The API idea seems like a good point happy to approve

matthewhardern (approved):

FayeCarter (approved):

PR #1063, opened 2021-06-21
f-media-element@v0.4.0 - modifications to support pre-line in content

matthewhardern:
This PR adds some mobile responsive features to allow media element to stack if stackedOnMobile is true. Also have added a `white-space: pre-line` to enable new lines to be observed. Plus fixed a bug where reverse prop would not work if not in stacked mode.

## Browsers Tested

- [x] Chrome (latest)

### DIFF
--- a/packages/components/molecules/f-media-element/CHANGELOG.md
+++ b/packages/components/molecules/f-media-element/CHANGELOG.md
@@ -3,6 +3,20 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+v0.3.1
+------------------------------
+*June 21, 2021*
+
+### Added
+- stackOnMobile prop to stack when on device sizes <=narrowMid
+- `white-space: pre-line` to enable newlines to be preserved when used in title or text
+- Additional unit tests for new features
+
+### Fixed
+- Styling fix to make sure reverse prop works when not in stacked mode
+
+
 v0.3.0
 ------------------------------
 *June 09, 2021*
--- a/packages/components/molecules/f-media-element/CHANGELOG.md
+++ b/packages/components/molecules/f-media-element/CHANGELOG.md
@@ -4,12 +4,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-v0.3.1
+v0.4.0
 ------------------------------
 *June 21, 2021*
 
 ### Added
-- stackOnMobile prop to stack when on device sizes <=narrowMid
+- stackWhenNarrow prop to stack when on device sizes <=narrowMid
 - `white-space: pre-line` to enable newlines to be preserved when used in title or text
 - Additional unit tests for new features
 
--- a/packages/components/molecules/f-media-element/package.json
+++ b/packages/components/molecules/f-media-element/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-media-element",
   "description": "Fozzie Media Element - provides ability to display text and an image in different orientations",
-  "version": "0.3.1",
+  "version": "0.4.0",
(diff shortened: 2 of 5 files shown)


packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -165,6 +171,43 @@ $font-sizes: (
...
+    }
+}
+
+/**
+ * Modifier – .c-mediaElement--stackMobile
+ *
+ * Applies flex direction column on <=narrow
+ */
+.c-mediaElement--stackMobile {
+    @include media('<=narrow') {
```
minrwhite: Does it not work here to stick a single `@include media('<=narrow') { ... }` expression around the `.c-mediaElement--stackMobile { ... }` one, to prevent having to invoke the same mixin multiple times?

matthewhardern: good point

packages/components/molecules/f-media-element/CHANGELOG.md
```diff
@@ -3,6 +3,20 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+v0.3.1
+------------------------------
+*June 21, 2021*
+
+### Added
+- stackOnMobile prop to stack when on device sizes <=narrowMid
```
ashleynolan: Should reference width by `narrow`/`mid` etc, as that's what we're checking for (someone could have a narrow screen on tablet/desktop for instance). So maybe `stackWhenNarrow`?

packages/components/molecules/f-media-element/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-media-element",
   "description": "Fozzie Media Element - provides ability to display text and an image in different orientations",
-  "version": "0.3.0",
+  "version": "0.3.1",
```
ashleynolan: Should be a minor bump

matthewhardern: agreed have made 0.4.0

packages/components/molecules/f-media-element/CHANGELOG.md
matthewhardern: `stackWhenNarrow` seemed a good choice

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -18,11 +19,12 @@
             <p :class="$style['c-mediaElement-text']">
                 {{ text }}
             </p>
-            <slot></slot>
+            <slot />
         </div>
         <div
             :class="[
                 $style['c-mediaElement-imgWrapper'],
+                'c-mediaElement-imgWrapper--override',
```
ashleynolan: What does this class do? Can't see a reference to it in the SCSS

ashleynolan (commented):
Looks good – just one super small q

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: This allows the user who uses this component to override the imageWrapper class, this is particularly useful if you want to make some changes to the way flexbox behaves. Example use case is on narrow screen sizes it would allow me to position the image differently to account for the narrow format, this use case being something I want to do on `f-offers` header component. This use case seems very specific so I thought by adding an override class it would still give me the benefits of using the component.

ashleynolan: Should this not just be a prop that controls the positioning of the image? Depends how specific you think the functionality is I guess but generally I'd try to do this through exposing props rather than classnames.

matthewhardern: I think for now I will remove until I can figure out a better way of dealing with this, as the stacking on narrow is more important atm. :)

minrwhite (approved):

ashleynolan (approved):

PR #1067, opened 2021-06-23
fozzie-components@v3.42.0 – Optimising CircleCI build

ashleynolan:
## Changed
- Optimises the build so that only packages that have been amended run in the build and are tested via Storybook.


JoshuaNg2332 (dismissed):

.circleci/config.yml
```diff
@@ -14,6 +15,26 @@ commands:
           fingerprints:
             - "0e:50:82:ad:81:17:a5:31:04:13:fd:6e:c7:72:c9:18"
 
+  detect_root_change:
+    description: Detects root package.json changes
+    steps:
+      - run: |
+          changes=`git diff --name-only master...$CIRCLE_BRANCH | grep -v '^packages/'`
```
siggerzz, line 22: As part of this PR, we want our npm scripts to only run against changed packages. However, if you change a file at the root of the monorepo, nothing would get triggered. 

This isn't ideal. We wanted a way to still run scripts against all packages if the root of the monorepo changes (global dependencies in root package.json for example...)


In this PR, the following files have been changed:

```
.circleci/config.yml
CHANGELOG.md 
package.json
packages/storybook/config/storybook/main.js
packages/storybook/config/storybook/story.config.js
```


In this scenario, the command does the following when running CI:


- determine what files have changed in the branch:
`git diff --name-only master...$CIRCLE_BRANCH`

- Print out the files that are in the root of the monorepo
- `| grep -v '^packages/'`

- Store the value in an env var called `$CHANGES`

The output of which would be:

```
CHANGELOG.md 
package.json
```

Because root files have changed, CI will run scripts against every package.

.circleci/config.yml
```diff
@@ -14,6 +15,26 @@ commands:
...
+
+          if [ "${#changes}" > 0 ];
```
siggerzz, line 24: if the length of `$CHANGES` > 0, set `ROOT_CHANGE=true` 



This is then used to determine if all packages should be built / listed / tested, or only changed packages.

.circleci/config.yml
siggerzz: Obviously this can be improved at some point to only check for certain root files.

We don't really want to be running a full CI build if `bear.png` changes 🤣

It's a starting point at least!

dandel10n (dismissed):

.circleci/config.yml
matthewhardern: Just one question what would happen when the base branch is a beta branch so for example if we make a beta branch for `f-checkout` then PR into that? Apologies if its a dumb question.

ashleynolan: It would match the changes made against the `master` branch the way it's setup at the moment, so if you're changing `f-checkout` (even when PR-ing into the beta branch) in that instance, it would run the tests and build for `f-checkout` each time, as it'll know that that package is changing.

I think that's probably what you would want in that instance?

matthewhardern: yea thats fine just checking what would happen

matthewhardern (approved):

DamianMullins (approved):

JoshuaNg2332 (approved):

PR #1072, opened 2021-06-24
f-checkout@v0.148.0 - Store modified address in localStorage.

kevinrodrigues:
When the user attempts to update the checkout address (the address specifically from LocalStorage) then we resave the edited address so if they navigate back to checkout from payments then the edited address is populated.

## Browsers Tested

- [x] Chrome (latest)


packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -356,9 +356,19 @@ export default {
...
+
                 if (addressService.doesAddressInStorageAndFormMatch(storedAddress, state.address)) {
                     addressCoords = [storedAddress.Field2, storedAddress.Field1];
                     commit(UPDATE_GEO_LOCATION, addressCoords);
+                } else if (window.localStorage) {
+                    const addressDetails = state.address;
+
+                    window.localStorage.setItem('je-full-address-details', JSON.stringify({
+                        PostalCode: addressDetails.postcode,
+                        Line1: addressDetails.line1,
```
charliearlie, line 368: Do we not also need to add the new coordinates we retrieve from the API into Field1 and Field 2 here as well?

kevinrodrigues: The coordinates come from Loqate, which is the call from the searchbox on home so we won't be able to update that at checkout right now. If the user completes payment then goes back to home for another order in the future and searches then the lat long gets pulled through. For it to work in checkout we need  a Loqate call to be added.

charliearlie: Oh okay, makes sense but then won't the stored address and stored coordinates mismatch? Because the coordinates we'll have will be for the previous address?

kevinrodrigues: Should clear it and only add the edited values to local storage.. Or have i missed anything?

simonsandersje (dismissed):

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -356,9 +356,19 @@ export default {
 
             if (isAddressInLocalStorage) {
                 const storedAddress = addressService.getAddressFromLocalStorage(false);
+
                 if (addressService.doesAddressInStorageAndFormMatch(storedAddress, state.address)) {
                     addressCoords = [storedAddress.Field2, storedAddress.Field1];
                     commit(UPDATE_GEO_LOCATION, addressCoords);
+                } else if (window.localStorage) {
```
zlatinivanov-je: Apologies if I don't see it clearly, but Line 357 checks `isAddressInLocalStorage` and on this line here we check for `window.localStorage` is there some opportunity to optimise this? Use the `addressService` for example.

zlatinivanov-je (commented):
Just a small comment to clarify. Other than that happy to approve

packages/components/organisms/f-checkout/src/store/checkout.module.js
kevinrodrigues: Fixed.

DamianMullins (approved):

simonsandersje (approved):

zlatinivanov-je (approved):

PR #1074, opened 2021-06-24
fozzie-components@v3.42.3 – Fix storybook deploy

ashleynolan:
## Fixed
- Storybook deploy should reference yarn (to call lerna indirectly).


ashleynolan:
Merging this as the build is currently broken and this will fix it. Any comments on it though lt me know 😄

PR #1079, opened 2021-06-28
f-form-field@1.16.0 - Fix hover state

FayeCarter:
### Fixed
- Hover state to be consistent across all field types.
- First grouped field to have correct `border-radius`.

| Before | After | 
| --- | --- |
ezgif com-gif-maker (2) |  ezgif com-gif-maker (3) |

### DIFF
--- a/packages/components/atoms/f-form-field/src/components/FormDropdown.vue
+++ b/packages/components/atoms/f-form-field/src/components/FormDropdown.vue
@@ -70,6 +70,9 @@ export default {
 </script>
 
 <style lang="scss" module>
+$form-input-bg                            : $color-container-default;
+$form-input-bg--hover                     : darken($form-input-bg, $color-hover-01);
+
 .c-formDropdown-icon {
     width: spacing(x1.5);
     position: absolute;
@@ -95,5 +98,9 @@ export default {
     /* Remove default styling */
     border: none;
     appearance: none;
+
+    &:hover {
+        background-color: $form-input-bg--hover;
+    }
 }
 </style>
--- a/packages/components/atoms/f-form-field/src/components/FormDropdown.vue
+++ b/packages/components/atoms/f-form-field/src/components/FormDropdown.vue
@@ -70,8 +70,7 @@ export default {
 </script>
 
 <style lang="scss" module>
-$form-input-bg                            : $color-container-default;
-$form-input-bg--hover                     : darken($form-input-bg, $color-hover-01);
+$form-input-bg--hover                     : darken($color-container-default, $color-hover-01);
 
 .c-formDropdown-icon {
     width: spacing(x1.5);
--- a/packages/components/atoms/f-form-field/src/components/FormField.vue
+++ b/packages/components/atoms/f-form-field/src/components/FormField.vue
@@ -294,6 +294,12 @@ $form-input-focus--boxShadow              : 0 0 0 2px $form-input-focus;
     & + & {
         margin-top: spacing(x2);
     }
+
+    &:hover {
+        .c-formField-field {
+            background-color: $form-input-bg--hover;
+        }
+    }
 }
 
     .c-formField-fieldWrapper {
@@ -317,9 +323,6 @@ $form-input-focus--boxShadow              : 0 0 0 2px $form-input-focus;
         background-clip: padding-box;
         padding: $form-input-padding;
 
-        &:hover {
-            background-color: $form-input-bg--hover;
-        }
 
         &:focus,
         &:active,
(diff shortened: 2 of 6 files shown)


JoshuaNg2332 (dismissed):

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -294,6 +294,12 @@ $form-input-focus--boxShadow              : 0 0 0 2px $form-input-focus;
     & + & {
         margin-top: spacing(x2);
     }
+
+    &:hover {
+        .c-formField-field {
```
ashleynolan: Looking at the video, this means the hover applies when hovering over any part of the outer wrapper?

This looks a bit strange to me – what's the reason for changing this? (would be the sort of change I'd user test to see if users think it's a bit confusing)

ashleynolan: Notice you've normalised it actually, so it's not a new addition. May have already checked, but do we know if it should be that the background always appears (when hovering labels) or should we have it so that it never changes the bg colour (so make it consistent the opposite way, if that makes sense).

minrwhite:
I think the visually consistent behaviour added here is great - provided the 'click' behaviour is also consistent. I'm assuming previously the `:hover` state was enabled when hovering over a label, because when you click on that label, it focuses on the associated input element. If you have an area which you can hover over, causes the background change, but that then doesn't focus the form element (or the first form element if it is in a set without labels) on a click in that location then that could be a jarring experience for users.

FayeCarter:
@ashleynolan and @benwhite-deltas, thanks for this! I've moved the hover back out of the wrapper now. @ross-gratton suggested it would be useful to move the text box description into the label, this way when you click the description text it will also move focus to the text area.
I've made these changes now if you're free to take a look, please?

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
```diff
@@ -95,5 +98,9 @@ export default {
     /* Remove default styling */
     border: none;
     appearance: none;
+
+    &:hover {
```
simonsandersje, line 102: Did you check if this needs updating for focus state?

simonsandersje (dismissed):

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
```diff
@@ -70,6 +70,9 @@ export default {
 </script>
 
 <style lang="scss" module>
+$form-input-bg                            : $color-container-default;
```
minrwhite: Should we be using this as the background colour for form elements when not `:hover`ed? I can't see this used anywhere apart from in the definition for `$form-input-bg--hover`, in which case, defining this separately is probably not necessary?

DamianMullins (dismissed):

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
FayeCarter: Good point 👍

FayeCarter: @benwhite-deltas I've now removed this, thanks! 😄

minrwhite (approved):

dandel10n (approved):

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
```diff
@@ -70,6 +70,8 @@ export default {
 </script>
 
 <style lang="scss" module>
+$form-input-bg--hover                     : darken($color-container-default, $color-hover-01);
```
ashleynolan, line 73: Small one, but if you want to share this across components, you can move the variables out into the [common.scss](https://github.com/justeat/fozzie-components/blob/master/packages/components/atoms/f-form-field/src/assets/scss/common.scss) file and then those get imported into all components to be used.

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -17,14 +17,19 @@
                 :is-inline="isInline"
                 :data-test-id="testId.label">
                 {{ labelText }}
+                <template #description>
+                    <p
```
ashleynolan, line 21: Does this render as a `p` tag inside a label? If so, this would be better as a span, as block-level elements like `<p>` inside a label isn't valid html.

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -17,14 +17,19 @@
...
+                        v-if="hasInputDescription"
+                        :class="[
+                            'u-spacingTop',
+                            'u-spacingBottom--large',
+                            $style['c-formField-labelDescription']
```
ashleynolan, line 26: Should be `c-formField-label-description` to match the naming scheme as it's a child element of `c-formField-label`

PR #1080, opened 2021-06-29
fozzie-components@v3.42.6 - Fix Storybook deploy

siggerzz:
Since adding the `--since origin master` check to build changes packages, we need to modify the CircleCI config to build all packages if the `deploy-storybook` CI job is running, otherwise it'll cause no packages to build, resulting in a failed deploy


DamianMullins (dismissed):

ashleynolan (dismissed):

.circleci/config.yml
```diff
@@ -34,18 +34,19 @@ commands:
           echo 'Changes detected:'
           echo $changes
 
-          if [[ $changes ]]; then
-            echo 'export ROOT_CHANGE=true' >> $BASH_ENV
+          if [[ $changes ] || [ $CIRCLE_BRANCH == "master" ]]; then
```
ashleynolan: What does this extra part do?

Looks like it needs to be `[[ $changes ]] || [[ $CIRCLE_BRANCH == "master" ]]`

.circleci/config.yml
```diff
@@ -34,18 +34,19 @@ commands:
           echo 'Changes detected:'
           echo $changes
 
-          if [[ $changes ]]; then
-            echo 'export ROOT_CHANGE=true' >> $BASH_ENV
+          if [[ $changes ]] || [ $CIRCLE_BRANCH == "master" ]; then
```
siggerzz, line 37: We need to run the build against all packages when in master, otherwise the visual regression baseline images won't get set 👍🏼

JoshuaNg2332 (approved):

DamianMullins (approved):

PR #1086, opened 2021-06-30
storybook@v0.37.0 - add optional chaining config modification

matthewhardern:
This PR fixes / allows babel plugins to work correctly within storybook stories, this enables the use of the optional chaining plugin on storybook serve and build. This approach taken modifies the `main.js` storybook config file to modify the webpack config to force it to make a `rootMode: 'upward'` check for babel config. The reason we have to do this is because babel does not play well with vue-cli-service running outside of root within a mono repo like ours, I have tried to get babel to work correctly, of which you can indeed get storybook to pick up on `babel.config.js` by placing it in the `config` directory. However it does not actually apply any of the plugins for some reason. Hence why this method has been chosen.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "root",
-  "version": "3.42.6",
+  "version": "3.42.7",
   "private": true,
   "scripts": {
     "build": "cross-env-shell \"lerna run $LERNA_ARGS build --stream\"",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "root",
-  "version": "3.42.6",
+  "version": "3.42.5",
```
DamianMullins:
```suggestion
  "version": "3.42.7",
```

matthewhardern: yep my bad

DamianMullins (approved):

dandel10n (approved):

PR #1087, opened 2021-06-30
f-media-element@v1.0.0 - changes to allow more custom control with flexbox using new flex prop

matthewhardern:
This PR adds a new prop called flex, this allows a lot finer control over the layout of the `f-media-element` at multiple screen sizes. I have also updated the styles to have new classes for flex example `c-flex-col-gt`. I have also added modifiers to front like `narrowMid:c-flex-col-gt` for example. Let me know if people would prefer a different approach.

### DIFF
--- a/packages/components/molecules/f-media-element/CHANGELOG.md
+++ b/packages/components/molecules/f-media-element/CHANGELOG.md
@@ -4,6 +4,19 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.5.0
+------------------------------
+*June 30, 2021*
+
+### Added
+- Added new prop called flex, which allows custom flex rules when implementing
+
+### Removed
+- removed stackOnNarrow as we now allow breakpoints to be passed into flex prop
+- removed stacked as we now allow breakpoints to be passed into flex prop
+- removed reverse as we now allow reverse to be passed into flex prop
+
+
 v0.4.0
 ------------------------------
 *June 21, 2021*
--- a/packages/components/molecules/f-media-element/CHANGELOG.md
+++ b/packages/components/molecules/f-media-element/CHANGELOG.md
@@ -4,9 +4,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-v0.5.0
+v1.0.0
 ------------------------------
-*June 30, 2021*
+*July 02, 2021*
 
 ### Added
 - Added new prop called flex, which allows custom flex rules when implementing
--- a/packages/components/molecules/f-media-element/src/FlexStyles.js
+++ b/packages/components/molecules/f-media-element/src/FlexStyles.js
@@ -1,0 +1,86 @@
+import { MODIFIER_OPPOSITES_RULES_MAP, MODIFIER_RULES_MAP } from './constants';
+
+class FlexStyles {
+    flexDirection = '--row';
+
+    flexReverse = '';
+
+    defaultFlexDirection = '--row';
+
+    defaultFlexReverse = '';
+
+    operator = '';
+
+    operatorOpposite = '';
+
+    mediaSize = '';
+
+    static returnReverseClassSegment (value) {
+        return value ? '--reverse' : '';
+    }
+
+    static returnDirectionClassSegment (value) {
+        return value ? '--col' : '--row';
+    }
+
+    get styles () {
+        const test = [
+            this._createDefaultStyle(),
+            this._createOppositeStyle(),
+            this._createStyle()
+        ];
+
+        console.log(test);
+
+        return test;
+    }
+
+    set column (value) {
+        this.flexDirection = FlexStyles.returnDirectionClassSegment(value);
+    }
+
+    set reverse (value) {
+        this.flexReverse = FlexStyles.returnReverseClassSegment(value);
+    }
+
+    set defaultColumn (value) {
+        this.flexDirection = FlexStyles.returnDirectionClassSegment(value);
+    }
+
+    set defaultReverse (value) {
+        this.flexReverse = FlexStyles.returnReverseClassSegment(value);
+    }
+
+    set rule ([operator, size]) {
+        this.operator = `--${MODIFIER_RULES_MAP[operator]}`;
+        this.operatorOpposite = `--${MODIFIER_OPPOSITES_RULES_MAP[operator]}`;
+        this.mediaSize = `--${size}`;
+    }
+
+    constructor ({
+        default: { column: defaultColumn, reverse: defaultReverse },
+        modifier
+    }) {
+        this.defaultColumn = defaultColumn;
+        this.defaultReverse = defaultReverse;
+        if (modifier) {
+            this.column = modifier?.column;
+            this.reverse = modifier?.reverse;
+            this.rule = modifier?.rule;
+        }
+    }
+
+    _createDefaultStyle () {
+        return `c-mediaElement${this.defaultFlexDirection}${this.defaultFlexReverse}`;
+    }
+
+    _createOppositeStyle () {
+        return `c-mediaElement${this.operatorOpposite}${this.mediaSize}${this.defaultFlexDirection}${this.defaultFlexReverse}`;
+    }
+
+    _createStyle () {
+        return `c-mediaElement${this.operator}${this.mediaSize}${this.flexDirection}${this.flexReverse}`;
+    }
+}
+
+export default FlexStyles;
--- a/packages/components/molecules/f-media-element/src/FlexStyles.js
+++ b/packages/components/molecules/f-media-element/src/FlexStyles.js
@@ -1,104 +1,19 @@
 import { MODIFIER_OPPOSITES_RULES_MAP, MODIFIER_RULES_MAP } from './constants';
 
 class FlexStyles {
-    flexDirection = '--row';
-
-    flexReverse = undefined;
-
     defaultFlexDirection = '--row';
 
     defaultFlexReverse = '';
 
-    operator = undefined;
+    flexDirection = '--row';
 
-    operatorOpposite = undefined;
+    flexReverse;
 
-    mediaSize = undefined;
+    operator;
 
-    /**
-     * Adds the specific part of css class string
-     * @param value
-     * @returns {string}
-     */
-    static returnReverseClassSegment (value) {
-        return value ? '--reverse' : '';
-    }
+    operatorOpposite;
 
-    /**
-     * Adds the specific part of css class string
-     * @param value
-     * @returns {string}
-     */
-    static returnDirectionClassSegment (value) {
-        return value ? '--col' : '--row';
-    }
-
-    /**
-     * Function creates a string (css class name) using the previously set class properties
-     * @returns {*}
-     * @private
-     */
-    static createStyle (...values) {
-        if (!values.some(value => value === undefined)) {
-            return values.reduce((a, c) => `${a}${c}`, 'c-mediaElement');
-        }
-        return undefined;
-    }
-
-    /**
-     * Compiles a list of css classes to be applied
-     * @returns {*[]}
-     */
-    get styles () {
-        return [
-            FlexStyles.createStyle(this.defaultFlexDirection, this.defaultFlexReverse),
-            FlexStyles.createStyle(this.operatorOpposite, this.mediaSize, this.defaultFlexDirection, this.defaultFlexReverse),
-            FlexStyles.createStyle(this.operator, this.mediaSize, this.flexDirection, this.flexReverse)
-        ].filter(value => value !== undefined);
-    }
-
-    /**
-     * Sets the flex direction css string part value
-     * @param value
-     */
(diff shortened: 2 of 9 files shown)


packages/components/molecules/f-media-element/src/constants.js
```diff
@@ -11,3 +11,19 @@ export const FONT_SIZE = {
     XL: 'xl',
     XXL: 'xxl'
 };
+
+export const BREAK_POINTS = {
```
zlatinivanov-je: I don't see where `BREAK_POINTS` is used.

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -74,69 +75,81 @@ export default {
...
-                case ALIGN.LEFT:
-                    return this.$style['c-mediaElement-content--left'];
-                case ALIGN.RIGHT:
-                    return this.$style['c-mediaElement-content--right'];
-                case ALIGN.CENTER:
-                    return this.$style['c-mediaElement-content--center'];
-                default:
-                    return this.$style['c-mediaElement-content--left'];
-            }
+            const alignment = Object.keys(ALIGN)
```
zlatinivanov-je, line 94: Sometimes I prefer `switch` statements, they appear to read a bit easier to me.

zlatinivanov-je (commented):
Just one point to clarify on the `BREAK_POINTS` use

packages/components/molecules/f-media-element/src/constants.js
matthewhardern: thanks have removed

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: I can agree to a point (as I originally added them), I agree when we have a limited set of cases they can indeed be easier to read, but in this case I am thinking ahead for what I would like to do next which is to allow alignments on top of the existing left,center and right. This is when switch statements begin to smell a bit for me, when your writing essentially the same thing out multiple times and you know that additional cases could be added in the future. This is where I think a simple object lookup via key works best as I just add more options into the object if I want in this case additional alignments.

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -145,87 +131,181 @@ export default {
...
+    /**
+    * Modifier – .c-mediaElement--gt--$name--col
+    * Modifier – .c-mediaElement--gt--$name--col--reverse
+    * Modifier – .c-mediaElement--gt--$name--row
+    * Modifier – .c-mediaElement--gt--$name--row--reverse
+    *
+    * Applies flex direction column on $name
+    */
+    @include media('>'+$name) {
+        .c-mediaElement--gt-#{$name}--col {
```
minrwhite: Missing hyphen, I think?
```suggestion
        .c-mediaElement--gt--#{$name}--col {
```

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -145,87 +131,181 @@ export default {
...
+
+    /**
+    * Modifier – .c-mediaElement--gte--$name--col
+    * Modifier – .c-mediaElement--gte--$name--col--reverse
+    * Modifier – .c-mediaElement--gte--$name--row
+    * Modifier – .c-mediaElement--gte--$name--row--reverse
+     *
+     * Applies flex direction column on $name
+     */
+    @include media('>='+$name) {
```
minrwhite, line 195: I feel like you could nest these another level deeper to avoid the repetition, but it might be already a bit too clever to abstract any further!

E.g. 
```scss
$operators: (
    'lte': '<=',
    /* ... */
)

@each $name, $value in $breakpoints {
    @each $operator, $operatorClassAbbreviation in $operators {
        @include media($operator+$name) {
            .c-mediaElement-col--#{$operatorClassAbbreviation}--#{$name} { /* ... */ }
            /* ... */
        }
    }
}
```

packages/components/molecules/f-media-element/src/FlexStyles.js
```diff
@@ -0,0 +1,82 @@
+import { MODIFIER_OPPOSITES_RULES_MAP, MODIFIER_RULES_MAP } from './constants';
```
minrwhite, line 1: Just, wow! 💯

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: yea thanks great catch.

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: Yea this was the same conversation I had with @ashleynolan tbh. It's a hard one because my developer head tells me it can be abstracted further, but from a readability point it may make it harder to understand. What's your thoughts?

packages/components/molecules/f-media-element/CHANGELOG.md
```diff
@@ -4,6 +4,19 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.5.0
```
ashleynolan:
```suggestion
v1.0.0
```

packages/components/molecules/f-media-element/CHANGELOG.md
```diff
@@ -4,6 +4,19 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+------------------------------
+*June 30, 2021*
```
ashleynolan: Probably worth a bump

```suggestion
*July 2, 2021*
```

packages/components/molecules/f-media-element/src/FlexStyles.js
```diff
@@ -0,0 +1,117 @@
...
+
+class FlexStyles {
+    flexDirection = '--row';
```
ashleynolan: Might be a stupid question (I don't use class syntax often).

Is there a reason these setup properties wouldn't be inside the constructor so that all of the initial values are grouped together?

Also, it makes more sense to me to have the constructor at the top of the class, rather than at the bottom as it is currently – just makes it a little clearer imo.

packages/components/molecules/f-media-element/src/FlexStyles.js
```diff
@@ -0,0 +1,117 @@
...
+    static returnReverseClassSegment (value) {
+        return value ? '--reverse' : '';
+    }
+
+    /**
+     * Adds the specific part of css class string
+     * @param value
+     * @returns {string}
+     */
+    static returnDirectionClassSegment (value) {
```
ashleynolan: Super NIT naming thing – we tend to prefix these kind of functions `get` rather than `return`.

Not a hard and fast rule, but I'd probably update these to be `getDirectionClassSegment` and above to `getReverseClassSegment`

packages/components/molecules/f-media-element/src/FlexStyles.js
```diff
@@ -0,0 +1,117 @@
...
+        return value ? '--col' : '--row';
+    }
+
+    /**
+     * Function creates a string (css class name) using the previously set class properties
+     * @returns {*}
+     * @private
+     */
+    static createStyle (...values) {
```
ashleynolan: Would `createClassname` be a more accurate name?

packages/components/molecules/f-media-element/src/FlexStyles.js
```diff
@@ -0,0 +1,117 @@
...
+            return values.reduce((a, c) => `${a}${c}`, 'c-mediaElement');
+        }
+        return undefined;
+    }
+
+    /**
+     * Compiles a list of css classes to be applied
+     * @returns {*[]}
+     */
+    get styles () {
```
ashleynolan: Same here – would `classnames` be better here?

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -145,87 +131,181 @@ export default {
...
         text: subheading-s,
     ),
-    fontSizeXXLarge: (
+    xxl: (
         title: heading-xxl,
         text: subheading-l,
     )
 );
 
+$flexMargin: spacing(x3);
```
ashleynolan: I'd call this `$mediaElement-content-margin`

packages/components/molecules/f-media-element/src/components/MediaElement.vue
ashleynolan: Yeh, I'm on the side of having this listed out (just as it doesn't impact performance in any way, as it compiles out to the larger code block), but interested in hearing people's thoughts.

packages/components/molecules/f-media-element/src/components/_tests/MediaElement.test.js
```diff
@@ -56,75 +56,6 @@ describe('MediaElement.vue', () => {
             expect(textWrapper.text()).toEqual(mockText);
         });
 
-        it('should apply stack class when stacked prop is true', () => {
```
ashleynolan, line 59: Should we have new tests for the flex class going in?

packages/components/molecules/f-media-element/src/FlexStyles.js
matthewhardern: Point 1 it's generally a rule that putting the class fields at the top allows you to document the object properties. I follow the [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes](url) for this. They quote the following "By declaring fields up-front, class definitions become more self-documenting, and the fields are always present."

Point 2 I agree constructor would be better suited at the top of the class. I will make that change.

packages/components/molecules/f-media-element/src/FlexStyles.js
matthewhardern: Yea makes sense happy to update

packages/components/molecules/f-media-element/src/FlexStyles.js
matthewhardern: agreed will name `createClassName`

packages/components/molecules/f-media-element/src/FlexStyles.js
matthewhardern: agreed `classNames`

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: yep will make the change.

packages/components/molecules/f-media-element/src/FlexStyles.js
ashleynolan: Makes sense. I would probably get rid of the extra line breaks between properties too

With the undefined properties, could they just be defined as

```
flexReverse;
operator;
…
```

matthewhardern: yea good point will make the change.

matthewhardern: Actually eslint requires the spacing?? but yes will make the change to remove undefined.

packages/components/molecules/f-media-element/src/components/_tests/MediaElement.test.js
matthewhardern: Yep have added tests to support the new prop

packages/components/molecules/f-media-element/src/FlexStyles.js
ashleynolan: Weird – what was the ESLint error out of interest? Can maybe look to update this in a separate PR then

packages/components/molecules/f-media-element/src/FlexStyles.js
```diff
@@ -0,0 +1,117 @@
...
+            FlexStyles.createClassName(this.operatorOpposite, this.mediaSize, this.defaultFlexDirection, this.defaultFlexReverse),
+            FlexStyles.createClassName(this.operator, this.mediaSize, this.flexDirection, this.flexReverse)
+        ].filter(value => value !== undefined);
+    }
+
+    /**
+     * Sets the flex direction css string part value
+     * @param value
+     */
+    set column (value) {
```
ashleynolan, line 77: Another dumb class question from me (won't hold up the PR, just I'm interested) – do the set functions get called when the Class is initialised? I couldn't see them getting called anywhere so I'm assuming so (which is why I thought I'd ask!)

ashleynolan (approved):

packages/components/molecules/f-media-element/src/FlexStyles.js
matthewhardern: Yea in this case they are fired when I set the values in the constructor. They can also be set when values are set in a method. They are fired whenever you do `this.column = true` for example, the set for `column` then fires and passes the value that is set in this example `true` you can then use this function to set other variables on the class and do anything with passed value.

packages/components/molecules/f-media-element/src/components/MediaElement.vue
matthewhardern: For now I will leave as is and we can come back and adjust if we are finding the repetition annoying when doing any future updates.

minrwhite (approved):

PR #1089, opened 2021-07-01
f-button@1.8.1 Fix link event handler

zlatinivanov-je:
Resolves an issue where `f-button` link event handlers were not bound
--------------

### Fixed

- `$listeners` to button link issue
 - When `v-on` is set to `f-button` with `href` e.g. link would not trigger

Checks
--------

## Browsers Tested

- [ ] Chrome (latest)
- [ ] Internet Explorer 11


dandel10n (dismissed):
As I understand this PR is waiting for #1055 to be merged first

ashleynolan:
Feel free to make this v1.8.1 as I'm re-working my PR a little based on feedback

minrwhite (dismissed):

DamianMullins (dismissed):

zlatinivanov-je:
Seems to consistently fail CircleCI, although the changes would not suggest to be the reason... 🆘

matthewhardern:
> Seems to consistently fail CircleCI, although the changes would not suggest to be the reason... 🆘

I have just seen this SOS so thought I would investigate, it looks like this is not a fault of yours at all.It's because some time ago in the `IconButton.stories.js` the code references the `f-vue-icons` package directly, instead it should reference `@justeat/f-vue-icons` and this should be added as a dev dependency in package.json. The reason I think this is breaking now is because we no longer build all packages as part of the CI process so it can no longer find it directly. If you update the file to reference icons correctly and add the devdependency it will work and build.

matthewhardern (approved):

DamianMullins (approved):

PR #1090, opened 2021-07-01
fozzie-components@v3.44.0 - Bundlewatch optimisation #globalconfig

siggerzz:
This PR:

- Adds a new `maxBundleSize` property to each `package.json` file where we want Bundlewatch checks.
- Adds a new prompt to the component generator asking if Bundlewatch is required, and if so, what the max bundle size should be.

- Only runs Bundlewatch against changed packages if `maxBundleSize` is present.


ashleynolan:
May need to update this a little as @xander-marjoram just merged in a PR on the bundlewatch config.

bundlewatch.config.js
```diff
@@ -1,48 +1,44 @@
...
+const { execSync } = require('child_process');
+let outputChangedPackages;
+
+const maxBundleSizeExistsForPackage = packageLocation => {
+    const { maxBundleSize } = require(`${packageLocation}/package.json`);
+    
+    return (maxBundleSize);
+}
+
+const getMaxSizeForPackage = packageLocation => {
```
xander-marjoram, line 6: What's the difference between these two methods?

bundlewatch.config.js
```diff
@@ -1,48 +1,44 @@
...
+        outputChangedPackages = execSync('npx lerna ls --since origin/master --json');
+    } catch (error) {
+        console.info('No changed packages found.');
+        process.exit(0);
+    }
+
+    const changedPackagesArray = JSON.parse(outputChangedPackages.toString());
+
+    let changedPackageLocations = [];
+    changedPackagesArray.forEach(package => changedPackageLocations.push(package.location));
```
xander-marjoram: Slight simplification
```suggestion
    const changedPackageLocations = changedPackagesArray.map(package => package.location);
```

packages/components/molecules/f-alert/package.json
```diff
@@ -25,6 +26,7 @@
     "fozzie"
   ],
   "scripts": {
+
```
xander-marjoram: Just wondering if there's a reason for all these new lines?

packages/tools/generator-component/generators/app/prompts.js
```diff
@@ -55,6 +55,20 @@ const prompts = [
...
         }
     },
+    {
+        message: 'Does your component require Bundlewatch checks?',
+        name: 'needsBundlewatch',
+        type: 'confirm',
+        default: true,
+    },
+    {
+        message: "What size limit do you want in place for your bundles (in kB)? e.g. '15'",
```
xander-marjoram, line 65: Is it possible to have a default value for this based on the type of component? Or set a very strict 5kB for new packages which can be increased later?

siggerzz: Will set 5kB as the default for now :)

xander-marjoram (approved):

siggerzz:
Shouldn't have to pull master in this time around as there are no visual regression differences

DamianMullins (approved):

PR #1092, opened 2021-07-02
f-statistics@v0.1.0 - Pivot MVP from ElasticSearch

tommcclean:
[The readme can be found here ](https://github.com/tommcclean/fozzie-components/blob/f-statistics%40v0.1.0/packages/services/f-statistics/README.md)

- This PR replaces `f-stat-client`, it is essentially a refactor where we push to somewhere different
- It isn't intended that we will push this way forever, this is an MVP prototype - so it might seem a little noddy right now.
- It is designed for Web Hosts to create the module and then provide it to any stat generators; like `f-http`, rather than relying on those generators to initialise it.


.yarnrc
```diff
@@ -3,3 +3,6 @@
 
 
 yarn-path ".yarn/releases/yarn-1.19.1.js"
+
+registry "https://registry.yarnpkg.com"
+"@justeat:registry" "http://packages.je-labs.com/npm/private-npm"
```
tommcclean, line 8: This won't work; looking for a workaround

PR #1094, opened 2021-07-02
f-analytics@v0.3.0 - Create a new service (mixin) to wrap GTM actions

oliversweb:
### Added
- Added `analytics.mixin.vue` to handle GTM/GA` platformData` push logic


packages/services/f-analytics/src/mixins/_tests/analytics.created.test.js
```diff
@@ -0,0 +1,119 @@
...
+                }
+            );
+
+            // Assert
+            expect(registerStoreModuleSpy).toHaveBeenCalled();
+            expect(prepareServersideAnalyticsSpy).toHaveBeenCalled();
+            expect(storeUpdatePlatformDataSpy).not.toHaveBeenCalled();
+        });
+
+        it('should set the \'serverside only\' \'platformData\' properties with appropriate defaults if serverside but values not available', () => {
```
kevinrodrigues: Doesn't have to change but you could use back ticks here instead of the escapes.

packages/services/f-analytics/src/mixins/_tests/analytics.methods.test.js
```diff
@@ -0,0 +1,218 @@
...
+                        localVue,
+                        mocks: {
+                            gtmSettings
+                        }
+                    }
+                );
+
+                // Assert
+                expect(preparePageSpy).toHaveBeenCalled();
+                expect(document.head.innerHTML).toContain('src="https://www.googletagmanager.com/gtm.js?id=GTM-123456A"');
```
kevinrodrigues, line 57: More of a suggestion but would it be cleaner to just check that the method returns true or falsey if `window.dataLayer` isn't defined? That way it can save the setup of checking for the structure of the innerHTML code. (e.g the GTM id could change and this test would fail)

packages/services/f-analytics/src/store/analyticsModule.js
```diff
@@ -0,0 +1,32 @@
...
+            jeUserPercentage: 0,
+            currency: '',
+            version: '',
+            instancePosition: ''
+        }
+    }),
+
+    actions: {
+        updatePlatformData: ({ commit }, platformData) => {
+            commit('updatePlatformData', platformData);
```
kevinrodrigues, line 23: Do we want a `mutations` constants file for this `updatePlatformData`

packages/services/f-analytics/src/mixins/analytics.mixin.vue
```diff
@@ -1,5 +1,166 @@
...
+            const platformData = { ...this.platformData };
+
+            platformData.name = this.$route.name;
+            platformData.appType = 'web';
+            platformData.applicationId = 7;
+            platformData.userAgent = navigator.userAgent || 'N/A';
+            platformData.branding = this.getBrand(this.$i18n.locale);
+            platformData.country = this.getCountry(this.$i18n.locale);
+            platformData.language = this.getLanguage(this.$i18n.locale);
+            platformData.currency = this.getCurrency(this.$i18n.locale);
```
FayeCarter: I _think_ lines 93 -102 can be re-written to something like 
```
const platformData = { 
           ...this.platformData,
           name: this.$route.name,
           appType: 'web',
           (same with the rest of these changes)
  };

packages/services/f-analytics/src/mixins/_tests/analytics.methods.test.js
oliversweb: I want to make sure the GTM markup is correct and not changed (broken) by anyone so that's why I check for specifics. The GTM id is mocked.

packages/services/f-analytics/src/store/analyticsModule.js
oliversweb: I thought it was a bit of overkill atm but if it extends into a more richer $store in the future then would be good to add then.

packages/services/f-analytics/src/mixins/analytics.mixin.vue
oliversweb: I tried that before but if any of the lines cause an exception then the error message only indicates the first line (the declaration line) and not the line that caused the error.  So for maintenance/debugging reasons I did it this way.

packages/services/f-analytics/src/mixins/_tests/analytics.created.test.js
oliversweb: I had that before but on the test output if the value between the back ticks is not a value that can be resolved it displays a warning indicating that it cannot find the value, e.g. Cannot find `serverside only`.  So did it this way.

kevinrodrigues (dismissed):

FayeCarter (approved):

kevinrodrigues (approved):

PR #1095, opened 2021-07-05
f-checkout@v0.152.0 Fix phone number prepopulation for returning guest user

pkedziora:
Fixes prepopulation of mobile phone number for guest users in some cases.

For guest users returning to checkout within first 60 minutes for a different checkout, the JWT auth token does not have phone number claim populated. While this issue is investigated for a long term solution, this fix retrieves the customer's phone number from getCustomerUrl endpoint in case all other sources for phone number did not produce it.

---

## UI Review Checks
No UI changes

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


oliversweb (dismissed):

SachaZvetelman:
Is there a reason to have the call to `getCustomer` inside the `getCheckout` action? I'd keep them as two separate actions, since they are two separate API calls, that, unless I'm missing something, are unrelated.

pkedziora:
@SachaZvetelman The reason I put it there is to keep it local with the other method that reads the phone number from the JWT token (resolveCustomerDetails)

SachaZvetelman:
To me, this breaks the Single Responsibility Principle, since a method responsible for getting the checkout data is now also responsible for getting the customer data as well.
I'd extract it into a separate function, personally.

The other method could be moved as well, but we can discuss/do that as part of a separate PR.

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -146,6 +177,9 @@ export default {
 
             resolveCustomerDetails(data, state);
 
+            // This call can be removed when newly created guest JWT token has phone number claim populated
+            await enrichPhoneNumber(data.customer, state, getCustomerUrl, timeout);
```
FayeCarter: would this need to be after line 183 as you haven't updated the state yet? Then you could also remove the data.customer part and just use what's in the stae

pkedziora: @FayeCarter no, it uses state in the same manner that resolveCustomerDetails uses it, just to retrieve the authToken for the api call

pkedziora: @FayeCarter the second bit with customer being there on the state, yes

pkedziora:
@SachaZvetelman Cool, I'll move it out

packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
```diff
@@ -378,7 +380,13 @@ describe('CheckoutModule', () => {
                 // Use a new copy per test so any mutations do not affect subsequent tests
                 checkoutDeliveryCopy = Object.assign(checkoutDelivery);
 
-                axios.get = jest.fn(() => Promise.resolve({ data: checkoutDeliveryCopy }));
+                axios.get = jest.fn(url => {
+                    if (url === 'http://localhost/customer') {
```
BanksySan: Repeated string `http://localhost/customer`.

BanksySan (dismissed):

packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
SachaZvetelman: This conditional in the `beforeEach` is an indicator that it should probably have a default value, in my humble opinion. You can then override it in the test(s) when you need something different.

i.e. (pseudocode):
```
beforeEach()
  axios.get = ... return Promise.resolve({ data: customer });


it('myTest')
  axios.get = ... return Promise.resolve({ data: checkoutDeliveryCopy });
```

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-checkout",
   "description": "Fozzie Checkout – Fozzie Checkout Component",
-  "version": "0.151.0",
+  "version": "0.154.0",
```
dandel10n:
```suggestion
  "version": "0.152.0",
```

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -884,6 +886,71 @@ describe('Checkout', () => {
...
+                });
+
+                // Act
+                const result = wrapper.vm.shouldLoadCustomer;
+
+                // Assert
+                expect(result).toBe(true);
+            });
+
+            it('should return `false` if customer is logged not logged in', () => {
```
dandel10n:
```suggestion
            it('should return `false` if customer is not logged in', () => {
```

dandel10n (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -392,6 +397,14 @@ export default {
                 (!this.address || !this.address.line1);
         },
 
+        /* If phone number is missing both from chckout api and from
+        * `state.AuthToken`, then retrieve the phone number from customer api
+        * This can happen for newly created guest */
+        shouldLoadCustomer () {
+            return this.isLoggedIn &&
```
SachaZvetelman, line 404: Super nit, but I would keep this in the same line.

dandel10n (approved):

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -302,6 +303,37 @@ export default {
...
+         * @param {Object} context - Vuex context object, this is the standard first parameter for actions
+         * @param {Object} payload - Parameter with the different configurations for the request.
+         */
+        getCustomer: async ({ commit, state, dispatch }, {
+            url,
+            timeout
+        }) => {
+            if (!state.customer || state.customer.mobileNumber) {
+                return;
+            }
```
SachaZvetelman, line 318: Isn't this validation redundant? Or in other words, if we have it here, do we need to validate this in other places?

packages/components/organisms/f-checkout/src/store/tests/checkout.module.test.js
pkedziora: Refactored

packages/components/organisms/f-checkout/src/store/checkout.module.js
pkedziora: Refactored

SachaZvetelman (approved):

packages/components/organisms/f-checkout/src/components/Checkout.vue
pkedziora: Will open separate PR

packages/components/organisms/f-checkout/src/store/checkout.module.js
pkedziora: Will open separate PR

PR #1096, opened 2021-07-05
f-offers@v0.6.0 - adding header component

matthewhardern:
Adding in the offers header with `Header.vue`, this uses the latest `f-media-element` to improve use of components while also improving the responsiveness of the header on lower resolutions. I have created tests for the expected copy on the header when both authenticated and unauthenticated. Also as part of this PR I have made a minor update to offers vuex module, in turn needing slight update to `SearchBox` unit test.

## UI Review Checks

Screenshot 2021-07-05 at 11 41 35

Screenshot 2021-07-05 at 11 41 46

Screenshot 2021-07-05 at 11 41 59

Screenshot 2021-07-05 at 11 42 09

Screenshot 2021-07-05 at 11 43 54

## Browsers Tested

- [x] Chrome (latest)


packages/components/organisms/f-offers/package.json
```diff
@@ -43,8 +43,10 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
+    "@justeat/f-globalisation": "1.0.0",
     "@justeat/f-services": "1.11.0",
-    "@justeat/f-globalisation": "1.0.0"
+    "axios": "0.21.1",
```
zlatinivanov-je, line 48: Thought we were moving towards `f-http`? I'll read on to see if I can spot the reason...

matthewhardern: I have been informed that atm we are not production ready with f-http by @tommcclean

packages/components/organisms/f-offers/src/components/NoOffersFound.vue
```diff
@@ -8,6 +8,7 @@
             image-url="https://d30v2pzvrfyzpo.cloudfront.net/a/hw/img/decoration/oi_no-results-image.svg"
             :image-align="imageAlign"
             :text-size="textSize"
+            stack-when-narrow
```
zlatinivanov-je, line 11: No other references found of `stack-when-narrow`

packages/components/organisms/f-offers/src/components/_tests/Header.test.js
```diff
@@ -0,0 +1,122 @@
...
+
+describe('Header.vue', () => {
+    describe('ssAuthenticated: false', () => {
+        let wrapper;
+
+        beforeEach(() => {
+            // Arrange
+            const getters = {
+                isAuthenticated: () => false,
+                friendlyName: () => 'Matthew'
```
zlatinivanov-je: Is `friendlyName` always expected with a defined with a value? Could it be `undefined` or `null` for example?

packages/components/organisms/f-offers/src/store/offers.module.js
```diff
@@ -0,0 +1,122 @@
...
+            friendlyName: 'Matthew',
+            firstName: '',
+            lastName: ''
+        },
+
+        brazeApiKey: undefined
+    }),
+
+    getters: {
+        isAuthenticated: state => !!state.authToken,
```
zlatinivanov-je: That would suggest the user is always authenticated since the state authToken has a default value.

packages/components/organisms/f-offers/src/store/offers.module.js
```diff
@@ -0,0 +1,122 @@
...
+        firstName: data?.customer?.firstName ? data?.customer?.firstName : tokenData?.given_name,
+        lastName: data?.customer?.lastName ? data?.customer?.lastName : tokenData?.family_name
+    };
+};
+
+export default {
+    namespaced: true,
+
+    state: () => ({
+        authToken: 'test',
```
zlatinivanov-je: Should that be set to `null` or `undefined` instead

matthewhardern: Your correct my mistake I have left that in while testing.

zlatinivanov-je (commented):
Mostly approved, with a few minor comments.

packages/components/organisms/f-offers/src/components/_tests/Header.test.js
matthewhardern: good point I will make sure I am accounting for this, thanks

packages/components/organisms/f-offers/src/store/offers.module.js
matthewhardern: yep should not have default value, I will change back to null

packages/components/organisms/f-offers/src/components/NoOffersFound.vue
matthewhardern: correct, I need to update this to the newer media element, but will be doing that as part of next PR.

matthewhardern:
> Mostly approved, with a few minor comments.

Thanks @zlatinivanov-je I will make the amendments.

packages/components/organisms/f-offers/src/components/_tests/SearchBox.test.js
```diff
@@ -7,7 +7,7 @@ import {
     ACTION_VALID_SEARCH,
     ACTION_INVALID_ADDRESS,
     ACTION_EMPTY_ADDRESS,
-    ACTION_UNKNOWN_ERROR
+    ACTION_UNKNOWN_ERROR, VUEX_MODULE_NAMESPACE_OFFERS_SEARCH
```
minrwhite:
```suggestion
    ACTION_UNKNOWN_ERROR,
    VUEX_MODULE_NAMESPACE_OFFERS_SEARCH
```

packages/components/organisms/f-offers/src/store/offers.module.js
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * @function resolveCustomerDetails
+ * @param  {object} data  - Api response object.
+ * @param  {string} authToken - auth token.
+ */
+const resolveCustomerDetails = (data, authToken) => {
+    const tokenData = jwtDecode(authToken);
+    return {
+        phoneNumber: data?.customer?.phoneNumber ? data?.customer?.phoneNumber : tokenData?.phone_number,
```
minrwhite: [Based on this](https://github.com/justeat/fozzie-components/blob/e6e58e52f8517c53bcfe255ba22c7c02b05dd3d3/packages/components/organisms/f-checkout/src/store/checkout.module.js#L57), do we also need to check `tokenData?.mobile_number` here?

packages/components/organisms/f-offers/src/store/offers.module.js
```diff
@@ -0,0 +1,122 @@
...
+            lastName: ''
+        },
+
+        brazeApiKey: undefined
+    }),
+
+    getters: {
+        isAuthenticated: state => !!state.authToken,
+        friendlyName: state => state.customer.friendlyName
+    },
```
minrwhite: Please can the order of keys in stores be reordered to `state`, `actions`, `mutations` and then `getters`? Currently we haven't yet defined any getters within stores in this repo, but in several internal consumer-facing components, this is the order that is used, so it would be good to be consistent here.

matthewhardern: Yea happy to make that change, apologies was unaware of the convention.

packages/components/organisms/f-offers/src/store/offers.module.js
matthewhardern: Yes you are correct, will make the addition

matthewhardern:
@benwhite-deltas, @zlatinivanov-je I have made the changes you suggested

packages/components/organisms/f-offers/src/store/offers.module.js
```diff
@@ -0,0 +1,123 @@
...
+/**
+ * @function resolveCustomerDetails
+ * @param  {object} data  - Api response object.
+ * @param  {string} authToken - auth token.
+ */
+const resolveCustomerDetails = (data, authToken) => {
+    const tokenData = jwtDecode(authToken);
+    return {
+        phoneNumber: data?.customer?.phoneNumber ? data?.customer?.phoneNumber :
+            (tokenData?.phone_number || tokenData?.mobile_number),
```
minrwhite, line 34: The place I linked before checks mobile_data then falls back to phone_number, could we reverse the order of these to be consistent, unless there's a reason it needs to be different here?
```suggestion
            (tokenData?.mobile_number || tokenData?.phone_number),
```

packages/components/organisms/f-offers/src/store/offers.module.js
```diff
@@ -118,5 +96,28 @@ export default {
...
+    mutations: {
+        [MUTATION_SET_AUTH_TOKEN]: (state, value) => {
+            state.authToken = value;
+        },
+
+        [MUTATION_SET_CUSTOMER_DATA]: (state, { friendlyName, firstName, lastName }) => {
+            state.customer = {
+                friendlyName,
+                firstName,
+                lastName
```
minrwhite, line 110: If these are not defined in the payload, does that mean that the state values could be set from `null` to `undefined`? If so, please could we set default values as `null` to ensure that an 'unset' value remains predictably as null?

minrwhite (approved):
A couple of small points, but happy if these get addressed in a follow-up PR.

packages/components/organisms/f-offers/src/store/offers.module.js
matthewhardern: yep good point will address in a follow up PR

packages/components/organisms/f-offers/src/store/offers.module.js
matthewhardern: again good point will address in follow up PR

zlatinivanov-je (approved):

PR #1097, opened 2021-07-05
f-feature-management@v0.1.0 - Skeleton implementation of feature management service

jnewc:
This adds a stub implementation of the feature management service, and related tests.

Tests are currently disabled while the implementation is being developed.

See README.md for more details.

---

## UI Review Checks


packages/services/f-feature-management/.eslintrc.js
```diff
@@ -0,0 +1,56 @@
...
+  overrides: [
+    {
+      files: ["*.json", "*.jsonc", "*.json5", "*.json6"],
+      parserOptions: {
+        parser: 'jsonc-eslint-parser',
+      }
+    }
+  ],
+  rules: {
+    // json-schema-validator plugin does not play nicely with the vuejs-accessibility plugin.
```
jnewc, line 15: As discussed @ashleynolan

packages/services/f-feature-management/README.md
```diff
@@ -0,0 +1,21 @@
...
+**This is a WIP**
+
+This service allows querying of feature flags as provided by our backend Feature Management service.
+
+## Usage
+
+To instantiate the SDK:
+
+```javascript
+import featureManagement from '@just-eat/je-feature-management'
```
xander-marjoram: Should it be this?
```suggestion
import featureManagement from '@justeat/f-feature-management';
```

packages/services/f-feature-management/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+const scope = 'je-my-team-scope'
```
xander-marjoram:
```suggestion
const scope = 'je-my-team-scope';
```

packages/services/f-feature-management/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@justeat/f-feature-management",
+  "version": "0.1.0",
+  "description": "Authentication helper to communicate with open apis",
```
xander-marjoram: This doesn't seem quite right

packages/services/f-feature-management/src/index.js
```diff
@@ -0,0 +1,34 @@
+function logValueRequest (func, key, value) {
```
xander-marjoram: We generally prefer the ES6 syntax:
```suggestion
const logValueRequest = (func, key, value) => {
```

packages/services/f-feature-management/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+export default function (scope) {
+    function fullKey (key) {
+        return (scope && `${scope}::${key}`) || key;
+    }
+    function forward (fn, key) {
+        return fn(fullKey(key));
+    }
+    return {
+        getBooleanValue: key => forward(getBooleanValue, key),
```
xander-marjoram: This pattern seems like it could be simplified (or at least its readability improved), let's talk through it tomorrow

packages/services/f-feature-management/test/endToEnd.test.js
```diff
@@ -0,0 +1,20 @@
+import featureManagement from '../src/index';
+import expectations from './data/e2e-expectations.json';
+
+const { getBooleanValue, getIntegerValue, getStringValue } = featureManagement('je-web-core');
```
xander-marjoram, line 4: You could probably use a dummy scope for the tests

packages/services/f-feature-management/test/data/simple-fmconfig.json
```diff
@@ -0,0 +1,45 @@
+{
```
xander-marjoram, line 1: Maybe not in this PR, but I feel like this folder could use its own README.

packages/services/f-feature-management/test/endToEnd.test.js
jnewc: It can actually be empty now as the e2e tests assume an empty scope

packages/services/f-feature-management/src/index.js
```diff
@@ -0,0 +1,34 @@
+function logValueRequest (func, key, value) {
+    console.log(`${func.name}: ${key} => ${value}`); // es-lint-disable
+}
+
+function getBooleanValue (key) {
```
gareththackeray, line 5: Given JS is dynamically typed I wonder if we need separate functions per return type.  Would make sense with TypeScript but not sure with JS.  Might be worth having an additional function for any type. (Just called `getValue()` I guess.)

Interestingly Optimizely have introduced a "generic" version: https://docs.developers.optimizely.com/full-stack/docs/get-feature-variable-javascript#section-generic

jnewc: It depends if/how we want to verify that the type is correct for a given key.

For the other SDKs, you'll get a `null`/`nil` value back, which is essentially declaring an error state (as type enforcement requires you to handle it), so the equivalent here would be something like throwing an error ... which we definitely don't want to do IMO.

If we're putting the onus on the consumer to always type check, I'm happy to reduce to one method (although, I'd place a reasonable bet someone will get burned by truthy checking an integer or string at some point).

gareththackeray: I think JS engineers are just used to that.  They'll use `=== true` most of the time I expect, if they are being strict. But OTOH they can embrace truthiness if standards permit.  E.g. empty string or 0 could mean "this is turned off" which is idiomatic for JS.  Maybe just include all of them, then people get the option?  Or we can wait and see what comments this PR gets 😄

xander-marjoram: I think I like there being different methods for each return type, as long as we guarantee that type. It would help with readability too and make sure the developer knows what kind of value they're working with.

If this were typescript, on the other hand, then I'd be happy with something like `getValue<T>`.

packages/services/f-feature-management/test/data/simple-fmconfig.json
jnewc: Ticketed

CHANGELOG.md
```diff
@@ -23,6 +23,12 @@ v3.44.3
 
 ## Fix
 - Fix Bundlewatch fails
+v3.45.0
+------------------------------
+*July 6, 2021*
+
+### Added
+- Feature Management service component (initial skeleton).
```
minrwhite: This should be added at the top.

packages/services/f-feature-management/README.md
```diff
@@ -0,0 +1,21 @@
...
+import featureManagement from '@justeat/f-feature-management';
+
+const scope = 'je-my-team-scope';
+const { getBooleanValue, getIntegerValue, getStringValue } = featureManagement(scope);
+
+// e.g.
+const myFeatureValue = getBooleanValue('my-feature-value');
+```
+
+*NOTE: if you're unsure what scope to use, please reach out to us in #team-experimentation-platform.*
```
minrwhite: As this is an external repo, it seems weird having an internal channel listed here; but I'm not sure what to suggest as an alternative.

packages/services/f-feature-management/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+        return (scope && `${scope}::${key}`) || key;
+    }
+    function forward (fn, key) {
+        return fn(fullKey(key));
+    }
+    return {
+        getBooleanValue: key => forward(getBooleanValue, key),
+        getIntegerValue: key => forward(getIntegerValue, key),
+        getStringValue: key => forward(getStringValue, key)
+    };
```
minrwhite: I'd agree with @xander-marjoram on this. My preference on initial glance would be to refactor fullKey to ES6 syntax and git rid of forward entirely, like so:
```suggestion
    const fullKey = key => scope
        // ternary preferred over && and || operators, and break to indented multi-line for readability
        ? `${scope}::${key}`)
        : key;

    return {
        getBooleanValue: key => getBooleanValue(fullKey(key)),
        getIntegerValue: key => getIntegerValue(fullKey(key)),
        getStringValue: key => getStringValue(fullKey(key))
    };
```

packages/services/f-feature-management/test/data/schema.json
```diff
@@ -0,0 +1,221 @@
...
+                        "additionalProperties": false,
+                        "properties": {
+                            "type": {
+                                "type": "string",
+                                "enum": ["IdHash"],
+                                "description": "The mechanism used for evaluation."
+                            },
+                            "detail": {
+                                "type": "object",
+                                "description": "Can vary depending on 'type'. Only IdHash defined at the moment.",
```
minrwhite, line 58: I'm assuming that when/if more than `IdHash` is defined then there'll be some schema validation based around a `oneOf` structure to mandate certain property requirements depending on which type is chosen?

packages/services/f-feature-management/test/endToEnd.test.js
```diff
@@ -0,0 +1,20 @@
...
+        describe(ex.description, () => {
+            const { value: expectedValue, experimentKey: expectedKey, experimentVariant: expectedVariant } = ex.expectation;
+            const feature = functionMap[ex.valueType](ex.feature);
+            expect(feature.value).toBe(expectedValue);
+            if (expectedKey && expectedVariant) {
+                expect(feature.experimentKey).toBe(expectedKey);
+                expect(feature.experimentVariant).toBe(expectedVariant);
+            }
+        });
+    });
```
minrwhite: There's a way of doing this idiomatically to jest, and it looks something like:
```suggestion
    describe.each(expectations)('Expectation %# - ', ({ description, expectation, feature, valueType }) => {
        it(description, () => {
            const { value: expectedValue, experimentKey: expectedKey, experimentVariant: expectedVariant } = expectation;
            const feature = functionMap[valueType](feature);
            expect(feature.value).toBe(expectedValue);
            if (expectedKey && expectedVariant) {
                expect(feature.experimentKey).toBe(expectedKey);
                expect(feature.experimentVariant).toBe(expectedVariant);
            }
        });
    });
```

packages/services/f-feature-management/test/getBooleanValue.test.js
```diff
@@ -0,0 +1,23 @@
...
+        expect(result.value).toBe(true);
+    });
+
+    describe('Get negative boolean value', () => {
+        const key = 'boolean-value-false';
+        const result = getBooleanValue(key);
+        expect(result.value).toBe(false);
+    });
+
+    describe('Get non-existent boolean value', () => {
```
minrwhite: Replace `describe()`s with `it()`s for individual test cases, [`describe()`s are more of a wrapper](https://archive.jestjs.io/docs/en/23.x/api#describename-fn).

minrwhite (dismissed):
Overall structure looks good - a few comments and opinions below:

packages/services/f-feature-management/test/data/schema.json
jnewc: Yes, that's correct. The existing schema only enforces the requirements for `IdHash`, and we will rework the schema if/when a new type is introduce (which will be infrequently, if ever).

packages/services/f-feature-management/test/endToEnd.test.js
jnewc: Thanks, this is much cleaner.

minrwhite (dismissed):

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v3.45.0
+------------------------------
+*July 6, 2021*
```
xander-marjoram: When you fix the conflicts, please can you bump this date too

packages/services/f-feature-management/src/index.js
jnewc: We've discussed outside of this thread and we're going to add both the generic and non-generic methods.

packages/services/f-feature-management/README.md
jnewc: I'll remove for now as we're in close contact with all teams that would have/need a scope.

packages/services/f-feature-management/test/endToEnd.test.js
```diff
@@ -6,11 +6,13 @@ const { getBooleanValue, getIntegerValue, getStringValue } = featureManagement('
...
-    describe.each(expectations)('Expectation %# - ', ({ description, expectation, feature, valueType }) => {
+    describe.each(expectations)('Expectation %# - ', ({
+        description, expectation, feature, valueType
+    }) => {
         it(description, () => {
             const { value: expectedValue, experimentKey: expectedKey, experimentVariant: expectedVariant } = expectation;
-            const feature = functionMap[valueType](feature);
-            expect(feature.value).toBe(expectedValue);
+            const featureValue = functionMap[valueType](feature);
+            expect(featureValue.value).toBe(expectedValue);
```
minrwhite, line 15: `featureValue.value` reads oddly to me. I'm assuming this is so you don't get a linting error for `feature` being redefined in a nested scope? I'd prefer reassigning `feature` to `featureName` above:
```
    describe.each(expectations)('Expectation %# - ', ({
        description, expectation, feature: featureName, valueType
    }) => {
```
and then continue to use `feature` within the test case.

packages/services/f-feature-management/test/getValue.test.js
```diff
@@ -0,0 +1,41 @@
+import featureManagement from '../src/index';
+
+const { getValue } = featureManagement('');
+
+describe.skip('Generic value retrieval', () => {
```
minrwhite, line 5: Is there a plan for these to be unmarked as `.skip`?

minrwhite (dismissed):

packages/services/f-feature-management/test/endToEnd.test.js
jnewc: Yeah, you're right about why I renamed it, and it will soon also contain the experiment name and variant for tracking too.

I like your suggestion as `feature` is still a better name for the model, so will change (in next PR).

packages/services/f-feature-management/test/getValue.test.js
jnewc: Yes, I'm somewhat doing TDD in the early stages. 

The e2e tests in particular will exercise almost the entire implementation once completed so they'll be un-skipped late in the process, although I may start posting the quantity that are passing on future PRs. 

I can start un-skipping the other tests a few PRs from now.

packages/services/f-feature-management/test/endToEnd.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+describe.skip('End-to-End Tests', () => {
+    describe.each(expectations)('Expectation %# - ', ({
+        description, expectation, feature, valueType
+    }) => {
+        it(description, () => {
+            const { value: expectedValue, experimentKey: expectedKey, experimentVariant: expectedVariant } = expectation;
+            const featureValue = functionMap[valueType](feature);
+            expect(featureValue.value).toBe(expectedValue);
+            if (expectedKey && expectedVariant) {
```
xander-marjoram, line 16: I think we should avoid conditional expectations. Happy for you to update in your next PR though.

jnewc: I suppose I could change it to a falsy check of the value. I think in most cases the expected fields would be `null` whereas the `featureValue` properties would be `undefined`.

Yes, will mull it over and change for next PR 👍

packages/services/f-feature-management/.eslintrc.js
jnewc: This has now been fixed for `eslint-plugin-vuejs-accessibility@v0.7.1` - will update in next PR and hopefully remove all these exceptions 🙌 

FYI @xander-marjoram

xander-marjoram (approved):

minrwhite (approved):

PR #1100, opened 2021-07-06
f-checkout@v0.155.0 - Validate onBlur for Email, Postcode and Mobile

oliversweb:
### Changed
- Email, Postcode and Mobile was validating on every keypress but now validating onBlur.


packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
...
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+v0.155.0
+------------------------------
+*July 6, 2021*
+
+### Changed
+- Email, Postcode and Mobile was validating on every keypress but now validating onBlur.
+- Added a max length of 50 on email.
```
xander-marjoram: Has this requirement come from anywhere in particular?

oliversweb: No but I added a max on all the other fields before and missed this one off, so thought would add something sensible now

packages/components/organisms/f-checkout/src/components/_tests/Guest.test.js
```diff
@@ -78,13 +78,15 @@ describe('Guest', () => {
 
         describe('isEmailValid ::', () => {
             it.each([
-                [true, true, false],
-                [false, true, true],
-                [false, false, false],
-                [false, false, true]
-            ])('should return %s if email.email = %s and `isFieldEmpty` returns %s for email field', (expected, email, isEmpty) => {
+                [true, true, true, false],
```
xander-marjoram, line 82: Do we need more cases now we've got an extra parameter?

packages/components/organisms/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -316,9 +320,13 @@ module.exports = class Checkout extends Page {
      *
      * @param {String} Field name
      * @param {String} Value to set
+     * @param {Boolean} Whether to tab out of the field or not
      */
-    setFieldValue (fieldName, value) {
+    setFieldValue (fieldName, value, tabOut = false) {
         this.fields[fieldName].input.setValue(value);
+        if (tabOut) {
+            this.fields[fieldName].input.setValue(['Tab']);
```
xander-marjoram: I'm a bit torn about whether this should live inside this function or be its own.
Also, small nit: you could assign `this.fields[fieldName].input` to a variable.

packages/components/organisms/f-checkout/test/component/f-checkout-guest.component.desktop.spec.js
```diff
@@ -46,4 +47,37 @@ describe('f-checkout "guest" component tests - @browserstack', () => {
         // Assert
         expect(checkout.getFieldValue(field).length).toEqual(maxlength);
     });
+
+    forEach([
+        ['jazz.man@tunetown.com'],
+        ['jazzman@tunetown.com']
+    ])
+    .it('should be validate email addresses ["%s"]', email => {
```
xander-marjoram:
```suggestion
    .it('should be valid email address: "%s"', email => {
```

packages/components/organisms/f-checkout/test/component/f-checkout-guest.component.desktop.spec.js
```diff
@@ -46,4 +47,37 @@ describe('f-checkout "guest" component tests - @browserstack', () => {
...
+    });
+
+    forEach([
+        ['@jazz.man@tunetown.com'],
+        ['.jazzman@tunetown.com'],
+        ['jazzmantunetown.com'],
+        ['jazzman@'],
+        ['jazzman']
+    ])
+    .it('should be invalidate email addresses ["%s"]', email => {
```
xander-marjoram:
```suggestion
    .it('should be invalid email address: "%s"', email => {
```

xander-marjoram: And is the `[]` around each email address necessary?

packages/components/organisms/f-checkout/test-utils/component-objects/f-checkout.component.js
xander-marjoram: Actually, I've had another idea, how about replacing the `tabOut` parameter with an `options` parameter, which could contain `tabOut`? Then you would call it like this, which I think is easier to read and makes extending it slightly easier

`checkout.setFieldValue('emailAddress', email, { tabOutOfField: true });`

packages/components/organisms/f-checkout/CHANGELOG.md
xander-marjoram: I feel like 50 is quite short, short enough where valid customer emails might be rejected. I think 100 is better, although email addresses can even be something like 250 characters long (or even longer).

xander-marjoram (dismissed):

kevinrodrigues (dismissed):

siggerzz:
@oliversweb Let me know if you'd like to go through these visual regression diffs together 👍🏼

xander-marjoram (approved):

kevinrodrigues (approved):

PR #1108, opened 2021-07-09
f-cookie-banner@v0.20.0 - Reopen link should not be visible in ANZ

simonsandersje:
The reopen link was appearing on tenants using the legacy banner, which was not the correct behaviour

Also fixes the incorrectly wired useGreyBackground prop


packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -165,6 +166,13 @@ export default {
          */
         legacyBanner () {
             return this.shouldShowLegacyBanner === null ? this.config.displayLegacy : this.shouldShowLegacyBanner;
+        },
+        /**
+         * Check if the reopen cookie link should use a grey background
+         * @returns {Boolean}
+         */
+        shouldUseGreyBackground () {
```
DamianMullins: Do you need this completed property? Seems fine to pass along the prop as you were before if you don't need any additional logic?

DamianMullins: Ah, is it so you can configure it in Storybook?

DamianMullins (dismissed):

minrwhite (dismissed):
To stay aligned with how we are passing other props (e.g. `shouldShowLegacyBanner`), I'd prefer if `useGreyBackground` prop became `shouldUseGreyBackground` on the external interface, and the internal computed property was called `useGreyBackground`. That said, that would be a breaking change, and so I'm happy if this is done later.

simonsandersje:
> To stay aligned with how we are passing other props (e.g. `shouldShowLegacyBanner`), I'd prefer if `useGreyBackground` prop became `shouldUseGreyBackground` on the external interface, and the internal computed property was called `useGreyBackground`. That said, that would be a breaking change, and so I'm happy if this is done later.

@benwhite-deltas Understood and changed - luckily, there are no consuming apps that use the prop yet

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
dandel10n: Sorry, but I don't get the reason for this computed prop still

dandel10n: Can't you just use one `shouldUseGreyBackground` prop without a computed one?

simonsandersje: @dandel10n Computed value removed

minrwhite (approved):

FayeCarter (approved):

PR #1113, opened 2021-07-14
f-form-field@v1.17.0 - updates to match PIE

FayeCarter:
### Changed
- Updates to match PIE designs
- Disabled label style

### Added
- `fieldSize` prop.

#### Small - hover
image

#### Medium (default)

| Before | After |
|--- | --- |
| image | Screenshot 2021-07-14 at 14 26 55 | 

#### Large - focus
image

#### Disabled

| Before | After |
|--- | --- |
| image | Screenshot 2021-07-14 at 14 26 15 |


packages/components/atoms/f-form-field/CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v1.17.0
+------------------------------
+*June 30, 2021*
```
ashleynolan: Needs a bump

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -15,6 +15,7 @@
                 :label-style="normalisedLabelStyle"
                 :label-for="uniqueId"
                 :is-inline="isInline"
+                :disabled="$attrs.disabled"
```
ashleynolan: Does this apply a disabled attribute to the form label? If so, I'm pretty sure that's invalid HTML.

If you need to style the label based on the input being disabled, I think you'll need to either have a conditional class apply to the outer form wrapper (but you might hit scope issues with CSS modules then due to the label being a separate component), or you could pass through `$attrs.disabled` as a prop value and then set a class directly on the label element then in the sub-component.

I think you'll need to pass it through as a prop, as I don't think the label will be able to style based on the outer scope of the component.

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -304,12 +327,21 @@ $form-input-focus--boxShadow              : 0 0 0 2px $form-input-focus;
...
+        @include form-field-size($form-input-height-medium, $form-input-padding-medium); //convert height to rem
+    }
+
+    .c-formField-field--large {
+        @include form-field-size($form-input-height-large, $form-input-padding-large); //convert height to rem
     }
 
     .c-formField-field {
         width: 100%;
+        box-sizing: border-box;
```
ashleynolan: Does the element not get this as default? (it should come through as part of the reset stylesheet I think, unless it gets overidden)

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -282,17 +291,31 @@ export default {
...
 $form-input-borderColour--focus           : $color-grey-50;
 $form-input-borderColour--invalid         : $color-support-error;
 $form-input-borderColour--disabled        : $color-disabled-01;
-$form-input-height                        : 46px; // height is 46px + 1px border = 48px
-$form-input-padding                       : spacing(x1.5) spacing(x2);
+
+$form-input-height-small                  : spacing(x5);
+$form-input-padding-small                 : spacing(x1) spacing(x2);
+
+$form-input-height-medium                 : spacing(x6);
```
ashleynolan: One thing worth checking with the height in the browser is if the height is 48px with the border included (and the same for small/large).

In the designs, the 48px includes the border inside that height, but the CSS box-model includes the border outside of the height, so you may end up with a form-field that's 50px height (if that makes sense).

We had a similar thing with buttons I think – would just mean you need to do factor in the border width when setting the height (so this could then be `spacing(x6) - ($form-input-borderWidth * 2)` for instance.

That would be why `$form-input-height` is set to `46px` in the old code though (see the comment `// height is 46px + 1px border = 48px`), so worth double checking in devtools and just updating in each size.

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -134,7 +137,13 @@ export default {
         labelStyle: {
             type: String,
             default: 'default',
-            validator: value => (VALID_LABEL_STYLES.indexOf(value) !== -1) // The prop value must match one of the valid input types
+            validator: value => (VALID_LABEL_STYLES.indexOf(value) !== -1) // The prop value must match one of the valid label types
+        },
+
+        fieldSize: {
```
ashleynolan, line 142: Would be good to add some tests for this (can be a separate PR though).

ashleynolan (changes requested):

packages/components/atoms/f-form-field/src/components/FormField.vue
FayeCarter: I will add for this PR, thanks!

packages/components/atoms/f-form-field/src/components/FormField.vue
FayeCarter: I'll change it to a prop. Thanks!

xander-marjoram:
While you're here doing PIE updates... How does this handle long text for the dropdown component? I had a chat with Marty from design the other day and he said it should be cut off before going under the caret, with something like `text-overflow: ellipsis;`

simonsandersje (dismissed):

kevinrodrigues (dismissed):

FayeCarter:
> While you're here doing PIE updates... How does this handle long text for the dropdown component? I had a chat with Marty from design the other day and he said it should be cut off before going under the caret, with something like `text-overflow: ellipsis;`

@xander-marjoram I've got a few PRs to do for this so I'll add it in in the next one

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -336,8 +369,7 @@ $form-input-focus--boxShadow              : 0 0 0 2px $form-input-focus;
             border-color: $form-input-borderColour--invalid;
         }
 
-        // Disabled state
-        &[disabled] {
+        .c-formField--disabled & {
```
ashleynolan: You should still be able to use the `disabled` attribute here (rather than using a class), as it was only on the label it wasn't valid HTML (and the disabled attribute should still be on the input?)

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -4,7 +4,8 @@
         :class="[
             $style['c-formField'], {
                 [$style['c-formField--invalid']]: hasError,
-                [$style['c-formField--grouped']]: isFieldGrouped
+                [$style['c-formField--grouped']]: isFieldGrouped,
+                [$style['c-formField--disabled']]: isDisabled
```
ashleynolan: Do you need this class? Not sure you'll need it when you change the CSS back to use `[disabled]` for the form input?

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
```diff
@@ -7,7 +7,10 @@
             :data-test-id="testId.icon" />
         <select
             :id="$attrs.id"
-            :class="$style['c-formDropdown-select']"
+            :class="[
+                $style['c-formDropdown-select'],
+                { [$style['c-formDropdown--disabled']]: isDisabled }
```
ashleynolan: Rather than a class on the dropdown, I'd apply it as a disabled attribute (as that is valid).

So `:disabled="isDisabled"` should work, and then just update the CSS to use the disabled attribute.

You likely won't need the prop either, as the `attrs` are passed through already, so you can just make a computed like you do on the `formField` component that looks at the `$attrs.disabled` value

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
```diff
@@ -100,4 +107,15 @@ export default {
         background-color: $form-input-bg--hover;
     }
 }
+
+.c-formDropdown--disabled {
```
ashleynolan: Should be able to change this to use `[disabled]` once the above is updated.

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
```diff
@@ -1,6 +1,7 @@
 <template>
     <div
         :class="$style['c-formDropdown']"
+        :disabled="attributes.disabled"
```
ashleynolan: Probably worth having a hangout about this one as think getting wires crossed and will be easier to go through it that way. Let me know when you're free.

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
```diff
@@ -1,13 +1,21 @@
 <template>
     <div
-        :class="$style['c-formDropdown']"
+        :class="[
+            $style['c-formDropdown'],
+            { [$style['c-formField--invalid']]: hasError }
```
ashleynolan: Looks like the styling is being applied to the `c-formField-field` so you could move this class down onto the `select` element and then change the selector so that it just targets the `c-formField--invalid` class?

ashleynolan (commented):
One small comment – then LGTM

ashleynolan (approved):

packages/components/atoms/f-form-field/src/components/_tests/FormField.test.js
```diff
@@ -254,6 +283,24 @@ describe('FormField', () => {
...
+                // Arrange
+                const propsData = {
+                    disabled
+                };
+
+                // Act
+                const wrapper = shallowMount(FormField, { propsData });
+
+                // Assert
+                expect(wrapper.vm.isDisabled).toEqual(expected);
```
matthewhardern, line 301: I am not a huge fan of testing internal implementation details, this leads to brittle tests. In fact its recommended not to do this in the vue test utils guide. https://vue-test-utils.vuejs.org/guides/#getting-started

matthewhardern: We should only be testing the public interface in my opinion.

matthewhardern: Apart from this it looks good, I will tick, but think we should double tink about testing things like this

matthewhardern (approved):

packages/components/atoms/f-form-field/src/components/_tests/FormField.test.js
FayeCarter: @matthewhardern I'm doing a few more tickets on form field, I'll go over the tests within the next tickets?

PR #1122, opened 2021-07-16
f-statistics@v0.1.0 - Initial MVP

tommcclean:
This PR is an initial MVP - we are using Just Log for now as a way of transporting logs; but this likely won't be the case following the MVP.

I renamed `f-stat-client` to `f-statistics` and rewired it from ElasticSearch to consuming an existing instance of Just log to transport the stat with proprietary fields. It accepts an existing instance because if it creates a new one in a site that already exists; logs are duplicated.

The purpose of this module is to be used by stat collectors like `f-http` and the RUM stats generator; so that they can publish statistics in a uniform way to the same place with the same key properties that can identify them.


packages/services/f-statistics/src/tests/index.test.js
```diff
@@ -0,0 +1,13 @@
+import StatisticsClient from '../index';
+
+describe('f-statistics', () => {
+    it('should be defined', async () => {
```
kevinrodrigues: Do you need async here? Same as below.

tommcclean: No; correcting now ta

kevinrodrigues (approved):

packages/services/f-statistics/src/defaultOptions.js
```diff
@@ -0,0 +1,6 @@
+const defaultConfig = {
```
zlatinivanov-je, line 1: Would it be better to throw an error instead of providing defaults?

packages/services/f-statistics/src/index.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable camelcase */
+import defaultOptions from './defaultOptions';
```
zlatinivanov-je, line 2: the export name is `defaultConfig`

packages/services/f-statistics/src/index.js
```diff
@@ -0,0 +1,23 @@
...
+
+export default class StatisticsClient {
+    constructor (justLogInstance, options = {}) {
```
zlatinivanov-je, line 5: Just a niggly bit `justLogInstance` could be named more generic...

zlatinivanov-je (commented):
Just a couple of small comments... Happy to approve.

siggerzz (approved):

packages/services/f-statistics/src/defaultOptions.js
tommcclean: Would rather provide defaults in this case because I would rather capture some data without the perfect configuration, than nothing at all; making it error would be problematic as it would error during interception which could cause the XHR to fail - this is more of a nice to have at this stage so shouldn't cause customer impact.

It is likely we will add more configuration options in later too - the defaults should all function correctly; but providing overrides would be use-case specific.

packages/services/f-statistics/src/index.js
tommcclean: At the moment, it has to be a provided just log instance; we might in future make it more variable; but at the moment its just a dumb wrapper that only works with that provider.

packages/services/f-statistics/src/index.js
tommcclean: Ill pick this one up on the next bump - won't cause an impact; but could be renamed to follow convention

zlatinivanov-je (approved):

PR #1130, opened 2021-07-20
f-braze-adapter@v4.0.0-beta.3

matthewhardern:
Changing braze adapter to no longer be a dynamic import as it prevents it being used inside another component within this library, instead the `@braze/sdk` is listed as a peer dependency and dev dependency. I am also using vite to bundle the package as it also now needs to be bundled to be used within the offers component within this library. Vite was chosen as it was being used by `f-services`. Also changed is the braze SDK version I have now upgraded this to the latest version. 

### DIFF
--- a/packages/services/f-braze-adapter/package.json
+++ b/packages/services/f-braze-adapter/package.json
@@ -1,10 +1,10 @@
 {
   "name": "@justeat/f-braze-adapter",
   "description": "Fozzie Metadata Service",
-  "version": "4.0.0-beta.2",
-  "main": "src/index.js",
+  "version": "4.0.0-beta.3",
+  "main": "dist/f-braze-adapter.umd.js",
   "files": [
-    "src"
+    "dist"
   ],
   "homepage": "https://github.com/justeat/fozzie-components/tree/master/packages/services/f-braze-adapter",
   "contributors": [
@@ -25,7 +25,8 @@
     "fozzie"
   ],
   "scripts": {
-    "prepublishOnly": "yarn lint:fix && yarn test",
+    "prepublishOnly": "yarn lint:fix && yarn test && yarn build",
+    "build": "vite build",
     "lint": "vue-cli-service lint --no-fix",
     "lint:fix": "yarn lint --fix",
     "test": "vue-cli-service test:unit",
@@ -35,7 +36,6 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "appboy-web-sdk": "2.5.2",
     "date-fns": "2.16.1",
     "lodash.findindex": "4.6.0",
     "lodash.orderby": "4.6.0",
@@ -43,15 +43,18 @@
     "lodash.uniq": "4.5.0"
   },
   "devDependencies": {
+    "@braze/web-sdk": "3.3.0",
     "@vue/cli-plugin-babel": "4.4.6",
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.4.6",
     "core-js": "3.6.5",
     "jest-extended": "0.11.5",
     "mockdate": "3.0.2",
-    "js-cookie": "2.2.1"
+    "js-cookie": "2.2.1",
+    "vite": "2.4.2"
   },
   "peerDependencies": {
-    "js-cookie": ">=2.2.1"
+    "js-cookie": ">=2.2.1",
+    "@braze/web-sdk": "3.3.0"
   }
 }
--- a/packages/services/f-braze-adapter/package.json
+++ b/packages/services/f-braze-adapter/package.json
@@ -3,6 +3,7 @@
   "description": "Fozzie Metadata Service",
   "version": "4.0.0-beta.3",
   "main": "dist/f-braze-adapter.umd.js",
+  "module": "dist/f-braze-adapter.es.js",
   "files": [
     "dist"
   ],
@@ -43,7 +44,7 @@
     "lodash.uniq": "4.5.0"
   },
   "devDependencies": {
-    "@braze/web-sdk": "^3.3.0",
+    "@braze/web-sdk": ">=3.3.0",
     "@vue/cli-plugin-babel": "4.4.6",
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.4.6",
@@ -55,6 +56,6 @@
   },
   "peerDependencies": {
     "js-cookie": ">=2.2.1",
-    "@braze/web-sdk": "^3.3.0"
+    "@braze/web-sdk": ">=3.3.0"
   }
 }
--- a/packages/services/f-braze-adapter/package.json
+++ b/packages/services/f-braze-adapter/package.json
@@ -44,7 +44,7 @@
     "lodash.uniq": "4.5.0"
   },
   "devDependencies": {
-    "@braze/web-sdk": ">=3.3.0",
+    "@braze/web-sdk": "^3.3.0",
     "@vue/cli-plugin-babel": "4.4.6",
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.4.6",
--- a/packages/services/f-braze-adapter/package.json
+++ b/packages/services/f-braze-adapter/package.json
@@ -43,7 +43,7 @@
     "lodash.uniq": "4.5.0"
   },
   "devDependencies": {
-    "@braze/web-sdk": "3.3.0",
+    "@braze/web-sdk": "^3.3.0",
     "@vue/cli-plugin-babel": "4.4.6",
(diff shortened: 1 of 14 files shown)


packages/services/f-braze-adapter/package.json
```diff
@@ -35,23 +36,25 @@
...
     "jest-extended": "0.11.5",
     "mockdate": "3.0.2",
-    "js-cookie": "2.2.1"
+    "js-cookie": "2.2.1",
+    "vite": "2.4.2"
   },
   "peerDependencies": {
-    "js-cookie": ">=2.2.1"
+    "js-cookie": ">=2.2.1",
+    "@braze/web-sdk": "3.3.0"
```
minrwhite: This should be a slightly more accepting version specifier, to allow consumers to use a later version if necessary
```suggestion
    "@braze/web-sdk": "^3.3.0"
```

packages/services/f-braze-adapter/src/services/_tests/BrazeDispatcher.node.test.js
```diff
@@ -22,8 +22,8 @@ describe('BrazeDispatcher › node', () => {
         expect.assertions(2);
 
         try {
-            const Dispatcher = new BrazeDispatcher(0);
-            await Dispatcher.configure({
+            // eslint-disable-next-line no-unused-vars
+            const Dispatcher = new BrazeDispatcher({
```
minrwhite, line 26: If you're expecting the throw to happen within the constructor, you can just return, as you're expecting the assertions above:
```suggestion
            return new BrazeDispatcher({
```

packages/services/f-braze-adapter/src/services/_tests/BrazeDispatcher.test.js
```diff
@@ -152,112 +160,114 @@ describe('instantiation', () => {
...
-
-                // Act
-                const promiseB = await brazeDispatcher.configure(enabledComponentParameters);
-
-                // Assert
-                expect(promiseA).toBe(promiseB);
-            });
-
-            it('should push an event to datalayer when userId set following initialisation', async () => {
+            // it.each([
```
minrwhite: Are these to be reinstated at some point? If not, I'd prefer if these were just deleted, rather than commenting out, and if so, please add a comment explaining what is happening with them.

packages/services/f-braze-adapter/src/services/_tests/BrazeDispatcher.test.js
```diff
@@ -152,112 +160,114 @@ describe('instantiation', () => {
...
-                await brazeDispatcher.configure(enabledComponentParameters);
-                appboy.requestContentCardsRefresh.mockReset();
-
-                // Act
-                await brazeDispatcher.configure(enabledComponentParameters);
-
-                // Assert
-                expect(appboy.requestContentCardsRefresh).not.toHaveBeenCalled();
-            });
+            // it('should not request a refresh if an update is already pending', async () => {
```
minrwhite: As above

packages/services/f-braze-adapter/src/services/_tests/BrazeDispatcher.node.test.js
matthewhardern: awesome thanks

packages/services/f-braze-adapter/src/services/_tests/BrazeDispatcher.test.js
matthewhardern: correct they can be removed

packages/services/f-braze-adapter/src/services/_tests/BrazeDispatcher.node.test.js
matthewhardern: actually eslint then complains about `Expected to return a value at the end of async arrow function` if I remove it so can't win either way

minrwhite (approved):

packages/services/f-braze-adapter/vite.config.js
```diff
@@ -0,0 +1,24 @@
+// vite.config.js
+const path = require('path');
+
+module.exports = {
+    build: {
+        target: 'es2015',
```
DamianMullins, line 6: Just a note for the future; we probably want to add this into `f-services` as well at some point.

packages/services/f-braze-adapter/package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "@justeat/f-braze-adapter",
   "description": "Fozzie Metadata Service",
-  "version": "4.0.0-beta.2",
-  "main": "src/index.js",
+  "version": "4.0.0-beta.3",
+  "main": "dist/f-braze-adapter.umd.js",
```
DamianMullins, line 5: You could also add a `"module": "dist/f-braze-adapter.es.js"` line to offer ESM to package consumers.

packages/services/f-braze-adapter/package.json
```diff
@@ -25,7 +25,8 @@
     "fozzie"
   ],
   "scripts": {
-    "prepublishOnly": "yarn lint:fix && yarn test",
+    "prepublishOnly": "yarn lint:fix && yarn test && yarn build",
+    "build": "vite build",
```
DamianMullins, line 30: 💘

packages/services/f-braze-adapter/package.json
```diff
@@ -35,23 +36,25 @@
...
     "jest-extended": "0.11.5",
     "mockdate": "3.0.2",
-    "js-cookie": "2.2.1"
+    "js-cookie": "2.2.1",
+    "vite": "2.4.2"
   },
   "peerDependencies": {
-    "js-cookie": ">=2.2.1"
+    "js-cookie": ">=2.2.1",
+    "@braze/web-sdk": "^3.3.0"
```
DamianMullins: You'll want to set a version range here, like the line above.

```suggestion
    "@braze/web-sdk": ">=3.3.0"
```

matthewhardern: thanks sorted

packages/services/f-braze-adapter/package.json
matthewhardern: good idea have added

packages/services/f-braze-adapter/package.json
```diff
@@ -35,23 +37,25 @@
...
   "dependencies": {
-    "appboy-web-sdk": "2.5.2",
     "date-fns": "2.16.1",
     "lodash.findindex": "4.6.0",
     "lodash.orderby": "4.6.0",
     "lodash.startswith": "4.2.1",
     "lodash.uniq": "4.5.0"
   },
   "devDependencies": {
+    "@braze/web-sdk": ">=3.3.0",
```
DamianMullins: Sorry, this one should stay as it was, with no range or caret — only the `peerDependencies` need that.

matthewhardern: my bad sorted it

packages/services/f-braze-adapter/vite.config.js
matthewhardern: Yes I will note it down

packages/services/f-braze-adapter/src/services/BrazeConsumerRegistry.js
```diff
@@ -52,9 +51,9 @@ class BrazeConsumerRegistry {
         this.consumers = [];
 
         // subscribe to event stream for braze cards and messages
-        dispatcherEventStream.subscribe(CONTENT_CARDS_EVENT_NAME, this.applyContentCardCallbacks);
-        dispatcherEventStream.subscribe(IN_APP_MESSAGE_EVENT_NAME, this.applyInAppMessageCallbacks);
-        dispatcherEventStream.subscribe(IN_APP_MESSAGE_EVENT_CLICKS_NAME, this.applyInAppMessageClickEventsCallbacks);
+        dispatcherEventStream.subscribe(CONTENT_CARDS_EVENT_NAME, cards => this.applyContentCardCallbacks(cards));
```
DamianMullins, line 54: Is this change functionally equivalent to what it was before?

matthewhardern: Well I thought it would make no difference if it was `this.applyContentCardCallbacks` vs `cards => this.applyContentCardCallbacks(cards)`, but it was not seeing `this` within the method and throwing `consumers of undefined` error. So changing it to this makes it work, I assume some scope difference, but tbh someone else maybe able to better explain why it happens.

DamianMullins: Arrow functions use the value of `this` in their lexical scope which could explain it? i.e. `this` references the class instance scope rather than the callback scope?

matthewhardern: that makes a lot of sense actually, thanks for the explanation.

DamianMullins (approved):

PR #1131, opened 2021-07-20
f-button@1.10.0 - f-button changes according to latest PIE designs

dandel10n:
### Added
- `trailing-icon` and `leading-icon` named slots for ability to add an icon before or after button text
- `hasIcon` prop for spacing and icon colours css classes to be added
- validation for `hasIcon` prop to be able to add only `leading`/`trailing`/false

### Changed
- Button paddings to be aligned with the PIE designs
- XSmall button font size to be aligned with the PIE designs

### Updated
- Fozzie version to v5.0.0-beta.10 which has the latest pie design token variables.

Primary:
Screenshot 2021-07-20 at 15 57 04
Screenshot 2021-07-20 at 15 57 13

Secondary:
Screenshot 2021-07-20 at 15 58 46
Screenshot 2021-07-20 at 15 58 52

Outline:
Screenshot 2021-07-20 at 15 59 36
Screenshot 2021-07-20 at 15 59 43

Ghost:
Screenshot 2021-07-20 at 16 00 38
Screenshot 2021-07-20 at 16 00 32


packages/components/atoms/f-button/CHANGELOG.md
```diff
@@ -4,6 +4,19 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v1.10.0
+------------------------------
+*July 20, 2021*
+
+### Added
+- `trailing-icon` and `leading-icon` named slots for ability to add an icon befor or after button text
```
ashleynolan:
```suggestion
- `trailing-icon` and `leading-icon` named slots for ability to add an icon before or after button text
```

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -14,19 +14,37 @@
...
-            :class="[
-                $style['o-btn-text'],
-                (isLoading ? $style['o-btn-text--hidden'] : '')
-            ]">
-            <slot />
+            :class="$style['o-button-content']">
+
+            <span
+                :class="[
+                    (hasLeadingIcon ? $style['o-btn-icon'] : ''),
```
ashleynolan: Would this work?

```suggestion
                    (hasLeadingIcon ? $style['o-btn-icon', 'o-btn-icon--leading'] : ''),
```

or if not, perhaps:

```suggestion
                    (hasLeadingIcon ? [$style['o-btn-icon'], $style['o-btn-icon--leading']] : ''),
```

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -14,19 +14,37 @@
...
+                :class="[
+                    $style['o-btn-text'],
+                    (isLoading ? $style['o-btn-text--hidden'] : '')
+                ]">
+                <slot />
+            </span>
+
+            <span
+                :class="[
+                    (hasTrailingIcon ? $style['o-btn-icon'] : ''),
```
ashleynolan: Same as above – guessing there is a syntax that will work here to apply both classes in one conditional?

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -145,7 +171,7 @@ $btn-default-borderRadius              : $border-radius;
 $btn-default-font-family               : $font-family-base;
 $btn-default-font-size                 : 'body-l';
 $btn-default-weight                    : $font-weight-bold;
-$btn-default-padding                   : 11px 1.5em 13px;
+$btn-default-padding                   : 12px 24px;
```
ashleynolan: In the past, we've need to have 1px higher on the bottom than the top to account for the way line-height is rendered. Is this still the case? (Would be worth comparing the before/after of the buttons to see the difference).

Good if we can get rid of this though – just wanted to check!

ashleynolan: Also, I'd use `spacing(x3)` for the left/right padding (same below on line 221 and then `spacing(x2)` on line 226 and `spacing()` on line 230)

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -192,14 +218,17 @@ $btn-disabled-bgColor                  : $color-disabled-01;
 $btn-disabled-textColor                : $color-content-disabled;
 
 $btn-sizeLarge-font-size               : 'heading-s';
-$btn-sizeLarge-padding                 : 13px 1.2em 15px;
+$btn-sizeLarge-padding                 : 14px 24px;
 $btn-sizeLarge-loading-color           : $color-content-interactive-primary;
 $btn-sizeLarge-loading-colorOpaque     : rgba($btn-sizeLarge-loading-color, $btn-default-loading-opacity);
+$btn-sizeLarge-iconHeight              : 23.5px;
```
ashleynolan: Might be worth checking with design comparing with a real icon design, but it looks like the icon size itself in the button designs for large icons is `18 x 18` which is inside a `24 x 24` bounding box (so padded out by 6px).

Medium is then 18x18 (with 24x24 bounding box), small is `15x15` (bounding box is 20px) and small is 12x12 (bounding box of 16px).

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -67,6 +85,14 @@ export default {
         isLoading: {
             type: Boolean,
             default: false
+        },
+        hasTrailingIcon: {
```
ashleynolan: I'd condense this into one prop `hasIcon` that can take values `false`, `trailing` or `leading`. That way you never get into a weird spot where people could set true for both of these props and you have to decide which classes to apply.

ashleynolan (dismissed):
Few comments – looks good though otherwise.

packages/components/atoms/f-button/src/components/Button.vue
dandel10n: I checked that with the equal spacing the button will be the same as on designs when with the current spacing it is not. I will change it to `spacing()` where applicable

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -14,19 +14,33 @@
...
+            <span
+                v-if="hasIcon && hasIcon === 'leading'"
+                :class="(hasIcon === 'leading' ? [$style['o-btn-icon'], $style['o-btn-icon--leading']] : '')">
+                <slot name='leading-icon' />
+            </span>
+
+            <span
+                :class="[
+                    $style['o-btn-text'],
+                    (isLoading ? $style['o-btn-text--hidden'] : '')
```
FayeCarter: Could the classes for these three spans be 
```
{ [$style['o-btn-icon'], $style['o-btn-icon--leading']] :   hasIcon === 'leading' }
{ [$style['o-btn-text--hidden']] : isLoading }
{ [$style['o-btn-icon'], $style['o-btn-icon--trailing']] : hasIcon === 'trailing' }
```

DamianMullins: I agree, I think they should be. Not sure where the ternary approach came in but we should be using the Vue syntax for conditional classes IMO.

dandel10n: It didn't allow me to use two classes in one list so I had to split them back in two separate checks

DamianMullins: Oh right, I didn't catch that there were two classes. In that case, they should be broken out into two separate lines I reckon. Also, the `hasIcon === 'leading'` & `hasIcon === 'trailing'` expressions should be moved to computed props to keep logic out of the template.

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -14,19 +14,36 @@
...
         <span
-            :class="[
-                $style['o-btn-text'],
-                (isLoading ? $style['o-btn-text--hidden'] : '')
-            ]">
-            <slot />
+            :class="$style['o-button-content']">
+
+            <span
+                v-if="hasIcon && hasIcon === 'leading'"
```
DamianMullins: `hasIcon === 'leading'` should be split out into a computed property to keep logic out of the template and also to reduce duplication.

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -14,19 +14,36 @@
...
+                :class="[$style['o-btn-text'], {
+                    [$style['o-btn-text--hidden']]: isLoading
+                }]">
+                <slot />
+            </span>
+
+            <span
+                v-if="hasIcon && hasIcon === 'trailing'"
+                :class="[{
+                    [$style['o-btn-icon']]: hasIcon === 'trailing',
```
DamianMullins: Same as the last comment, `hasIcon === 'trailing'` should be split out into a computed property.

packages/components/atoms/f-button/src/components/Button.vue
dandel10n: Great suggesting, thanks, now everything looks much tidier!

DamianMullins (approved):

FayeCarter (approved):

packages/components/atoms/f-button/src/components/_tests/Button.test.js
```diff
@@ -120,6 +120,36 @@ describe('Button', () => {
...
+
+            it.each([
+                'leading',
+                'trailing',
+                false
+            ])('should not throw an error when "hasIcon" is set to %s', hasIcon => {
+                // Arrange
+                const propsData = {
+                    hasIcon,
+                    isIcon: false
```
ashleynolan, line 132: Just out of interest – what currently happens if `isIcon` is set to true and hasIcon is set to `leading` or `trailing`?

Does an error get thrown, or does the component attempt to append an icon to the icon button?

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -135,6 +166,10 @@ export default {
             if (!VALID_BUTTON_SIZES.includes(this.buttonSize)) {
                 throw new TypeError(`buttonSize is set to "${this.buttonSize}", but it can only be one of the following buttonSizes: "${VALID_BUTTON_SIZES.join('", "')}"`);
             }
+
+            if (!VALID_BUTTON_ICON_POSITION.includes(this.hasIcon)) {
+                throw new TypeError(`hasIcon is set to "${this.hasIcon}", but it can only be one of the following buttonSizes: "${VALID_BUTTON_ICON_POSITION.join('", "')}"`);
```
ashleynolan, line 171: Super small thing – but should probably say:

"but it can only be one of the following **values**"

rather than `buttonSizes`

PR #1132, opened 2021-07-20
f-form-field@v1.19.0 - Add icon to valid fields

FayeCarter:
### Added
- Icon slots for valid input types.
- Tests to cover changes.

### TextFields

image

### Dropdown

image


packages/components/atoms/f-form-field/src/assets/scss/common.scss
```diff
@@ -22,31 +22,45 @@ $form-input-borderColour--invalid       : $color-support-error;
 
 $form-input-focus--boxShadow            : 0 0 0 2px $color-focus;
 
-$form-input-height-small                : spacing(x5);
-$form-input-padding-small               : spacing(x1) spacing(x2);
+$form-inputHeight--small                : spacing(x5);
```
ashleynolan: I'd probably keep the names the same as before, as matches the way the other variables are named above.

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
```diff
@@ -83,6 +89,22 @@ export default {
 </script>
 
 <style lang="scss" module>
+$form-dropdown-iconPosition--small              : 17px;
+$form-dropdown-iconPosition--medium             : 21px;
+$form-dropdown-iconPosition--large              : 25px;
+
+.c-formDropdown-padding-small {
+    @include icon-position($form-dropdown-iconPosition--small);
```
ashleynolan: Does this just control the arrow spacing from the bottom of the dropdown?

If so, I wouldn't tie this into the `icon-position` mixin as it's clearer just setting these as `bottom: 17px` etc.  I think it makes sense when you're switching between left and right, but not so much for this where it'll always be using `bottom`

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -85,6 +88,32 @@
                 :data-test-id="`${testId.label}--inline`">
                 {{ labelText }}
             </form-label>
+
+            <span
+                v-if="hasLeftIcon"
```
ashleynolan: I would keep the naming to match the designs as much as possible.

So I'd change `left` > `leading` and `right` > `trailing` so that these would then become `hasLeadingIcon` and `hasTrailingIcon` (and the slots would then change to be `icon-leading` and `icon-trailing`) and the classnames the same as well.

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -260,6 +290,18 @@ export default {
 
         isDisabled () {
             return this.$attrs.disabled === 'disabled';
+        },
+
+        isValidIconField () {
+            return VALID_ICON_INPUT_TYPES.includes(this.inputType);
```
ashleynolan, line 297: Would be good to throw an error if someone is trying to use an icon on an invalid field type.

I've done something similar [when validating props](https://github.com/justeat/fozzie-components/blob/master/packages/components/atoms/f-button/src/components/Button.vue#L130-L133) – this is sightly different, but I wonder if you could set a watcher on this computed prop and if it's invalid, throw a type error.

Would mean you can not really worry about checking for `isValidIconField` then below as it'll have thrown an error and won't render at all until resolved.

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -260,6 +290,18 @@ export default {
...
+        },
+
+        hasLeftIcon () {
+            return this.$slots['icon-left'] && this.isValidIconField;
+        },
+
+        hasRightIcon () {
+            return this.$slots['icon-right'] && this.isValidIconField && !this.isDropdown;
```
ashleynolan: Same here – would be good to throw an error if someone tries to use a right icon with a dropdown – otherwise this handles the error but doesn't tell the consuming application why it's not rendering which can be a bit confusing.

ashleynolan (changes requested):
Few comments from me that would be good to add in/update

packages/components/atoms/f-form-field/src/assets/scss/common.scss
FayeCarter: 🤦‍♀️ I changed it back and forth about 8 times cos I couldn't decide which one to go with 😂

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
```diff
@@ -96,7 +107,21 @@ export default {
     }
 }
 
+.c-formDropdown-iconPosition-small {
```
matthewhardern: should this not be `--small`

packages/components/atoms/f-form-field/src/assets/scss/common.scss
```diff
@@ -22,31 +22,45 @@ $form-input-borderColour--invalid       : $color-support-error;
...
 
 @mixin form-field-size($height, $padding) {
     @include rem(height, $height);
     padding: $padding;
 }
 
+@mixin icon-position($size, $position: 'bottom') {
+    svg {
+        @if $position == 'leading' {
+            left: $form-input-iconPosition;
```
ashleynolan: Just looking at how this gets it's values, it's a little bit confusing atm. It takes a param of `size`, but for leading and trailing, ignores the `size` param and just uses the outer scoped `$form-input-iconPosition` (which is the same as the `size` passed in?).

Also, it looks like in some instances (when bottom is set) this helps control the size of the icon, but when leading/trailing is set, then size isn't needed at all?

May be worth catching up on hangout, as I think you can likely simplify this a little. I think either the size param needs renaming (and then you can use it when trailing or leading position is set) or need to rejig it so that size becomes an optional parameter and just use the globally set variables (so don't bother passing them in for leading or trailing icons).

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -342,4 +408,48 @@ export default {
...
+    .c-formField-icon--small {
+        @include icon-position($form-input-iconPosition--small);
+    }
+
+    .c-formField-icon--medium {
+        @include icon-position($form-input-iconPosition--medium);
+    }
+
+    .c-formField-icon--large {
+        @include icon-position($form-input-iconPosition);
```
ashleynolan: I would be tempted to make the vertical and horizontal positions separate variables – as if one changes at some point in the future, we may want to change it independent of one another (so the left/right position may change, but the vertical large position may stay the same, or vice versa).

Something like `$form-input-icon-horizontalIndent` and `$form-input-icon-verticalIndent` would make sense I think?

Also, I'd make then setup the vertical variables so that the default is medium (as in the designs), so:

```scss
$form-input-icon-verticalIndent                  : 15px;
$form-input-icon-verticalIndent--small           : 11px;
$form-input-icon-verticalIndent--large           : 19px;

```

ashleynolan (changes requested):
Couple of small comments, otherwise looks good 👍

packages/components/atoms/f-form-field/src/assets/scss/common.scss
FayeCarter: That's way better, thanks Ash

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -342,4 +409,48 @@ export default {
...
+
+    .c-formField-icon--disabled {
+        svg {
+            path {
+                fill: $color-content-disabled;
+            }
+        }
+    }
+
+    .c-formField-icon--small {
```
ashleynolan, line 433: I think you could potentially get rid of the extra classnames for these sizes, and use the formfield size class like this:

```scss
    .c-formField-icon {
        .c-formField-field--small ~ & {
            @include indent-icon('bottom', $form-input-icon-verticalIndent--small);
        }
        .c-formField-field--medium ~ & {
            @include indent-icon('bottom', $form-input-icon-verticalIndent);
        }
        …
    }
```

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
```diff
@@ -2,16 +2,20 @@
     <div
         :data-test-id="testId.container">
         <caret-icon
-            :class="$style['c-formDropdown-icon']"
+            :class="[
+                $style['c-formDropdown-icon'],
+                $style[`c-formDropdown-iconIndent--${fieldSize}`]
```
ashleynolan: You could do the same as below for these size classes, but you'd need to move the `caret-icon` element to be after the select element (so not sure if that's possible, or if it'd break the styling already setup for that)

packages/components/atoms/f-form-field/src/components/FormField.vue
FayeCarter: This sadly doesn't work for the dropdown because it's no longer a sibling

dandel10n (approved):
Looks good to me, just don't forget to update README with the new props

FayeCarter:
> Looks good to me, just don't forget to update README with the new props

Thanks, I'll do a full update of the README when all the changes are done 😄

packages/components/atoms/f-form-field/src/components/FormField.vue
ashleynolan: You should be able to change this for this component, as the styles are duplicated inside the drop down component separately?

FayeCarter: I _think_ that might just be for the caret in the dropdown `c-formDropdown-icon`? This is for the leading/trailing icon

ashleynolan (approved):

PR #1141, opened 2021-07-22
#trivial - remove testing eventData

dandel10n:
### Removed
- Some eventData values which were added for testing specific case


DamianMullins (dismissed):

dandel10n:
Changes will be added to another PR

PR #1142, opened 2021-07-22
fozzie-components@3.50.0 - Add names for unnamed Circle CI steps

xander-marjoram:
### Added
- Names for unnamed Circle CI steps.

| Before | After |
| --- | --- |
| 20210722T174839 | 20210723T115414 |

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -270,7 +270,7 @@ jobs:
     steps: # a collection of executable commands
       - checkout # special step to check out source code to working directory
       - detect_root_change
-      - echo_root
+      # - echo_root
       - restore_cache:
           name: Restore Yarn Package Cache
           keys:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,117 +18,135 @@ commands:
   detect_root_change:
     description: Detects root file changes
     steps:
-      - run: |
-          set -e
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -18,117 +18,135 @@ commands:
...
+            echo 'export RUN_ALL=false' >> $BASH_ENV
+              echo 'Root files not changed'
+            fi
 
-  echo_root:
-    description: Detects root package.json changes
-    steps:
-      - run: |
-          echo $RUN_ALL
+  # echo_root:
```
xander-marjoram, line 47: This step doesn't seem to do anything yet but I've left it in (but commented out) so it can be implemented later.

DamianMullins (approved):

kevinrodrigues (approved):

.circleci/config.yml
ashleynolan: Will need to check with @siggerzz but this may be linked to the checks that run if a root change is found.

It equally might just echo out the variable being used (which is what it looks like), so although not crucial to the actual build, may be something we want for debug purposes?

ashleynolan:
🚀 ⭐ for this change – was starting to hurt my brain looking at the builds 🤣

PR #1144, opened 2021-07-23
f-button@1.10.1 vertical alignment fix for icon button

dandel10n:
### Removed
- Extra wrapper around button text to fix `isIcon` button vertical alignment.

### Removed
- Unused f-button and f-card component dev dependencies

### DIFF
--- a/packages/components/organisms/f-loyalty/CHANGELOG.md
+++ b/packages/components/organisms/f-loyalty/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+Latest (to be added to next release)
+------------------------------
+*July 23, 2021*
+
+### Removed
+- Unused f-button and f-card component dev dependencies
+
+
 v0.2.0
 ------------------------------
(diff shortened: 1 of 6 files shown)


packages/components/organisms/f-loyalty/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
 
 
+Latest (to be added to next release)
+------------------------------
+*July 23, 2021*
+
+### Removed
+- Unused f-button and f-card component dev dependencies
+
+
```
minrwhite, line 14: I think they are here for use in a later PR, can you confirm please, @matthewhardern ? If so, is there a problem with them being included here for the time being, @dandel10n ?

ashleynolan (approved):

packages/components/organisms/f-loyalty/CHANGELOG.md
dandel10n: At the moment we have an issue with component styles overwriting each other in storybook when different versions of one component are used in other components. To fix this we need to be sure that the latest version of the component is used everywhere. So until these components are actually used I decided to remove them from here to be sure that by the time they are actually used here they are on the latest version.

FayeCarter (approved):

PR #1147, opened 2021-07-26
f-analytics@0.4.0 - Add route to feature name map.

kevinrodrigues:
Maps the route to a feature name

# Screenshot:

Screenshot 2021-07-26 at 12 25 02

## Browsers Tested

- [x] Chrome (latest)


packages/services/f-analytics/src/mixins/analytics.mixin.vue
```diff
@@ -93,7 +97,7 @@ export default {
         prepareAnalytics () {
             const platformData = { ...this.platformData };
 
-            platformData.name = this.$route.name;
+            platformData.name = MAP_ROUTE_TO_FEATURE_NAME[this.$route.name];
```
dandel10n: Should we leave the option to add `this.$route.name` if the name is not in the `MAP_ROUTE_TO_FEATURE_NAME` object?

kevinrodrigues: Ah yeah that's a good shout. 👍

kevinrodrigues: Fixed.

FayeCarter (approved):

dandel10n (approved):

zlatinivanov-je (approved):

PR #1149, opened 2021-07-27
f-content-cards@v5.0.0-beta.1 - Adding in the new braze Adapter and making braze sdk peer dependency

matthewhardern:
Adding in the brazeAdapter v4.0.0-beta.3, making changes to the main ContentCards.js file to adjust to utilise the new braze adapter. Updated unit tests to account for the change in instantiation of the new braze adapter. Added @braze/sdk as a peer dependency and dev dependency. Also updated the vue.config.js to exclude the @braze/sdk from being bundled in final package (being included in dev dependency it requires this). 

### DIFF
--- a/packages/components/organisms/f-content-cards/package.json
+++ b/packages/components/organisms/f-content-cards/package.json
@@ -52,10 +52,10 @@
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.1.1",
     "vue": "2.x",
-    "@braze/web-sdk": "3.3.0"
+    "@braze/web-sdk": ">=3.3.0"
   },
   "devDependencies": {
-    "@braze/web-sdk": "3.3.0",
+    "@braze/web-sdk": "^3.3.0",
     "@justeat/f-braze-adapter": "4.0.0-beta.3",
     "@justeat/f-vue-icons": "1.2.0",
     "@justeat/f-wdio-utils": "0.2.0",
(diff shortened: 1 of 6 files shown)


packages/components/organisms/f-content-cards/package.json
```diff
@@ -51,17 +51,19 @@
   },
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.1.1",
-    "vue": "2.x"
+    "vue": "2.x",
+    "@braze/web-sdk": "3.3.0"
```
minrwhite: Let's make the `@braze/web-sdk` `peerDependencies` requirement a bit more permissive, to allow for minor version bumps:
```suggestion
    "@braze/web-sdk": "^3.3.0"
```

packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -208,41 +208,39 @@ export default {
          * @return {Promise<void>}
```
minrwhite: This is no longer correct, I think?
```suggestion
         * @return {void}
```
We can probably remove the `await` in the `mounted()` hook, as well.

packages/components/organisms/f-content-cards/package.json
matthewhardern: changed to `>=` as per the braze adapter way of doing it

minrwhite (approved):

PR #1150, opened 2021-07-27
f-feature-management@v0.2.0 - Implement Feature Management SDK

gareththackeray:
Implemented the logic for FM SDK and tested it against our x-platform tests.
How exactly we will load the configuration will likely differ between implementations.
For now we have bypassed the question of how that will work by requiring it to be passed in when creating the object.
There are plenty of questions outstanding around scopes, config polling, shared / common scopes, SSR vs client etc etc.
The idea in this commit is to get the working logic in; it is not yet ready for use.


packages/services/f-feature-management/package.json
```diff
@@ -35,11 +37,15 @@
...
     "jest": "25.4.0",
     "jest-teamcity": "1.7.0",
     "uglifyjs-webpack-plugin": "2.2.0",
     "webpack": "4.42.1",
     "webpack-cli": "3.3.11"
   },
-  "peerDependencies": {}
+  "peerDependencies": {},
+  "dependencies": {
+    "crypto-js": "^4.1.1"
```
ashleynolan: Can you remove the `^` – just helps with locking versions between machines:

```suggestion
    "crypto-js": "4.1.1"
```

packages/services/f-feature-management/src/calculateFractions.js
```diff
@@ -0,0 +1,81 @@
...
+function byteArrayToWordArray (ba) {
+    const wa = [];
+    for (let i = 0; i < ba.length; i++) {
+        wa[(i / 4) | 0] |= ba[i] << (24 - 8 * i);
+    }
+
+    return CryptoJS.lib.WordArray.create(wa, ba.length);
+}
+
+function wordToByteArray (word, length) {
```
ashleynolan, line 30: Can you add JSDoc comments to any new functions? ([for example](https://github.com/justeat/fozzie-components/blob/master/packages/services/f-http/src/createClient.js) – doesn't need to be in this commit, but added as part of this work).

Looks like you've got them in some places below, but not everywhere.

packages/services/f-feature-management/src/index.js
```diff
@@ -1,38 +1,107 @@
...
+import trackExperiment from './trackExperiment';
+import { logger, setLogger } from './logger';
+
 const logValueRequest = (func, key, value) => {
-    console.log(`${func.name}: ${key} => ${value}`); // es-lint-disable
+    logger.logInfo(`${func.name}: ${key} => ${value}`);
 };
 
+
+function getValue (key) {
```
ashleynolan: JSDoc comments here please

packages/services/f-feature-management/src/index.js
```diff
@@ -1,38 +1,107 @@
...
+    logValueRequest(getValue, key, null);
```
ashleynolan: I'd probably consider abstracting some of the additional functions out of the `index.js` into modules you import in – I'm a bit of a fan of keeping the main entry point as small as possible, just as I think it makes the base module more readable/understandable for anyone new to the code.

ashleynolan (commented):
Few comments just from a quick look. Will take another sweep over the JS when I get chance though too

ashleynolan:
@gareththackeray One other thing I noticed – can you remove the JIRA IDs from the commit titles? InfoSec don't like us using them on public repo's, because…reasons 🤷  😄

packages/services/f-feature-management/src/calculateFractions.js
gareththackeray: Done. I had previously only added to exported or interesting functions.

packages/services/f-feature-management/src/index.js
gareththackeray: Done

gareththackeray:
Replaced with https://github.com/justeat/fozzie-components/pull/1152

PR #1151, opened 2021-07-27
f-loyalty@v0.3.0 - adding header component and tests

matthewhardern:
Adding in the header component for f-loyalty. Unit tests added to support the new header, plus put back f-button and f-card removal as they were added in at the base to be available for future components as part of f-loyalty. If this is an issue comment below.

## Browsers Tested

- [x] Chrome (latest)

## UI - Screenshots

Screenshot 2021-07-28 at 14 47 29
Screenshot 2021-07-28 at 14 47 44
Screenshot 2021-07-28 at 14 47 55
Screenshot 2021-07-28 at 14 48 07


ashleynolan:
@matthewhardern Could you add a screenshout of the changes – just helps with visualising it (for me anyway 😄 )

packages/components/organisms/f-loyalty/src/components/Header.vue
```diff
@@ -0,0 +1,139 @@
...
+
+<script>
+import BreadCrumbs from '@justeat/f-breadcrumbs';
+import MediaElement from '@justeat/f-media-element';
+import '@justeat/f-media-element/dist/f-media-element.css';
+import '@justeat/f-breadcrumbs/dist/f-breadcrumbs.css';
+import debounce from 'lodash.debounce';
+import { FONT_SIZE } from '@justeat/f-media-element/src/constants';
+
+const REQUIRED_WINDOW_WIDTH = 600;
```
ashleynolan, line 29: May be worth a comment as to what this is used for, as it's not obvious without going through the related code.

packages/components/organisms/f-loyalty/src/components/Header.vue
```diff
@@ -0,0 +1,139 @@
...
+    </div>
+</template>
+
+<script>
+import BreadCrumbs from '@justeat/f-breadcrumbs';
+import MediaElement from '@justeat/f-media-element';
+import '@justeat/f-media-element/dist/f-media-element.css';
+import '@justeat/f-breadcrumbs/dist/f-breadcrumbs.css';
+import debounce from 'lodash.debounce';
+import { FONT_SIZE } from '@justeat/f-media-element/src/constants';
```
ashleynolan: This seems strange that you need to import in the font_size to then pass it back through to the mediaElement component above.

Doesn't need to be changed necessarily as part of this PR, but I would probably provide an abstraction so that `mediaElement` can take in a prop identifier (like `small/medium/large` for the fontSize), and then the media element just links that ID to the actual value as part of the component?

packages/components/organisms/f-loyalty/src/components/Loyalty.vue
```diff
@@ -68,13 +73,11 @@ export default {
...
+    flex-direction: column;
     justify-content: center;
-    min-height: 80vh;
-    width: 80vw;
+    width: 100%;
     margin: auto;
-    border: 1px solid $color-red;
-    font-family: $font-family-base;
-    @include font-size(heading-m);
+    background-color: $color-grey-10;
```
ashleynolan: Should use `$color-background-default` here.

packages/components/organisms/f-loyalty/src/components/Header.vue
```diff
@@ -0,0 +1,139 @@
...
+        handleResize () {
+            this.windowWidth = process.browser && window.innerWidth;
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+
+$responsive-mobileViewBreakpoint: '<=narrowMid';
```
ashleynolan: Not sure if you're planning on using this elsewhere in future PRa, but while it's only used in one place atm, seems like a premature abstraction.

ashleynolan (commented):
Looks good – couple of small comments

packages/components/organisms/f-loyalty/src/components/Header.vue
ashleynolan: Just had another look at this – it does look like this prop is a string already?

If so, I'd probably just pass through the string rather than import it from the component and then pass it back through. For me, it's expected that consumers of the component would need to know how to define the props (may be worth a second opinion, but I think just passing in `xl` is fine for this?)

matthewhardern: your correct its my mistake, just used to typing that constant. Its not correct as I am not exporting those constants as part of f-media-element package.

matthewhardern: I will make the change

packages/components/organisms/f-loyalty/src/components/Header.vue
matthewhardern: agreed

packages/components/organisms/f-loyalty/src/components/Header.vue
matthewhardern: Yes will note that its used for changing image based on window width.

matthewhardern:
> @matthewhardern Could you add a screenshout of the changes – just helps with visualising it (for me anyway 😄 )

Yep, have attached to description.

packages/components/organisms/f-loyalty/CHANGELOG.md
```diff
@@ -4,12 +4,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-Latest (to be added to next release)
+v0.3.0
 ------------------------------
-*July 23, 2021*
+*July 20, 2021*
```
ashleynolan: This date should likely be bumped

packages/components/organisms/f-loyalty/src/components/Header.vue
```diff
@@ -0,0 +1,137 @@
...
+            return this.windowWidth >= REQUIRED_WINDOW_WIDTH
+                ? `${TEMP_IMG_PATH}stampcards-header-${this.$t('percentage')}.svg`
+                : `${TEMP_IMG_PATH}stampcards-header-small.svg`;
+        },
+
+        /**
+         * Sets the media element font size based on the current screen size
+         * @returns {string}
+         */
+        getFontSizeBasedOnScreenSize () {
```
ashleynolan, line 74: Does this get used atm? Can't see where it's called atm

matthewhardern: good spot, I missed adding it to media element.

packages/components/organisms/f-loyalty/src/components/Header.vue
```diff
@@ -0,0 +1,137 @@
...
+    @include media('<mid') {
+        padding: #{$layout-margin--mid}px;
+    }
+
+    @include media('<narrow') {
+        padding: #{$layout-margin--narrow}px;
+    }
+}
+
+.c-loyaltyHeader {
```
ashleynolan, line 126: Super NIT, but as this comes first in the source order I'd put this above the styles for `.c-loyaltyHeader-container`

ashleynolan (approved):

packages/components/organisms/f-loyalty/src/tenants/en-GB.js
```diff
@@ -1,6 +1,13 @@
 const messages = {
     locale: 'en-GB',
-    text: 'I am a VLoyalty Component (GB)'
+    text: 'I am a VLoyalty Component (GB)',
```
DamianMullins, line 3: Will this be visible to customers?

matthewhardern: not atm, I will be adding things here as I build this component up, so until all the bits are built then it won't be seen by anyone

DamianMullins (approved):

packages/components/organisms/f-loyalty/src/components/Header.vue
matthewhardern: will add to next PR as you suggested

PR #1153, opened 2021-07-28
f-feature-management@0.2.0 - Implement Feature Management SDK

xander-marjoram:
Creating this PR for @gareththackeray in place of #1150 and #1152
> Implemented the logic for FM SDK and tested it against our x-platform tests.
> How exactly we will load the configuration will likely differ between implementations.
> For now we have bypassed the question of how that will work by requiring it to be passed in when creating the object.
> There are plenty of questions outstanding around scopes, config polling, shared / common scopes, SSR vs client etc etc.
> The idea in this commit is to get the working logic in; it is not yet ready for use.


packages/services/f-feature-management/src/FeatureManager.js
```diff
@@ -0,0 +1,123 @@
...
+     * Gets feature and tracks any experiments.  Checks it's a boolean before returning.
+     * @param {string} key
+     * @returns Boolean or null
+     */
+    getBooleanValue (key) {
+        logValueRequest(this.getBooleanValue, key, true);
+
+        const value = this.getValue(key);
+
+        if (value === true || value === false) {
```
xander-marjoram, line 82:
```suggestion
        if (typeof value === 'boolean') {
```

packages/services/f-feature-management/src/FeatureManager.js
```diff
@@ -0,0 +1,123 @@
...
+        return null;
+    }
+
+    /**
+     * Gets feature and tracks any experiments.  Checks it's a string before returning.
+     * @param {string} key
+     * @returns String or null
+     */
+    getStringValue (key) {
+        logValueRequest(this.getStringValue, key, '');
```
xander-marjoram, line 112: Are these calls to `logValueRequest` a bit premature? They all seem to be running before the value is fetched

packages/services/f-feature-management/src/evaluateDefaultValue.js
```diff
@@ -0,0 +1,20 @@
+import evaluateRestriction from './evaluateRestriction';
+
+/**
+ * Returns the first default value where the contex matches the value's restrictions.
```
xander-marjoram: Nit: `context`

packages/services/f-feature-management/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.2.0
+------------------------------
+*July 27, 2021*
```
xander-marjoram: Nit: Might as well update this date when you do other changes

packages/services/f-feature-management/src/FeatureManager.js
gareththackeray: Well, Jack added them in... I don't really know how useful they are.  Not sure why it makes a difference if they are before or after the meat though?

packages/services/f-feature-management/src/FeatureManager.js
gareththackeray: Someone on stack overflow reckoned it was quicker.  Obviously that will have no impact on anyone.  Happy to change but not sure it's an improvement?

packages/services/f-feature-management/src/evaluateRestriction.js
```diff
@@ -0,0 +1,220 @@
+import { logger } from './logger';
+
+/**
+ * Checks that the restriction object has at most one operator specified
+ * @param {object} restriction. Throws if not.
```
xander-marjoram: I think this "Throws if not" is in the wrong place

packages/services/f-feature-management/src/evaluateRestriction.js
```diff
@@ -0,0 +1,220 @@
...
+}
+
+/**
+ * Utility function to return true if predicate passes for every item in the array.
+ * @param {array} array
+ * @param {function} predicate - map from the array items
+ * @returns Boolean
+ */
+function all (array, predicate) {
+    const passes = array.filter(i => predicate(i));
```
xander-marjoram: There's already a built-in JS array method called [`every`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) that you could try using?

packages/services/f-feature-management/src/evaluateRestriction.js
```diff
@@ -0,0 +1,220 @@
...
+    return passes.length === array.length;
+}
+
+/**
+ * Utility function to return true if predicate passes for at least one item in the array.
+ * @param {array} array
+ * @param {function} predicate - map from the array items
+ * @returns Boolean
+ */
+function any (array, predicate) {
```
xander-marjoram: There's also the array method [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

gareththackeray: TIL

packages/services/f-feature-management/src/evaluateRestriction.js
gareththackeray: TIL

packages/services/f-feature-management/src/evaluateRestriction.js
```diff
@@ -0,0 +1,220 @@
...
+}
+
+/**
+ * Compares two version numbers of format x0.x1...xn, where xi are positive integers.
+ * 1.2.0 > 1.2.
+ * @param {string} valueA
+ * @param {string} valueB
+ * @returns 1 if A is larger, 0 if equal, -1 if B is larger
+ */
+function compareVersionNumbers (valueA, valueB) {
```
xander-marjoram, line 60: Are there unit tests for this method?

xander-marjoram: (And others in this file)

packages/services/f-feature-management/src/index.js
```diff
@@ -1,38 +1,10 @@
...
-    };
+import createFeatureManagementInstance from './createFeatureManagementInstance';
+
+/**
+ * Returns an instance of Feature Management.
+ * @param {object} settings - json and contextGetter must be set
+ * @returns Feature Management object to allow querying of features.
+ */
+export default function (settings) {
+    return createFeatureManagementInstance(settings);
```
xander-marjoram, line 9: Is it up to the consuming application how often this is updated? Assuming it'll end up being once per request rather than keeping it the same between users?

packages/services/f-feature-management/src/logger.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Default logger implementation.  The interface matches that used by various Fozzie components.
+ */
+const logger = {
+    logError (logMessage, store, logPayload = {}) {
+        if (console.error) {
```
xander-marjoram: You might want to call this file `clientLogger` or something, and maybe checking that `console` also exists so it doesn't error if you try and access a property when console is undefined.

xander-marjoram: I see the analytics below is also TODO so I'm assuming this is the same

packages/services/f-feature-management/src/evaluateRestriction.js
gareththackeray: Updated

packages/services/f-feature-management/src/evaluateRestriction.js
gareththackeray: No. It is somewhat tested by the e2e tests, but no specific unit tests.  If I get time I'll add some in for this but otherwise we can ticket it up?

packages/services/f-feature-management/src/index.js
gareththackeray: What do you mean by "updated"?

packages/services/f-feature-management/src/FeatureManager.js
xander-marjoram: I'm only asking because looking at the definition for `getValueRequest`, the third parameter is the value so I wondered why we're submitting the default values for all of the types rather than the value that is actually returned. If it doesn't matter then all cool.

packages/services/f-feature-management/src/evaluateRestriction.js
gareththackeray: Added

packages/services/f-feature-management/src/evaluateIdHashLogic.js
```diff
@@ -0,0 +1,92 @@
...
+
+    // 2. if no matching rule, return null - calling code will revert to defaults
+    if (!matchingRule) {
+        return null;
+    }
+
+    // 3. compute the audienceFraction and variantFraction, based on the relevant id and prefix for the experiment
+    const { audienceFraction, variantFraction } = calculateFractions(idHashLogicDetail.prefix + getIdToHash(idHashLogicDetail, context));
+
+    // 4. if the user is not caught in traffic allocated to the experiment, return null - calling code will rever to defaults
```
xander-marjoram: Nit: revert

packages/services/f-feature-management/src/index.js
xander-marjoram: I guess calling `initFeatures`, if the features are updated

packages/services/f-feature-management/test/endToEnd.test.js
```diff
@@ -1,21 +1,51 @@
...
+    logWarn(){},
+    logInfo(){}
+}
+
+const fm = featureManagement({
+    json: JSON.stringify(config),
+    logger: logger
+});
+
+const functionMap = { bool: fm.getBooleanValue.bind(fm), int: fm.getIntegerValue.bind(fm), string: fm.getStringValue.bind(fm) };
```
xander-marjoram, line 19: As you've done it, the answer is probably "yes", but is the `bind`ing necessary here?

packages/services/f-feature-management/test/endToEnd.test.js
```diff
@@ -1,21 +1,51 @@
...
             const featureValue = functionMap[valueType](feature);
-            expect(featureValue.value).toBe(expectedValue);
+            expect(featureValue).toBe(expectedValue);
+
+            
             if (expectedKey && expectedVariant) {
-                expect(feature.experimentKey).toBe(expectedKey);
-                expect(feature.experimentVariant).toBe(expectedVariant);
+
+                expect(trackExperiment).toHaveBeenCalled()
```
xander-marjoram: Nit: missing semicolon, might be worth re-running the linter to check if there's anything else

packages/services/f-feature-management/test/endToEnd.test.js
gareththackeray: It's because the functions get pulled from their context (which is methods on a class) into this "map".  So when you run it `this` is undefined.  It might be that a) there is a better way of enforcing that they are bound somehow or b) we should avoid classes and this's.  But I think we can work through that as we start to iterate on it and integrate it with consumers.

packages/services/f-feature-management/test/endToEnd.test.js
gareththackeray: Strange I very much ran the linter, as does the commit hook.  Maybe it doesn't care about test files?

packages/services/f-feature-management/src/index.js
gareththackeray: That's something we'll need to think about in due course.  I expect we'll have a poller at some point.  But just sidestepping all those questions for this PR.

packages/services/f-feature-management/src/createFeatureManagementInstance.js
```diff
@@ -0,0 +1,25 @@
+import { init as initFeatures } from './featureGetter';
+import { setContextGetter } from './contextGetter';
+import { setLogger } from './logger';
+import FeatureManager from './FeatureManager';
+
+const instance = null;
```
tommcclean: I can't see where this is ever set

packages/services/f-feature-management/src/contextGetter.js
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * @returns Context required to evaluate features at current time current request, e.g. country.
+ */
+function getContext () {
+    if (contextGetter) {
+        return contextGetter();
+    }
+
+    return null;
```
tommcclean: Meh
```suggestion
    return (contextGetter)
      ? contextGetter()
      : null;
```

packages/services/f-feature-management/src/logger.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Default logger implementation.  The interface matches that used by various Fozzie components.
+ */
+const logger = {
+    logError (logMessage, store, logPayload = {}) {
```
tommcclean, line 5: ℹ️ - Just a reminder when you productionise this bit; pass a property to the payload (or add in these methods)
```suggestion
    logError (logMessage, store, logPayload = {}) {
      logPayload = {
        ...logPayload,
        tags: 'featureManagement'
      };
```

packages/services/f-feature-management/src/FeatureManager.js
```diff
@@ -0,0 +1,123 @@
...
+            return null;
+        }
+    }
+
+    /**
+     * Gets feature and tracks any experiments.  Checks it's a boolean before returning.
+     * @param {string} key
+     * @returns Boolean or null
+     */
+    getBooleanValue (key) {
```
tommcclean, line 77: Is the developer going to know what type to retrieve?

packages/services/f-feature-management/src/createFeatureManagementInstance.js
```diff
@@ -0,0 +1,25 @@
...
+ * @returns Singleton FeatureManagement object
+ */
+export default function (settings) {
+    if (instance) return instance;
+
+    if (settings.logger) {
+        setLogger(settings.logger);
+    }
+
+    initFeatures(settings.json);
```
tommcclean, line 20: What produces the JSON here - I can see it gets passed in, does the client application create it?

packages/services/f-feature-management/src/createFeatureManagementInstance.js
gareththackeray: Oops good spot

packages/services/f-feature-management/src/createFeatureManagementInstance.js
gareththackeray: We're basically sidestepping that decision for this initial PR.  But yes if it were to be used as it currently is, the application would pass it in.  For the potential SearchWeb prototype integration this would actually work fine because the .net server would have loaded it and could write it into the page.

packages/services/f-feature-management/src/FeatureManager.js
gareththackeray: We shall see!  I actually think they'll normally just call `getValue()`.

packages/services/f-feature-management/src/logger.js
gareththackeray: Shall I just commit your suggestion then?

packages/services/f-feature-management/src/FeatureManager.js
gareththackeray: If you have an opinion the typed ones being counterproductive please yell and we could look at removing them.

packages/services/f-feature-management/src/logger.js
tommcclean: Nah just keep it in mind for now; that was all

packages/services/f-feature-management/src/FeatureManager.js
tommcclean: Its a bit hard to say at the moment as I can't picture the whole thing in my head very clearly yet; I guess if there is a value in having the specific typings then its cool. Just trying to think as a consumer would I actually now which one to call

packages/services/f-feature-management/src/createFeatureManagementInstance.js
tommcclean: Okay cool makes sense; yeah doesn't seem wise to stay that way beyond prototype

packages/services/f-feature-management/test/endToEnd.test.js
ashleynolan: It should lint the test files too (just checking it locally, it runs on auto-save in VSCode for me or the test files in this service).

ashleynolan (commented):
Looks good – happy to approve this once the other comments have been addressed. Can't see anything extra from my side.

packages/services/f-feature-management/src/FeatureManager.js
gareththackeray: I'm not sure.  It's not a very javascripty way to think.  @jnewc put them in in the first place so may have an opinion.  In any case no consumers yet so we can change before it gets to primetime.

gareththackeray: We actually discussed it at the time https://github.com/justeat/fozzie-components/pull/1097#discussion_r664699137

tommcclean (dismissed):

ashleynolan (approved):

tommcclean (approved):

PR #1154, opened 2021-07-28
f-form-field@v1.21.0 - Create affixed field

FayeCarter:
### Added
- `AffixedFormField` component.
- Tests to cover changes.
 
Screenshot 2021-07-29 at 09 56 09


packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -364,6 +397,28 @@ export default {
...
+                throw new TypeError('Form field is set to have a "suffix" and "trailingIcon", only one can be displayed');
+            }
+
+            if (this.prefix.length > 3) {
+                throw new TypeError(`Form field is set to have a "prefix" of ${this.prefix.length} characters long. Prefix has a max length of 3 characters`);
+            }
+
+            if (this.suffix.length > 3) {
+                throw new TypeError(`Form field is set to have a "suffix" of ${this.suffix.length} characters long. Suffix has a max length of 3 characters`);
+            }
```
ashleynolan: As these are validating against a single prop value (rather than checking against two values) you can add this as a prop validator where the prop is initialised ([see this](https://vuejs.org/v2/guide/components-props.html#Prop-Validation))

packages/components/atoms/f-form-field/src/components/FormLabel.vue
```diff
@@ -9,7 +9,9 @@
                 [$style['c-formField-label--inline']]: isInline,
                 [$style['c-formField-label--disabled']]: isDisabled
             }
-        ]">
+        ]"
+        @mouseenter="toggleHoverState"
+        @mouseleave="toggleHoverState">
```
ashleynolan: Can delete this – must have missed this when deleting the hover class stuff.

packages/components/atoms/f-form-field/src/components/FormLabel.vue
```diff
@@ -58,6 +60,12 @@ export default {
         hasLabelText () {
             return this.$slots.default && !!this.$slots.default[0].text.length;
         }
+    },
+
+    methods: {
+        toggleHoverState () {
+            this.$emit('change-hover-state');
```
ashleynolan: Same here – can delete this method now

ashleynolan (commented):
Looks good – couple of small comments but then LGTM

packages/components/atoms/f-form-field/src/components/FormLabel.vue
FayeCarter: think it was me 🤦

ashleynolan (dismissed):

packages/components/atoms/f-form-field/src/components/_tests/FormField.test.js
```diff
@@ -389,6 +394,94 @@ describe('FormField', () => {
...
+                    // Act
+                    wrapper = mount(FormField, { propsData });
+                });
+
+                it('should return false`', () => {
+                    // Assert
+                    expect(wrapper.vm.isAffixedField).toBe(false);
+                });
+
+                it('wrapper should display `affixedFormField`', () => {
```
kevinrodrigues: Minor but could you change this to read `should display 'affixedFormField' wrapper` please.

packages/components/atoms/f-form-field/src/components/FormFieldAffixed.vue
```diff
@@ -0,0 +1,189 @@
...
+            :id="$attrs.id"
+            ref="input"
+            :class="$style['c-formField-field']"
+            :disabled="attributes.disabled"
+            v-bind="attributes"
+            :value="value"
+            :data-test-id="testId.input"
+            @change="updateInput">
+
+        <div
```
kevinrodrigues: Should these be buttons elements?

packages/components/atoms/f-form-field/src/components/FormFieldAffixed.vue
```diff
@@ -0,0 +1,189 @@
...
+            :class="[
+                $style['c-formField-affix'],
+                $style['c-formField-prefix']
+            ]"
+            :data-test-id="testId.prefix"
+            @click="addFocus">
+            {{ prefix }}
+        </div>
+
+        <div
```
kevinrodrigues: Same as above?

packages/components/atoms/f-form-field/src/components/FormFieldAffixed.vue
```diff
@@ -0,0 +1,189 @@
...
+
+    .c-formField-prefix {
+        order: 1;
+    }
+    .c-formField-suffix {
+        order: 3;
+    }
+
+
+.c-formField--affixed--medium {
```
matthewhardern: Could this be reduced to something like this?
```
$sizes: (small:  $form-input-height--small, medium: $form-input-height, large: $form-input-height--large);

@each $value, $size in $sizes {
    .c-formField--affixed--#{$size} {
        height: $value

        .c-formField-affix {
            padding: $value;
        }
    }
}
```

packages/components/atoms/f-form-field/src/assets/scss/common.scss
```diff
@@ -23,14 +23,16 @@ $form-input-borderColour--invalid       : $color-support-error;
 
 $form-input-focus--boxShadow            : 0 0 0 2px $color-focus;
 
+$form-input-padding--side              : spacing(x2);
+
 $form-input-height                      : spacing(x6);
-$form-input-padding                     : spacing(x1.5) spacing(x2);
+$form-input-padding                     : 11px $form-input-padding--side 13px;
```
matthewhardern, line 29: I am not a fan at all of have random px values in our code, seems like these spacings should come from PIE?

packages/components/atoms/f-form-field/src/components/FormFieldAffixed.vue
ashleynolan: These aren't interactive – just visual prefix/suffix elements that display on top of the form field

packages/components/atoms/f-form-field/src/assets/scss/common.scss
ashleynolan: Yeh, agree these look more random – the way centring works within the form-field to get things looking centred, the vertical padding (which is usually 12px top and bottom) actually needs to be 11px and 13px so that it actually lines up with the icon properly.

This could definitely come from PIE eventually, but at the moment these values don't exist in design tokens to do this.

matthewhardern: Ok makes sense, just note to come back and look at this in future

packages/components/atoms/f-form-field/src/components/FormFieldAffixed.vue
matthewhardern: would they not be better as a `span` ?

matthewhardern (approved):

ashleynolan (approved):

PR #1155, opened 2021-07-29
f-form-field@v2.0.0 - Update props and documentation

FayeCarter:
### Remove
- **Breaking Change**: `labelStyle` prop.
- **Breaking Change**: `Inline` and `InlineNarrow` styles.

### Added
- `shouldShowLabelText` prop.
- Documentation.

### Changed
- **Breaking Change**: Replaced `description` slot and `hasInputDescription` prop with `labelDescription` prop.


packages/components/atoms/f-form-field/CHANGELOG.md
```diff
@@ -4,6 +4,21 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+*August 5, 2021*
+
+### Remove
+- `labelStyle` prop.
+- `Inline` and `InlineNarrow` styles.
+- `hasInputDescription` prop.
+
+### Added
+- `shouldShowLabelText` prop.
+- `labelDescription` prop.
```
ashleynolan: I would put this under a `Changed` section and put that you've removed the description slot in favour of a `labelDescription` prop.

Also, anything that is considered "Breaking" between versions (like this) is good to add *Breaking* to the front of, so that anyone using the component and upgrading can see what they might need to watch out for. So here I would add that before this label change and the InlineNarrow and labelStyle changes above.

packages/components/atoms/f-form-field/README.md
```diff
@@ -72,15 +72,29 @@ The props that can be defined are as follows (if any):
...
 | `dropdownOptions` | `Array` | `null` | The options to be displayed in the dropdown. |
 | `isGrouped` | `Boolean` | `false` | When `true` will remove margin between all grouped form fields. |
 | `minNumber` | `Number` or `undefined` | `undefined` | Sets the value of the `min` property when `inputType` is `number` |
 | `maxNumber` | `Number` or `undefined` | `undefined` | Sets the value of the `max` property when `inputType` is `number` |
-| `hasInputDescription` | `Boolean` | `false` | When `true` will add the ability to insert extra html element under form label to the component via slot. |
+| `labelDescription` | `String` | `'` | Extra text that will display underneath the label text. |
+| `labelDetails` | `String` | `'` | Details for the label that will display on the right of label text. |
+| `assistiveText` | `String` | `'` | Text to assist the user that will display underneath the input field. |
+| `prefix` | `String` | `'` | Prefix will display on the left of the input field. |
+| `suffix` | `String` | `'` | Suffix will display on the left of the input field. |
```
simonsandersje: Will both 'prefix' and 'suffix' display on the left?

FayeCarter: nope 🤦

packages/components/atoms/f-form-field/README.md
```diff
@@ -72,15 +72,29 @@ The props that can be defined are as follows (if any):
...
 | `dropdownOptions` | `Array` | `null` | The options to be displayed in the dropdown. |
 | `isGrouped` | `Boolean` | `false` | When `true` will remove margin between all grouped form fields. |
 | `minNumber` | `Number` or `undefined` | `undefined` | Sets the value of the `min` property when `inputType` is `number` |
 | `maxNumber` | `Number` or `undefined` | `undefined` | Sets the value of the `max` property when `inputType` is `number` |
-| `hasInputDescription` | `Boolean` | `false` | When `true` will add the ability to insert extra html element under form label to the component via slot. |
+| `labelDescription` | `String` | `'` | Extra text that will display underneath the label text. |
+| `labelDetails` | `String` | `'` | Details for the label that will display on the right of label text. |
+| `assistiveText` | `String` | `'` | Text to assist the user that will display underneath the input field. |
+| `prefix` | `String` | `'` | Prefix will display on the left of the input field. |
+| `suffix` | `String` | `'` | Suffix will display on the right of the input field. |
```
ashleynolan:
```suggestion
| `labelDescription` | `String` | `''` | Extra text that will display underneath the label text. |
| `labelDetails` | `String` | `''` | Details for the label that will display on the right of label text. |
| `assistiveText` | `String` | `''` | Text to assist the user that will display underneath the input field. |
| `prefix` | `String` | `''` | Prefix will display on the left of the input field. |
| `suffix` | `String` | `''` | Suffix will display on the right of the input field. |
```

packages/components/atoms/f-form-field/README.md
```diff
@@ -72,15 +72,29 @@ The props that can be defined are as follows (if any):
 | `locale` | `String` | `''` | Sets locale for I18n. |
 | `labelText` | `String` | `''` | The text that will be displayed in the form field label. |
 | `inputType` | `String` | `text` | The type of input needed. <br>Options: `text`, `email`, `password`, `radio`, `checkbox`, `dropdown`, `number`, `tel`, `textarea`  |
-| `labelStyle` | `String` | `'default'` | Defines where the label will be rendered in relation to the form field. <br>Options: <br>`default` - Displays the label above the form field, <br>`inline` Displays the label inside the form field,<br>`inlineNarrow` Displays the label above the form field when in web/tablet. Displays the label inside the form field when in mobile.<br>|
+| `shouldShowLabelText` | `Boolean` | `true` | Hides the label if set to `false` |
```
ashleynolan: Super NIT – I'd probably order these props alphabetically, so that it's easier to scan if people are looking for info on a specific prop

packages/components/atoms/f-form-field/README.md
```diff
@@ -72,15 +72,29 @@ The props that can be defined are as follows (if any):
...
 
+
+### Slots
+
+Available slots that can be used.
+
+| Slot Name  | Type  | Description |
+| ----- | ----- | ----------- |
+| `icon-leading` | Icon | Will display an icon inside the input field on the left hand side. |
+| `icon-trailing` | Icon | Will display an icon inside the input field on the right hand side. |
```
ashleynolan, line 97:
```suggestion
| `icon-trailing` | Icon | Will display an icon inside the input field on the right hand side. |

```

packages/components/atoms/f-form-field/src/components/FormLabel.vue
```diff
@@ -6,7 +6,6 @@
         :class="[
             $style['o-form-label'],
             $style['c-formField-label'], {
-                [$style['c-formField-label--inline']]: isInline,
```
ashleynolan, line 9: You can delete the classes to do with this now – like `.c-formField-label--inline` and the associated selectors with this class in it (line 86 and a couple more below that)

packages/components/atoms/f-form-field/stories/AffixedFormField.stories.js
```diff
@@ -27,8 +30,8 @@ export const AffixedFormFieldComponent = () => ({
         hasError: {
             default: boolean('hasError', false)
         },
-        fieldSize: {
-            default: select('Field Size', VALID_FIELD_SIZES, DEFAULT_FIELD_SIZE)
+        shouldShowLabelText: {
```
ashleynolan, line 33: Can be a separate PR, but should there be a couple of tests for this, or do the current tests cover this new prop?

packages/components/atoms/f-form-field/stories/FormField.stories.js
```diff
@@ -81,10 +80,8 @@ export const FormFieldComponent = () => ({
             :rows="7"
             :maxlength="200"
             :labelDetails="labelDetails"
-            :assistiveText="assistiveText"
-            :has-input-description="hasInputDescription">
-                Here is a bit more text to show
-        </form-field>`
+            :labelDescription="labelDescription"
```
ashleynolan, line 83: Same as above – are the current tests sufficient for the new prop rather than what was testing the slot previously?

packages/components/atoms/f-form-field/stories/FormField.stories.js
```diff
@@ -81,10 +80,8 @@ export const FormFieldComponent = () => ({
...
+            :assistiveText="assistiveText"/>`
 });
 
-FormFieldComponent.storyName = 'f-form-field';
+FormFieldComponent.storyName = 'Form Field';
```
ashleynolan, line 87: May be better as a separate PR (or raise an engineering excellence ticket for someone to pick up separately), but I think we should likely match the stories to the Figma naming and how they are split out.

So this would then be called `Text Input – Default` and then for the Affixed and Icon types, it would be `Text Input – Affixed` and `Text Input – icon`.

We should also create separate story files for dropdown, textarea and radio/checkbox components I think, as it's not obvious that those components are a part of the Default Form Field component otherwise just from looking at the nav structure – I think if I was new to the repo, it would be easy to presume that those didn't exist.

Probably enough there though to be a separate ticket in EE though!

FayeCarter: I'll create a ticket for it

packages/components/atoms/f-form-field/stories/AffixedFormField.stories.js
FayeCarter: I've added tests now 👍

simonsandersje (approved):

kevinrodrigues (approved):

PR #1157, opened 2021-08-02
f-cookie-banner@v0.21.0: Reopen link to bottom of window

simonsandersje:
Reopen link should appear at the bottom of the window, even when the main content is less than 100% height


matthewhardern (approved):

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -334,6 +337,13 @@ export default {
                     break;
                 }
             }
+        },
+
+        /**
+         * Check to see if we need to absolute position reopen link
+         */
+        isBodyHeightLessThanWindowHeight () {
```
DamianMullins, line 345: Do you need to re-run this if the window is resized?

DamianMullins (approved):

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
simonsandersje: The only time I can reproduce the steps needed to require this work is when you manage to hit a URL for a search that results in an error on SERP. I can't imagine anyone resizing this page. They will probably be straight on the back button, or clicking on the JE logo.

PR #1163, opened 2021-08-03
f-http@v0.6.1 - Fixes and Improvements for Integration Testing

tommcclean:
v0.6.1
*August 03, 2021*

### Added
- Support for resetting mocks between tests

### Changed
- Updated and simplfiied documentation

### Fixed
- Corrections to a few class names
- Removed async from the Mock Factory

### DIFF
--- a/packages/services/f-http/README.md
+++ b/packages/services/f-http/README.md
@@ -26,14 +26,14 @@ This package exposes methods for interacting with restful services, it may abstr
 
 ## Benefits (Soon)
 - _Opt-in ability to use dynamic timeouts_
-- _Opt-in automatic providing of diagnostic headers, such as je-conversation
-- _Opt-in automatic providing of headers, such as accept-tenant
+- _Opt-in automatic providing of diagnostic headers, such as je-conversation_
+- _Opt-in automatic providing of headers, such as accept-tenant_
 
 ## Usage
 
 ### Installation
 
-Install the module using npm or yarn
+Install the module using npm or Yarn
 
 ```sh
 yarn add @justeat/f-http
@@ -43,7 +43,7 @@ yarn add @justeat/f-http
 ### Initialisation
 Ideally the package should be initialised by your website and the httpClient placed in context or a prototype, rather than initialising it in each individual component or every time you make a request.
 
-#### Initialise an httpClient
+#### Initialise an `httpClient`
 ```js
 import httpModule from '@justeat/f-http';
 
@@ -94,7 +94,7 @@ export default {
 ```
 
 ### Setting Authorisation Token
-You can optionally set an authorisation token globally so that all requests provide it
+You can globally set the authorisation token so that all requests provide it
 
 ```js
 // Some event happened that means we now have a token
@@ -163,4 +163,4 @@ put | PUT a resource | resource URL _[string]_, body _[object]_, headers _[array
 patch | PATCH a resource | resource URL _[string]_, body _[object]_, headers _[array]_
 delete | DELETE a resource | resource URL _[string]_, headers _[array]_
 setAuthorisationToken | Set the authorisation token for all requests | authorisationToken _[string]_
-readConfiguration | Returns the configuration which has been provided | None
+readConfiguration | Returns the provided options | None
--- a/packages/services/f-http/README.md
+++ b/packages/services/f-http/README.md
@@ -15,26 +15,25 @@ Javascript HTTP client for interacting with restful services
 [![Coverage Status](https://coveralls.io/repos/github/justeat/f-http/badge.svg)](https://coveralls.io/github/justeat/f-http)
 [![Known Vulnerabilities](https://snyk.io/test/github/justeat/f-http/badge.svg?targetFile=package.json)](https://snyk.io/test/github/justeat/f-http?targetFile=package.json)
 
-This package exposes methods for interacting with restful services, it may abstract any number of popular NPM packages which provide the ability to perform GET, PUT, POST, PATH, DELETE requests; while also adding lots of additional benefits.
+This package exposes methods for interacting with restful services, it may abstract any number of popular NPM packages which provide the ability to perform GET, PUT, POST, PATH, DELETE requests; while also adding some additional benefits.
 
 ## Benefits (Now)
 - Easy configuration of reusable clients which can be retrieved from context
 - Enables us to switch to alternative HTTP packages when desired
 - Sensible defaults, with the ability to override everything
 - Ability to set authorisation tokens for all requests for specific clients
-- Ability to name each client, so output from it is grouped and labelled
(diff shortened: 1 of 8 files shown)


JoshuaNg2332 (dismissed):

packages/services/f-http/README.md
```diff
@@ -15,26 +15,25 @@ Javascript HTTP client for interacting with restful services
...
-- Ability to name each client, so output from it is grouped and labelled
-- Automatically collect stats showing how long real API calls take
+- Ability to automatically collect stats showing how long real API calls take
 
 ## Benefits (Soon)
-- _Opt-in automatic logging of errors_
 - _Opt-in ability to use dynamic timeouts_
-- _Opt-in automatic providing of diagnostic headers, such as Conversation ID_
+- _Opt-in automatic providing of diagnostic headers, such as je-conversation
+- _Opt-in automatic providing of headers, such as accept-tenant
```
BanksySan: Missing closing `_` for the italics.

packages/services/f-http/README.md
```diff
@@ -15,26 +15,25 @@ Javascript HTTP client for interacting with restful services
...
 
 ## Usage
 
 ### Installation
 
-Install the module using npm or Yarn:
+Install the module using npm or yarn
```
BanksySan: _Yarn_ should have a capital first letter here (_npm_ still shouldn't).

packages/services/f-http/README.md
```diff
@@ -44,53 +43,25 @@ yarn add @justeat/f-http
 ### Initialisation
 Ideally the package should be initialised by your website and the httpClient placed in context or a prototype, rather than initialising it in each individual component or every time you make a request.
 
-#### Initialise in a Nuxt Plugin (Works on Server and Client)
+#### Initialise an httpClient
```
BanksySan: Whilst your here, _httpClient_ should be either _HTTP client_ or `httpClient`.

packages/services/f-http/README.md
```diff
@@ -112,28 +83,24 @@ If you would rather create the HTTPClient when you use it, that's fine too; it j
...
+    const httpClient = new httpModule.CreateClient(configuration);
 
     const result = await httpClient.get('/todos/1');
   }
 }
 ```
 
 ### Setting Authorisation Token
-If you need to call an authorised endpoint, you could pass the authorisation token through as a header in every request; but if you group clients correctly you may be able to benefit from applying it once for the lifetime of that client.
+You can optionally set an authorisation token globally so that all requests provide it
```
BanksySan: Should be either:

> You can optionally optionally set the authorisation token to be global...

or:

> You can globally set the authorisation token...

packages/services/f-http/README.md
```diff
@@ -203,4 +163,4 @@ put | PUT a resource | resource URL _[string]_, body _[object]_, headers _[array
 patch | PATCH a resource | resource URL _[string]_, body _[object]_, headers _[array]_
 delete | DELETE a resource | resource URL _[string]_, headers _[array]_
 setAuthorisationToken | Set the authorisation token for all requests | authorisationToken _[string]_
-readConfiguration | Returns the configuration which has been applied | None
+readConfiguration | Returns the configuration which has been provided | None
```
BanksySan: Would _Returns the current configuration._ or _Returns the active configuration._ be better?

packages/services/f-http/src/index.js
```diff
@@ -5,8 +5,8 @@ import httpVerbs from './httpVerbs';
 /**
  *  @class httpModule
  *  @type {Object}
- *  @property {function} createClient Create a new HTTP client - accepts options.
- *  @property {mockFactory} mockFactory Exposes functions to setup mocking of the underlying HTTP connection
+ *  @property {object} CreateClient Create a new HTTP client - accepts options.
```
BanksySan, line 8: Can we define what options are valid?

packages/services/f-http/src/mockFactory.js
```diff
@@ -15,7 +15,7 @@ export default class StatClient {
      * @param {object} statusCode - The status code to return
      * @param {object} data - The data to return
      */
-    async setupMockResponse (method, url, requestData, statusCode, data) {
+    setupMockResponse (method, url, requestData, statusCode, data) {
```
BanksySan, line 18: Did you mean to remove `async`?

BanksySan (dismissed):

packages/services/f-http/src/mockFactory.js
xander-marjoram: It doesn't look like anything is being `await`ed so this looks okay to me.

tommcclean: Yeah this was intentional

packages/services/f-http/src/index.js
tommcclean: These are documented in the readme; thats a pattern we use for all of the services

JoshuaNg2332 (approved):

packages/services/f-http/README.md
```diff
@@ -151,31 +118,26 @@ const wrapper = mount(MyComponent, {
...
 ```
 
-### Integration Testing Guidance
-F-Http exposes a way to create a mockClient, this enables you to mock the underlying HTTPProvider with responses.
+### Integration / System Testing Guidance
+The module exposes a way to create a MockClient, so you can mock the underlying API with pre-configured responses.
 
 ```js
-import httpModule from '@justeat/f-http';
+import { httpVerbs, MockFactory, CreateClient } from '@justeat/f-http';
```
DamianMullins, line 125: Is this the recommended way to import the functions now? Asking as the examples above are using the `httpModule.CreateClient` syntax.

tommcclean: I don't think we have a recommended way at the moment; mostly trying both out until one seems like its better than another, it works either way.

I think I am leaning towards the destructured approach simply because it reduces the code footprint by a tiny amount, but I haven't settled yet.

It seems better to destructure in tests where we use all of the exports; but no real difference in actual web hosts when wiring up the actual client). Ill probably revisit the way we document that once we have completed an MVP as the package isn't production ready yet so not anticipating people will use it yet.

Additionally; I think ill rename the export CreateClient if / before we recommend the package for use, as it makes more sense to use the class name `httpClient` given it has changed from a function a few versions ago.

kevinrodrigues (approved):

packages/services/f-http/README.md
DamianMullins: I think if you're using named imports and you want to create a variable to contain everything then you have to use the `import * as httpModule` syntax. 

The nice thing with named exports is that if you have the npm package configured correctly then it'll allow for tree-shaking and only the used functions will be included in your final JS bundle.

PR #1164, opened 2021-08-03
f-registration@v0.59.0 - Upgrade f-http to 0.6.1

tommcclean:
**f-registration**
v0.59.0
*August 04, 2021*

### Changed
- Updated Integration Tests to work with new f-http version
- Use $http from context instead of creating a new HTTP Client
- Updated version of `f-button (Carried over).

<hr>

**Storybook**
v0.39.0
*August 04, 2021*

### Added
- Upgrade to f-http@v0.6.1


JoshuaNg2332 (dismissed):

packages/storybook/context/http.context.js
```diff
@@ -1,9 +1,9 @@
 import Vue from 'vue';
-import http from '@justeat/f-http';
+import httpModule from '@justeat/f-http';
 
 // Refer to f-http documentation for usage
 export default () => {
-    const httpClient = http.createClient();
+    const httpClient = new httpModule.CreateClient();
```
matthewhardern, line 6: Could we not export the http as a const also so we don't have to pass it through to service functions?

packages/components/organisms/f-registration/src/components/Registration.vue
```diff
@@ -470,7 +470,8 @@ export default {
                     registrationSource: 'Native',
                     marketingPreferences: []
                 };
-                await RegistrationServiceApi.createAccount(this.createAccountUrl, registrationData, this.tenant);
+
+                await RegistrationServiceApi.createAccount(this.$http, this.createAccountUrl, registrationData, this.tenant);
```
matthewhardern, line 474: See below seems annoying to have to pass `this.$http` to the service this way

packages/storybook/context/http.context.js
tommcclean: How do you mean? Making it a global?

The module doesn't act as a singleton, because we may want different clients (for example Smart Gateway could come baked with the base url so it doesn't need to be provided to underlying components anymore etc), its also allowing us to inject our stats transporter in once and any configuration which can now be standardised - rather than having to wire it up in every component (which isn't possible at the transportation tech can only be initialised once)

packages/components/organisms/f-registration/src/components/Registration.vue
tommcclean: What would you recommend instead? If we performed our API calls in Vuex then we wouldn't need to; its because we are using JS files outside of the Vue scope.

packages/storybook/context/http.context.js
matthewhardern: thanks for the explanation, as per our convo makes more sense.

packages/components/organisms/f-registration/src/components/Registration.vue
matthewhardern: without being able to use singleton can see this would not be easy to do, so for now might be best as you said to get it working

JoshuaNg2332 (approved):

matthewhardern (approved):

simonsandersje (approved):

PR #1166, opened 2021-08-04
f-checkout@v0.168.0 - Added tab loop tests to f-checkout accessibility. #globalconfig

JoshuaNg2332:
v0.167.0
*August 5, 2021*

### Added
- Modal specific tab looping tests to `axe-accessibility.spec.js` that support `dinein`, `collection` and `delivery` variants of checkout that support both authenticated and guest states.

Failing test

### DIFF
--- a/packages/components/organisms/f-checkout/test-utils/component-objects/f-checkout.component.js
+++ b/packages/components/organisms/f-checkout/test-utils/component-objects/f-checkout.component.js
@@ -1,4 +1,4 @@
-const Page = require('@justeat/f-wdio-utils/src/page.object');
+const Page = require('../../../../../services/f-wdio-utils/src/page.object');
 const {
     CHECKOUT_COMPONENT,
     ORDER_TIME_DROPDOWN,
@@ -14,6 +14,7 @@ const {
     PRE_ORDER_WARNING,
     CHECKOUT_ERROR_MESSAGE,
     RETRY_BUTTON,
+    CLOSE_MODAL,
     DUP_ORDER_GO_TO_HISTORY_BUTTON,
     ERROR_PAGE_COMPONENT,
     ERROR_PAGE_HEADING,
@@ -47,6 +48,8 @@ module.exports = class Checkout extends Page {
     get checkoutErrorMessage () { return $(CHECKOUT_ERROR_MESSAGE); }
 
     get errorMessageRetry () { return $(RETRY_BUTTON); }
+
+    get closeMessageModal () { return $(CLOSE_MODAL); }
 
     get errorMessageDupOrderGoToHistory () { return $(DUP_ORDER_GO_TO_HISTORY_BUTTON); }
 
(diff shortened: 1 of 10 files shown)


packages/components/organisms/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -1,4 +1,4 @@
-const Page = require('@justeat/f-wdio-utils/src/page.object');
+const Page = require('../../../../../services/f-wdio-utils/src/page.object');
```
JoshuaNg2332, line 1: This will be addressed in a 2nd PR once this has merged so that I can publish the updated f-wdio-utils to npm.

packages/components/organisms/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -169,6 +172,32 @@ module.exports = class Checkout extends Page {
         return this.errorMessageDupOrderGoToHistory.click();
     }
 
+    getCheckoutErrorTabOrder () {
```
kevinrodrigues, line 175: This one and the one below is quite similar, could they be made into one method instead?

packages/components/organisms/f-checkout/test/accessibility/axe-accessibility.spec.js
```diff
@@ -65,4 +67,236 @@ describe('Accessibility tests', () => {
...
+        checkout.populateGuestCollectionCheckoutForm(addressInfo);
+        checkout.goToPayment();
+
+        // Assert
+        const result = checkout.getCheckoutErrorTabOrder();
+        expect(result.actual).toEqual(result.expected);
+    });
+
+    forEach([
+        ['delivery', 'restaurant-not-taking-orders'],
```
kevinrodrigues, line 156: Could these potentially be added to the forEach above where we just have collection? (Line 125)

packages/components/organisms/f-checkout/test-utils/component-objects/f-checkout.component.js
JoshuaNg2332: Currently this and the method below are serving 2 different expected tab orders. Which is why they are split into 2 methods. We had the decision to either have the expected tab loop order defined in the test or have them defined in the method and found that to be a neater solution due to the fact that multiple tests would be using the same tab loop order.
This is why they are 2 methods.

packages/components/organisms/f-checkout/test/accessibility/axe-accessibility.spec.js
JoshuaNg2332: They can't as delivery and collection have different form fields for the guest checkout state.
When you try to have them together and give it more fields to fill out than the page has, the test fails, citing that it can't find the relevant fields to populate. (e.g. address is not required in collection).

kevinrodrigues (dismissed):

kevinrodrigues (approved):

dandel10n (approved):

PR #1178, opened 2021-08-09
f-analytics@0.8.0 - Added global method to Push GTM ad-hoc events

oliversweb:
### Added
- Added push ad-hoc event method to global var

### Changed
- Moved the preparation of the server side plaform data to a plugin.
- Refactored the `Store` to allow store serverside events to be stored until clientside.


packages/services/f-analytics/src/mixins/analytics.mixin.vue
```diff
@@ -62,13 +66,18 @@ export default {
...
-            this.updatePlatformData(platformData);
+            this.pushPlatformData(platformData);
         },
 
         prepareUserData (authToken) {
             const userData = { ...this.userData };
 
-            userData['a-UserId'] = this.$cookies.get('je-auser') || undefined;
+            userData['a-UserId'] = undefined;
+            // TODO - Read manually to reduce need on global '$cookies'
```
dandel10n, line 41: By "read manually" do you mean pass it as a prop the same way as authToken?

packages/services/f-analytics/src/store/_tests/analytics.module.test.js
```diff
@@ -1,71 +1,361 @@
...
+                // Act
+                pushPlatformData({ commit, state }, modifieldState.platformData);
+
+                // Assert
+                expect(commit).toHaveBeenLastCalledWith('pushPlatformData', modifieldState.platformData);
+            });
+
+            it('should push the `platformData` if clientside & dataLayer present', () => {
+                // Arrange
+                const windowsPushSpy = jest.fn();
```
dandel10n: Can this be:
```
const windowsPushSpy = jest.spyOn(window.dataLayer, 'push');
```
as you did in other tests?

packages/services/f-analytics/src/store/_tests/analytics.module.test.js
```diff
@@ -1,71 +1,361 @@
...
+                        push: windowsPushSpy
+                    }
+                }));
+
+                // Act
+                pushPlatformData({ commit, state }, modifieldState.platformData);
+
+                // Assert
+                expect(windowsPushSpy).toHaveBeenLastCalledWith({ platformData: { ...modifieldState.platformData } });
+                expect(state.platformData).toEqual(modifieldState.platformData);
```
dandel10n: You have separate tests to check that mutations update the state, I would remove these checks from actions to keep tests focused on actions only

packages/services/f-analytics/src/store/analytics.module.js
```diff
@@ -25,24 +32,58 @@ export default {
...
     mutations: {
-        pushPlatformData: (state, platformData) => {
-            state.platformData = platformData;
+        [PUSH_PLATFORM_DATA]: (state, platformData) => {
+            state.platformData = { ...state.platformData, ...platformData };
         },
-        pushUserData: (state, userData) => {
+
+        [PUSH_USER_DATA]: (state, userData) => {
             state.userData = userData;
```
dandel10n: This should be also changed:
`state.userData = { ...state.userData, ...userData };`

packages/services/f-analytics/src/store/_tests/analytics.module.test.js
oliversweb: no, because I do an `expect` on it in the Assert section but I need to assign it to the window mock below, so by creating it separately I can do both.

packages/services/f-analytics/src/mixins/analytics.mixin.vue
oliversweb: No, once moved into the service class there is a helper method `getCookie()` which doesn't rely on anything other than the request, so we then don't have to 🤞 that `$cookies` is available.

packages/services/f-analytics/src/plugins/analytics.plugin.js
```diff
@@ -1,8 +1,40 @@
...
+        const parts = value.split(`; ${name}=`);
+        if (parts.length === 2) {
+            return parts.pop().split(';').shift();
+        }
+    }
+
+    return undefined;
+};
+
+const mapServersideValues = (store, req, options) => {
```
AndyMayJE: We will need to extend this after 0.7.0 to grab the httpStatusCode and dispatch to store pageData

packages/services/f-analytics/src/mixins/_tests/analytics.created.test.js
```diff
@@ -22,7 +22,7 @@ describe('Analytics', () => {
     describe('created ::', () => {
         let component;
         let prepareServersideAnalyticsSpy;
-        let storeUpdatePlatformDataSpy;
+        let pushUpdatePlatformDataSpy;
```
dandel10n: Should this become `pushPlatformDataSpy`?

packages/services/f-analytics/src/mixins/_tests/analytics.methods.test.js
```diff
@@ -138,7 +138,7 @@ describe('Analytics', () => {
             let wrapper;
 
             component.mixins[0].created = jest.fn(() => true);
-            const storeUpdateUserDataSpy = jest.spyOn(component.mixins[0].methods, 'updateUserData');
+            const pushUpdateUserDataSpy = jest.spyOn(component.mixins[0].methods, 'pushUserData');
```
dandel10n:
```suggestion
            const pushUserDataSpy = jest.spyOn(component.mixins[0].methods, 'pushUserData');
```

packages/services/f-analytics/src/plugins/analytics.plugin.js
oliversweb: Yes, we have a ticket for that

AndyMayJE (dismissed):

packages/services/f-analytics/src/mixins/analytics.mixin.vue
```diff
@@ -62,13 +66,18 @@ export default {
             platformData.language = COUNTRY_INFO[this.$i18n.locale].language;
             platformData.currency = COUNTRY_INFO[this.$i18n.locale].currency;
 
-            this.updatePlatformData(platformData);
+            this.pushPlatformData(platformData);
```
dandel10n: Do we still need this preparePlatformData() method if you moved it to service in this PR? Sorry, maybe I missed something

oliversweb: Left it there to reduce merge issues because @AndyMayJE relies on the method atm, so if I remove it in my pr then I cannot see what logic I need to re-add to the service method.

oliversweb: Just until @AndyMayJE merges his pr because he has logic to add to this area and if I delete it on my pr then merge hell :\

packages/services/f-analytics/src/store/_tests/analytics.module.test.js
```diff
@@ -1,94 +1,395 @@
...
+                    branding: 'test-branding',
+                    country: 'zu',
+                    language: 'ze',
+                    currency: 'zud'
+                };
+                const expected = { ...currentState.platformData, ...clientsidePlatformData };
 
                 // Act
-                updatePlatformData({ commit }, updatedPlatformData);
+                mutations.pushPlatformData(currentState, clientsidePlatformData);
```
dandel10n: For consistency, could this be `mutations[PUSH_PLATFORM_DATA]` a on line 302?

packages/services/f-analytics/src/store/_tests/analytics.module.test.js
```diff
@@ -1,94 +1,395 @@
...
+                    conversationId: '460cc3a8-83f7-4e80-bb46-c8a69967f249',
+                    requestId: '6cbe6509-9122-4e66-a90a-cc483c34282e',
+                    orientation: 'Landscape',
+                    display: 'wide'
+                };
+                const expected = { ...currentState.pageData, ...clientsidePageData };
 
                 // Act
-                updateUserData({ commit }, updatedUserData);
+                mutations.pushPageData(currentState, clientsidePageData);
```
dandel10n: The same here

packages/services/f-analytics/src/store/analytics.module.js
```diff
@@ -35,30 +43,64 @@ export default {
...
             display: ''
-        }
+        },
+        events: []
     }),
 
     actions: {
-        updatePlatformData: ({ commit }, platformData) => {
-            commit('pushPlatformData', platformData);
+        [PUSH_PLATFORM_DATA]: ({ commit, state }, platformData) => {
```
dandel10n: Actions shouldn't use `mutation-types` values. And as far as I know actions should be responsible only for manipulation the data and calling the mutation. So we need to move datalayer push from it. I would call the action `setPlatformData` or `updatePlatformData` as action won't be responsible for a push.

```suggestion
        setPlatformData: ({ commit, state }, platformData) => {
```

packages/services/f-analytics/src/store/analytics.module.js
```diff
@@ -35,30 +43,64 @@ export default {
...
+            if (platformData) {
+                commit(PUSH_PLATFORM_DATA, platformData);
+            }
+
+            if (isDataLayerPresent()) {
+                window.dataLayer.push({ platformData: { ...state.platformData } });
+            }
+        },
+
+        [PUSH_USER_DATA]: userData => {
```
dandel10n: As we don't manipulate the state here we shouldn't have this action. This logic should live in the methods

packages/services/f-analytics/src/store/analytics.module.js
```diff
@@ -35,30 +43,64 @@ export default {
...
+            if (userData && isDataLayerPresent()) {
+                window.dataLayer.push({ userData: { ...userData } });
+            }
         },
-        updateUserData: ({ commit }, userData) => {
-            commit('pushUserData', userData);
+
+        [PUSH_PAGE_DATA]: ({ commit, state }, pageData) => {
```
dandel10n: The same as for platform data here

packages/services/f-analytics/src/store/analytics.module.js
```diff
@@ -35,30 +43,64 @@ export default {
...
+            }
+
+            if (isDataLayerPresent()) {
+                window.dataLayer.push({ pageData: { ...state.pageData } });
+            }
         },
-        updatePageData: ({ commit }, pageData) => {
-            commit('pushPageData', pageData);
+
+        [PUSH_EVENT]: ({ commit, state }, event) => {
```
dandel10n: The same as for platform data here. I would split this into two actions: one for updating eventList and one for clearing it and would handle dataLayer push in the methods

dandel10n (dismissed):

packages/services/f-analytics/src/mixins/_tests/analytics.methods.test.js
```diff
@@ -151,80 +106,123 @@ describe('Analytics', () => {
...
                 });
             });
-            it('should set the correct pageData', () => {
+
+            it('should not push pageData to datalayer if datalayer not present', async () => {
+                // Arrange - window state
+                windowsPushSpy = jest.fn();
+                const originalWindow = { ...window };
+                const windowSpy = jest.spyOn(global, 'window', 'get');
+                windowSpy.mockImplementation(() => ({
```
kevinrodrigues, line 207: Do you need this one since the next line (213) seems to override this `mockImplementation`

packages/services/f-analytics/src/plugins/analytics.plugin.js
```diff
@@ -1,7 +1,47 @@
...
+        }
+    }
+
+    return undefined;
+};
+
+const prepareServersideValues = (store, req, res, options) => {
+    // Only available serverside
+    if (typeof (window) === 'undefined') {
+        // Platform Data
```
kevinrodrigues: Do we need this comment?

packages/services/f-analytics/src/plugins/analytics.plugin.js
```diff
@@ -1,7 +1,47 @@
...
+        if (process.env.justEatEnvironment) platformData.environment = process.env.justEatEnvironment;
+        if (process.env.FEATURE_VERSION) platformData.version = process.env.FEATURE_VERSION;
+        if (process.env.INSTANCE_POSITION) platformData.instancePosition = process.env.INSTANCE_POSITION;
+
+        const userPercent = getCookie('je-user_percentage', req);
+        if (userPercent) platformData.jeUserPercentage = userPercent;
+
+        store.dispatch(`${options.namespace}/${UPDATE_PLATFORM_DATA}`, platformData);
+
+        // Page Data
```
kevinrodrigues: Do we need this comment?

packages/services/f-analytics/src/services/_tests/analytics.service.test.js
```diff
@@ -0,0 +1,118 @@
...
+        jest.clearAllMocks();
+    });
+
+    it('should instantiate a new instance', () => {
+        // Act
+        const instance = () => new AnalyticService(store, req, options);
+
+        // Assert
+        expect(instance).not.toThrowError();
+        expect(instance).not.toBeNull();
```
kevinrodrigues: I could be wrong here but would `.toBeDefined()` be better here since you could get an undefined result which would pass here?

packages/services/f-analytics/src/mixins/_tests/analytics.methods.test.js
oliversweb: I need to create a spy on it; so I can prove it has not been called, before I implement the second mock to control the flow.

kevinrodrigues (approved):

dandel10n (approved):

PR #1180, opened 2021-08-10
f-checkout@v0.169.0 - add age verification

FayeCarter:
### Added
- Age Verification Page.
- Tests to cover changes.

### Removed
- Some eventData values which were added for testing specific case

### Changed
- `f-form-field` version.

(Stacked dropdowns will be replaced by a date picker soonish 🤞 )
image


packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -371,6 +377,10 @@ export default {
             'userNote'
         ]),
 
+        shouldShowAgeVerification () {
```
charliearlie, line 379: Can remove this until my work is merged in. 🙂

zlatinivanov-je (commented):
LGTM, just wonder if `input type="date"` was discussed at all... 
Also not clear on date validation, does not seem to be implemented.

FayeCarter:
> LGTM, just wonder if `input type="date"` was discussed at all...
> Also not clear on date validation, does not seem to be implemented.

So I'm aiming to create a date picker as a new component after this is done.
The validations are also set up as another ticket. 👍

packages/components/organisms/f-checkout/src/components/AgeVerification.vue
```diff
@@ -0,0 +1,160 @@
...
+        :card-heading="$t(`ageVerification.heading`)"
+        data-test-id="checkout-age-verification-component">
+        <p
+            :class="$style['c-checkout-ageVerification-description']"
+            data-test-id="checkout-age-verification-description">
+            {{ $t(`ageVerification.description`) }}
+        </p>
+
+        <div>
+            <form-field
```
ashleynolan, line 16: Are these form field wrapped in a `form` element at any point? (hard to tell when split into components.

If not, should likely change the outer `div` to a `form` ideally so that it's better semantically.

Not sure on the requirements for this form, but would it be beneficial to have this submit to an endpoint? Just wondering if it's fully client-side validated, how easy it would be for people to get around it potentially (but may not even be an issue if that's accepted risk from product).

packages/components/organisms/f-checkout/src/components/_tests/AgeVerification.test.js
```diff
@@ -0,0 +1,231 @@
...
+                    // Act
+                    wrapper.vm.initDateOfBirth();
+                });
+
+                it('should update `selectedDate.day` with "1"', () => {
+                    // Assert
+                    expect(wrapper.vm.selectedDate.day).toEqual('1');
+                });
+
+                it('should update `selectedDate.month` with "0"', () => {
```
ashleynolan: Hard to know without context, but why does this get set to `0`? Does this represent no selected value when it's displayed on screen to the user? If so, may be worth adding a bit more description to the test title (especially as the day gets set to `1`?)

packages/components/organisms/f-checkout/src/components/_tests/AgeVerification.test.js
```diff
@@ -0,0 +1,231 @@
...
+                    // Assert
+                    expect(wrapper.vm.selectedDate.month).toEqual('0');
+                });
+
+                it('should update `selectedDate.year` with the current year', () => {
+                    // Assert
+                    // expect(dateNowSpy).toHaveBeenCalled();
```
ashleynolan: Is this line needed ?

packages/components/organisms/f-checkout/src/components/_tests/AgeVerification.test.js
```diff
@@ -0,0 +1,231 @@
...
+                    store: createStore({
+                        ...defaultCheckoutState,
+                        customer: {
+                            dateOfBirth: new Date('2010', '9', '2')
+                        }
+                    })
+                });
+            });
+
+            it('should call `updateDateOfBirth` with `selectedDate`', () => {
```
ashleynolan, line 213: I'm not a test guru by any means, but rather than testing that it calls `updateDateofBirth` (which you can do too), would it not be better to test that the state itself has been updated to the new date? (as that's the desired result)

packages/components/organisms/f-checkout/src/components/AgeVerification.vue
FayeCarter: I'll change it to a form, thanks.

The button will make a call to updateCheckout which should allow it to get properly validated.

packages/components/organisms/f-checkout/src/components/_tests/AgeVerification.test.js
FayeCarter: Nope, thanks

packages/components/organisms/f-checkout/src/components/_tests/AgeVerification.test.js
FayeCarter: I can do, but `handleAgeVerifcation` just calls the action in the store that then calls the mutation that updates the state 🤯 .  So I thought you wouldn't test that the state changes because `handleAgeVerifcation` doesn't directly change the state?

packages/components/organisms/f-checkout/src/components/_tests/AgeVerification.test.js
FayeCarter: Thanks, I'll add some description. Annoyingly I'm assuming dates are stored as an array, so when you do `getMonth()` January = 0, Feb =1 etc.

FayeCarter: @ashleynolan I've updated the test description and added a comment, Is that any better?

packages/components/organisms/f-checkout/src/components/AgeVerification.vue
```diff
@@ -0,0 +1,164 @@
...
+            this.updateDateOfBirth(dateOfBirth);
+            this.$emit('verify-age');
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+.c-checkout-ageVerification-description {
+    margin: spacing(x2) 0 spacing(x2);
```
ashleynolan:
```suggestion
    margin: spacing(x2) 0;
```

packages/components/organisms/f-checkout/test-utils/component-objects/f-checkout-selectors.js
```diff
@@ -54,7 +54,7 @@ export const FIELDS = {
         typeError: '[data-test-id="error-address-postcode-type-error"]'
     },
     userNote: {
-        input: '[data-test-id="formfield-textarea"]',
+        input: '[data-test-id="formfield-Note-textarea"]',
```
ashleynolan: Super NIT – can `Note` be lowercase, or does it get auto-populated somewhere (could be one to update in a separate PR).

ashleynolan (commented):
Couple of super small things, then LGTM

packages/components/organisms/f-checkout/src/components/_tests/AgeVerification.test.js
ashleynolan: Looks good 👍

packages/components/organisms/f-checkout/test-utils/component-objects/f-checkout-selectors.js
FayeCarter: It can be updated, It's auto-populated but it's just because we've put `name=Note` instead of `name=note` so I'll change it :)

packages/components/organisms/f-checkout/src/components/AgeVerification.vue
```diff
@@ -0,0 +1,164 @@
...
+                year: ''
+            }
+        };
+    },
+
+    computed: {
+        ...mapState(VUEX_CHECKOUT_MODULE, ['customer']),
+
+        days () {
+            return Array.from({ length: 31 }, (_, index) => {
```
matthewhardern, line 90: Could this give 31 days on 30 day months plus feb is 28 or 29 days?

packages/components/organisms/f-checkout/src/components/AgeVerification.vue
```diff
@@ -0,0 +1,164 @@
...
+            this.selectedDate.month = this.customer.dateOfBirth?.getMonth().toString() || this.months[0].value.toString();
+            this.selectedDate.year = this.customer.dateOfBirth?.getFullYear().toString() || this.years[0].text;
+        },
+
+        selectionChanged (selection, type) {
+            this.selectedDate[type] = selection;
+        },
+
+        handleAgeVerifcation () {
+            const dateOfBirth = new Date(this.selectedDate.year, this.selectedDate.month, this.selectedDate.day);
```
matthewhardern, line 147: if the above is true this will output the next available that has a 31 day month if its a 30 day month.

matthewhardern: based On outside communication I understand this will be resolved in a later phase and won't be touching production in this state.

packages/components/organisms/f-checkout/src/components/AgeVerification.vue
matthewhardern: see other comment

matthewhardern (approved):

charliearlie (dismissed):

charliearlie (approved):

PR #1181, opened 2021-08-10
f-footer@v4.30.1 - Republish with correct tag

simonsandersje:
Previous publish of the legacy f-footer missed the "legacy" tag


simonsandersje:
Tags fixed without this PR:
https://gist.github.com/benjie/faf8268b955c6f0b438f23193bac9939

PR #1182, opened 2021-08-11
f-checkout@0.170.0 - Age verification check

charliearlie:
### Added
- Check on basket API call to see if age verification prompt needs to be shown

If the Basket API call returns a restricted prompt for alcohol, we should show the user an age verification prompt, which this enables.


packages/components/organisms/f-checkout/src/checkout-issues.js
```diff
@@ -44,6 +44,11 @@ const checkoutIssues = {
         shouldRedirectToMenu: false
     },
 
+    DATE_OF_BIRTH_REQUIRED: {
```
FayeCarter, line 47: Get and update checkout have`AGE_VERIFICATION_REQUIRED` and `DATE_OF_BIRTH_REQUIRED` as issues, do we need to check for both?

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -413,7 +413,11 @@ export default {
         },
 
         shouldShowCheckoutForm () {
-            return !this.isLoading && !this.errorFormType;
+            return !this.isLoading && !this.errorFormType && !this.shouldShowAgeVerificationForm;
+        },
+
+        shouldShowAgeVerificationForm () {
+            return this.errors.some(error => error.code === 'DATE_OF_BIRTH_REQUIRED');
```
FayeCarter: can we get this from checkout issues?

packages/components/organisms/f-checkout/src/checkout-issues.js
charliearlie: I was told it'll be `DATE_OF_BIRTH_REQUIRED`  but we will also need to handle `AGE_VERIFICATION_FAILED` when the user enters an invalid DOB.

packages/components/organisms/f-checkout/src/components/Checkout.vue
charliearlie: Checkout issues doesn't hold the code of the error. The code is the key. But I have just realised this does need changing because as basket is another domain, this will never become false. I will have to go with the original idea of having it in data.

charliearlie: Actually we could remove the error when the user enters a correct DOB? 🤷‍♂️

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1159,6 +1159,60 @@ describe('Checkout', () => {
...
+                });
+            });
+
+            describe('when the `DATE_OF_BIRTH_REQUIRED` issue does not exist in errors', () => {
+                let wrapper;
+                beforeEach(() => {
+                    // Arrange && Act
+                    wrapper = shallowMount(VueCheckout, {
+                        store: createStore({
+                            ...defaultCheckoutState
```
FayeCarter: I don't think you need this or the cookie mock?

charliearlie: I've removed the cookie mock but doesn't all mounting of the checkout require the mocked store?

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -225,9 +225,17 @@ export default {
...
                     id: data.BasketId,
                     total: data.BasketSummary.BasketTotals.Total
-                }
+                },
+                ageRestricted: data.BasketSummary.Prompts.Restrictions.some(restriction => restriction.Type === 'Alcohol')
             };
 
+            // This logic is temporary while a new endpoint is built for fulfilment status.
+            // We can't call GET checkout as a guest so we're now having to use the basket response for age restrictions until then
+            if (basketDetails.ageRestricted && (tenant === 'au' || tenant === 'nz')) {
```
DamianMullins, line 242: Is this how we do feature toggling in Fozzie components?

packages/components/organisms/f-checkout/src/constants.js
```diff
@@ -33,6 +33,8 @@ export const ANALYTICS_ERROR_CODE_SET_ORDER_TIME = 'setOrderTime';
 export const ANALYTICS_ERROR_CODE_INVALID_ORDER_TIME = 'invalidOrderTime';
 
 export const DEFAULT_CHECKOUT_ISSUE = 'DEFAULT_CHECKOUT_ISSUE';
+export const DOB_REQUIRED_ISSUE = 'DATE_OF_BIRTH_REQUIRED';
```
DamianMullins, line 36: Looks like the string value for this constant is used in a few places, might be worth importing and using this in place instead? Means less to search for and update in the future if the value ever changes.

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
FayeCarter: It does but you don't need the `{...defaultState}` part just `store: createStore()`

packages/components/organisms/f-checkout/src/store/checkout.module.js
charliearlie: I would actually like to know how to do feature toggling for another feature I've built (split notes) but this isn't it. In this case, I'm just checking if it's AU/NZ because we're using the response from basket to determine if the age verification should be shown while we await a new checkout endpoint. The checkout endpoint will do the tenant checking for us, whereas I believe Basket will return the age restricted prompt regardless of country (although i may be wrong with that)

DamianMullins: Yeah sure, I think in some of the other modules the toggling is done via a config file where you set variables per tenant, pass in the current tenant from the consuming app, and read the relevant config out and use that throughout the components - that way each tenants config is in a self-contained file.

Another approach could be similar to how menu does feature toggling where we have a JS feature toggle package that accepts the current env, tenant etc, then reads in some JSON and can tell if a feature should be enabled or not.

As the first approach is already in use I'd suggest going with that one, maybe in a future PR if it's a lot of work?

charliearlie: Would love to do it in a future PR, maybe even in my existing Split Notes PR which I'm about to revisit. Can I approach you with any questions I may have? 👀

FayeCarter (dismissed):

packages/components/organisms/f-checkout/src/store/checkout.module.js
DamianMullins: Sure, give me a shout any time (apart from next week as I'm on leave!).

packages/components/organisms/f-checkout/src/components/_tests/Checkout.test.js
```diff
@@ -1159,6 +1160,52 @@ describe('Checkout', () => {
...
+                    });
+                });
+
+                it('should return true', () => {
+                    // Assert
+                    expect(wrapper.vm.shouldShowAgeVerificationForm).toBe(true);
+                });
+            });
+
+            describe('when the `DATE_OF_BIRTH_REQUIRED` issue does not exist in errors', () => {
```
DamianMullins: Can use the `DOB_REQUIRED_ISSUE` constant here.

packages/components/organisms/f-checkout/src/checkout-issues.js
DamianMullins: Can use the `DOB_REQUIRED_ISSUE` constant here?

```suggestion
    [DOB_REQUIRED_ISSUE]: {
```

DamianMullins (dismissed):

FayeCarter (dismissed):

FayeCarter (dismissed):

DamianMullins (dismissed):

FayeCarter (dismissed):

DamianMullins (dismissed):

charliearlie:
Now the build is failing because of a selector snapshot?

FayeCarter (approved):

DamianMullins (approved):

PR #1183, opened 2021-08-11
f-http@v0.8.1 - Added new script to build f-http for storybook

siggerzz:


packages/services/f-http/package.json
```diff
@@ -31,6 +31,7 @@
     "build": "vite build",
     "lint": "eslint \"!(dist)/**/*.{js,vue}\"",
     "lint:fix": "yarn lint --fix",
+    "storybook:build": "vite build",
```
ashleynolan, line 34: Could this potentially hit a race condition with the `storybook:build` script being run in the storybook package?

I'm guessing you need this to complete before that other script builds?

siggerzz: Nope. f-http will always get built first because it's a dependency of Storybook. (See the use of `--include-dependencies` in root package.json 👍🏼

JoshuaNg2332 (dismissed):

FayeCarter (dismissed):

DamianMullins (approved):

FayeCarter (approved):

PR #1184, opened 2021-08-12
f-checkout@v0.172.0 - Add age verification validations

FayeCarter:
### Added
- Error Message to Age Verification page.
- 
Screenshot 2021-08-11 at 13 22 27


packages/components/organisms/f-checkout/src/components/_tests/AgeVerification.test.js
```diff
@@ -79,6 +79,57 @@ describe('AgeVerification', () => {
...
+            ])('should return %s when dateOfBirth is set to %s', (expectedValue, dateOfBirth) => {
+                // Arrange
+                wrapper = shallowMount(AgeVerification, {
+                    i18n,
+                    localVue,
+                    store: createStore(),
+                    computed: {
+                        userDateOfBirth () {
+                            return dateOfBirth;
+                        }
```
DamianMullins: You can shorten this using an arrow function.

```suggestion
                        userDateOfBirth: () => dateOfBirth
```

packages/components/organisms/f-checkout/src/components/_tests/AgeVerification.test.js
```diff
@@ -79,6 +79,57 @@ describe('AgeVerification', () => {
...
+            ])('should return %s when hasSelectedDateOfBirth is set to %s and isValidAge is %s', (expected, hasSelectedDateOfBirth, isValidAge) => {
+                // Arrange
+                wrapper = shallowMount(AgeVerification, {
+                    i18n,
+                    localVue,
+                    store: createStore(),
+                    computed: {
+                        isValidAge () {
+                            return isValidAge;
+                        }
```
DamianMullins: You can shorten this using an arrow function.

```suggestion
                        isValidAge: () => isValidAge
```

packages/components/organisms/f-checkout/src/components/AgeVerification.vue
```diff
@@ -37,6 +37,18 @@
                 :value="selectedDate.year"
                 @input="selectionChanged($event, 'year')" />
 
+            <div
+                ref="AgeVerificationError"
+                tabindex="-1">
+                <error-message
```
dandel10n, line 43: Can the ref and tabindex be straight on error-message component without the wrapping div?

FayeCarter: Annoyingly no, If I have it on them, the screen reader tries to read the error icon rather than the error message

dandel10n: Hmm, maybe we need to fix this in f-error-message so the icon is ignored by screen readers. But it is not a blocker for your PR, just annoying that we need a wrapper when it should work without it, right?

packages/components/organisms/f-checkout/src/components/AgeVerification.vue
```diff
@@ -59,6 +71,11 @@
 <script>
 import Card from '@justeat/f-card';
 import '@justeat/f-card/dist/f-card.css';
+<<<<<<< HEAD
```
charliearlie: Merge conflict?

packages/components/organisms/f-checkout/src/components/AgeVerification.vue
```diff
@@ -123,6 +139,21 @@ export default {
...
+
+        userDateOfBirth () {
+            return new Date(this.selectedDate.year, this.selectedDate.month, this.selectedDate.day);
+        },
+
+        isValidAge () {
+            const currentDate = new Date();
+            const minimumAge = new Date((currentDate.getFullYear() - 18), currentDate.getMonth(), currentDate.getDate());
+
+            return this.userDateOfBirth <= minimumAge;
```
charliearlie: In this case, comparing the two dates like this wouldn't work as they are not numbers, I don't think? Could potentially use their timestamps?

charliearlie: image

FayeCarter: sorted, thanks!

packages/components/organisms/f-checkout/src/components/AgeVerification.vue
FayeCarter: Agreed, I'll add a ticket in

dandel10n (dismissed):

DamianMullins (dismissed):

charliearlie (approved):

DamianMullins (approved):

PR #1210, opened 2021-08-23
fozzie-components@v3.53.0 - CircleCI. updates to improve build reliability

matthewhardern:
This PR adds changes to our Circle CI config to improve our build reliability by utilising caching and splitting up the build process.

The config changes have the following effects:

1. We now save chunks of cache based on the atoms, molecules and organisms directories. 
2. The cache for each of these directories is saved against a key which is a combination of branch and commit hash
3. When a PR is opened the config will check if it needs to run all, in the case for master this is always. 
4. If build all is TRUE then it will build all packages and then cache the resulting dist directories (for now these directories are hard coded, this will change, but for first version we will run with hard coded directories until a better solution is found). This will happen always on master branch
5.  If build all is FALSE then it will attempt to recover the branch cache for the PR branch, if it is the first build attempt, then it will use the master cache as the base cache.
6. It will then build only the changed components, I have updated this so that it uses `master...HEAD` instead of origin/master as this prevents building things that have changed in master but not your branch.
7. Once the build is complete it then saves an updated cache again against the PR branch to be used in later steps or if changes are pushed to the branch
8. Once the build is complete and linting has been performed 2 other steps are kicked off `unit_tests` and `browser_tests`. These steps then use the build cache from the previous build process and perform the relevant tests. These steps are run in parallel.

Anything I have missed that you have questions about please leave a comment.


oliversweb (dismissed):

.circleci/config.yml
```diff
@@ -106,47 +209,29 @@ commands:
             then
               yarn storybook:build
             else
-              yarn storybook:build-changed
+              yarn <storybook:build-changed
```
siggerzz: Is this intentional?

.circleci/config.yml
```diff
@@ -106,47 +209,29 @@ commands:
...
-          name: Run Percy visual regression tests for updated packages
-          command: |
-            if [ "${RUN_ALL}" == "true" ];
-            then
-              yarn test:visual
-            else
-              yarn test:visual --since origin/master
-            fi
+      - run_command:
+          command_description: un Percy visual regression tests for packages
```
siggerzz: `Run Percy visual regression tests for packages`

.circleci/config.yml
```diff
@@ -267,26 +352,56 @@ jobs:
...
+      LERNA_ARGS: --concurrency 1
+    steps:
+      - checkout
+      - detect_root_change
+      - cache_checksum
+      - restore_node_modules_cache
+      - install_node_dependencies
+      - restore_cache_dist_directories
       - unit_integration_tests
+      - run:
```
siggerzz, line 391: I'd do this after the build. No point running the tests if bundlewatch fails 👍🏼

siggerzz (changes requested):
Couple minor comments!

CHANGELOG.md
```diff
@@ -3,6 +3,15 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+v3.53.0
+------------------------------
+*August 12, 2021*
```
kevinrodrigues: 23rd.

.circleci/config.yml
matthewhardern: no good catch

ashleynolan (approved):

kevinrodrigues (approved):

siggerzz (approved):

PR #1214, opened 2021-08-24
f-checkout@v0.173.3 Improve logging

BanksySan:
Added extra items to the logged data when a user attempts to submit an invalid checkout form.

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/Checkout.vue
+++ b/packages/components/organisms/f-checkout/src/components/Checkout.vue
@@ -378,6 +378,10 @@ export default {
             'userNote'
         ]),
 
+        ...mapState(VUEX_CHECKOUT_ANALYTICS_MODULE, [
+            'changedFields'
+        ]),
+
         wasMobileNumberFocused () {
             return this.$v.customer.mobileNumber.$dirty;
         },
@@ -988,9 +992,18 @@ export default {
                 error: ANALYTICS_ERROR_CODE_INVALID_MODEL_STATE
             });
 
+            const expandedData = {
+                ...this.eventData,
+                enteredPostcode: this.address?.postcode,
+                location: this.$cookies.get('je-location'),
+                locationUk: this.$cookies.get('je-location-uk'),
+                changedFields: this.changedFields,
+                isPostcodeChanged: this.changedFields.includes('addressPostcode')
+            };
+
             this.logInvoker({
                 message: 'Checkout Validation Error',
-                data: { ...this.eventData, validationState },
+                data: { ...expandedData, validationState },
                 logMethod: this.$logger.logWarn
             });
         },
--- a/packages/components/organisms/f-checkout/CHANGELOG.md
+++ b/packages/components/organisms/f-checkout/CHANGELOG.md
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.173.3
+------------------------------
+*August 24, 2021*
+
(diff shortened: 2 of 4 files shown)


SachaZvetelman (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -988,9 +992,18 @@ export default {
...
+            const expandedData = {
+                ...this.eventData,
+                enteredPostcode: this.address?.postcode,
+                location: this.$cookies.get('je-location'),
+                locationUk: this.$cookies.get('je-location-uk'),
+                changedFields: this.changedFields,
+                isPostcodeChanged: this.changedFields.includes('addressPostcode')
+            };
+
             this.logInvoker({
```
charliearlie, line 1004: I believe there's a test for this which may need updating. Including mocking cookies.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -988,9 +992,18 @@ export default {
                 error: ANALYTICS_ERROR_CODE_INVALID_MODEL_STATE
             });
```
charliearlie, line 993: Should we remove the invalid model state log as well @SachaZvetelman ?

packages/components/organisms/f-checkout/src/components/Checkout.vue
charliearlie: `checkout.test.js` from line 3452 🙂

packages/components/organisms/f-checkout/src/components/Checkout.vue
SachaZvetelman: That's a really good question. I'll need to confirm with Jack. Maybe leave it for the time being and we can remove it later so this PR doesn't get delayed.

charliearlie: Yep, I just mentioned it as it was visible here! I'll take a note to remember to follow up.

BanksySan: I'd rather keep the PR focused, I'm quite happy to create a new one though is there's tidy-up outside of this.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "root",
-  "version": "3.53.1",
```
SachaZvetelman, line 3: Do we need to bump this now when we modify one of our components? I didn't know.

package.json
```diff
@@ -1,6 +1,6 @@
...
+  "version": "3.53.2",
```
SachaZvetelman: Do we need to bump this now when we modify one of our components? I didn't know.

CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v3.53.2
+------------------------------
+*August 24, 2021*
+
+### Added
+- f-checkout logging items to aid debugging conversion drop.
+
```
SachaZvetelman: Same here.

package.json
BanksySan: I assumed do.  It's probably sensible to increment the minor version as it's a different codebase but not a change in functionality.

SachaZvetelman: We only bump the version and update the CHANGELOG of the root when we modify something at the root level. If we only modify components, we bump and update the CHANGELOGs of said components. So in this case I think we need to revert this change.

charliearlie (approved):

tommcclean (approved):

PR #1215, opened 2021-08-24
f-button@v1.11.0 - Allow for conditional nested slot content.

DamianMullins:
### Added
- `hasNestedContent` computed property that checks whether button content should be nested inside span tags.
- Secondary default slot to be used when content is not nested.
- Unit tests for `hasNestedContent` computed property.

### Changed
- Wrap nested content only when necessary.
- Use Vue object syntax to add optional classes.

### DIFF
--- a/packages/components/atoms/f-button/src/components/Button.vue
+++ b/packages/components/atoms/f-button/src/components/Button.vue
@@ -3,11 +3,12 @@
         :is="componentType"
         :class="[
             $style['o-btn'],
-            (isIcon ? $style['o-btn--icon'] : ''),
             $style[`o-btn--${buttonType}`],
-            $style[`o-btn--size${buttonSizeClassname}`],
-            (isFullWidth ? $style['o-btn--fullWidth'] : ''),
-            (isLoading ? $style['o-btn--loading'] : '')
+            $style[`o-btn--size${buttonSizeClassname}`], {
+                [$style['o-btn--icon']]: isIcon,
+                [$style['o-btn--fullWidth']]: isFullWidth,
+                [$style['o-btn--loading']]: isLoading
+            }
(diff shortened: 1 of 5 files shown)


ashleynolan (approved):

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -3,44 +3,51 @@
...
+                    v-if="hasTrailingIcon"
+                    :class="[$style['o-btn-icon'], $style['o-btn-icon--trailing']]"
+                    data-test-id="button-trailing-icon">
+                    <slot name='trailing-icon' />
+                </span>
             </span>
+        </template>
 
+        <template v-else>
             <slot />
```
dandel10n, line 49: I think we still want `o-button-content` class to be applied on a wrapper around the slot even if `hasNestedContent === false`

DamianMullins: To keep the content centered? I guess I could use the same computed property to apply the `o-button-content` class to the button itself?

For a bit of context, the issue I'm aiming to resolve here is that on the menu we need to centre content and the nested spans would mean that we have to nest 3-4 levels deep and we'd like to avoid that if possible.

DamianMullins: Checking this, the default styles appear to centre-align content without that class applied. Also, none of the visual tests changed so I assumed the changes wouldn't break any existing layouts?

ashleynolan: I think it _should_ be ok without the class – if the visual regression tests aren't failing then should be ok.

dandel10n (approved):

PR #1220, opened 2021-08-26
f-services@1.13.0 - Update UK postcode validation to match ConsumerWeb

charliearlie:
As part of the new Checkout, we have noticed that postcodes are not accepted if the postcode contains whitespace. This is to rectify that by using the regex used in ConsumerWeb.

### DIFF
--- a/packages/components/organisms/f-checkout/CHANGELOG.md
+++ b/packages/components/organisms/f-checkout/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.175.0
+*August 25, 2021*
+
+### Changed
+- Map checkout update request now trims whitespace from postcode
+
+
+
 v0.174.0
 *August 25, 2021*
 
--- a/packages/components/organisms/f-checkout/CHANGELOG.md
+++ b/packages/components/organisms/f-checkout/CHANGELOG.md
@@ -11,7 +11,6 @@ v0.175.0
 - Map checkout update request now trims whitespace from postcode
 
 
-
 v0.174.0
 *August 25, 2021*
 
--- a/packages/components/organisms/f-checkout/CHANGELOG.md
+++ b/packages/components/organisms/f-checkout/CHANGELOG.md
@@ -4,13 +4,6 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-v0.175.0
-*August 25, 2021*
-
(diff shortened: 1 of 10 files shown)


packages/services/f-services/src/validations.js
```diff
@@ -31,7 +31,7 @@ const getFormValidationState = $v => {
 
 const POSTCODE_REGEX = {
     // https://stackoverflow.com/questions/164979/uk-postcode-regex-comprehensive#164994
```
SachaZvetelman: This comment is no longer relevant, I believe.

packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
 
 
+v0.175.0
+*August 25, 2021*
+
+### Changed
+- Map checkout update request now trims whitespace from postcode
+
+
+
```
SachaZvetelman: Super nit: one too many :D

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -384,6 +384,9 @@ export default {
 
         updateAddressDetails ({ commit, dispatch }, payload) {
             const [field] = Object.keys(payload);
+            Object.keys(payload).forEach(k => {
+                payload[k] = typeof payload[k] === 'string' ? payload[k].trim() : payload[k];
```
SachaZvetelman: I might be wrong, but aren't these `string` every time?
Do we want to trim every bit of the address? (I assume "yes" is the answer).

```
const [field, value] = Object.keys(payload);
...
...
commit(UPDATE_FULFILMENT_ADDRESS, value.trim());
```

charliearlie: I thought we could just trim the postcode but I guess there's no harm in trimming all lines?

And I was just being overly cautious because I know some tenants have fully numerical postcodes. Happy to change though, because they will all be stored in strings. I just have trust issues with JavaScript.

SachaZvetelman: As long as we're passing all these fields as strings (which we should), I think we should be safe.

SachaZvetelman: And yeah, let's trim all the address fields.

tommcclean: What about this https://vuejs.org/v2/guide/forms.html#trim

charliearlie: Would that suggestion work while using f-form-field @tommcclean? Because the value is set through an action?

JoshuaNg2332 (dismissed):

DamianMullins (approved):

kevinrodrigues (approved):

PR #1224, opened 2021-08-27
f-cookie-banner@0.24.0 - Fix cookie position.

kevinrodrigues:
Moved method to computed so the banner can be repositioned when the consuming app loads.

###Before:
Screenshot 2021-08-27 at 09 36 16

### After:
Screenshot 2021-08-27 at 09 36 27

## Browsers Tested

- [x] Chrome (latest)

### DIFF
--- a/packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
+++ b/packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
@@ -149,7 +149,7 @@ export default {
         const copy = localeConfig.messages;
         const consentCookieName = 'je-cookieConsent';
         const legacyConsentCookieName = 'je-banner_cookie';
-        const reopenLinkToBottom = this.isBodyHeightLessThanWindowHeight();
+        const reopenLinkToBottom = this.isBodyHeightLessThanWindowHeight;
 
         return {
             config: { ...localeConfig },
@@ -170,6 +170,17 @@ export default {
          */
         legacyBanner () {
             return this.shouldShowLegacyBanner === null ? this.config.displayLegacy : this.shouldShowLegacyBanner;
+        },
+
+        /**
+         * Check to see if we need to absolute position reopen link.
+         * * @returns {Boolean}
+         */
+        isBodyHeightLessThanWindowHeight () {
+            if (typeof window === 'object') {
+                return window.innerHeight - document.body.offsetHeight > 0;
+            }
+            return false;
         }
     },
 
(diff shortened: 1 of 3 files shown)


AlixRichards (commented):
Looks good - are there any tests that need to be added/amended for this?

SachaZvetelman:
Maybe a stupid question, but how does this behave if the user resizes their screen? Does it work well? Does it trigger the computed prop recalculation thousands of times?

kevinrodrigues:
> Looks good - are there any tests that need to be added/amended for this?

There's no tests previously, will add them tho.

kevinrodrigues:
> Maybe a stupid question, but how does this behave if the user resizes their screen? Does it work well? Does it trigger the computed prop recalculation thousands of times?

This picks up the window inner height property so it won't trigger again unless something is added / loaded to the DOM again. The window resize event would be different, that would then trigger multiple times without a throttle in place.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -170,6 +170,17 @@ export default {
...
         legacyBanner () {
             return this.shouldShowLegacyBanner === null ? this.config.displayLegacy : this.shouldShowLegacyBanner;
+        },
+
+        /**
+         * Check to see if we need to absolute position reopen link.
+         * * @returns {Boolean}
+         */
+        isBodyHeightLessThanWindowHeight () {
+            if (typeof window === 'object') {
```
DamianMullins, line 180: Will this ever be recalculated? As far as I'm aware, as the `window` & `document` objects aren't reactive there is nothing in here to trigger a re-calc?

kevinrodrigues: Do you mean if the page grew dynamically? Seems to work ok in checkout, the problem we had was that the page whilst loading would inject the cookie banner first in the wrong position, with this change it now recalculates the location.

kevinrodrigues:
> Looks good - are there any tests that need to be added/amended for this?

@AlixRichards I will add tests in a separate PR as this issue is currently impacting conversion we believe so we want to get this related asap.

AlixRichards (approved):

SachaZvetelman (approved):
As long as you've tested resizing (both the window and adding/removing components from the DOM dynamically), I'm happy with this 😄

kevinrodrigues:
> > Looks good - are there any tests that need to be added/amended for this?
> 
> @AlixRichards I will add tests in a separate PR as this issue is currently impacting conversion we believe so we want to get this related asap.

https://github.com/justeat/fozzie-components/pull/1233

PR #1225, opened 2021-08-27
Dedupe component type & name #trivial

dbernar1:
The component name and type was explicitly being passed to the Page object constructor, but it was always the same values for a given component.

I've pulled out constructors for each component which provide the component type and name so that individual tests do not have to.

Thank you for the review.

p.s. any feedback is welcome, it's my first PR here.


dbernar1:
You can review this PR commit-by-commit most easily probably. Each commit contains the change for a single component.

dbernar1:
If possible, IMO, it would be best if we did not perform any of the actions suggested by the Danger bear. 

Since the changes are only for tests, I do not perceive any benefit to a version bump. A changelog entry does not seem to make sense without a version bump. The PR touches several components, so the PR title convention does not fit.

We could do it if the tool requires it, but I don't see it as providing a benefit. What do you think?

Thank you.

siggerzz (dismissed):

siggerzz:
Thanks for this!

xander-marjoram (dismissed):

DamianMullins (dismissed):

DamianMullins (dismissed):

dandel10n (approved):

xander-marjoram (approved):

PR #1232, opened 2021-09-01
f-checkout@0.177.0 - Switch geolocation warning to error

charliearlie:
### Changed
- Gelocation warning to error for more details in Sentry


simonsandersje (dismissed):

packages/components/organisms/f-checkout/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.177.0
+------------------------------
+*September 1, 2021*
+
+### Changed
+- Gelocation warning to error for more details in Sentry
+
```
kevinrodrigues, line 13: Minor, should be an extra line break here :)

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -372,9 +372,11 @@ export default {
             }
 
             if (!addressCoords && state.authToken) {
-                const { data } = await addressGeocodingApi.getGeoLocation(url, postData, timeout, state);
+                const response = await addressGeocodingApi.getGeoLocation(url, postData, timeout, state);
```
DamianMullins: You could still destructure here and use `data` in the check below?

charliearlie: The error we're getting is 
```
Invalid attempt to destructure non-iterable instance.
In order to be iterable, non-array objects must have a [Symbol.iterator]() method.
``` 

so just trying to avoid that and confirm that it's the response from the server causing it. Although, I'm fairly sure axios always returns the data object but not 100% sure on that.

packages/components/organisms/f-checkout/CHANGELOG.md
charliearlie: Added

kevinrodrigues (dismissed):

packages/components/organisms/f-checkout/src/store/checkout.module.js
DamianMullins: And it's failing on this line? I'd expect to see that error if you were restructuring using square brackets ().

Looks to me like [it'd be the mutation throwing that error](https://github.com/justeat/fozzie-components/blob/b313ca208377eb4da9498183cc6cb1a0a3e93435/packages/components/organisms/f-checkout/src/store/checkout.module.js#L566) as it's attempting to destrucure the lat/lng values — if you pass an invalid value it'll throw.

In any case, adding the `if` guard here should fix it, you could put the `data` restructure back and just do the check against `data?.geometry`?

DamianMullins: You could also move the guard into the mutation itself and only destructure the lat/lngs if the payload has a value. This could be a better approach as the mutation is called in a couple of places.

DamianMullins: One more thing, you can probably write a unit test for the mutation that will throw that non-iterable error in its current state.

charliearlie: Thanks @DamianMullins - for some reason I didn't even consider the mutation. I will update the PR now.

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -561,11 +561,14 @@ export default {
             state.orderId = orderId;
         },
 
-        [UPDATE_GEO_LOCATION]: (state, [lng, lat]) => {
-            state.geolocation = {
-                latitude: lat,
-                longitude: lng
-            };
+        [UPDATE_GEO_LOCATION]: (state, coords) => {
```
DamianMullins, line 564: Might be worth writing some unit tests to verify this new behaviour?

charliearlie: There is one test to see if the state is not updated if the coordinates are not an array which I assume covers it?

DamianMullins (approved):

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -374,7 +374,7 @@ export default {
             if (!addressCoords && state.authToken) {
                 const { data } = await addressGeocodingApi.getGeoLocation(url, postData, timeout, state);
 
-                commit(UPDATE_GEO_LOCATION, data.geometry.coordinates);
+                commit(UPDATE_GEO_LOCATION, data?.geometry.coordinates);
```
BanksySan, line 377: Could this throw a null reference exception?

BanksySan (changes requested):

BanksySan (approved):

PR #1243, opened 2021-09-03
f-consumer-oidc@1.2.0 - Add vite to resolve correct build output.

kevinrodrigues:
Convert to `Vite` in-order to fix build output which was previously outputting the wrong file.

### Dist output after:

Screenshot 2021-09-03 at 13 08 37

### Todo:

Publish & test in coreweb.


packages/services/f-consumer-oidc/webpack.config.js
```diff
@@ -1,37 +0,0 @@
...
-        globalObject: 'this'
-    },
-    optimization: {
-        minimize: true,
-        minimizer: [new UglifyJsPlugin({
-            include: /\.min\.js$/
-        })]
-    },
-    externals: {
-        'js-cookie': 'js-cookie'
```
DamianMullins, line 34: Do you need to do anything in Vite to handle externalising peer deps?

kevinrodrigues: Ah that's a good catch! I'll check. Thanks

kevinrodrigues: Should be fixed now 👍

DamianMullins (approved):

simonsandersje (approved):

PR #1245, opened 2021-09-06
f-checkout@0.180.0 - Add Administrative Area input for AU

charliearlie:
### Added
- `administrativeArea` input for AU


packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -135,11 +163,19 @@ export default {
...
+
         /*
         * Checks that postcode is a valid postcode and that the field is not empty.
         */
         isAddressPostcodeValid () {
             return (!this.$v[VALIDATIONS.address].postcode.$dirty || this.$v[VALIDATIONS.address].postcode.isValidPostcode) && !this.isAddressPostcodeEmpty;
+        },
+
+        shouldShowAdministrativeArea () {
+            return this.tenant === 'au';
```
zlatinivanov-je: Should this include NZ

zlatinivanov-je: When I saw `State` mentioned below realised what that is! Assuming comment irrelevant, resolving.

zlatinivanov-je (dismissed):

packages/components/organisms/f-checkout/src/components/Address.vue
```diff
@@ -135,11 +163,19 @@ export default {
...
+        },
+
         /*
         * Checks that postcode is a valid postcode and that the field is not empty.
         */
         isAddressPostcodeValid () {
             return (!this.$v[VALIDATIONS.address].postcode.$dirty || this.$v[VALIDATIONS.address].postcode.isValidPostcode) && !this.isAddressPostcodeEmpty;
+        },
+
+        shouldShowAdministrativeArea () {
```
FayeCarter: Is it worth moving this into checkout and passing it through as a prop. Then the validations bit [here](https://github.com/justeat/fozzie-components/pull/1245/files#diff-3f7bda367e6ae2f5c141559c1aec5722d6606ee4d895f7d44e9837c20f7efb3cR1140) can use it too?

charliearlie: Yeah, nothing wrong with that. I'll update.

packages/components/organisms/f-checkout/src/components/_tests/Address.test.js
```diff
@@ -30,7 +34,9 @@ const $v = {
 };
 
 describe('Address', () => {
-    const propsData = {};
+    const propsData = {
+        tenant: 'au'
```
BanksySan: Should we be including `nz` as well?

packages/components/organisms/f-checkout/src/store/tests/__snapshots__/checkout.module.test.js.snap
```diff
@@ -3,6 +3,7 @@
 exports[`CheckoutModule mutations :: UPDATE_STATE :: should update state with delivery response. 1`] = `
 Object {
   "address": Object {
+    "administrativeArea": undefined,
```
BanksySan, line 6: Why do we instantiate something as `undefined`?

packages/components/organisms/f-checkout/test/component/f-checkout-delivery.component.desktop.spec.js
```diff
@@ -5,30 +5,60 @@ const Checkout = require('../../test-utils/component-objects/f-checkout.componen
...
+            expect(checkout.isPostcodeTypeErrorDisplayed()).toBe(false);
+        });
     });
 
-    it('should enable a user to submit a postcode with correct characters', () => {
-        // Arrange
-        const addressInfo = {
-            postcode: 'AR51 1AA'
-        };
+    describe('au tenant', () => {
```
BanksySan, line 37: Do we want to support `nz` too?

BanksySan (changes requested):

packages/components/organisms/f-checkout/src/components/_tests/Address.test.js
charliearlie: Nope. Only AU has a required state field.

packages/components/organisms/f-checkout/test/component/f-checkout-delivery.component.desktop.spec.js
charliearlie: I will have to double check but as mentioned above, AU is the only country that requires the administrative Area field on the API

FayeCarter (approved):

packages/components/organisms/f-checkout/src/store/tests/__snapshots__/checkout.module.test.js.snap
charliearlie: I think because it doesn't exist so the snapshot needs to do something with it. The administrativeArea will only exist if the user inputs it.

BanksySan (approved):

DamianMullins (approved):

PR #1248, opened 2021-09-07
f-analytics@0.13.0 - Read pageData.conversationid from cookie

AndyMayJE:
### Changed
- Read pageData.conversationId from x-je-conversation cookie
- Updated analytics.mapper functions to return new objects

### DIFF
--- a/packages/services/f-analytics/src/services/_tests/analytics.service.test.js
+++ b/packages/services/f-analytics/src/services/_tests/analytics.service.test.js
@@ -348,20 +348,20 @@ describe('Analytic Service ::', () => {
             }
         );
 
-        it('should set conversationId and requestId values when provided', () => {
+        it('should set requestId value when provided', () => {
             // Arrange
             const expected = {
                 ...defaultState.pageData,
                 group: 'test-feature-name',
                 name: 'test-page-name',
-                conversationId: '460cc3a8-83f7-4e80-bb46-c8a69967f249',
+                conversationId: undefined,
                 requestId: '6cbe6509-9122-4e66-a90a-cc483c34282e',
                 orientation: 'Landscape',
                 display: 'mid'
             };
 
             // Act
-            service.pushPageData({ pageName: expected.name, conversationId: expected.conversationId, requestId: expected.requestId });
+            service.pushPageData({ pageName: expected.name, requestId: expected.requestId });
 
             // Assert
             expect(windowsPushSpy).toHaveBeenCalledWith({ pageData: { ...expected } });
@@ -399,6 +399,29 @@ describe('Analytic Service ::', () => {
 
             // Act
             service.pushPageData({ pageName: expected.name, customFields: { custom1: 'one', custom2: 'two' } });
+
+            // Assert
+            expect(windowsPushSpy).toHaveBeenCalledWith({ pageData: { ...expected } });
+        });
+
+        it('should set the conversationId value when x-je-conversation cookie exists', () => {
+            // Arrange
+            const expected = {
+                ...defaultState.pageData,
+                group: 'test-feature-name',
+                name: 'test-page-name',
+                conversationId: 'f0740341-4369-437d-bcce-735a71ee5b78',
+                requestId: undefined,
+                orientation: 'Landscape',
+                display: 'mid'
+            };
+
+            get = jest.fn();
+            when(get).calledWith('x-je-conversation').mockReturnValue(expected.conversationId);
+            Cookies.mockImplementation(() => ({ get }));
+
+            // Act
+            service.pushPageData({ pageName: expected.name });
 
             // Assert
             expect(windowsPushSpy).toHaveBeenCalledWith({ pageData: { ...expected } });
--- a/packages/services/f-analytics/src/services/_tests/analytics.service.test.js
+++ b/packages/services/f-analytics/src/services/_tests/analytics.service.test.js
@@ -353,15 +353,13 @@ describe('Analytic Service ::', () => {
             const expected = {
                 ...defaultState.pageData,
                 group: 'test-feature-name',
-                name: 'test-page-name',
-                conversationId: undefined,
                 requestId: '6cbe6509-9122-4e66-a90a-cc483c34282e',
                 orientation: 'Landscape',
                 display: 'mid'
             };
 
             // Act
-            service.pushPageData({ pageName: expected.name, requestId: expected.requestId });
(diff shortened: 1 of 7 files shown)


oliversweb (dismissed):

packages/services/f-analytics/src/services/_tests/analytics.service.test.js
```diff
@@ -348,20 +348,20 @@ describe('Analytic Service ::', () => {
             }
         );
 
-        it('should set conversationId and requestId values when provided', () => {
+        it('should set requestId value when provided', () => {
             // Arrange
             const expected = {
                 ...defaultState.pageData,
                 group: 'test-feature-name',
```
xander-marjoram: Just wondering if it's necessary to pass most of these values in. Could they not just be the defaults, and we only override the values that are pertinent to this particular test?

packages/services/f-analytics/src/services/_tests/analytics.service.test.js
```diff
@@ -348,20 +348,20 @@ describe('Analytic Service ::', () => {
...
                 name: 'test-page-name',
-                conversationId: '460cc3a8-83f7-4e80-bb46-c8a69967f249',
+                conversationId: undefined,
```
oliversweb: I think you can remove this line as the `defaultState.pageData` sets the convo id to undefined 👍

packages/services/f-analytics/src/services/analytics.mapper.js
```diff
@@ -159,20 +159,21 @@ export const mapUserData = ({ userData, authToken, req } = {}) => {
...
-    httpStatusCode
+    httpStatusCode,
+    req
 } = {}) => {
     pageData.group = featureName;
     if (pageName) pageData.name = pageName;
+    const conversationId = getCookie('x-je-conversation', req);
     if (conversationId) pageData.conversationId = conversationId;
     if (requestId) pageData.requestId = requestId;
     if (httpStatusCode) pageData.httpStatusCode = httpStatusCode;
```
xander-marjoram: I know it's not from this PR, but small nit on line 180, `diplay`

xander-marjoram: Also not a massive fan of this method mutating the input object rather than returning a new object, like
```javascript
const conversationId = getCookie('x-je-conversation', req);

return {
    ...pageData,
    group: featureName,
    conversationId: conversationId || pageData.conversationId,
    ... etc, etc
}
```

packages/services/f-analytics/src/services/_tests/analytics.service.test.js
xander-marjoram: I did think this too, but I thought the fact it was explicitly undefined might make the tests a bit easier to understand.

packages/services/f-analytics/src/services/_tests/analytics.service.test.js
```diff
@@ -409,19 +407,17 @@ describe('Analytic Service ::', () => {
             const expected = {
                 ...defaultState.pageData,
                 group: 'test-feature-name',
-                name: 'test-page-name',
                 conversationId: 'f0740341-4369-437d-bcce-735a71ee5b78',
-                requestId: undefined,
-                orientation: 'Landscape',
-                display: 'mid'
+                display: 'mid',
```
xander-marjoram: And again here; are the `display` and `orientation` values relevant to this test?

packages/services/f-analytics/src/services/analytics.mapper.js
```diff
@@ -171,14 +171,18 @@ export const mapPageData = ({
...
     const conversationId = getCookie('x-je-conversation', req);
-    if (conversationId) pageData.conversationId = conversationId;
-    if (requestId) pageData.requestId = requestId;
-    if (httpStatusCode) pageData.httpStatusCode = httpStatusCode;
     const displaySize = getDisplaySize();
-    if (displaySize) pageData.display = displaySize;
     const orientation = getOrientation();
-    if (orientation) pageData.orientation = orientation;
+
+    return {
```
xander-marjoram: \*chefs kiss\*

packages/services/f-analytics/src/services/_tests/analytics.service.test.js
oliversweb: This is the default state `...defaultState.pageData` and is used elsewhere in various tests, so he could create a new expected state which is pretty much what he is doing here.

packages/services/f-analytics/src/services/_tests/analytics.service.test.js
oliversweb: But the test is only focusing on the `requestid` and has little to do with the convo id at this point.

packages/services/f-analytics/src/services/_tests/analytics.service.test.js
oliversweb: Maybe not but as we use the real helper to calc these values are we mock the display above we have to provide valid values and not rely on the default.

packages/services/f-analytics/src/services/analytics.mapper.js
oliversweb: @xander-marjoram I'm not a fan of overloading the `return`, now when stepping thru as a dev you have to exit the method before you can inspect the final value of `pageData`.

oliversweb: Plus now in this file we have one method passing by value and the others passing by ref which is not ideal

oliversweb (approved):

packages/services/f-analytics/src/services/_tests/analytics.service.test.js
```diff
@@ -331,77 +331,90 @@ describe('Analytic Service ::', () => {
...
             };
 
             // Act
-            service.pushPageData({ pageName: expected.name, httpStatusCode: expected.httpStatusCode });
+            service.pushPageData({ httpStatusCode: expected.httpStatusCode });
 
             // Assert
-            expect(windowsPushSpy).toHaveBeenCalledWith({ pageData: { ...expected } });
+            expect(windowsPushSpy).toHaveBeenCalledWith(expect.objectContaining({
+                pageData: expect.objectContaining({
```
xander-marjoram, line 377: Oh, this works even when it's nested? Neat.

xander-marjoram (approved):

oliversweb (approved):

PR #1249, opened 2021-09-07
f-braze-adapter@v4.0.0-beta.5 - adding in logging for better visibility

matthewhardern:
Adding updated logging to support a more refined logging for the adapter, while taking into account the new registry system, with logs now specifying the section of the adapter and a key indicating the consumer of the adapter. Tags have been added so that logs can be easily separated for display purposes.  variables are prefixed with logging so it obvious the keys are for logging purposes. I have also put a temp function in during the instantiation of the consumer class, because the dispatcher is not available during this time as the consumer has not been registered yet. I will look into a better solution at a later date.


packages/services/f-braze-adapter/src/services/BrazeConsumer.js
```diff
@@ -32,9 +34,17 @@ class BrazeConsumer {
         callbacks = {},
         interceptInAppMessages = {},
         interceptInAppMessageClickEvents = {},
-        loggerCallbacks = {},
-        customFilters = []
+        logger = (type, message, data) => console.log(type, message, data),
+        customFilters = [],
+        key: $key
```
minrwhite: Is it worth using `consumerKey` for this, given that there only seems to be one usage here, within a template string? The usage of `$key` looks a bit odd to me due to the recurrance of `$` in the tempate string, as well the potential confusion when mixing usage of `$` prefix for member variables.
```suggestion
        key: consumerKey
```

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
```diff
@@ -32,9 +34,17 @@ class BrazeConsumer {
...
     }) {
+        this.$logger = logger;
+        // key for logging
+        this.$consumerKey = `BrazeAdapter--consumer--${$key}`;
```
minrwhite:
```suggestion
        this.$consumerKey = `BrazeAdapter--consumer--${consumerKey}`;
```

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
```diff
@@ -109,11 +141,10 @@ class BrazeConsumer {
 
     /**
      * Function that loops over all loggerCallbacks and calls them
```
minrwhite: This description needs updating

minrwhite (commented):
A couple of comments. It would be good to have test assertions that the log messages are sent to the event stream as expected in different scenarios, if this is possible. I'd understand if you'd prefer to do this in a later PR though.

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
matthewhardern: Agreed see the confusion there, will make the change

matthewhardern:
> A couple of comments. It would be good to have test assertions that the log messages are sent to the event stream as expected in different scenarios, if this is possible. I'd understand if you'd prefer to do this in a later PR though.

I agree It could do with some coverage to test the logging output occurs as expected. We do currently have coverage for the dispatcher in areas, so I will follow up in a later PR if thats ok to complete coverage for the logs.

minrwhite (approved):

minrwhite (approved):

packages/services/f-braze-adapter/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v4.0.0-beta.5
+------------------------------
+*September 07, 2021*
+
+### Added
+- Finner detailed log events for greater error and event visibility
```
tommcclean: NIT
```suggestion
- Finer detailed log events for greater error and event visibility
```

PR #1258, opened 2021-09-10
f-searchbox@4.0.0-beta.34 - Move submit events later in flow.

DamianMullins:
### Changed
- Moved `clearAddressValue` later in `submit` function so that address values are not cleared before being set in the location cookie.
- Moved `onCustomSubmit` later in `submit` function so that the location cookie is set before consuming applications fire the `onSubmit` event.

### DIFF
--- a/packages/components/molecules/f-searchbox/src/components/Form.vue
+++ b/packages/components/molecules/f-searchbox/src/components/Form.vue
@@ -328,8 +328,6 @@ export default {
 
             if (this.isValid === true) {
                 this.setErrors([]);
-                this.clearAddressValue(this.shouldClearAddressOnValidSubmit);
-                onCustomSubmit(this.onSubmit, this.addressValue, e);
                 this.verifyHasPostcodeChanged();
 
                 if (this.service.isAutocompleteEnabled) {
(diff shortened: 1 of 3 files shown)


kevinrodrigues (approved):

packages/components/molecules/f-searchbox/src/components/Form.vue
```diff
@@ -328,8 +328,6 @@ export default {
 
             if (this.isValid === true) {
                 this.setErrors([]);
-                this.clearAddressValue(this.shouldClearAddressOnValidSubmit);
```
minrwhite, line 331: Should we duplicate this before the [`return false;`](https://github.com/justeat/fozzie-components/pull/1258/files#diff-65aa0d48592087454609523d61236ae899bd7efda26d0f5caf5323837dac1745R340) line, so that this behaviour remains consistent? Or is that not necessary?

DamianMullins: yeah, good shout — have updated.

minrwhite (approved):

PR #1266, opened 2021-09-13
f-development-context@v1.1.0 - + Storybook@v0.43.0 - Add in mock for `$gtm`

oliversweb:
### Added
- Add $gtm (aka f-analytics) into Context so Mono Repo components can do this.$gtm.pushEvent({ ... }});
___
### Added
- Extended f-development-context Package to support $gtm (aka f-analytics)


packages/services/f-development-context/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-development-context",
   "description": "A package to provide development context options for Storybook and WebDriver",
-  "version": "1.0.1",
+  "version": "1.2.0",
```
tommcclean: NIT: but probably 1.1 - otherwise looks good

tommcclean (approved):

packages/storybook/context/analytics.context.js
```diff
@@ -0,0 +1,80 @@
...
+export default () => {
+    const gtm = {
+        pushPlatformData: ({
+            featureName, locale, customFields
+        } = {}) => {
+            let data = { ...platformData, name: featureName, language: locale };
+            if (customFields) {
+                data = { ...data, ...customFields };
+            }
+            console.log('pushPlatformData', { platformData: { ...data } });
```
kevinrodrigues, line 52: Do we need these console.logs?

oliversweb: Sanity Check : Yes so you at least check in storybook that some form of output is being produced from you code (albeit not actually a GTM event)

kevinrodrigues (approved):

PR #1273, opened 2021-09-15
f-button@2.1.0 - Update f-button styles.

DamianMullins:
### Changed
- Moved `position` style to loading class in order to prevent allow consuming apps to control how the button is displayed on the page.


packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -275,9 +275,6 @@ $btn-icon-sizeXSmall-iconSize          : 18px;
 }
 
 .o-btn {
-    position: relative;
-    display: inline-block;
```
ashleynolan, line 279: I think line 279 and 280 may be used in the icon buttons for alignment – it doesn't look like there are any visual regression checks for this atm though, so may need to check storybook and make sure they don't look wonky without this.

DamianMullins: Yeah, from my testing the styles I removed didn't appear to have any effect on the icon positioning. I'm currently integrating the button component into menu and I might not actually need this change — I've adjusted the order that the CSS is imported so that I can better control the cascade.

PR #1277, opened 2021-09-16
f-feature-management-vue@v0.1.0 - Vue 2/3 wrapper for f-feature-management

gareththackeray:
Added f-feature-management-vue module. This allows reactive use of f-feature-management within Vue 2/3 (using vue-demi for x-compatibility).

Any thoughts welcome on the usefulness or otherwise of this module.


packages/services/f-feature-management-vue3/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "@vue/cli-plugin-eslint": "3.9.2",
+    "babel-loader": "8.1.0",
+    "eslint-plugin-json-schema-validator": "1.0.48",
+    "eslint-plugin-jsonc": "1.3.1",
+    "jest": "25.4.0",
+    "jest-teamcity": "1.7.0",
+    "vite": "2.4.4",
+    "@vue/reactivity": "3.2.11"
+  },
+  "peerDependencies": {
```
gareththackeray: I've made these peers in an attempt to keep bundle size down in case multiple components take on a dependency on this package.  I _think_ this makes sense?

packages/services/f-feature-management-vue3/vite.config.js
```diff
@@ -0,0 +1,11 @@
+import path from 'path';
+
+export default {
+    build: {
+        target: 'es2015',
+        lib: {
+            entry: path.resolve(__dirname, 'src/index.js'),
+            name: 'f-feature-management-vue3'
+        }
+    }
```
DamianMullins: I _think_ if you want to exclude your `peerDependencies` from the final bundle you'll need to add some additional config here, [like we have in the `f-consumer-oidc` package](https://github.com/justeat/fozzie-components/blob/5272c2f44f47b59be3c75e267729f89208db2058/packages/services/f-consumer-oidc/vite.config.js#L11-L21).

You should be able to compare the diff between the resulting bundle before and after making that change to check if the code is excluded.

DamianMullins:
@gareththackeray Is there going to be a vue 2 version alongside this? Also, is there a recommended way to test this all works as expected as I don't think we're using Vue 3 anywhere at JET right now?

Also, this raises another question, when we're building or converting to Vue 3 components in the future, do we want to go with explicit "Vue3" references in package names or just assume that one major version will be Vue 2 whilst another will be Vue 3. Or something else 🤔  I'll start a thread up if this hasn't been discussed already.

packages/services/f-feature-management-vue3/README.md
```diff
@@ -0,0 +1,25 @@
+# Feature Management for Vue 3
+
+This service wraps [f-feature-management](../f-feature-management) for use with Vue 3, allowing features to be queried in a reactive way (so they refresh when new config is loaded) if required.  Non-reactive querying is also possible.
+
+## Usage
+
+The `settings` object is documented in the main [f-feature-management](../f-feature-management) module. What this module adds is that the return "Feature Management" object has a property called `reactive`. If you call `featureManagement.reactive.getValue()` from a Vue 3 context, it will be reactive and will be re-evaluated when new config is loaded by the Feature Management SDK.
+
+Calling `getValue()` from the base "Feature Managment" object will not be reactive and the value will remain until you explicitly call it again.
```
xander-marjoram:
```suggestion
Calling `getValue()` from the base "Feature Management" object will not be reactive and the value will remain until you explicitly call it again.
```

packages/services/f-feature-management-vue3/package.json
xander-marjoram: I definitely agree that `f-feature-management` should be a peer dependency, but I'm not sure if having the reactivity package as one too is worthwhile, especially if it's already a `devDependency`.

gareththackeray:
> @gareththackeray Is there going to be a vue 2 version alongside this? 

Yep - I said as much in the PR description 😆  

>Also, is there a recommended way to test this all works as expected as I don't think we're using Vue 3 anywhere at JET right now?

Yeah it's an interesting question.  I'm sure it'd be possible to set up a fully automated test that checks the Vue reactivity. I will create a ticket for that.  

For a manual test, I have a working PoC.  I'll change it to use this lib once I've published it; for the time being it's basically just this code.

> Also, this raises another question, when we're building or converting to Vue 3 components in the future, do we want to go with explicit "Vue3" references in package names or just assume that one major version will be Vue 2 whilst another will be Vue 3. Or something else 🤔 I'll start a thread up if this hasn't been discussed already.

~It's not something I've discussed no.  I think this lib might be different to most components.  The Feature Management portal uses Vue3 but has lots of Fozzie components.  This one is special because Vue2 needs the `Vue.set()` function to ensure reactivity whereas Vue3 doesn't have that available (although I believe there is some sort of shim lib available).~ Edit: not true; FM is using Vue2; I was mistaken.

packages/services/f-feature-management-vue3/package.json
gareththackeray: I only made it a `devDependency` because the tests require it (and didn't work till I added it).  It is used at runtime so it can't only be a `devDependency`.

DamianMullins: I think `devDependencies` is correct for an npm package as it'll be included in the final bundle rather than used in the consuming application. 

Putting it in `dependencies` means that when you run `yarn install` in a consuming app, the dep will be downloaded but not necessarily used.

DamianMullins:
@gareththackeray I should really read the PR description first! 🤦🏻 

There is [vue-demi](Vue demi) which you might be able to use in order to offer Vue 2 & Vue 3 versions from a single package. I'd be interested to see if that would work.

Lastly, POC looks interested, though we probably don't want to link to our internal Github on our public repos!

packages/services/f-feature-management-vue3/package.json
DamianMullins: > I only made it a `devDependency` because the tests require it (and didn't work till I added it). It is used at runtime so it can't only be a `devDependency`.

Yeah, `peerDependencies` should go in `devDependencies` and then excluded in the build config so that we can run the code in isolation or in unit tests.

packages/services/f-feature-management-vue3/src/createFeatureManagementForVue3.js
```diff
@@ -0,0 +1,45 @@
...
+    const originalOnUpdated = settings.onUpdated;
+
+    const overriddenSettings = {
+        ...settings,
+        onUpdated: () => {
+            if (typeof originalOnUpdated === 'function') {
+                originalOnUpdated();
+            }
+
+            /* eslint-disable no-use-before-define */
```
xander-marjoram: Are these eslint comments necessary now?

gareththackeray: I couldn't work out a better way to do it TBH.

packages/services/f-feature-management-vue3/test/createFeatureManagementForVue3.test.js
```diff
@@ -0,0 +1,53 @@
...
+    });
+
+    it('calling reactive.getValue() should delegate to function on base', async () => {
+        //Arrange
+        const mockFm = createFeatureManagement();
+        mockFm.getValue.mockReturnValue('return value');
+        
+        //Act
+        const fm = await createFeatureManagementForVue3({});
+        const val = fm.getValue('parameter');
```
xander-marjoram: Should this be like this, given the description of the test?
```suggestion
        const val = fm.reactive.getValue('parameter');
```

packages/services/f-feature-management-vue3/src/createFeatureManagementForVue3.js
xander-marjoram: I mean it looks like the function _is_ defined before being used (I'm assuming that it wasn't when these comments were added), so I was wondering if these were still required

packages/services/f-feature-management-vue3/test/createFeatureManagementForVue3.test.js
gareththackeray: Ah yes. Hope it still works!

packages/services/f-feature-management-vue3/src/createFeatureManagementForVue3.js
gareththackeray: it's referring to `featureManagement`, which is defined lower down.

gareththackeray: Actually it was finding me a genuine bug 🤦 .  Have fixed now.

gareththackeray:
> @gareththackeray I should really read the PR description first! 🤦🏻
> 
> There is [vue-demi](Vue demi) which you might be able to use in order to offer Vue 2 & Vue 3 versions from a single package. I'd be interested to see if that would work.
> 
> Lastly, POC looks interested, though we probably don't want to link to our internal Github on our public repos!

@DamianMullins I've put vue-demi in now and got it working with the PoC. Thanks for that.

packages/services/f-feature-management-vue/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+The `settings` object is documented in the main [f-feature-management](../f-feature-management) module. What this module adds is that the return "Feature Management" object has a property called `reactive`. If you call `featureManagement.reactive.getValue()` from a Vue context, it will be reactive and will be re-evaluated when new config is loaded by the Feature Management SDK.
+
+Calling `getValue()` from the base "Feature Management" object will not be reactive and the value will remain until you explicitly call it again.
+
+So for example if you have:
+
+```javascript
+  data: () => {
+      featureManagement: createFeatureManagementForVue3({/*settings*/})
```
xander-marjoram:
```suggestion
      featureManagement: createFeatureManagementForVue({/*settings*/})
```

packages/services/f-feature-management-vue/README.md
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }
+```
+
+## vue-demi
+[vue-demi](https://github.com/vueuse/vue-demi) provides some shims around Vue 2 / 3 to help libs be compatible with either.
+
+For this library we are using the `set()` shim (only exists in Vue2).  There is a shim over `reactive()` for Vue2, which relies on the [composition-api](https://github.com/vuejs/composition-api) and converts the object to a Vue2 proxy (rather than creating a new proxy object, as `reactive()` does in Vue3).
+
+We are not making use of this however, and indeed have not made `@vue/compositionapi` a peer dependency.  The use of the Feature Management instance within a Vue2 app will do this for us, and as long we we use the Vue2 `set()` function all should be well.
```
xander-marjoram:
```suggestion
We are not making use of this however, and indeed have not made `@vue/compositionapi` a peer dependency.  The use of the Feature Management instance within a Vue2 app will do this for us, and as long we use the Vue2 `set()` function all should be well.
```

packages/services/f-feature-management-vue/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@justeat/f-feature-management-vue",
+  "version": "0.1.0",
+  "description": "Service for querying feature values provided by Feature Management, made reactive for Vue3.",
```
xander-marjoram: Maybe update to include something about Vue 2 as well

packages/services/f-feature-management-vue/src/createFeatureManagementForVue.js
```diff
@@ -0,0 +1,50 @@
...
+                originalOnUpdated();
+            }
+            setReactiveProperty(featureManagement);
+        }
+    };
+
+    featureManagement = await createFeatureManagement(overriddenSettings);
+
+    // if Vue2 is being used reactive will be null/undefined
+    if (reactive) {
```
xander-marjoram, line 41: So maybe asking the obvious... but is it invalid to try and configure the SDK in a reactive way for Vue 2? I didn't get that impression from the readme

gareththackeray: Vue 3 reactivity works by swapping out the original object for a proxy, whereas Vue 2 works by adding reactivity behaviour to the object.

So for Vue 3 it's important we get hold of the reactive proxy otherwise we'll be setting properties on the original object, which is of no consequence to Vue.  (Note: Vue tracks the generated proxies i.e. there is only one per object.)

With Vue 2 we simply need to call `Vue.set()` to set properties, and as long as everything is set up for reactivity on that object, changes will be reactive.  But the `reactive()` function is not a thing for Vue2.

Hope that makes sense?

DamianMullins: You could be ultra-safe here and use the `isVue2` boolean that `vue-demo` offers.

gareththackeray: Do you think that's safer?  For me it's saying "is there a `reactive` import? is so, use it".  Seems just as safe to me?

DamianMullins: For anyone not familiar with the differences between both frameworks it would be clearer. Personally, I had no idea that `reactive` would be null if using Vue 2 and would not have known if I wasn\t reviewing this PR!

gareththackeray: This is true, and is why I wrote the comment on the line above.

xander-marjoram (dismissed):

packages/services/f-feature-management-vue/.eslintrc.js
```diff
@@ -0,0 +1,56 @@
...
+    }
+  ],
+  rules: {
+    // json-schema-validator plugin does not play nicely with the vuejs-accessibility plugin.
+    // Github issue open here: https://github.com/ota-meshi/eslint-plugin-json-schema-validator/issues/76
+    // For now, disable all vuejs-accessibility rules as this service does not contain any vue dependencies.
+    //
+    // Add further rules whenever the following error is encountered:
+    //     `Error while loading rule 'vuejs-accessibility/*': context.parserServices.defineTemplateBodyVisitor is not a function`
+    'vuejs-accessibility/alt-text': 'off',
```
xander-marjoram, line 21: These shouldn't be necessary for services, but if they came out of the template then we should update that from our side

packages/services/f-feature-management-vue3/vite.config.js
gareththackeray: Done and checked filesize.

xander-marjoram (approved):

DamianMullins (approved):

PR #1282, opened 2021-09-20
f-content-cards@v6.0.0-beta.1 - Adding logging

matthewhardern:
This PR adds logging via the `this.$logger` to content cards component. It covers the instantiation of the braze adapter and whether its successful or not. Covering the number of cards returned to the user and time since mount to returning these cards (later to use f-statistics). Also added is tags as a prop so consuming page can assign itself to the outputted logs. I have also removed logger callback and tests as the braze adapter now just gets passed the `this.$logger` so test coverage is now covered by `this.$logger`.  Added mocks to the existing tests to allow logger to work.


packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -233,11 +250,34 @@ export default {
...
-                    loggerCallbacks: {
-                        logger: this.handleLogging(this.$logger)
-                    }
+                    logger: this.$logger,
+                    tags: this.tags
                 });
+
+                this.$logger.logInfo(
+                    `Content Cards (setupMetadata) - BrazeAdapter successfully initialised. Key: (${this.loggingKey})`,
+                    this.$store,
```
zlatinivanov-je, line 263: Not sure how much data is in `this.$store` and how/where it gets sent, just a note on being sensible.

zlatinivanov-je: Also there seems to be some repetition in the log data, have you considered extracting that to reduce repetition?

zlatinivanov-je (commented):
Just a couple of considerations. Let me know what you think

packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -233,11 +250,34 @@ export default {
...
+                    {
+                        data: {
+                            key: this.loggingKey
+                        },
+                        tags: this.tags
```
minrwhite: Can we keep the order of keys in the extra data argument consistent between invocations to the various `logger.log*` methods, please? Either change it here as so, to match the others:
```suggestion
                        tags: this.tags,
                        data: {
                            key: this.loggingKey
                        }
```
or ideally, make all invocations have alphabetical order of extra log data - though can accept this is changing every other invocation so happy if you just want to do this one.

minrwhite (commented):
One minor comment but overall looks good

minrwhite:
Also - should CI checks be running against this?

packages/components/organisms/f-content-cards/src/components/ContentCards.js
matthewhardern: Thanks for the comments, on point 1 its a requirement of the logger plugin to pass the whole store object, so I have to do that. point 2 agree I will make the abstraction.

matthewhardern:
> Also - should CI checks be running against this?

its hidden under all checks have passed

minrwhite:
> its hidden under all checks have passed

🤦🏽  Thanks!

packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -320,12 +318,11 @@ export default {
                         `Content Cards (metadataContentCards) - Successfully received content cards. Key: (${this.loggingKey})`,
                         this.$store,
                         {
-                            tags: this.tags,
+                            ...this.defaultLoggingData,
+                            Count: cards.length,
                             data: {
-                                key: this.loggingKey,
                                 source: CARDSOURCE_METADATA
-                            },
```
minrwhite: Does this merge cleanly into the destructured `this.defaultLoggingData`, or simply replace the `data` value with one that doesn't include the key? I'd suspect it does the latter, unless you do something like:
```suggestion
                            data: {
                                ...this.defaultLoggingData.data,
                                source: CARDSOURCE_METADATA
```

matthewhardern: indeed my mistake

matthewhardern: fixed :)

minrwhite (approved):

zlatinivanov-je (approved):

PR #1286, opened 2021-09-21
f-cookie-banner@v2.2.0 - add tab loop

simonsandersje:
Add a tab loop to keep keyboard users within the modal

## Browsers Tested

- [x] Chrome (latest)

### DIFF
--- a/packages/components/organisms/f-cookie-banner/CHANGELOG.md
+++ b/packages/components/organisms/f-cookie-banner/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v2.2.0
+------------------------------
+*September 21, 2021*
+
+### Added
+- Banner now has tab loop to keep keyboard navigation users within the banner
+
+
 v2.1.0
 ------------------------------
 *September 17, 2021*
--- a/packages/components/organisms/f-cookie-banner/package.json
+++ b/packages/components/organisms/f-cookie-banner/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-cookie-banner",
   "description": "Fozzie Cookie Banner – Cookie Banner",
-  "version": "2.1.0",
+  "version": "2.2.0",
   "main": "dist/f-cookie-banner.umd.min.js",
   "maxBundleSize": "30kB",
   "files": [
--- a/packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
+++ b/packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
@@ -51,7 +51,9 @@
                     </div>
                 </div>
 
-                <div :class="$style['c-cookieBanner-cta']">
+                <div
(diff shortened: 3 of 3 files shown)


DamianMullins (commented):
As you're using the mega-modal it [should provide you the key-trapping feature automatically when the `open` method is called](https://github.com/justeat/fozzie-components/blob/9159b3a417db7efd9b5fa3577061e0302b2e0a1f/packages/components/molecules/f-mega-modal/src/components/MegaModal.vue#L212-L240). Seems that part isn't triggering for some reason though 🤔

simonsandersje:
@DamianMullins I saw that the tab loop functionality is available in mega modal, but as we don't use the open() method, the keyActions() method isn't called. Refactoring the cookie banner to use the open() method seemed like a lot of work to just get the key-trap functionality

kevinrodrigues:
Hi @DamianMullins @simonsandersje, I've taken a look at this bug ticket (picked it up from Simon to help out). I looked through the cookie banner code and there's quite a bit of logic that depends on the `shouldHideBanner` variable, which means in-order to make this work in the correct way, using `f-mega-modal`'s behaviour, we need to refactor `f-cookie-banner` quite a bit. My suggestion is to allow this PR to go through as a temporary fix for the accessibility issue we have and I will create a ticket so this is corrected, making `f-cookie-banner` use the in-built capabilities that `f-mega-modal` provides. Let me know your thoughts on that?

kevinrodrigues (approved):

zlatinivanov-je (approved):

zlatinivanov-je:
> Hi @DamianMullins @simonsandersje, I've taken a look at this bug ticket (picked it up from Simon to help out). I looked through the cookie banner code and there's quite a bit of logic that depends on the `shouldHideBanner` variable, which means in-order to make this work in the correct way, using `f-mega-modal`'s behaviour, we need to refactor `f-cookie-banner` quite a bit. My suggestion is to allow this PR to go through as a temporary fix for the accessibility issue we have and I will create a ticket so this is corrected, making `f-cookie-banner` use the in-built capabilities that `f-mega-modal` provides. Let me know your thoughts on that?

Shall we create an issue or record the described above somewhere so it could be picked up at some point

kevinrodrigues:
> > Hi @DamianMullins @simonsandersje, I've taken a look at this bug ticket (picked it up from Simon to help out). I looked through the cookie banner code and there's quite a bit of logic that depends on the `shouldHideBanner` variable, which means in-order to make this work in the correct way, using `f-mega-modal`'s behaviour, we need to refactor `f-cookie-banner` quite a bit. My suggestion is to allow this PR to go through as a temporary fix for the accessibility issue we have and I will create a ticket so this is corrected, making `f-cookie-banner` use the in-built capabilities that `f-mega-modal` provides. Let me know your thoughts on that?
> 
> Shall we create an issue or record the described above somewhere so it could be picked up at some point

Hi @zlatinivanov-je I have added some info to the original ticket (not sure if we can link internal tickets on a public repo) but can ping that to you :) hope that helps? In terms of the refactor ticket I will create that today and it will be refined by the web team hopefully next week.

DamianMullins:
@kevinrodrigues Yeah that sounds fine, I know we've looked at this functionality in the past (can't remember which component) and it was a little tricky.

PR #1287, opened 2021-09-21
f-analytics@v0.15.0 - Preserve state when registering module server & client side

oliversweb:
### Changed
- Always attempt to register the store but always preserve state, if present.
- Removed the redundant `groupName` prop from pushPageData().

### DIFF
--- a/packages/services/f-analytics/src/services/analytics.mapper.js
+++ b/packages/services/f-analytics/src/services/analytics.mapper.js
@@ -185,7 +185,7 @@ export const mapUserData = ({ userData, authToken, req } = {}) => {
  * Maps various static/computed variables to the PageData.
  *
  * @param {object} pageData - A reference to the current PageData instance
- * @param {string} featureName - The name of the feature
+ * @param {string} groupName - The name of the group
  * @param {string} pageName - The name of the page
  * @param {string} requestId - The current request Id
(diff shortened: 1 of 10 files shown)


tommcclean (approved):

FayeCarter (approved):

packages/services/f-analytics/src/services/analytics.mapper.js
```diff
@@ -103,13 +104,14 @@ const mapUserAgent = req => {
...
 
+    /* eslint-disable dot-notation */
     const mappedData = {
         ...platformData,
         jeUserPercentage: userPercent || platformData.jeUserPercentage,
-        environment: process.env.justEatEnvironment || platformData.environment,
-        version: process.env.FEATURE_VERSION || platformData.version,
-        instancePosition: process.env.INSTANCE_POSITION || platformData.instancePosition,
-        isPilot: process.env.IS_PILOT || platformData.isPilot
+        environment: process.env['justEatEnvironment'] || platformData.environment,
```
kevinrodrigues, line 111: What's the error out of interest?

oliversweb: No error just doesn't evaluate but 'bracket notation' works ... read up a bit and it seems 'bundling' can cause issues with 'dot notation'.

PR #1290, opened 2021-09-22
f-analytics@v0.17.0 - Removed redundant Plugin code

oliversweb:
### Changed
- Removed the plugin and moved logic to in to Service constructor, onus now on consumer to declare global in a local plugin


ashleynolan (dismissed):

packages/services/f-analytics/src/services/_tests/analytics.service.test.js
```diff
@@ -16,55 +17,256 @@ describe('Analytic Service ::', () => {
...
+        windowSpy.mockRestore();
+        storeDispatchSpy.mockRestore();
         jest.clearAllMocks();
     });
 
-    it('should instantiate a new instance', () => {
-        // Act
-        const instance = () => new AnalyticService(store, options);
+    describe('When creating a new instance', () => {
+        it('should instantiate a new instance', () => {
```
kevinrodrigues: Maybe this should read..
`should not throw error when instance is created with valid option types` ?

packages/services/f-analytics/src/services/_tests/analytics.service.test.js
```diff
@@ -16,55 +17,256 @@ describe('Analytic Service ::', () => {
...
     let windowsPushSpy;
     let get;
-
-    const mockWindow = ({ winWidth = 667, winHeight = 375 } = {}) => {
-        windowsPushSpy = jest.fn();
-        windowCopy = { ...window };
+    let registerStoreModuleSpy;
+
+    const mockWindow = ({
+        winWidth = 667,
```
kevinrodrigues, line 28: Do you need to create these? You could just pass the value in directly to line 45 & 46 instead? that way you dont need the spy or to return anything

kevinrodrigues (dismissed):

kevinrodrigues (approved):

ashleynolan (approved):

PR #1297, opened 2021-09-24
f-performance@0.1.0

ashleynolan:
Creation of the base package for f-performance

### DIFF
--- a/packages/services/f-performance/package.json
+++ b/packages/services/f-performance/package.json
@@ -1,0 +1,51 @@
+{
+  "name": "@justeat/f-performance",
+  "description": "Fozzie Performance - RUM performance metric collector",
+  "version": "0.1.0",
+  "main": "dist/f-performance.umd.min.js", 
+  "files": [
+    "dist" 
+  ],
+  "homepage": "https://github.com/justeat/fozzie-components/tree/master/packages/services/f-performance",
+  "contributors": [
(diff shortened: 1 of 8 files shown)


packages/services/f-performance/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "license": "Apache-2.0",
+  "engines": {
+    "node": ">=10.0.0"
+  },
+  "keywords": [
+    "fozzie"
+  ],
+  "scripts": {
+    "prepublishOnly": "yarn lint && yarn test && yarn build",
+    "lint": "vue-cli-service lint",
```
DamianMullins, line 29: Might be more of a thing for the generator project but maybe we should consider offering a Vite version of the npm packages and tasks, as plain JS projects don't really need Vue-CLI to handle things like building, linting etc - I'm thinking of projects like `f-consumer-oidc`, `f-http`, etc.

DamianMullins (dismissed):

sonnyprince (dismissed):

packages/services/f-performance/package.json
tommcclean: ^ yeah agreed, basically none of our services use Vue CLI anymore

ashleynolan: @tommcclean Are you ok to create an EE ticket for this? I'm not sure what needs updating to put the service build on Vite myself (can pull this in then in the next couple of weeks)

PR #1302, opened 2021-09-27
f-checkout@1.3.0 - Alter update checkout request for age verification

charliearlie:
### Added
- New function to map update checkout request

### Changed
- Map update checkout data before calling `handleUpdateCheckout`

### DIFF
--- a/packages/components/organisms/f-checkout/src/components/AgeVerification.vue
+++ b/packages/components/organisms/f-checkout/src/components/AgeVerification.vue
@@ -116,7 +116,8 @@ export default {
 
         months () {
             return new Array(12).fill(0).map((_, i) => {
-                const date = new Date(`${i + 1}/1`);
+                const year = new Date().getFullYear();
+                const date = new Date(year, i);
 
                 return {
                     text: date.toLocaleDateString(undefined, { month: 'long' }),
--- a/packages/components/organisms/f-checkout/src/components/AgeVerification.vue
+++ b/packages/components/organisms/f-checkout/src/components/AgeVerification.vue
@@ -116,8 +116,7 @@ export default {
 
         months () {
             return new Array(12).fill(0).map((_, i) => {
-                const year = new Date().getFullYear();
-                const date = new Date(year, i);
+                const date = new Date(new Date().getFullYear(), i);
 
                 return {
                     text: date.toLocaleDateString(undefined, { month: 'long' }),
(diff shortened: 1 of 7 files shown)


packages/components/organisms/f-checkout/src/components/AgeVerification.vue
```diff
@@ -116,7 +116,8 @@ export default {
 
         months () {
             return new Array(12).fill(0).map((_, i) => {
-                const date = new Date(`${i + 1}/1`);
+                const year = new Date().getFullYear();
```
dandel10n: Do you need this constant here?

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -1100,6 +1098,23 @@ export default {
                 await this.loadAvailableFulfilment();
                 this.availableFulfilmentTimesKey++;
             }
+        },
+
+        mapCheckoutUpdateRequest (ageVerificationOnly) {
```
dandel10n: Can we rename this function to avoid the confusion between  `mapCheckoutUpdateRequest` and `mapUpdateCheckoutRequest`

packages/components/organisms/f-checkout/src/components/AgeVerification.vue
charliearlie: I do not. Have removed and got this value in the instantiation of the date below.

packages/components/organisms/f-checkout/src/components/Checkout.vue
charliearlie: Changed. Naming variables/functions is the hardest part of the job 🙈

FayeCarter (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -709,6 +697,16 @@ export default {
             window.location.assign(`${this.paymentPageUrlPrefix}/${this.orderId}`);
         },
 
+        /**
+         * Call update checkout with only the user's DOB for age verification
+         * This is to avoid creating too many side effects with the original mapper for update checkout
+         */
+        async verifyCustomerAge () {
+            const data = this.mapCheckoutUpdateRequest(true);
```
zlatinivanov-je: The argument `true` here looks a bit odd... have you considered setting a default param on the method?

BanksySan (dismissed):

packages/components/organisms/f-checkout/src/components/Checkout.vue
charliearlie: I agree. I've changed it to an `options` object parameter

FayeCarter (approved):

BanksySan (approved):

PR #1304, opened 2021-09-29
fozzie-components@4.3.0 - Don't run visual regression tests for draft / "wip" PRs

xander-marjoram:
- Added conditional logic (via `visual-regression-preflight.js` script to determine whether or not to run the visual regression tests.
- Added `pipeline_trigger@v1` action which calls the CircleCI API with a token (a Github secret configured in the repo settings) and the branch name (e.g., `pull/1304/head`) to re-run the build.
  - This action is **manually bundled** by `@vercel/ncc` to include the dependencies inline. This is necessary because Github actions cannot pull repos from npm (apart from the official Github action packages, I believe), and the alternative was adding a `yarn install && yarn build` to the action which would make it take several minutes.
- Added a `label-watcher` workflow which calls the `pipeline_trigger` action (unless the PR is a Draft or it has the `wip` label applied).
- Re-runs the PR build when the `wip` label is removed from a PR, this time with the visual regression tests.

---

In my opinion, the downsides to this are:
- The `pipeline_trigger` action (if updated) needs to be manually re-built and the output file committed. This shouldn't be necessary very often.
- This could slow down the PR process, as removing the `wip` label will re-run the whole build which can take an hour or so. People may be inclined to remove it straight away.
- The automatic labeler actions don't seem to be working yet (perhaps this PR would need to be merged first?), so this may rely on PR authors adding the `wip` label themselves.


jamieomaguire (dismissed):

visual-regression-preflight.js
```diff
@@ -0,0 +1,19 @@
...
+    const json = await response.json();
+    const details = {
+        draft: json.draft,
+        labels: json.labels.map(l => l.name)
+    };
+
+
+    // If PR is draft or has given labels, skip visual regression tests to save credits
+    const shouldRunTests = !details.draft && !LABELS_TO_STOP.some(label => details.labels.includes(label));
+    console.log(shouldRunTests);
```
jamieomaguire, line 19: should this val be returned?

xander-marjoram: This PR isn't quite ready for review, but I'll answer anyway 😄 
This seems to be the way to "return" a value from a node script so that it can be read by the shell that runs it, see here:
https://github.com/justeat/fozzie-components/pull/1304/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R227

There may be better ways but this is just a spike atm

jamieomaguire: Ah cheers! I was being nosey ;)

ashleynolan (commented):
Would it be possible to add just the automatic label adding as a separate PR, to make sure this works potentially? I think checking this works before we merge everything else is pretty key for this workflow (as otherwise, we're relying on people manually doing this, which IMO won't happen)

Otherwise, I think it looks good. It would be nice to just run the Visual regression tests once the label is removed – will this be possible once we have CircleCI approved and have split out the tasks potentially?

xander-marjoram:
@ashleynolan That's a good idea, I'll make a separate PR with the labeler.

On your second point, I think it may be possible if the Visual Regression tests are pulled out into their own workflow ([see this article here](https://support.circleci.com/hc/en-us/articles/360050351292-How-to-trigger-a-workflow-via-CircleCI-API-v2)), although it would also require the initial yarn setup too, but I think we already have a cache set up so we can load them from there.

LTurns (commented):
This is awesome! 👏 I wonder if we could use a similar approach for retriggering the danger build - maybe with a title-watcher?

siggerzz (dismissed):
Great job! 🚀 

As mentioned by the comments, we can take this further once we have the CircleCI plan to split the build tasks into separate jobs and create new workflows that get triggered dependant on the state of the PR 👍🏼

.circleci/config.yml
```diff
@@ -223,9 +223,16 @@ commands:
...
-          command_name: "test:visual"
+      - run:
+          name: Run Percy visual regression tests for packages
+          command: |
+            val=$(node ./visual-regression-preflight.js)
+            if [ $val == "true" ];
+            then
+              yarn test:visual
+            else
+              echo Skipping visual regression tests
```
BanksySan: Could this message include why they are being skipped?  Something like `Skipping visual regression tests as PR is either draft or WIP.`

.github/wip-labeler.yml
```diff
@@ -0,0 +1,6 @@
+# Add 'wip' label to all new PRs
+wip:
```
BanksySan: This could be confusing if people aren't expecting `WIP` to be added by default.

.github/actions/pipeline_trigger@v1/dist/index.js
```diff
@@ -0,0 +1,13640 @@
+/******/ (() => { // webpackBootstrap
```
BanksySan, line 1: This file looks generally mental.  Is it meant to be checked in?

BanksySan (changes requested):

.github/actions/pipeline_trigger@v1/dist/index.js
xander-marjoram: Yes, although I agree it's not ideal. From PR description:

> - This action is manually bundled by `@vercel/ncc` to include the dependencies inline. This is necessary because Github actions cannot pull repos from npm (apart from the official Github action packages, I believe), and the alternative was adding a `yarn install && yarn build` to the action which would make it take several minutes.

.github/wip-labeler.yml
xander-marjoram: I suppose so, but it shouldn't take long to get used to.

ashleynolan: I guess once we've finalised the full workflow for this, we could add a section in the docs that explain the PR process and include what labels get added when opening a PR.

BanksySan (approved):

.github/wip-labeler.yml
xander-marjoram: Yeah, we could potentially add it to the PR template

ashleynolan (approved):

PR #1305, opened 2021-09-30
f-footer@v5.0.1 Add gender pay gap report link

simonsandersje:
v5.0.1
*September 30, 2021*

### Added
- Add link to "Gender Pay Gap 2019-2021" report for UK

### DIFF
--- a/packages/components/organisms/f-footer/src/tenants/en-GB.js
+++ b/packages/components/organisms/f-footer/src/tenants/en-GB.js
@@ -243,6 +243,13 @@ export default {
                     gtm: 'click_about_modern_slavery_statement'
                 },
                 {
+                    url: '/CmsAssets/media/Files/Common/legal/gender-pay-gap/gender-pay-gap-2019-2021',
+                    text: 'Gender Pay Gap 2019-2021',
+                    target: '_blank',
+                    rel: 'noopener',
+                    gtm: 'click_about_gender-pay-gap-2019-2021'
+                },
+                {
                     url: 'https://secure.ethicspoint.eu/domain/media/en/gui/106834/index.html',
                     text: 'Ethics hotline',
                     target: '_blank',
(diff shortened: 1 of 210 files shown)


packages/components/organisms/f-footer/src/tenants/en-GB.js
```diff
@@ -242,6 +242,13 @@ export default {
                     rel: 'noopener',
                     gtm: 'click_about_modern_slavery_statement'
                 },
+                {
+                    url: '/CmsAssets/media/Files/Common/legal/gender-pay-gap/gender-pay-gap-2019-2021',
```
BanksySan, line 246: Is the casing correct?  `Files`?

packages/components/organisms/f-footer/src/tenants/en-GB.js
```diff
@@ -242,6 +242,13 @@ export default {
...
+                    text: 'Gender Pay Gap 2019-2021',
+                    target: '_blank',
```
BanksySan, line 248: Could we use a named target instead of `_blank` so that multiple clicks doesn't open more than one window?

BanksySan (changes requested):

minrwhite (approved):

packages/components/organisms/f-footer/src/tenants/en-GB.js
simonsandersje: This is the link that legal have given us and the link works. Since the deadline is Monday, I don't think we have time to spend too long on clarifying with them

packages/components/organisms/f-footer/src/tenants/en-GB.js
simonsandersje: I'll raise a ticket to check whether we need to do this for footer links (short deadline)

DamianMullins (approved):

packages/components/organisms/f-footer/src/tenants/en-GB.js
ashleynolan: I think as this is copying the current link format in the footer, this should be raised as a separate ticket so we can make that change as an Accessibility Excellence Ticket.

@simonsandersje @BanksySan Can you catch up to create this ticket at some point (if you both think it's worth doing)

PR #1308, opened 2021-09-30
f-restaurant-card@0.1.0 - generate component

jamieomaguire:
Initial generated template for the new `f-restaurant-card` component.

This is going to act as a shared component between any pages (such as Search and Home) that currently have their own implementation of the same restaurant card.

Visual examples of the current implementation across pages below:

Screenshot 2021-10-04 at 09 24 24

Screenshot 2021-10-04 at 09 25 03
Screenshot 2021-10-04 at 09 25 49

Screenshot 2021-10-04 at 09 26 01


dandel10n (approved):

packages/components/molecules/f-restaurant-card/README.md
```diff
@@ -0,0 +1,127 @@
...
+    components: {
+        RestaurantCard
+    }
+}
+```
+
+If you are using Webpack, you can import the component dynamically to separate the `restaurant-card` bundle from the main `bundle.client.js`:
+
+```js
+import '@justeat/f-restaurant-card/dist/f-restaurant-card.css';
```
zlatinivanov-je, line 54: Shall we add a comment to suggest for components used in more than one location to have the styles imported in a _once, single place in the host app_. e.g. Host app `main.scss`

zlatinivanov-je (approved):
LGTM
Just a small suggestion. Possibly more relevant to components such as `f-button` etc. that are likely to be used in more than one location.

packages/components/molecules/f-restaurant-card/README.md
jamieomaguire: I'd suggest making that change in the Yeoman generator! Sounds like a good idea. I plan to update this readme once we start properly adding code to the component and will include this suggestion :)

PR #1309, opened 2021-10-03
#trivial Incorporated Page.withQuery into a param for Page.load

dbernar1:
Simplifies loading a page with the correct config. Removes need for instance data in Page objects.

Old way:
```
button = new Button();
button.withQuery('knob-Button Type', 'primary');
button.withQuery('knob-Button Size', 'medium');
button.load();
```
New way:
```
button.load({
    'Button Type': 'primary',
    'Button Size': 'medium',
});
```

### DIFF
--- a/packages/components/atoms/f-button/test-utils/component-objects/f-button.component.js
+++ b/packages/components/atoms/f-button/test-utils/component-objects/f-button.component.js
@@ -13,12 +13,12 @@ module.exports = class Buttons extends Page {
         super.open(url);
     }
 
-    load (type = 'action') {
-        if (type === 'action') {
-            super.load(this.actionComponent);
-        } else {
-            super.load(this.linkComponent);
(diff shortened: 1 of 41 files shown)


zlatinivanov-je (approved):
Maybe consider breaking it down to separate components, even if we have two approaches during the transition to make it easier to review and reduce risk on missing something.

ashleynolan:
@dbernar1 Is this PR still needed? If so, can you resolve the conflicts (or close it if not)

dbernar1:
I'm not sure whether a refactoring is ever needed, but I think the code is better with this change.

I may make the update when I grab a chance, but I'm not sure when I will have the time.

Thanks.

siggerzz:
Apologies for letting this PR get stale. We've got a ticket on the backlog to implement this change.

PR #1310, opened 2021-10-04
fozzie-components@v4.1.0 – Cache Component in Circle CI #trivial

jamieomaguire:
Adds the new f-restaurant-card component to the cached components list in Circle CI

Also updates the basic component test to use the newer patterns other molecules use.


CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+Latest (to be added to next release)
```
ashleynolan:
```suggestion
v4.1.0
```

Just as we don't 'release' the root, we just version changes for documentation purposes

ashleynolan: You'll need to bump the package.json version number too

jamieomaguire: ah, maybe I misunderstood but I thought this is what @dandel10n was asking me to do! Will update to the suggestion :)

ashleynolan (dismissed):

AlixRichards (dismissed):

CHANGELOG.md
dandel10n: I thought that we don't need to bump the version for this kind of change...

ashleynolan: For changes to the root config, I bump on any change (as it's not being released anyway – it's just to catalog the change).

But for small changes in a releasable package that don't actually need to be released, then we don't need to bump the version every time.

dawidchar (approved):

ashleynolan (approved):

PR #1321, opened 2021-10-05
f-test-data-creator@v0.2.0 - Added products to basket API #trivial

LTurns:
v0.2.0
*October 8, 2021*

### Added
- Capability to add product items to a basket in `basketApi` - enabling us to complete our e2e tests with the correct happy path


JoshuaNg2332 (dismissed):

packages/services/f-test-data-creator/src/clients/basketApi.js
```diff
@@ -9,6 +9,9 @@ module.exports = class BasketServiceApi {
     }
 
     async createBasketForUserAsync (basketInfo, timeout = 5000) {
+        const timeElapsed = Date.now();
+        const today = new Date(timeElapsed);
```
SachaZvetelman: I'm not sure I fully understand what we're trying to do here :) Could you please clarify?

packages/services/f-test-data-creator/src/clients/basketApi.js
```diff
@@ -20,20 +23,22 @@ module.exports = class BasketServiceApi {
...
-            Deals: [],
-            MenuGroupId: basketInfo.menuId,
-            OrderDetails: {
-                Location: {
-                    GeoLocation: {
-                        Latitude: 0,
-                        Longitude: 0
-                    },
-                    ZipCode: this.postcode
+            restaurantSeoName: basketInfo.restaurantSEO,
```
SachaZvetelman: Should this be `basketInfo.restaurantSeo`?

packages/services/f-test-data-creator/src/services/menuService.js
```diff
@@ -3,22 +3,38 @@ const MenuServiceApi = require('../clients/menuApi');
...
 
-        console.log(`Successfully returned menu for service type: ${serviceType} and restaurant ${restaurantSEO}`);
+        const products = result[0].Categories.filter(category => this.products[serviceType].includes(category.Id));
 
-        return result[0];
+        if (products === undefined) {
+            throw new Error(`No products available for ${serviceType}`);
+        }
+
+        const product = products[0].ItemIds.filter(id => this.items[serviceType].includes(id)).toString();
```
SachaZvetelman: If you're trying to find one element only, normally it's better to use `find` instead of `filter`. `find` will stop whenever it finds an element matching your criteria, whereas `filter` will go through all the elements, meaning that if you have a big collection, it may be less performant.

packages/services/f-test-data-creator/src/clients/basketApi.js
zlatinivanov-je: seen that coming up as `restaurantSEO` in other places, but a good point the prop name here then looks inconsistent.

packages/services/f-test-data-creator/src/clients/basketApi.js
JoshuaNg2332: So for generating a basket, it requires the date and time as part of the basket information.
So I believe this is why these are created.

zlatinivanov-je: Will it make sense to have that in own method, as I am guessing it will be used in a few places?

SachaZvetelman: If you need the current date and time, can't you do something like `new Date().toLocaleString();`?

DamianMullins: We do `new Date().toISOString()` when configuring basket requests in menu.

SachaZvetelman: Or that, yes 😄

JoshuaNg2332: Ok. I'll address that now. Thank you for the suggestions! :)

zlatinivanov-je (approved):

SachaZvetelman (approved):
Approved! As a side note, since it wasn't introduced in this PR, is there any reason for us using `.catch()` instead of `await` and `try-catch`, which we normally favour?

JoshuaNg2332:
> Approved! As a side note, since it wasn't introduced in this PR, is there any reason for us using `.catch()` instead of `await` and `try-catch`, which we normally favour?

Not sure personally. 
This will have to be a question for @LTurns I think.

PR #1322, opened 2021-10-06
f-compatibility@v0.1.0 - IE11 Browser compatibility check

simonsandersje:
Initial scaffolding for a new service to check browser compatibility.

Although this is designed in the first instance to check for IE11 to give a browser deprecation warning, I will try to develop it so it can be repurposed in future for other compatibility checks.


DamianMullins (approved):

DamianMullins:
@simonsandersje One thing you _could_ use for the compatibility check is [`bowser`](https://github.com/lancedikson/bowser) — we use it in menu to detect supported browsers when pushing client-side logs. 

It uses a small bit of config that it checks against the user-agent. If any of that sounds useful, give me a shout and we can go through it all together.

simonsandersje:
Will close for now while I investigate using Bowser as per @DamianMullins 's comment

PR #1323, opened 2021-10-06
f-cookie-banner@3.2.0 - Adding a nameSuffix prop.

SachaZvetelman:
v3.2.0
*October 06, 2021*

### Added
- New prop `nameSuffix` so teams or 3rd parties using this component can save the user's consent under a different cookie name. This allows the cookie banner to work on multiple sub-domains easily.


packages/components/organisms/f-cookie-banner/src/components/_tests/CookieBanner.test.js
```diff
@@ -210,17 +210,21 @@ describe('CookieBanner', () => {
...
                     computed: {
-                        legacyBanner: () => false
+                        legacyBanner: () => false,
+                        consentCookieName: () => cookieName
                     }
                 });
-                CookieHelper.set('je-cookieConsent', cookieValue);
+
+                jest.spyOn(CookieHelper, 'get').mockReturnValue(cookieValue);
+                // CookieHelper.set(cookieName, cookieValue);
```
DamianMullins: Do you need this line?

packages/components/organisms/f-cookie-banner/src/components/_tests/CookieBanner.test.js
```diff
@@ -210,17 +210,21 @@ describe('CookieBanner', () => {
...
                     propsData,
                     computed: {
-                        legacyBanner: () => false
+                        legacyBanner: () => false,
+                        consentCookieName: () => cookieName
                     }
                 });
-                CookieHelper.set('je-cookieConsent', cookieValue);
+
+                jest.spyOn(CookieHelper, 'get').mockReturnValue(cookieValue);
```
DamianMullins: One thing I like to do (and this might just be personal preference) is to set `jest.mock('js-cookie');` at the start of the test module, then you can set up mocks directly on `js-cookie` like this `cookieHelper.get.mockReturnValue(cookieValue);`.

It also means that the npm package is fully mocked and you don't accidentally end up calling a real method in your tests.

Feel free to ignore if you don't like that approach or if there is a convention used in this package already.

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -181,6 +183,18 @@ export default {
...
         legacyBanner () {
             return this.shouldShowLegacyBanner === null ? this.config.displayLegacy : this.shouldShowLegacyBanner;
+        },
+
+        /**
+		 * Return the cookie consent name with the suffix if provided.
+		 * @returns {String}
+		 */
+        consentCookieName () {
+            const baseCookieName = 'je-cookieConsent';
```
zlatinivanov-je: Shall we have the value stored in a config file, as it is referenced in a few locations. e.g. tests etc, to reduce the risk of error.

zlatinivanov-je (commented):
Ons small suggestion hope to reduce risks.

DamianMullins (approved):

dandel10n (approved):
Looks good, please make sure to re-build fozzie-components from the root level before publishing this version to npm

PR #1326, opened 2021-10-06
fozzie-components@4.3.0 - Add automatic labelling workflow

xander-marjoram:
### Added
- Labeler action
  - Automatically applies `wip` label to new PRs
  - Applies category labels based on whether the PR affects `atoms`, `molecules`, `services`, etc.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,24 @@
+name: "Pull Request Labeler"
+on:
+- pull_request_target
+
+jobs:
+  wip:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/labeler@v3
+      with:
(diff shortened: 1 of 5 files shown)


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,24 @@
+name: "Pull Request Labeler"
+on:
+- pull_request_target
```
xander-marjoram, line 3: I _think_ this is the reason it's not running yet.

https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target

`pull_request_target` runs from the base context of the PR rather than the merge commit, but it's recommended for things like this so we may need to merge it to see if it works.

ashleynolan (approved):

siggerzz (approved):

LTurns (approved):

PR #1331, opened 2021-10-07
f-restaurant-card@v0.2.0 - basic component structure

jamieomaguire:
Much of this PR is placeholder, however some bits added:

### Added
- Wrapper component to handle multiple component versions
- Passing props and slots through wrapper component
- Placeholder template and logic to component
- Wrapper component unit tests

### Note:
Please disregard the HTML and CSS - this is by no means what will actually be used in the component and is merely placeholder for now.

---
Screenshot 2021-10-07 at 15 48 21

## Browsers Tested

- [X] Chrome (latest)


packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -0,0 +1,124 @@
...
+        },
+        url: {
+            type: String,
+            default: null
+        },
+        logo: {
+            type: String,
+            default: null
+        },
+        isTemporarilyOffline: {
```
sonnyprince: should this component know about availability states? could we pass in different meta to show (this restaurant is offline text) and have a prop to "disable" the card (add opacity)

jamieomaguire: for sure, the props here are by no means final. I think just passing a prop to disable is more than likely how we'll do it. Most current references to the offline state drive things like displaying images/ratings/dishes etc which now all be slot driven now

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -0,0 +1,124 @@
...
+
+            <!-- Cuisines -->
+            <!-- START ERROR BOUNDARY -->
+            <slot name="cuisines" />
+            <!-- END ERROR BOUNDARY -->
+
+
+            <!-- New label -->
+            <!-- START ERROR BOUNDARY -->
+            <slot name="new" />
```
ashleynolan: I'd be tempted to make this a bit more descriptive, and call it `newLabel` or `newIndicator`? (whatever makes sense)

packages/components/molecules/f-restaurant-card/src/components/restaurantCardVersions.js
```diff
@@ -0,0 +1,7 @@
+import RestaurantCardV1 from './RestaurantCard.v1.vue';
+
+const components = {
+    v1: RestaurantCardV1
```
ashleynolan, line 4: This is super interesting – would be good to know (probably will need to test when you've built it out more) how much this impacts the size of the component. Mainly as that may determine how many versions you want running at once

ashleynolan (approved):

ashleynolan:
Not a blocker for this PR, but I'd probably start pulling together the README as you add props and other bits, as it'll make that task a little easier then when you want to put a PR in for that.

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: good idea, done! :)

packages/components/molecules/f-restaurant-card/src/components/restaurantCardVersions.js
jamieomaguire: Yeah that's a very good point!

jamieomaguire: will do some bundlewatch comparisons once there's more to this component

sonnyprince (approved):

PR #1332, opened 2021-10-07
f-ie-compatibility@v0.1.0 IE11 Browser warning banner

simonsandersje:
Scaffold created by yo generator for the banner telling Internet Explorer users that we no longer support IE


packages/components/atoms/f-ie-compatibility/README.md
```diff
@@ -0,0 +1,127 @@
+<div align="center">
+
+# f-ie-compatibility
```
ashleynolan, line 3: Can we drop the `ie` from the name? Just as the intention is that this will display on any browser that doesn't support the features that are not detected

PR #1340, opened 2021-10-11
f-compatibility@v0.1.0 IE11 Warning banner

simonsandersje:
Yo generator scaffold for f-compatibility - a banner to warn users when they using a non-supported browser


DamianMullins (dismissed):

simonsandersje:
Ash has requested that this be built as a service

PR #1343, opened 2021-10-11
f-button@3.1.0 - button stories refactoring

dandel10n:
### Changed
- Add spacing to buttons following a paragraph tag. This matches the current behaviour in Fozzie.
- Removed the background from the link button in disabled state.
- `f-vue-icons` package version bump.
- Button story now shows all the types and sizes of the button component (primary, secondary, outline and ghost).
- Changed knobs to controls for switching button props.
- Separate Demo story to still show one button in isolation.
- Separate story for Link Button to be able to run tests properly.
- Updated the tests in line with the changes.

Regular button:
Screenshot 2021-10-11 at 16 19 16

Mobile view:
Screenshot 2021-10-18 at 14 32 24

Link button:
Screenshot 2021-10-11 at 16 19 58

### Changed
- Enabled expanded version of controls tab to show descriptions and default values of properties.


packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -27,6 +27,11 @@
                     [$style['o-btn-content--hidden']]: isLoading
                 }]">
 
+                <span
```
DamianMullins: Looks like this element is already present above, should it be here twice?

dandel10n: Yeah, bad rebase, fixed

packages/components/atoms/f-button/stories/Button.stories.js
```diff
@@ -1,82 +1,104 @@
...
     components: {
         FButton,
-        EyeglassIcon
+        CollectionIcon
     },
+    props: Object.keys(argTypes),
+    template: `
+    <div>
+        <div
+            style="display: flex; flex-flow: row nowrap; align-items: center; justify-content: space-around; margin-bottom: 16px;"
```
ashleynolan: How does this look at smaller widths in storybook? (just wondering do the buttons wrap so it's a vertical list?)

ashleynolan: Also, rather than applying these styles inline, it would be good to add them to an external CSS file so that we could potentially reuse them (I'm guessing these styles will eventually be useful for icon buttons etc).

You could add a file [imported in `preview.js`](https://github.com/justeat/fozzie-components/blob/master/packages/storybook/config/storybook/preview.js) and just make sure the class naming is really specific to the documentation (so `.docs-buttonRow` or something? an then similarly, add a `.docs-buttonRow > button` to add the margin below?)

dandel10n: Actually, I will update to use grid and utility fozzie classes instead :+1:

dandel10n: @ashleynolan I added a screenshot with mobile view and changed the stories to use grid and utility classes

DamianMullins (dismissed):

dandel10n:
I splitted fozzie version bump into [a separate PR](https://github.com/justeat/fozzie-components/pull/1374). I will rebase this one after another one is merged

FayeCarter (dismissed):

packages/components/atoms/f-button/stories/Button.stories.js
```diff
@@ -1,82 +1,104 @@
...
-
-        href: {
-            default: text('href', '')
-        },
-
-        hasIcon: {
-            default: select('Add icon to the button text', VALID_BUTTON_ICON_POSITION, false)
-        }
+ButtonComponent.argTypes = {
+    disabled: {
```
FayeCarter: could these go in a shared/default argTypes file?

DamianMullins (dismissed):

DamianMullins:
@dandel10n It looks like [there's an issue with the f-button desktop spec](https://app.circleci.com/pipelines/github/justeat/fozzie-components/10610/workflows/ccd113e3-9ab9-4b00-b0e7-2d98f8288e7b/jobs/14537) according to CI?

ashleynolan:
One thing I was thinking (that I've seen in a few other storybooks) is that even though we're creating a page for all the buttons (more for our visual regression tests) we could potentially still have a storybook page which is called 'demo' so that people can change all the various props for one button.

Might be outside the scope of this JIRA, but I would make another ticket that links back to the old code being removed in this PR, as I think we'd likely reuse a lot of the old story code for that type of page?

dandel10n:
> One thing I was thinking (that I've seen in a few other storybooks) is that even though we're creating a page for all the buttons (more for our visual regression tests) we could potentially still have a storybook page which is called 'demo' so that people can change all the various props for one button.
> 
> Might be outside the scope of this JIRA, but I would make another ticket that links back to the old code being removed in this PR, as I think we'd likely reuse a lot of the old story code for that type of page?

@ashleynolan I added Demo story to this PR as it is easier to do now while I am on it than leaving it for later

ashleynolan:
@dandel10n Nice one 👍

packages/components/atoms/f-button/stories/Button.stories.js
dandel10n: Thanks, I extracted argTypes into its own file :+1:

DamianMullins (dismissed):

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -612,7 +616,7 @@ $btn-icon-sizeXSmall-buttonSize        : 32px;
 .o-btn--link {
     border: 0;
     background-color: transparent;
-    padding: 0;
+    padding: 0 !important;
```
ashleynolan: What's the reason for the important (just wondering what styles trump this one)

packages/storybook/config/storybook/scss-loader.scss
```diff
@@ -1,5 +1,6 @@
 @import "@justeat/fozzie/src/scss/fozzie";
 @include reset();
 @include typography();
+@include grid();
```
ashleynolan: I'd be a bit careful adding this potentially – it'll load in the old classes for every component in storybook, which I probably wouldn't do here (if you're just doing it for some structure in how multiple components are displayed?).

I would add a separate CSS file for any styles like this that you only load into stories that you need it for (or include it as part of the [preview-head file](https://github.com/justeat/fozzie-components/blob/master/packages/storybook/config/storybook/preview-head.html)?)

ashleynolan (commented):
Couple of small comments

packages/storybook/config/storybook/scss-loader.scss
dandel10n: These styles won't be added to component bundles, they will just become available in storybook. `grid()` is part of base fozzie styles and is added by both `$includeBaseFramework` and `$includeMinimalFramework` so I thought that it would make sense to reuse them in storybook as well instead of adding separate classes.

packages/components/atoms/f-button/src/components/Button.vue
dandel10n: `.o-btn--link` gets overwritten by size class modifiers. You can see this in [prod storybook](https://vue.pie.design/?path=/story/components-atoms-f-button--button-component) if you choose link button large or small/xsmall

packages/components/atoms/f-button/stories/Demo.stories.js
```diff
@@ -0,0 +1,83 @@
...
+                <collection-icon />
+            </template>
+            Label
+            <template #trailing-icon>
+                <collection-icon />
+            </template>
+        </f-button>`
+});
+
+DemoButtonComponent.argTypes = {
```
FayeCarter, line 42: could you use `...ButtonArgTypes` here to save having to add buttonType/isLoading etc?

FayeCarter (dismissed):

packages/components/atoms/f-button/src/components/Button.vue
ashleynolan: Ok – I'd probably add a comment next to it saying that too. Makes sense though 👍

packages/storybook/config/storybook/scss-loader.scss
ashleynolan: Yep – get that it isn't added to the actual bundle, but the purpose of this set of styles is to replicate the consuming application styles that get set (i.e. CoreWeb) which grid isn't imported in, so that's why I think it would be good to move this so that it's more 'storybook specific'.

The `$includeBaseFramework` and `$includeMinimalFramework` are in fozzie to support older microsites too
 – eventually, it would be good to remove these once we move verticals over 100% to components. I'm hoping we'd be able to deprecate the grid classes at some point as I'm encouraging anyone writing new styles to use CSS grid instead now too.

dandel10n: Ok, I will create some grid classes for storybook then

DamianMullins (approved):

ashleynolan (approved):

packages/components/atoms/f-button/src/components/Button.vue
DamianMullins: The `important` declaration here is causing issues when I add this to menu as I then need to add `!important` to almost all of them as they require some padding to make the click area large enough on small screens.

@dandel10n is there a way we can increase specificity here without having to use `important`?

ashleynolan: If that's the case, I'd just reference the sizing classes and target them here to increase the specificity.

PR #1351, opened 2021-10-13
f-cookie-banner@3.3.0 - Adding a `domain` prop

SachaZvetelman:
v3.3.0
*October 13, 2021*

### Added
- `domain` prop to allow the consumer to specify which hosts can receive this component's cookie.

### DIFF
--- a/packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
+++ b/packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
@@ -147,6 +147,11 @@ export default {
         nameSuffix: {
             type: String,
             default: ''
+        },
+
+        domain: {
+            type: String,
+            default: null
         }
     },
 
@@ -346,7 +351,8 @@ export default {
         setCookieBannerCookie (cookieValue) {
             CookieHelper.set(this.consentCookieName, cookieValue, {
                 path: '/',
-                expires: this.cookieExpiry
+                expires: this.cookieExpiry,
+                domain: this.domain
             });
         },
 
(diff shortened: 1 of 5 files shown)


FayeCarter (approved):

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -346,7 +351,8 @@ export default {
         setCookieBannerCookie (cookieValue) {
             CookieHelper.set(this.consentCookieName, cookieValue, {
                 path: '/',
-                expires: this.cookieExpiry
+                expires: this.cookieExpiry,
+                domain: this.domain
```
simonsandersje, line 355: How does js-cookie handle "null" for the domain value?

SachaZvetelman: I've tested this, and basically it behaves as if you didn't specify it (i.e. exactly as it works at the moment).
The way I tested this was by creating two entries on my `hosts` file, `app1.subdomain.com` and `app2.subdomain.com`, both pointing to `localhost`.
Then I went to the cookie banner component in each of these domains, and if I accepted the cookie banner in one of them, the cookie banner on the other one continues to be displayed. However, if I modify the story file and pass `subdomain.com` as the `domain`, when accepting the cookie banner in one, the other domain does not show the cookie banner.

I hope it makes sense, @simonsandersje, but please let me know if it doesn't 👍

simonsandersje (approved):

PR #1355, opened 2021-10-13
f-cookie-banner@3.4.0: Add requestAnimationFrame to ResizeObserver callback

AndyMayJE:
### Changed
- ResizeObserver to use requestAnimationFrame and watcher to ensure ResizeObserver does not throw `ResizeObserver loop limit exceeded` error.

## Example

| `isBodyHeightLessThanWindowHeight` | Screenshot |
| --- | --- |
| `true` | image |
| `false` | image |

## Browsers Tested

- [x] Chrome (latest)
- [x] FireFox 93 (mac)
- [x] Safari 15 (mac) 
- [x] iPhone SE
- [x] iPhone 11
- [x] Chrome Galaxy S10 

Tested locally using yalc in CoreWeb, ConsumerHelpWebApp


zlatinivanov-je (commented):
I have no issue approving this PR, but wanted to question if there is an other way in achieving the end result? Could it be done without the use of `ResizeObserver`?  
It may help to add a bit more context in the description, e.g. when does the Cookie banner positioning changes or needs to change.

AndyMayJE:
> I have no issue approving this PR, but wanted to question if there is an other way in achieving the end result? Could it be done without the use of `ResizeObserver`? It may help to add a bit more context in the description, e.g. when does the Cookie banner positioning changes or needs to change.

Thanks. The original solution to absolute position the re-open link when the body is smaller than the window lead to problems. This caused the link to sometimes be incorrectly positioned blocking content as it didn't account for content changes and load times in the parent page. 

The ResizeObserver allows us to watch changes in body height as content is dynamically loaded or changes. By design, it meets the purpose required. Obviously, if there are alternative solutions we suggested we should investigate and compare.

Personally, I would question if this logic and functionality really has a positive impact on the end-user. Should we leave the position alone, rather than making assumptions about the consuming page.

zlatinivanov-je (dismissed):

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -212,22 +221,37 @@ export default {
...
+                    this.bodyHeight = height;
+                });
+            };
+            this.bodyObserver = new ResizeObserver(resizeCallback);
+            const bodyElement = document.documentElement || document.body;
+            this.bodyObserver.observe(bodyElement);
+        },
+
+        /**
+         * Triggered vy watch.bodyHeight, calculates if body height is less than window height to absolutley position the reopen link
```
xander-marjoram: I won't ask you to change this now, but if there are any other changes then feel free to update these at the same time

```suggestion
         * Triggered by watch.bodyHeight, calculates if body height is less than window height to absolutely position the reopen link
```

packages/components/organisms/f-cookie-banner/src/components/_tests/CookieBanner.test.js
```diff
@@ -535,4 +559,57 @@ describe('CookieBanner', () => {
...
+                    localVue,
+                    i18n
+                });
+
+                await wrapper.setData({ bodyHeight: 500 });
+
+                expect(spyMethod).toBeCalled();
+            });
+
+            it('should not call updateIsBodyHeightLessThanWindowHeight when data.bodyHeigth newVal === oldVal', async () => {
```
xander-marjoram: Nit
```suggestion
            it('should not call updateIsBodyHeightLessThanWindowHeight when data.bodyHeight newVal === oldVal', async () => {
```

xander-marjoram (dismissed):
LGTM. Just a couple of typos which I'll let you decide how/if to handle.
Although if you do update them then the changelog date could be bumped too.

xander-marjoram (dismissed):

simonsandersje (dismissed):

simonsandersje (approved):

xander-marjoram (approved):

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -212,22 +221,37 @@ export default {
...
-            this.bodyObserver.observe(bodyElement);
+        if (typeof window === 'object' && this.shouldAbsolutePositionReopenLink && !this.legacyBanner) {
+            this.createResizeObserver();
         }
     },
 
     methods: {
         /**
-         * Triggered by <body> ResizeObserver calculates if the body height less than the window
+         * Creates <body> ResizeObserver to watch for body content height changes, triggers watch.bodyHeight method
```
ashleynolan, line 231: Sorry for the post-merge comment – was just interested to know what the issue was being solved.

So was this issue that the cookie-banner wasn't always positioned at the bottom of the page because content was still being loaded in and that led to an error?

PR #1356, opened 2021-10-14
f-checkout@2.5.1 and other packages version bumps

dandel10n:
### Changed
- Updated version of `f-button`.

### Changed
- Updated version of `f-button`.

### Changed
- Updated version of `f-button`, `f-alert`, `f-mega-modal`.

### Changed
- Bump `f-mega-modal` version to clean f-button styles.

### Changed
- Updated version of `f-button`.


yarn.lock
```diff
@@ -2197,6 +2197,11 @@
   dependencies:
     window-or-global "^1.0.1"
 
+"@justeat/f-mega-modal@3.0.0":
+  version "3.0.0"
+  resolved "https://registry.yarnpkg.com/@justeat/f-mega-modal/-/f-mega-modal-3.0.0.tgz#2e635096c2a6a8fd7da308479a2126748716dcbf"
+  integrity sha512-ZbAgBrqn3a2furQ5F0FLffbsMZar9aIDsRG/2uWnsv6uY0KIHgJi08kH2XK5NzDzckHih2uMN37JiYpQMw6rNQ==
+
```
SachaZvetelman, line 2204: Is this correct since you've updated `f-mega-modal` to `3.0.1`?

SachaZvetelman (approved):

DamianMullins (approved):

yarn.lock
DamianMullins: It could be because `f-mega-modal` is also used in `f-searchbox` & `f-offers`?

SachaZvetelman: Perhaps, but shouldn't then this part in `yarn.lock` be there already? 🤔 
I might be wrong.

dandel10n: Not sure why it was removed previously, but as yarn.lock rebuilds on every `yarn` I wouldn't worry about it.

DamianMullins: I believe an entry isn't created in the lock file when the version matches whatever is set in`package.json` for `f-mega-modal` (or any other package). 

Once there are versions that don't match an entry will be added here.

@dandel10n do you also need to bump the version in `f-searchbox` & `f-offers` for the modal?

PR #1359, opened 2021-10-14
f-restaurant-card@v0.3.0 - container styling

jamieomaguire:
### Added
- Basic component styling
### Updated
- Component HTML to use new classes, attributes and fewer elements
- Storybook args updated to use new component props contract

Notes: 
- The list item variant will break down to default at the specified media query.
- Need to confirm with design what to do when bg image is missing

### Default
Screenshot 2021-10-21 at 11 05 41

### List Item
Screenshot 2021-10-21 at 11 05 32

### Default in SearchWeb
Screenshot 2021-10-21 at 11 30 26

### List Item in SearchWeb
Screenshot 2021-10-21 at 11 23 27

## Browsers Tested

- [X] Chrome (latest)
- [X] Mobile - IPhone 13 Mini, Safari


packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,149 @@ export default {
...
+        min-height: 96px;
+    }
+}
+
+@mixin restaurantCard-tile {
+  display: block;
+  position: relative;
+  cursor: pointer;
+
+  &.c-restaurantCard--img {
```
sonnyprince: do we need this specificity? or could this be `&-withImg` - also less confusing as theres also `c-restaurantCard-img`

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,149 @@ export default {
...
+      padding: spacing(x2);
+      padding-top: spacing(x10) * 2;
+  }
+
+  &:not(.c-restaurantCard--img) {
+      padding-top: spacing() * 3.5;
+  }
+
+  .c-restaurantCard-img {
```
sonnyprince: can we use `&` instead of repeating `c-restaurantCard` on every class?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,149 @@ export default {
...
+    height: 228px;
+    left: 0;
+    right: 0;
+    top: 0;
+    bottom: spacing(x8);
+    z-index: -1;
+    position: absolute;
```
sonnyprince: if this is absolutely positioned, should we need the bottom and right offsets?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,149 @@ export default {
...
     display: flex;
-    justify-content: center;
-    margin: auto;
-    border: 1px solid purple;
-    font-family: $font-family-base;
-    @include font-size(heading-m);
-}
 
+    &-logo {
+        border: 0.5px solid $logo-borderColor;
```
sonnyprince: are .5 borders okay cross browser?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,149 @@ export default {
...
+        }
+
+        // less padding if img
+        .c-restaurantCard--img & {
+            padding-left: spacing(x5);
+        }
+    }
+
+    &--list-item {
+        @include media('>mid') {
```
sonnyprince: this is really cool!

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,149 @@ export default {
...
+$logo-borderRadius                        : $radius-rounded-b;
+$logo-borderColor                         : $color-border-default;
+
+@mixin card-frame {
+    z-index: 1;
+    background: $card-bgColor;
+    box-shadow: 0 6px 8px rgba($color-black, 0.02), 0 1px 20px rgba($color-black, 0.08), 0 3px 6px -1px rgba($color-black, 0.08);
+    border-radius: $card-borderRadius;
+    position: relative;
+    min-height: 114px;
```
sonnyprince: Understand these heights would've come from design - but could we do something a little more fluid and let the heights be based on the content?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,149 @@ export default {
...
+  display: block;
+  position: relative;
+  cursor: pointer;
+
+  &.c-restaurantCard--img {
+      padding: spacing(x2);
+      padding-top: spacing(x10) * 2;
+  }
+
+  &:not(.c-restaurantCard--img) {
```
sonnyprince: this okay x-browser?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -1,29 +1,33 @@
...
             <!-- Logo image -->
             <img
-                src=""
-                alt=""
+                :src="logoUrl"
+                alt="TODO - Determine appropriate alt"
                 width="50"
                 height="50"
                 loading="lazy"
-                data-test-id="restaurant_logo">
```
sonnyprince, line 26: we might still need this :)

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,147 @@ export default {
...
+        },
         // feature flags
         flags: {
             type: Object,
             default: () => ({})
         }
+    },
+    computed: {
+        displayModeModifier () {
+            return this.tileMode ? 'c-restaurantCard--tile' : 'c-restaurantCard--list-item';
```
sonnyprince: 🔥

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -1,29 +1,33 @@
...
-                alt=""
-                loading="lazy"
-                data-test-id="restaurant-cuisine-image">
-
+    <a
+        :href="url"
+        :class="[
+            $style['c-restaurantCard'],
+            $style[displayModeModifier],
+            { [$style['c-restaurantCard--img']]: !!imgUrl }]"
```
sonnyprince: this might make more sense here:
https://github.com/justeat/fozzie-components/pull/1359/files#r729153437

Having `c-restaurantCard--withImg` instead of`c-restaurantCard--img` - lil more descriptive

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
1itvinka: I would repeat `c-restaurantCard` on every class, because it's easier to navigate the code this way, when you look for a specific class.  @ashleynolan do we have any written policies on this?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
1itvinka: I would avoid unnecessary nesting when possible.

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -1,29 +1,33 @@
...
+            aria-label="TODO - Determine appropriate label" />
+
+        <!-- card content -->
+        <div :class="$style['c-restaurantCard-content']">
             <!-- Logo image -->
             <img
-                src=""
-                alt=""
+                :src="logoUrl"
+                alt="TODO - Determine appropriate alt"
```
DamianMullins: I think an empty `alt` attribute is appropriate here.

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -1,29 +1,33 @@
...
+        data-test-id="restaurantCard-component"
+        @click="$emit('restaurant-card-clicked')">
+
+        <!-- background image -->
+        <div
+            v-if="imgUrl"
+            :class="$style['c-restaurantCard-img']"
+            :style="`background-image: url(${imgUrl});`"
+            role="img"
+            aria-label="TODO - Determine appropriate label" />
```
DamianMullins: I'm not sure you need to describe this as it's a decorative image.

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
DamianMullins: The general guidance is as you say, @1itvinka — whilst it's nice to use the ampersand in this way in SCSS, in our code we've decided to stick with being more explicit with our class definitions. It makes finding classes and their uses much easier when searching in an IDE for example, and we'd like to keep this consistent across all of our codebases.

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
DamianMullins: Agree, these classes should be at the top-level and use the `c-restaurantCard` prefix.

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: the rename is definitely a good idea - will update now

jamieomaguire: and you're right - the specificity there isn't needed!

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: yup that sounds reasonable, I'll update to follow the established convention

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: you're correct - I just need to set width to 100%

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: hmm, possibly not for previous versions of a few browsers we support. will look into it

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: no idea why I removed this... adding back in

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: updated!

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: maybe we don't - unsure the best practice here for an empty div with a bg image and screenreaders

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: changed to 1px thickness (not sure where .5 came from) and removed nested syntax

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: updated

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: removed label

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: happy to remove the min-height. then it's height is entirely driven by content

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: Ok so I think my approach to styling this was backwards before. I've updated to style without an image first, then extend the padding when an img modifier class is present. It removes the need for this :not() selector and feels a lot more sensible. Cheers!

packages/components/molecules/f-restaurant-card/CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.3.0
+------------------------------
+*October 14, 2021*
+
+### Added
+- Basic component styling
+### Updated
```
ashleynolan: Super NIT:

```suggestion
- Basic component styling

### Updated
```

packages/components/molecules/f-restaurant-card/CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
...
+- Component HTML to use new classes, attributes and fewer elements
+- Storybook args updated to usenew component props contract
```
ashleynolan:
```suggestion
- Storybook args updated to use new component props contract
```

dawidchar (dismissed):

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
ashleynolan: Just a naming scheme thing – I'd use `c-restaurantCard--listItem`as we use dashes for hierarchy rather than for spaces

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,144 @@ export default {
...
+    }
+
+    .c-restaurantCard-content {
+        padding: spacing(x2);
+        @include card-frame;
+    }
+
+    .c-restaurantCard-logo {
+        left: 50%;
+        top: -25px;
```
ashleynolan: Is the idea that this is vertically aligned? If so, you could use a similar translateY trick to get it into the centre of the background image, rather than px

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,144 @@ export default {
...
+$tile-imgSpacing                          : spacing(x10) * 2;
+$img-border-radius                        : $radius-rounded-c;
+$img-width                                : 156px;
+$logo-borderRadius                        : $radius-rounded-b;
+$logo-borderColor                         : $color-border-default;
+
+@mixin card-frame {
+    z-index: 1;
+    background: $card-bgColor;
+    box-shadow: 0 6px 8px rgba($color-black, 0.02), 0 1px 20px rgba($color-black, 0.08), 0 3px 6px -1px rgba($color-black, 0.08);
```
ashleynolan: I assume this wouldn't get complained about in our linting rules, but if not, I'd structure this as:

```suggestion
    box-shadow: 0 6px 8px rgba($color-black, 0.02),
                0 1px 20px rgba($color-black, 0.08),
                0 3px 6px -1px rgba($color-black, 0.08);
```

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,144 @@ export default {
...
 <style lang="scss" module>
+$card-bgColor                             : $color-container-default;
+$card-borderRadius                        : $radius-rounded-c;
+$tile-imgSpacing                          : spacing(x10) * 2;
+$img-border-radius                        : $radius-rounded-c;
+$img-width                                : 156px;
+$logo-borderRadius                        : $radius-rounded-b;
+$logo-borderColor                         : $color-border-default;
+
+@mixin card-frame {
```
ashleynolan: Small thing – could you use the card component for this? Could make this change in a future PR, but that would just mean that it stays in sync with the default card styling (and you could extend it if needed then)

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -1,30 +1,34 @@
...
-                alt=""
-                loading="lazy"
-                data-test-id="restaurant-cuisine-image">
-
+    <a
+        :href="url"
+        :class="[
+            $style['c-restaurantCard'],
+            $style[displayModeModifier],
+            { [$style['c-restaurantCard--with-img']]: !!imgUrl }]"
```
ashleynolan: Small naming thing – we tend to try and use `has` or `is` for stateful naming like this where it makes sense to (just a consistency thing)

```suggestion
            { [$style['c-restaurantCard--hasImg']]: !!imgUrl }]"
```

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +94,144 @@ export default {
...
+
+.c-restaurantCard--list-item.c-restaurantCard--with-img {
+    @include media('<=mid') {
+        padding-top: $tile-imgSpacing;
+    }
+}
+
+.c-restaurantCard--tile.c-restaurantCard--with-img {
+    padding-top: $tile-imgSpacing;
+}
```
ashleynolan: Just to group these a little better under the with/hasImg modifier:

```suggestion
.c-restaurantCard--hasImg {
    &.c-restaurantCard--listItem {
        @include media('<=mid') {
            padding-top: $tile-imgSpacing;
        }
    }

    &.c-restaurantCard--tile {
        padding-top: $tile-imgSpacing;
    }
}
```

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: I can definitely do this in a future PR - I'll know more in the next day or so but in the stage 3 designs, it doesn't really seem to use a card anymore, or at least a transparent one. I'll try and pull in the card component in the next styling PR (which should be soon)

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +106,105 @@ export default {
...
+      @include media('>mid') {
+        min-height: 125px; // TODO: agree with design
+        height: 100%;
+        grid-column: 1/2;
+      }
+  }
+}
+
+.c-restaurantCard-content {
+  padding: spacing() 0 spacing() 0;
```
ashleynolan:
```suggestion
  padding: spacing() 0;
```

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +106,105 @@ export default {
...
+  border: 1px solid $logo-borderColor;
+  border-radius: $logo-borderRadius;
+  top: spacing(x2);
+  left: spacing(x2);
+  position: absolute;
+}
+
+// placeholder styles - will be extracted to subcomponent
+.c-restaurantCard-dish {
+  margin: 5px 0 5px 0;
```
ashleynolan:
```suggestion
  margin: 5px 0;
```

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -93,32 +106,105 @@ export default {
...
+  background: lightgreen;
+  padding: 1rem;
+  border: 2px dashed green;
+  border-radius: $card-borderRadius / 2;
```
ashleynolan, line 202: I hink dividing this, gives you a border-radius that isn't set in our design language (looks like this would be `6px`?). 

Would be better to use one of [the available aliases](https://unpkg.com/@justeat/pie-design-tokens@1.1.0/dist/jet.scss) (as should either be 4px or 8px border radius I would think here?

jamieomaguire: Oh yeah definitely - this was just some placeholder and by no means actually going to be used - I just wanted something green and _sort_ of getting across how it will eventually look. We will completely rewrite this class in a future PR.

dawidchar (approved):

PR #1361, opened 2021-10-15
f-button@3.0.3 - f-seleton-loader@2.1.0 - f-user-message@3.1.0 - Design team feedback

dandel10n:
### Changed
- Each button size vertical padding reduced by 1px to make button height 56px/48px/40px/32px instead of 58px/50px/42px/34px.

### Changed
- Rounded corners for all the skeleton blocks.

Before:
Screenshot 2021-10-15 at 14 56 32

After:
Screenshot 2021-10-15 at 14 55 22
f-user-message@3.1.0

### Changed
- Background and text colour to be in line with default warning message.

Before:
Screenshot 2021-10-15 at 14 57 48

After:
Screenshot 2021-10-15 at 14 54 55

### DIFF
--- a/packages/components/atoms/f-button/CHANGELOG.md
+++ b/packages/components/atoms/f-button/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v3.0.3
+------------------------------
+*October 15, 2021*
+
+### Changed
+- Each button size vertical padding reduced by 1px to make button height 56px/48px/40px/32px instead of 58px/50px/42px/34px.
+
+
 v3.0.2
 ------------------------------
 *October 13, 2021*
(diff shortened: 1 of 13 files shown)


1itvinka (approved):

DamianMullins (approved):

packages/components/atoms/f-button/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v3.0.3
+------------------------------
+*October 15, 2021*
+
+### Changed
+- Each button size vertical padding reduced by 1px to make button height 56px/48px/40px/32px instead of 58px/50px/42px/34px.
```
ashleynolan, line 12: I'm just checking storybook – it looks like the large button already is 56px (and then it has a 1px transparent border, which means devtools says it has a 58px height).

I think we've been back and forth a few times on button and form-field heights with design in the past, so probably worth catching up about this (and probably writing down whatever we do somewhere, cause I always forget!)

ashleynolan (changes requested):

ashleynolan:
I would likely break these changes out into separate PRs per component. Just because otherwise, any comments on one of the component changes will hold up the others going through.

PR #1362, opened 2021-10-15
f-navigation-links@v0.2.0 - Added styling and tests

oliversweb:
### Added
- Component layout and styling

___

image


packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -2,31 +2,91 @@
...
-        I am a NavigationLinks Component (GB)
+        <ul>
+            <li
+                v-for="({ Id, url, name, selected }, i) in links"
+                :key="`${i}_link`">
+                <span
+                    v-if="selected"
+                    :data-test-id="Id"
+                    tabindex="0"
+                    :class="[$style['c-hightlighted-link']]">
```
ashleynolan: I'd recommend adding classes to hook onto the elements for the styling – it just makes the styles less brittle if the markup ever were to change and it also helps define the hierarchy between the styles really clearly.

Also – do you need the outer div? I think you could just remove that and have the `ul` at the root.

So for this, I'd structure it something like this:

```
<ul class="c-navigationLinks">
    <li class="c-navigationLinks-item">
        <span
            if…
            class="c-navigationLinks-link c-navigationLinks-link--active">
            text
        </span>
        <a class="c-navigationLinks-link">
            text
        </a>
    </li>
</ul>
```

I think then in your styles below you should just be able to simplify the styles between the `span` and `a` as they have a shared classname, and then just add the `active` (or `highlight`) styles onto the modifier classname.

Happy to chat this through on Monday too

ashleynolan (changes requested):

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -2,31 +2,91 @@
...
+                text-decoration: none;
+            }
+        }
+    }
+
+    .c-hightlighted-link {
+        border-left: 2px solid $color-orange;
+        font-weight: $font-weight-bold;
+
+        a {
```
ashleynolan: Atm, it doesn't look like there would ever be an `a` inside the `c-hightlighted-link` tag? It looks like an either or based on the if?

packages/components/molecules/f-navigation-links/src/components/_tests/NavigationLinks.test.js
```diff
@@ -1,10 +1,64 @@
...
+
         expect(wrapper.exists()).toBe(true);
     });
+
+    it('should display the list elements correctly for each `link` supplied', () => {
+        const wrapper = shallowMount(NavigationLinks, { propsData });
+
+        const links = wrapper.find('[data-test-id="navigationLinks"] ul');
+        expect(links).toMatchSnapshot();
+    });
```
ashleynolan, line 69: Not sure if this is planned for a future PR, but it would be good to have some tests around what happens if the data passed into the prop is of an incorrect structure. So what would happen if I passed empty or no values for `url`, `name` or `selected` (what are the valid states and default states?).

packages/components/molecules/f-navigation-links/README.md
```diff
@@ -69,15 +67,34 @@ There may be props that allow you to customise its functionality.
...
+]
+ ```
 
 ### Events
 
 The events that can be subscribed to are as follows (if any):
 
 | Event | Description |
 | ----- | ----------- |
+None
```
ashleynolan: I'd just delete this section if there aren't any events.

packages/components/molecules/f-navigation-links/README.md
```diff
@@ -69,15 +67,34 @@ There may be props that allow you to customise its functionality.
...
-| Prop  | Type  | Default | Description |
-| ----- | ----- | ------- | ----------- |
+| Prop  | Type  | Default | Required | Description |
+| ----- | ----- | ------- | ----- | ----------- |
+| links | array | [] | true | An array of links objects (_see example below_) |
+
+```js
+[
+  {
+      Id: 'link1',            // Test Data Id
```
SachaZvetelman: `Id` should be `id` 👍

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,78 @@
 <template>
-    <div
+    <ul
+        v-if="links.length"
         :class="$style['c-navigationLinks']"
         data-test-id="navigationLinks">
-        I am a NavigationLinks Component (GB)
-    </div>
+        <li
+            v-for="({ Id, url, name, selected }, i) in links"
```
SachaZvetelman: Same as above, it should be `id`.

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,78 @@
...
+            :key="`${i}_link`"
```
SachaZvetelman: I might be wrong, but I think `:key="i"` should be sufficient.

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,78 @@
...
+            &:focus,
+            &:hover {
+                border-left: 2px solid $color-orange;
+                text-decoration: none;
+            }
+        }
+
+        .c-navigationLink--active {
+            border-left: 2px solid $color-orange;
+            font-weight: $font-weight-bold;
```
SachaZvetelman: Where possible, can you please use `spacing()` to define margins and paddings?

packages/components/molecules/f-navigation-links/src/components/_tests/NavigationLinks.test.js
```diff
@@ -1,10 +1,71 @@
...
+    ]
+};
+
 describe('NavigationLinks', () => {
     it('should be defined', () => {
-        const propsData = {};
         const wrapper = shallowMount(NavigationLinks, { propsData });
+
         expect(wrapper.exists()).toBe(true);
+        expect(wrapper.props().links).toEqual(propsData.links);
```
SachaZvetelman: I personally don't see the value of this assertion here, since we're basically testing that Vue knows how to receive `props`. But I might be wrong, so I'm keen to get your opinion as well on this.

SachaZvetelman (commented):
Looking good! Some minor comments only :)

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
DamianMullins: Using `id` would probably be best (as long as it's unique), that way if the number or order of items in the array changes then Vue will be able to use that to better update the DOM elements. By using the index Vue has no way of telling which elements can stay the same etc.

DamianMullins: Also, is the casing right for `id`? I see @SachaZvetelman mentioned above about casing of `id`?

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,78 @@
...
+    <ul
+        v-if="links.length"
         :class="$style['c-navigationLinks']"
         data-test-id="navigationLinks">
-        I am a NavigationLinks Component (GB)
-    </div>
+        <li
+            v-for="({ id, url, name, selected }, i) in links"
+            :key="i"
+            :class="$style['c-navigationLink-item']">
```
ashleynolan:
```suggestion
            :class="$style['c-navigationLinks-item']">
```

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,78 @@
...
+            <span
+                v-if="selected"
+                :data-test-id="id"
+                tabindex="0"
+                :class="[$style['c-navigationLink-link'], $style['c-navigationLink--active']]">
```
ashleynolan: So the base class is `c-navigationLinks-link` here and the modifier is indicated by the double dash which would be the active state `c-navigationLinks-link--active`. (Just a naming thing but can be easier to explain sometimes in person, so give me a shout if this doesn't make any sense!)

```suggestion
                :class="[$style['c-navigationLinks-link'], $style['c-navigationLinks-link--active']]">
```

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,78 @@
...
+                {{ name }}
+            </span>
+            <v-link
+                v-else
+                :data-test-id="id"
+                :has-text-decoration="false"
+                :href="url"
+                :class="$style['c-navigationLink-link']">
```
ashleynolan:
```suggestion
                :class="$style['c-navigationLinks-link']">
```

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,78 @@
...
+                text-decoration: none;
+            }
+        }
+
+        .c-navigationLink--active {
+            border-left: 2px solid $color-orange;
+            font-weight: $font-weight-bold;
+        }
+    }
+}
```
ashleynolan: Better not to nest these as it will just make the styles more specific. So you could just do:
```suggestion
}

.c-navigationLinks-item {
    display: inline-block;
    width: 100%;
}

.c-navigationLinks-link {
    display: block;
    padding: spacing() 0 spacing() spacing(x2);
    border-left: 2px solid $color-grey-30;
    color: $color-grey;
 
    &:focus,
    &:hover {
        border-left: 2px solid $color-orange;
        text-decoration: none;
    }
 }

.c-navigationLinks-link--active {
        border-left: 2px solid $color-orange;
        font-weight: $font-weight-bold;
    }
}

```

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,79 @@
...
-    </div>
+        <li
+            v-for="({ id, url, name, selected }, i) in links"
+            :key="i"
+            :class="$style['c-navigationLinks-item']">
+            <span
+                v-if="selected"
+                :data-test-id="id"
+                tabindex="0"
+                :class="[$style['c-navigationLinks-link'], $style['c-navigationLinks--active']]">
```
ashleynolan: Missed the naming change to the second class:

```suggestion
                :class="[$style['c-navigationLinks-link'], $style['c-navigationLinks-link--active']]">
```

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,79 @@
...
+        &:hover {
+            border-left: 2px solid $color-orange;
+            text-decoration: none;
+        }
+    }
+
+    .c-navigationLinks--active {
+        border-left: 2px solid $color-orange;
+        font-weight: $font-weight-bold;
+    }
```
ashleynolan: This block can be unnested like the code block I posted before? (and rather than use `li`, can hook onto `c-navigationLinks-item`)

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,76 @@
...
-    <div
+    <ul
+        v-if="links.length"
         :class="$style['c-navigationLinks']"
         data-test-id="navigationLinks">
-        I am a NavigationLinks Component (GB)
-    </div>
+        <li
+            v-for="({ id, url, name, selected }, i) in links"
+            :key="i"
```
SachaZvetelman, line 8: I think in the end @DamianMullins recommended using the `id` instead, as long as it's unique, which I think was a good suggestion from him.

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,76 @@
...
+    list-style-type: none;
+    margin: 0;
+    padding: 0;
+}
+
+.c-navigationLinks-item {
+    display: inline-block;
+    width: 100%;
+}
+.c-navigationLinks-link {
```
SachaZvetelman, line 61: Nit: empty line here, please :)

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
```diff
@@ -1,32 +1,76 @@
...
+    padding: spacing() 0 spacing() spacing(x2);
+    border-left: 2px solid $color-grey-30;
+    color: $color-grey;
+
+    &:focus, &:hover {
+        border-left: 2px solid $color-orange;
+        text-decoration: none;
+    }
+}
+.c-navigationLinks-link--active {
```
SachaZvetelman, line 73: Nit: empty line here, please :)

SachaZvetelman (dismissed):
Approved with some minor comments 👍

packages/components/molecules/f-navigation-links/src/components/NavigationLinks.vue
oliversweb: No guarantee as it's supplied by the consumer

SachaZvetelman (approved):

SachaZvetelman (approved):

packages/components/molecules/f-navigation-links/src/components/_tests/NavigationLinks.test.js
oliversweb: I could add a prop validator on a later pr 👍

ashleynolan (approved):

PR #1363, opened 2021-10-18
f-contact-preferences@0.3.0 - Initial response and mapping

xander-marjoram:
### Added
- Initial mapping and hard-coded mock

The axios request is commented out because the endpoint is not yet open to CORS, it can be tested within the SPA once the endpoint has CORS enabled.

The number of fieldsets is dynamic depending on the preferences response, and the copy is i18n-ified.


packages/components/organisms/f-contact-preferences/package.json
```diff
@@ -37,7 +37,8 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-services": "1.7.0"
+    "@justeat/f-services": "1.7.0",
+    "axios": "0.21.1"
```
DamianMullins: I've noticed we bundle the entire axios lib into a few of our components & services — we should be adding them as `peerDependency` and allowing the consuming application to pass a single version in instead.

When you add axios as a peer dep you'll need to give a range in the value also i.e. `"axios": ">=0.21.1"` otherwise we're saying the consuming app _must_ be using a specific version.

Alongside that you need to tell webpack to externalise the library in the `vue.config.js` file:

```js
configureWebpack: {
    externals: [
        'axios'
    ]
}
```

packages/components/organisms/f-contact-preferences/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "@justeat/f-contact-preferences",
   "description": "Fozzie Contact Preferences - Fozzie user contact preferences form component",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "main": "dist/f-contact-preferences.umd.min.js",
-  "maxBundleSize": "15kB",
+  "maxBundleSize": "20kB",
```
SachaZvetelman, line 6: How do we determine this number?

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -3,54 +3,47 @@
...
+                                <br>({{ copy[key].emailDescription }})
+                            </template>
+                        </span>
+                    </label>
+
+                    <label>
+                        <input
+                            type="checkbox"
+                            :disabled="!sms.enabled"
+                            :checked="sms.value">
```
SachaZvetelman, line 34: Can `sms` (or `email` ) be `null`?

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -62,34 +55,97 @@
...
         }
     },
+
     data () {
         const locale = globalisationServices.getLocale(tenantConfigs, this.locale, this.$i18n);
         const localeConfig = tenantConfigs[locale];
 
         return {
-            copy: { ...localeConfig }
+            copy: { ...localeConfig },
```
SachaZvetelman: Maybe I'm not seeing it, but is there any reason to spread the `localeConfig` which I assume is an object? Cannot we just do `copy: localeConfig` at this point?

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -62,34 +55,97 @@
...
+        try {
+            // const { data } = await axios.get(this.getPreferencesUrl, {
+            //     headers: {
+            //         'Content-Type': 'application/json',
+            //         ...(this.authToken && {
+            //             Authorization: `Bearer ${this.authToken}`
+            //         })
+            //     },
+            //     timeout: this.preferencesTimeout
+            // });
```
SachaZvetelman: Remember to remove before we merge :)

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -62,34 +55,97 @@
...
+                        sms: false,
+                        sort: 0
+                    }
+                ]
+            };
+
+            // this.$emit('preferences-get-success');
+            this.preferences = mapToPreferencesViewModel(data).preferences;
+        } catch (error) {
+            // TODO: WCB-2244
```
SachaZvetelman: Please remove 😉

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
xander-marjoram: I get it, but why? It's still a work-in-progress and this is the code that will be used eventually

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
xander-marjoram: No reason other than quicker to extend, I guess!
This is what comes straight out of the component generator.

packages/components/organisms/f-contact-preferences/package.json
xander-marjoram: There's no process defined for it yet. Really it should be as small as possible, but 16kB seems to be its current minimum size, unless f-services shrinks (or has shrunk since v1.7.0).

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
SachaZvetelman: Do you mean this PR is a WIP or the component is?
If it's going to be used, I don't mind, but the way I see it is that if we can remove unnecessary things from commits/PRs, it means each commit/PR has only the relevant changes to that commit/PR.
If you want to keep it, fine by me.

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
xander-marjoram: As far as I'm aware, no, it should be a boolean

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
xander-marjoram: Ah yes. Kind of irritating because this is best practice elsewhere!

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
DamianMullins: I did wonder this myself actually. For perf it's probably best to remove the spread if you're not extending it, though it's probably not having much of a negative effect.

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
xander-marjoram: Have removed

SachaZvetelman (dismissed):

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -62,34 +55,85 @@
...
         const localeConfig = tenantConfigs[locale];
 
         return {
-            copy: { ...localeConfig }
+            copy: localeConfig,
+            preferences: []
         };
     },
+
+    async mounted () {
```
kevinrodrigues: Is the async required here? Or is this for the catch later on?

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -62,34 +55,85 @@
...
         CardComponent,
         FButton
     },
+
     props: {
+        authToken: {
+            type: String,
+            default: ''
+        },
+        getPreferencesUrl: {
```
kevinrodrigues: Would it make more sense just calling this `preferencesUrl`?

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -3,54 +3,47 @@
...
-                </label>
-            </fieldset>
+            <div
+                v-for="{ email, key, sms } in preferences"
+                :key="key">
+                <h2 :class="$style['c-contactPreferences-subtitle']">
+                    {{ copy[key].subtitle }}
+                </h2>
+                <fieldset :class="$style['c-contactPreferences-fieldset']">
+                    <label>
```
kevinrodrigues, line 14: Is it worth pulling this markup out into a separate component as the email and sms markup looks very similar?

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
SachaZvetelman: I personally prefer `getPreferencesUrl`, since it's the URL for the `getPreferences` action, let's say 😬

tommcclean: Can you just request the prop as the base endpoint url. So if its smart gateway, smartGatewayBaseUrl etc. And build the rest of the link inside your component. I want to phase away from every URL being passed as an individual prop constructed in coreweb

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -62,34 +55,85 @@
...
+        },
+        getPreferencesUrl: {
+            type: String,
+            required: true
+        },
         locale: {
             type: String,
             default: ''
+        },
+        preferencesTimeout: {
```
tommcclean: I would argue not to even include this tbh; we so rarely ever use it and it is just as easy to update the component directly.

packages/components/organisms/f-contact-preferences/package.json
```diff
@@ -40,7 +40,8 @@
     "@justeat/f-services": "1.7.0"
   },
   "peerDependencies": {
-    "@justeat/browserslist-config-fozzie": ">=1.2.0"
+    "@justeat/browserslist-config-fozzie": ">=1.2.0",
+    "axios": ">=0.21.1"
```
tommcclean: 💡 - This might be a good opportunity to use `this.$http` and get free capture of response timings

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
xander-marjoram: Yes, ideally this would be using the checkbox from f-form-field but it's not ready yet. We've got a ticket in to-do for updating it, and then it can be used here.

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
xander-marjoram: Do we even want to refer to smartGateway? How about just apiBaseUrl?

packages/components/organisms/f-contact-preferences/package.json
xander-marjoram: Will remove axios for now, f-http can be added later when we need it 👍🏻

SachaZvetelman (dismissed):

packages/components/organisms/f-contact-preferences/stories/ContactPreferences.stories.js
```diff
@@ -11,10 +11,13 @@ export default {
...
 export const ContactPreferencesComponent = () => ({
-    components: { ContactPreferences },
+    components: {
+        ContactPreferences
+    },
     props: {
     },
-    template: '<contact-preferences />'
+    template: `<contact-preferences
+        get-preferences-url="/consumer/preferences" />`
```
tommcclean: I am guessing this is different now; might wanna swoop around on your next pr

tommcclean (dismissed):

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -3,54 +3,47 @@
...
-                    <span>
-                        {{ copy.newsAndOffers.email }}
-                    </span>
-                </label>
-            </fieldset>
+            <div
+                v-for="{ email, key, sms } in preferences"
+                :key="key">
+                <h2 :class="$style['c-contactPreferences-subtitle']">
+                    {{ copy[key].subtitle }}
```
DamianMullins, line 11: One for the future maybe; you could create a child component here that accepts either `copy[key]` itself or its properties as props. That way you can avoid the repeated calls to `copy[key]`.

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -62,34 +55,80 @@
...
         };
     },
+
+    mounted () {
+        try {
+            const data = {
+                preferencesVersionViewed: 0,
+                preferences: [
+                    {
+                        displayName: 'orderUpdates',
```
DamianMullins, line 107: Are the additional keys here, like `displayName`, to be used in the future?

packages/components/organisms/f-contact-preferences/vue.config.js
```diff
@@ -5,6 +5,11 @@ const sassOptions = require('../../../../config/sassOptions')(rootDir);
 
 // vue.config.js
 module.exports = {
+    configureWebpack: {
+        externals: [
+            'axios'
```
DamianMullins: You can remove this config if you are no longer using `axios`.

packages/components/organisms/f-contact-preferences/src/services/mapping/index.js
```diff
@@ -0,0 +1,21 @@
+const mapToPreferencesViewModel = data => ({
+    preferences: data.preferences.map(p => ({
```
DamianMullins: Might be personal preference, but I like to restructure the properties when mapping like this i.e.

```js
preferences: data.preferences.map(({ key, email, sms }) => ({
```

packages/components/organisms/f-contact-preferences/src/components/ContactPreferences.vue
xander-marjoram: This is just a copy of the expected response at the moment. Once we're reading it from the API directly, the mapper will handle the required properties, and we can forget about the rest 😄

DamianMullins: Gotcha.

packages/components/organisms/f-contact-preferences/src/services/mapping/index.js
xander-marjoram: Good shout, I'll destructure `data` too.

DamianMullins (approved):

kevinrodrigues (approved):

PR #1372, opened 2021-10-18
f-button@3.2.0 - change of button height

dandel10n:
### Changed
- Button border style definition moved to button types.
- Each button size vertical padding reduced by 1px to make button height 56px/48px/40px/32px instead of 58px/50px/42px/34px.

### Removed
- Outline css styles as focus styles applied with `box-shadow`.


FayeCarter (dismissed):

packages/components/atoms/f-button/CHANGELOG.md
```diff
@@ -4,6 +4,32 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v3.2.0
+------------------------------
+*October 18, 2021*
+
+### Changed
+- Button border style definition moved to button types.
+- Each button size vertical padding reduced by 1px to make button height 56px/48px/40px/32px instead of 58px/50px/42px/34px.
```
ashleynolan, line 13: Just re-adding my comment from the other PR 😆 

> I'm just checking storybook – it looks like the large button already is 56px (and then it has a 1px transparent border, which means devtools says it has a 58px height).

> I think we've been back and forth a few times on button and form-field heights with design in the past, so probably worth catching up about this (and probably writing down whatever we do somewhere, cause I always forget!)

ashleynolan (changes requested):

packages/components/atoms/f-button/CHANGELOG.md
dandel10n: I now changed the way we add borders to have consistent height between buttons without visual borders and with visual border (outline button).

ashleynolan: Have caught up with @dandel10n now – going to move the background and border style setting into a mixin that is commented so that it's a bit more manageable (and so we can comment it for anyone coming to these styles)

ashleynolan (dismissed):

packages/components/atoms/f-button/CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v3.2.0
+------------------------------
+*October 22, 2021*
+
+### Changed
+- Button border style definition moved to button types. Introdused `background-color()` mixin to apply the same colour for the background and for the border colour.
```
xander-marjoram: Nit: typo
```suggestion
- Button border style definition moved to button types. Introduced `background-color()` mixin to apply the same colour for the background and for the border colour.
```

packages/components/atoms/f-button/src/assets/scss/common.scss
```diff
@@ -5,3 +5,8 @@ $theme: 'jet' !default;
 
 @import  '~@justeat/fozzie/src/scss/fozzie';
 
+// mixin to aply the same colour for the border and for background of the button
```
xander-marjoram: Nit: typo
```suggestion
// mixin to apply the same colour for the border and for background of the button
```

packages/components/atoms/f-button/CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v3.2.0
+------------------------------
+*October 22, 2021*
```
xander-marjoram: (Might as well bump the date while you're at it too 😄)

xander-marjoram (approved):

FayeCarter (approved):

PR #1375, opened 2021-10-19
f-promotions-showcase@0.1.0 🌟 Creates initial version from generator

minrwhite:
### Added
- Initial version of promotions-showcase component - created from generator
- Added circleci dist cache config

The intention is that this component becomes a place that particular promotional messaging can be housed, with flexibility to support different needs depending on the consuming component. The initial goal for this is the Menu page in order to surface the different offers/offer types that the restaurant supports, and as is currently rolling out for apps. An example of how this will look can be seen here:
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,6 +70,7 @@ commands:
             - packages/components/molecules/f-breadcrumbs/dist
             - packages/components/molecules/f-media-element/dist
             - packages/components/molecules/f-mega-modal/dist
+            - packages/components/molecules/f-promotions-showcase/dist
             - packages/components/molecules/f-restaurant-card/dist
             - packages/components/molecules/f-searchbox/dist
             - packages/components/molecules/f-skeleton-loader/dist
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v5.0.1
+ ------------------------------
+ *October 20, 2021*
+
+### Added
+- `f-promotions-showcase` to Circle CI cache
+
+
 v5.0.0
 ------------------------------
  *October 21, 2021*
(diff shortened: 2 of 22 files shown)


dandel10n:
@benwhite-deltas could you please add more details about the component to the description (the purpose, maybe screenshots of what it will look like etc)?

FayeCarter (dismissed):

ashleynolan:
This feels like it could be more like an organism than a molecule (as it could potentially include a number of other components?). Would be interested to know your thinking though @benwhite-deltas

minrwhite:
My thinking on the taxonomy of this was it's of similar functionality and size to a modal, which is one of the examples given for a molecule (and indeed `f-mega-modal is a molecule). I was swithering on it though, so happy to move to an organism if you have strong feelings on it, @ashleynolan ?

ashleynolan:
@benwhite-deltas I've not got strong feelings – looking at your PR just made me think we probably need a better definition though as it's hard with components that are borderline like this.

Just doing a quick read around, Brad Frost (the creator of Atomic Design) says:

> My general rule of thumb is that if you break a component down and get basic tags (such as an image, heading, paragraph, button, etc), it’s a molecule. If you break the component down and get smaller components (social share lists, cards, media object, etc) it’s an organism

So maybe, we should look at this as a definition going forward (I want to rejig some of the molecules and organisms when we look at moving the `page` level components to their own folder, so will likely use the above as a good guide). 

With the above – does it feel like this would be more of a molecule or an organism potentially against that definition?

ashleynolan:
@benwhite-deltas Just FYI – I think this feels like a molecule still to me, but I'm not sure what you would anticipate it would be made up of, so you may have a better idea. Equally, we can always recategorise if we feel it gets more like an organism in future.

minrwhite:
Based on that definition, I would 100% go with molecule then, @ashleynolan - the PoC I have locally doesn't use smaller components barring svg logos imported as vue components (though is too big in it's own right to be an atom) but instead takes configuration that it uses to construct it's contents.

jamieomaguire (approved):

dandel10n (approved):

PR #1376, opened 2021-10-19
f-statistics@v0.4.0 - Ability to provide properties during initialisation

tommcclean:
We want to add properties during initialisation from the web host, which will go up with every stat. For things like experiment / feature buckets, or code versions.

Non breaking and optional.

https://github.com/tommcclean/fozzie-components/blob/f-statistics%40v0.4.0/packages/services/f-statistics/README.md

```
v0.4.0
*October 19, 2021*

### Added
- Ability to pass in custom properties during initialization which stick around
- Above properties are now published with every statistic
```

### DIFF
--- a/packages/services/f-statistics/CHANGELOG.md
+++ b/packages/services/f-statistics/CHANGELOG.md
@@ -3,6 +3,15 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.4.0
+------------------------------
+*October 19, 2021*
+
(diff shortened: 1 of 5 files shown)


xander-marjoram:
FYI the github labeler action doesn't work when you create a PR from a fork

tommcclean:
> FYI the github labeler action doesn't work when you create a PR from a fork

Ah okay was wondering what was going on there, is it needed? Doesn't seem to be a required check

xander-marjoram:
@tommcclean For a service, it's probably not needed, but for visual components it's the action that triggers the Percy screenshots so probably better not to work off a fork, especially in that case

tommcclean:
Moved off fork

PR #1377, opened 2021-10-20
f-statistics@v0.4.0 - Ability to provide properties during initialisation

tommcclean:
We want to add properties during initialisation from the web host, which will go up with every stat. For things like experiment / feature buckets, or code versions.

Non breaking and optional.

https://github.com/tommcclean/fozzie-components/blob/f-statistics%40v0.4.0/packages/services/f-statistics/README.md

```
v0.4.0
*October 19, 2021*

### Added
- Ability to pass in custom properties during initialization which stick around
- Above properties are now published with every statistic
```

### DIFF
--- a/packages/services/f-statistics/src/index.js
+++ b/packages/services/f-statistics/src/index.js
@@ -21,8 +21,8 @@ export default class StatisticsClient {
 
     publish (message, statisticPayload) {
         this.justLogInstance.info(message, {
-            ...this.basePayload,
-            ...statisticPayload
+            ...statisticPayload,
+            ...this.basePayload
         });
     }
 }
--- a/packages/services/f-statistics/src/index.js
+++ b/packages/services/f-statistics/src/index.js
@@ -8,25 +8,25 @@ export default class StatisticsClient {
     };
 
     constructor (justLogInstance, options = {}, basePayload = {}) {
-        this.basePayload = {
-            ...basePayload,
-            ...this.basePayload
-        };
-
         this.configuration = {
             ...defaultOptions,
             ...options
         };
 
+        this.basePayload = {
+            ...basePayload,
+            ...this.basePayload,
+            je_feature_for: this.configuration?.featureName || 'Unspecified',
+            je_environment: this.configuration?.environment || 'Unspecified'
+        };
+
         this.justLogInstance = justLogInstance;
     }
 
-    publish (message, payload) {
+    publish (message, statisticPayload) {
         this.justLogInstance.info(message, {
-            je_feature_for: this.configuration?.featureName || 'Unspecified',
-            je_environment: this.configuration?.environment || 'Unspecified',
             ...this.basePayload,
-            ...payload
+            ...statisticPayload
         });
     }
 }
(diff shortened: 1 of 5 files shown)


packages/services/f-statistics/src/tests/index.test.js
```diff
@@ -6,8 +7,118 @@ describe('f-statistics', () => {
         expect(StatisticsClient).toBeDefined();
     });
 
-    it('should define expected method', () => {
+    describe('constructor ::', () => {
+        it('should expose base payload when none is provided', () => {
+            // Arrange
+            const expectedPayload = {
+                je_feature: 'f-statistics',
```
xander-marjoram: I'd maybe pull this out into a `baseExpectedPayload` and then for some of the later tests you can do stuff like this:
```javascript
const expectedPayload = {
    ...baseExpectedPayload,
    a_test_property: 'this is a test'
};

File: packages/services/f-statistics/src/tests/index.test.js [hunk 2]
```diff
@@ -6,8 +7,116 @@ describe('f-statistics', () => {
...
-        expect(new StatisticsClient().publish).toBeDefined();
+            expect(new StatisticsClient().publish).toBeDefined();
+        });
+
+        it('should publish logs with expected properties', () => {
+            // Arrange
+            const statisticsClient = new StatisticsClient(justLogMock, null, null);
+
+            const expectedPayload = {
+                je_environment: 'test',
```
xander-marjoram: Nit: Looks like you could you use the `basePayload` here too?

packages/services/f-statistics/src/tests/index.test.js
```diff
@@ -6,8 +7,116 @@ describe('f-statistics', () => {
...
+            // Assert
+            expect(publications[0]).toEqual(expectedPayload);
+        });
+
+        it('should publish logs with additional base payload properties when provided', () => {
+            // Arrange
+            const statisticsClient = new StatisticsClient(justLogMock, null, { a_base_payload_property: 'This is a test' });
+
+            const expectedPayload = {
+                je_environment: 'test',
```
xander-marjoram: And here

xander-marjoram (dismissed):

packages/services/f-statistics/src/index.js
```diff
@@ -2,22 +2,27 @@
...
-    publish (message, payload) {
+    publish (message, statisticPayload) {
         this.justLogInstance.info(message, {
-            je_feature: 'f-statistics',
-            je_feature_for: this.configuration?.featureName || 'Unspecified',
-            je_logType: 'client-stats',
-            je_environment: this.configuration?.environment || 'Unspecified',
-            ...payload
+            ...this.basePayload,
+            ...statisticPayload
```
minrwhite: Are we okay with allowing keys from the basePayload (`je_feature` et al) to potentially be overridden here?

tommcclean: I kind of don't mind it; but it makes sense to invert this as I already do this in the constructor with the base - so consistency I guess.

minrwhite (approved):

xander-marjoram (approved):

PR #1380, opened 2021-10-20
f-checkout@v3.9.0  Prevent submission of user details on age verification.

BanksySan:
Sending an update after age verification causes errors as the other fields are required.

This removed that behaviour, the age is simply stored and the whole user state is stored when the checkout form is submitted.


packages/components/organisms/f-checkout/src/components/_tests/AgeVerification.test.js
```diff
@@ -308,9 +308,9 @@ describe('AgeVerification', () => {
                     expect(updateDateOfBirthSpy).toHaveBeenCalledWith(date);
                 });
 
-                it('should emit `CheckoutVerifyAge`', () => {
+                it('should emit not `CheckoutVerifyAge`', () => {
```
FayeCarter: if you've taken the emit out you can probably delete this test

BanksySan: Because it's been changed to explicitly not commit at that stage I thought that the test to assert that was worth it.  Tests serve two purposes, they confirm functionality and they document expected behaviour.

packages/components/organisms/f-checkout/src/components/Checkout.vue
```diff
@@ -1087,10 +1076,11 @@ export default {
 
         getMappedDataForUpdateCheckout (options = { ageVerificationOnly: false }) {
             const { ageVerificationOnly } = options;
```
FayeCarter: I think cos you've deleted the `verifyCustomerAge` method you will never call this with `{ ageVerificationOnly: true }` so you can get rid of the ternary?

If you're doing that you can then probably get rid of this whole method and just call `mapUpdateCheckoutRequest` from wherever else this is called

packages/components/organisms/f-checkout/src/store/checkout.module.js
```diff
@@ -439,6 +442,7 @@ export default {
 
         updateDateOfBirth: ({ commit }, dateOfBirth) => {
             commit(UPDATE_DATE_OF_BIRTH, dateOfBirth);
+            commit(CLEAR_DOB_ERROR, {});
```
FayeCarter: I think we might need a test for this and the mutation below

FayeCarter (dismissed):

FayeCarter (dismissed):

BanksySan (commented):
LGTM.

FayeCarter (approved):

charliearlie (approved):

PR #1382, opened 2021-10-20
generator-component@2.3.2 - Modify generator to allow services to be created correctly.

kevinrodrigues:
### Changed
- Template package.json so it outputs the correct settings for services.
- index.js (within src directory) which gives us the `isService` setting available within the config.

### Added
- Vite to the template folder for use with services.

### Removed
- f-services from service specific generation.

### Screenshot (services directory):
Screenshot 2021-10-20 at 15 51 35

### Screenshot (services package.json generation):

Screenshot 2021-10-20 at 15 54 44
Screenshot 2021-10-20 at 15 54 54

### Screenshot (components directory - to making sure it's the same as before):
Screenshot 2021-10-20 at 15 56 14

### Screenshot (components package.json - to making sure it's the same as before)
Screenshot 2021-10-20 at 15 56 26
:


packages/tools/generator-component/generators/app/templates/__package__.json
```diff
@@ -42,21 +46,24 @@
...
+<% if(config.isComponent) { %>    "@justeat/f-wdio-utils": "0.3.0",
+    "node-sass-magic-importer": "5.3.2",
     "@samhammer/vue-cli-plugin-stylelint": "2.0.1",
     "@vue/cli-plugin-babel": "4.4.6",
     "@vue/cli-plugin-eslint": "3.9.2",
     "@vue/cli-plugin-unit-jest": "4.4.6",
-    "@justeat/f-wdio-utils": "0.3.0",
-    "@vue/test-utils": "1.0.3"<% if(config.isComponent) { %>,
-    "node-sass-magic-importer": "5.3.2"<% } %>
+    "@vue/test-utils": "1.0.3"<% } else { %>    "vite": "2.6.9",
```
DamianMullins, line 64: Think the `vite` part needs to be on the next line?

DamianMullins: Ah, is it there to prevent a blank line?

ashleynolan (approved):

DamianMullins (approved):

packages/tools/generator-component/generators/app/templates/__package__.json
kevinrodrigues: Yeah was leaving odd gaps if it wasn't in the right area :| this sorts it.

PR #1383, opened 2021-10-21
f-restaurant-card@v0.4.0 - Add Error Boundary Wrapper Injection

dawidchar:
This PR adds the ability to inject error boundaries into the component, error boundaries can be used to capture errors within 
a subcomponent, such that if a subcomponent like the ratings fails, it does not take down the whole restaurant component but is contained within the boundary. The error boundary then handles logging the error.

The plan is to eventually move the error boundary into fozzie as a more generic component, however this will be a separate piece of work in the future.

---

## Browsers Tested

- [X] Chrome (latest)
- [ ] Internet Explorer 11
- [X] Mobile (Please list device/browser – Ideally one iPhone model and one Android)


jamieomaguire (approved):

dandel10n (approved):
Label watcher fails because the PR is made from a fork. Because we now use github actions where we trigger the Circle CI pipeline, the secrets are not passed to the runner when a workflow is triggered from a forked repository. 

To fix this PR you can add original repo as another remote to your fork:
`git add remote fozzie https://github.com/justeat/fozzie-components.git`
And then, from the branch
`git push fozzie`

jamieomaguire:
@dandel10n sorry I spotted this after merging - but yes we will make sure to just branch off of upstream in future and no longer use forks

PR #1384, opened 2021-10-21
f-breadcrumbs@2.1.0 - Render <span> not <a> if no URL is given

xander-marjoram:
### Changed
- If no URL is given, render `<span>` instead of `<a>` or `<router-link>`.

### Added
- Unit test coverage

### DIFF
--- a/packages/components/molecules/f-breadcrumbs/src/components/Breadcrumbs.vue
+++ b/packages/components/molecules/f-breadcrumbs/src/components/Breadcrumbs.vue
@@ -14,7 +14,7 @@
                     :key="`${index}_separator`"
                     :class="$style['c-breadcrumbs-item']">
                     <router-link
-                        v-if="routerLink"
+                        v-if="routerLink && url"
                         :to="url"
                         :class="[
                             $style['c-breadcrumbs-link'],
@@ -23,7 +23,7 @@
                         {{ name }}
                     </router-link>
                     <a
-                        v-else
+                        v-else-if="url"
                         :href="url"
                         :class="[
                             $style['c-breadcrumbs-link'],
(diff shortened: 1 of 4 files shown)


kevinrodrigues (approved):

AndyMayJE (approved):

packages/components/molecules/f-breadcrumbs/src/components/Breadcrumbs.vue
```diff
@@ -23,14 +23,22 @@
...
-                        v-else
+                        v-else-if="url"
                         :href="url"
                         :class="[
                             $style['c-breadcrumbs-link'],
                             linkActiveClass(index)
                         ]">
                         {{ name }}
                     </a>
+                    <span
```
ashleynolan, line 34: Out of interest – does this need to be in a `span` at all? Could you not add the default styles to the `li` and then just add the link tag if a URL is present?

xander-marjoram: Potentially not, although I think it makes the code a little more readable and sits nicely with the if/else-if/else pattern.

DamianMullins: You could use `template` in place of the `span` if you want to keep the if/else pattern?

ashleynolan: Yeh, that's what I was thinking. For me, markup should only be there if it's needed semantically or for hooking styles onto, so I'd look to remove this for a `template` tag if possible.

xander-marjoram: I'll test it out

PR #1385, opened 2021-10-21
fozzie-components@v5.0.0 – Adding directories for templates and page components

ashleynolan:
### Added
 - Component folders for `templates` and `pages`.
 - Templates are for layout components that are intended to handle page-level layout decisions (for example, a sideNav layout such as on our account pages). Templates don't contain any logic around the actual set of pages, they just handle the overall layout structure and styling.
 - Pages are for our largest organisms, whereby they are a collection of components that make up the main body of a page. For example, `f-registration` and `f-checkout` are both examples of `page` components.

 ### Changed
 - Config for CircleCI, Lerna, yarn workspaces and danger.js updated to include new folder paths.
 - Moved `f-registration` into `pages` directory. Will look to move other organisms one-by-one by liaising with each team working on these components.


.circleci/config.yml
```diff
@@ -100,10 +100,21 @@ commands:
             - packages/components/organisms/f-header/dist
             - packages/components/organisms/f-loyalty/dist
             - packages/components/organisms/f-offers/dist
-            - packages/components/organisms/f-registration/dist
             - packages/components/organisms/f-status-banner/dist
             - packages/components/organisms/f-takeawaypay-activation/dist
 
+      # TODO: Uncomment this once we have a template component
```
zlatinivanov-je, line 106: Would this break if left un-commented? Just thinking when the first template is created, the person will need to remember to un-comment. Will it be obvious then?

zlatinivanov-je (commented):
One question to clarify

.circleci/config.yml
ashleynolan: I wasn't sure if it would break – which is why I commented it out 😆  I know we're going to be adding one in the next week or so though, so I'll make sure this gets uncommented when that happens 👍

zlatinivanov-je: Have you tried having it uncommented? That way will be off your hands once merged ;)

zlatinivanov-je (approved):
Happy to approve

.circleci/config.yml
matthewhardern: when this first component is added to this they would need to add cache line anyway atm so they would see this

matthewhardern (approved):

kevinrodrigues (approved):

.github/component-labeler.yml
```diff
@@ -8,6 +8,12 @@ molecules:
 organisms:
 - 'packages/components/organisms/**/*'
 
+templates:
```
xander-marjoram, line 11: Just FYI I'm not sure what happens if the label doesn't already exist, so I've just made a couple for you:
https://github.com/justeat/fozzie-components/issues/labels

ashleynolan: Thanks!

PR #1386, opened 2021-10-21
Refactor of f-feature-management

jnewc:
A few changes for the Feature Management service:
 * Allow injection of axios client
 * Delegate tracking to integrating client instead of calling data layer directly (allows tracking w/ SSR)
 * Separate config file loading from initialisation, and return config to service consumers to allow more flexible caching
 * Test fixes / updates / cleanup

---


packages/services/f-feature-management/src/index.js
```diff
@@ -5,6 +5,6 @@ import createFeatureManagementInstance from './createFeatureManagementInstance';
  * @param {object} settings - json and contextGetter must be set
  * @returns Feature Management object to allow querying of features.
  */
-export default function (settings) {
-    return createFeatureManagementInstance(settings);
+export default function (settings, axiosClient) {
```
xander-marjoram: You should probably add a line to the JSDoc (comment above the function) to include the axios client

packages/services/f-feature-management/src/trackExperiment.js
```diff
@@ -1,26 +1,30 @@
+import isFunction from 'is-function';
```
xander-marjoram: Is this package necessary?

packages/services/f-feature-management/test/configStore.test.js
```diff
@@ -1,151 +1,144 @@
...
 
-    //Act
-    await loadFromCdn(newSettings);
+        jest.advanceTimersByTime(100000);
 
-    //Assert
-    expect(fetch).toHaveBeenCalledWith(`https://test.com/config/v1/${cdnSettings.scope}/${cdnSettings.environment}-${cdnSettings.key}`);
-  });
+        // Assert
+        expect(mockAxios.get).toHaveBeenCalledTimes(2);
```
xander-marjoram: This doesn't read quite right, IMO, I would expect it to say 1?

If it's carrying over state between the tests, you could try running `jest.resetAllMocks()` in the `afterEach`.

packages/services/f-feature-management/test/configStore.test.js
```diff
@@ -1,151 +1,144 @@
...
-    await loadFromCdn(newSettings);
+        // Assert
+        expect(mockAxios.get).toHaveBeenCalledWith(`https://test.com/config/v1/${cdnSettings.scope}/${cdnSettings.environment}-${cdnSettings.key}`);
+    });
 
-    //Assert
-    expect(fetch).toHaveBeenCalledTimes(2);    
+    it('should honour different pollInterval', async () => {
+        // Arrange
+        const newSettings = { ...cdnSettings, host: 'https://test.com', pollInterval: 60000 };
```
xander-marjoram: Multiline for legibility, please

packages/services/f-feature-management/test/getBooleanValue.test.js
```diff
@@ -0,0 +1,23 @@
+import featureManagement from '../src/index';
+
+const { getBooleanValue } = featureManagement('');
+
+describe.skip('Boolean value retrieval', () => {
```
xander-marjoram: Should these still be skipped?

packages/services/f-feature-management/test/getIntegerValue.test.js
```diff
@@ -0,0 +1,29 @@
+import featureManagement from '../src/index';
+
+const { getIntegerValue } = featureManagement('');
+
+describe.skip('Integer value retrieval', () => {
```
xander-marjoram: Same question

packages/services/f-feature-management/test/getStringValue.test.js
```diff
@@ -0,0 +1,35 @@
+import featureManagement from '../src/index';
+
+const { getStringValue } = featureManagement('');
+
+describe.skip('String value retrieval', () => {
```
xander-marjoram: Same question

packages/services/f-feature-management/test/getValue.test.js
```diff
@@ -0,0 +1,41 @@
+import featureManagement from '../src/index';
+
+const { getValue } = featureManagement('');
+
+describe.skip('Generic value retrieval', () => {
```
xander-marjoram: Same question

packages/services/f-feature-management/src/index.js
xander-marjoram: We could probably be even more generic so that it's called `httpClient` instead of `axiosClient`, to encourage the use of `f-http`.

xander-marjoram:
Also, it's less important for services, but FYI, some of our GitHub Actions don't work when triggered from forks, so we recommend PRs via branches on the justeat repo.

packages/services/f-feature-management/src/trackExperiment.js
jnewc: I thought this was something intended to be used transitively but I can't see it anywhere else. Will revert to `typeof(x) === 'function'` 👍

packages/services/f-feature-management/test/configStore.test.js
jnewc: It's called once by `executeSUT` and then once when polling occurs at ~60s, so I'm fairly certain it should be two. However, the previous one (L63) looks like it should be `1`.

packages/services/f-feature-management/test/getBooleanValue.test.js
jnewc: I think these may have slipped back in when I merged the latest as I'm pretty sure Gareth moved these elsewhere. Will investigate and remove if possible.

packages/services/f-feature-management/test/configStore.test.js
jnewc: I've moved the `executeSUT` calls out of `beforeEach` and in to the individual tests. This reduces the number of calls made and makes the tests more readable.

packages/services/f-feature-management/test/getBooleanValue.test.js
jnewc: Removed as these are covered by e2e. Same for all below.

jnewc:
Closing. Moving this to a local branch and reopening

PR #1389, opened 2021-10-25
f-link@2.1.0 - Support router-link

xander-marjoram:
### Added
- `storybook-vue-router` setup to work with `f-link`.

### Added
- Support for `router-link`.
- `isRouterLink` prop.

### Changed
- `href` from implicit attribute to (required) prop.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/atoms/f-link/README.md
+++ b/packages/components/atoms/f-link/README.md
@@ -71,7 +71,9 @@ The props that can be defined are as follows (if any):
 
 | Prop  | Type  | Default | Description |
 | ----- | ----- | ------- | ----------- |
-| `isExternalSite` | `String` | `false` | Sets aria description to 'Opens and external site' or 'Opens and external site in a new window/screen/tab' depending on target of link.|
+| `href` | `String` | This is a required prop | The URL or path of the link. For a `router-link` this will be mapped to the `to` attribute. |
+| `isExternalSite` | `Boolean` | `false` | Sets aria description to 'Opens and external site' or 'Opens and external site in a new window/screen/tab' depending on target of link.|
+| `isRouterLink` | `Boolean` | `false` | Determines whether to configure the link as a Vue Router router-link. |
 | `isBold` | `Boolean` | `false` | Sets link text to bold. |
 | `hasTextDecoration` | `Boolean` | `true` | Adds underline to link text. |
 | `isFullWidth` | `Boolean` | `false` | Link set as full width. |
@@ -126,5 +128,3 @@ cd ./fozzie-components/packages/components/atoms/f-link
 yarn test-component:chrome
 ```
 ## Documentation to be completed once module is in stable state.
-
-
--- a/packages/storybook/CHANGELOG.md
+++ b/packages/storybook/CHANGELOG.md
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.46.0
+------------------------------
+*October 25, 2021*
+
+### Added
+- `storybook-vue-router` setup to work with `f-link`.
+
+
 v0.45.1
 ------------------------------
 *October 22, 2021*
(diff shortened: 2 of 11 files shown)


oliversweb (approved):

packages/storybook/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.46.0
+------------------------------
+*October 25, 2021*
+
+### Added
+- `storybook-vue-router` setup to work with `f-link`.
```
dandel10n, line 12: Nice! Should we clean up an import af actual Router from a f-button story now?

kevinrodrigues (approved):

packages/storybook/CHANGELOG.md
xander-marjoram: @dandel10n Yeah, I don't see why not!

packages/components/atoms/f-link/README.md
```diff
@@ -71,7 +71,9 @@ The props that can be defined are as follows (if any):
 
 | Prop  | Type  | Default | Description |
 | ----- | ----- | ------- | ----------- |
-| `isExternalSite` | `String` | `false` | Sets aria description to 'Opens and external site' or 'Opens and external site in a new window/screen/tab' depending on target of link.|
+| `href` | `String` | This is a required prop | The URL or path of the link. For a `router-link` this will be mapped to the `to` attribute. |
+| `isExternalSite` | `Boolean` | `false` | Sets aria description to 'Opens and external site' or 'Opens and external site in a new window/screen/tab' depending on target of link.|
+| `isRouterLink` | `Boolean` | `false` | Determines whether to configure the link as a Vue Router router-link. |
```
DamianMullins, line 76: f-button automatically detects router links and configures them accordingly, is that something we could here as well rather than having to set a prop?

ashleynolan: Is this possible? Would be cool to avoid the prop if that's possible

DamianMullins: It looks like [`f-button` looks at the `$attrs` property](https://github.com/justeat/fozzie-components/blob/8db7dd716e86cc80988e6f48e8e5a616f08e9982/packages/components/atoms/f-button/src/components/Button.vue?_pjax=%23repo-content-pjax-container#L123-L133) to differentiate between the different types of components.

xander-marjoram: I think it's six of one and half-a-dozen of the other.

You either pass in a different attribute (`href` vs `to`) for a `router-link`, or you could pass in the same attribute 
(`href`) and an optional boolean flag (`isRouterLink`).

I actually preferred this latter approach because it's easier to understand what the code is doing. If we're treating an attribute as a prop we might as well make it one, `$attrs` can be a bit of a black box, and the errors it gives aren't as good as something like `"missing prop 'href'"`.

It's more semantic as the resulting HTML either way will be along the lines of `<a href="/" />`.

DamianMullins: yeah, I get what you mean to a point. One of the reasons I asked about removing the prop is because with just the single prop when setting up a router-link style link, you'd have to pass something like this `:href="{ name: 'my-route' }"`. That doesn't read right to me, IMO `href` should always be a string value.

I've just been converting menu to use `f-button` and, to me, it was super clear that when you use `href` you're dealing with a link and when you use `to` you're using a router link.

My other issue with this approach is that we now have two similar components that use two different patterns, it would be good to keep them consistent in order to avoid confusion.

xander-marjoram: Ok. I'd just come off working on the breadcrumb component and that had a nice Boolean handy flag called `routerLink` and I thought it would be better to match it up with that.

The distinction between `href` always being a string and `to` being either a string or an object makes sense.

PR #1390, opened 2021-10-25
f-breadcrumbs@3.0.0 - Added ability to add a background to breadcrumbs.

kevinrodrigues:
Note: If you are using a background (i.e hasBackgroundEnabled = true, type of style) you will need to add the prop to where you render the component otherwise you will lose the background style.

### Added
- `hasBackgroundEnabled` prop which adds a layer of style to the default version.
- Unit test coverage.

### Screenshot (hasBackgroundEnabled = false / default)

Screenshot 2021-10-25 at 11 12 21

### Screenshot (hasBackgroundEnabled = true)
Screenshot 2021-10-25 at 11 14 16

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)


kevinrodrigues:
@matthewhardern FYI as i believe you use a background in offers.

packages/components/molecules/f-breadcrumbs/README.md
```diff
@@ -55,6 +55,16 @@
...
+## Configuration
+
+### Props
+
+The props that can be defined are as follows:
+
+| Prop  | Type  | Default | Description |
+| ----- | ----- | ------- | ----------- |
+| `links` | `Array - object` | - | Defines a list of links to be rendered by f-breadcrumbs |
+| `hasBackgroundEnabled` | `Boolean` | false | Defines if the breadcrumb can have a background or not |
```
xander-marjoram: Can, or will?

packages/components/molecules/f-breadcrumbs/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-breadcrumbs",
   "description": "Fozzie Bread Crumbs – Provides clickable paths back to previous pages",
-  "version": "2.1.0",
+  "version": "3.0.0",
```
xander-marjoram, line 4: Are these breaking changes?

kevinrodrigues: Yeah, visually the component won't look as intended if the consumer of the component does not actively add the `hasBackgroundEnabled` prop. `Offers` requires the background for their breadcrumbs so for them it will break if they upgrade but do not add this prop.

xander-marjoram (dismissed):

packages/components/molecules/f-breadcrumbs/src/components/Breadcrumbs.vue
```diff
@@ -62,70 +70,99 @@ export default {
...
+                : '';
         }
     }
 };
 </script>
 
 <style lang="scss" module>
 
-$breadcrumbs-text-colour: $color-content-light;
+$breadcrumbs-text-colour: $color-grey-70;
```
minrwhite: IMO it would be good to split this into a `background`/`nobackground` definition pair, keep the defined reference to the pie token `$color-content-light` and add a new reference to `$color-content-default` for the `nobackground` colour?

minrwhite (commented):
One suggestion but otherwise LGTM

packages/components/molecules/f-breadcrumbs/README.md
```diff
@@ -55,6 +55,16 @@
...
     ```
+## Configuration
+
+### Props
+
+The props that can be defined are as follows:
+
+| Prop  | Type  | Default | Description |
+| ----- | ----- | ------- | ----------- |
+| `links` | `Array - object` | - | Defines a list of links to be rendered by f-breadcrumbs |
```
oliversweb, line 66: You might want to add an example of what the `object` in the array needs to look like, e.g.:
```
[
        {
            name: 'Home',
            url: '/',
            routerLink: false
        },
        {
            name: 'Accounts',
            url: '/accounts',
            routerLink: true
        },
        {
            name: 'Preferences'
        }
]
```

packages/components/molecules/f-breadcrumbs/src/components/Breadcrumbs.vue
ashleynolan: Always best to use alias tokens if possible, as they're less brittle than global tokens.

I think this one could use `$color-content-default` for instance (and I agree with @benwhite-deltas 
suggestion too, that there should be a variable for both states)

packages/components/molecules/f-breadcrumbs/src/components/Breadcrumbs.vue
```diff
@@ -62,70 +70,99 @@ export default {
...
 }
-.c-breadcrumbs-text {
+
+.c-breadcrumbs-text,
+.c-breadcrumbs-link {
     color: $breadcrumbs-text-colour;
     font-weight: $breadcrumbs-not-active-font-weight;
+
+    .c-breadcrumbs-list--hasBackground & {
+        color: $color-white;
```
ashleynolan: As @benwhite-deltas mentioned above – I'd make a variable for this text color and then use an alias token for this value (and update the reference here and below)

packages/components/molecules/f-breadcrumbs/src/components/Breadcrumbs.vue
kevinrodrigues: Just got round to fixing this. Let me know if it is as intended now.

packages/components/molecules/f-breadcrumbs/src/components/Breadcrumbs.vue
kevinrodrigues: Fixed.

packages/components/molecules/f-breadcrumbs/README.md
kevinrodrigues: Fixed.

minrwhite (dismissed):

packages/components/molecules/f-breadcrumbs/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v3.0.0
+------------------------------
+*October 25, 2021*
+
+### Added
+- `hasBackgroundEnabled` prop which adds a layer of style to the default version.
```
DamianMullins: This might be me being pedantic but I'd change this to `hasBackground`?

kevinrodrigues: Fixed. I actually had that like that before i opened the PR then decided to change it (not sure why).

xander-marjoram (approved):

minrwhite (approved):

DamianMullins (approved):

PR #1393, opened 2021-10-25
f-compatibility@v0.2.0 - IE11 Warning service with banner

simonsandersje:
v0.2.0
*October 25, 2021*

### Added
- Using Bowser to check for incompatible browsers
- Data for unsupported browsers (currently only IE11)
- Content for English, Spanish and Italian tenants
- Display banner for unsupported browsers (currently only IE11)

## UI Review Checks

f-compatibility - it
f-compatibility - uk
f-compatibility - es

## Browsers Tested

- [x] Chrome (latest)
- [x] Internet Explorer 11

- [x] FireFox (latest)
- [x] Edge (latest)


packages/services/f-compatibility/src/index.js
```diff
@@ -1,21 +1,76 @@
...
-
 /**
 * @overview Service to check browser compatibilities and warn users with incompatible browsers
 *
 * @module f-compatibility
 */
 
+import { UNSUPPORTED_BROWSERS, BROWSER_DATA } from './constants';
+
+const Bowser = require('bowser/bundled');
```
DamianMullins: I'd lowercase this variable if it isn't being used to create an instance with `new`.

packages/services/f-compatibility/src/index.js
```diff
@@ -1,21 +1,76 @@
...
+
+/* eslint-disable func-names */
+
 /**
- *  @class compatibility
- *  @type {Object}
+ *  @function attachEvents
  */
-const compatibility = {
+const attachEvents = function () {
```
DamianMullins, line 16: Can you use arrow functions in this module? If you're not using `this` you should be able to switch no problem.

DamianMullins: I think ESLint usually these up, maybe it's not running as expected?

simonsandersje: IE11 was failing on these when I used arrow functions

packages/services/f-compatibility/src/index.js
simonsandersje: Done

packages/services/f-compatibility/src/index.js
DamianMullins: Weird, but it's OK with `const`?

packages/services/f-compatibility/src/index.js
```diff
@@ -1,21 +1,76 @@
...
+    attachEvents();
+};
 
+/**
+ *  @function unsupportedBrowserAction
+ */
+const unsupportedBrowserAction = function (unsupportedBrowser) {
+    const browserData = BROWSER_DATA[unsupportedBrowser];
+    if (browserData.displayBanner) {
+        const tenant = document.getElementsByTagName('html')[0].lang;
```
minrwhite: Hopefully wouldn't get into this position to use it, but it might be worth falling back to a default tenant here:
```suggestion
        const tenant = document.getElementsByTagName('html')[0].lang || 'en-GB';
```

packages/services/f-compatibility/src/index.js
```diff
@@ -1,21 +1,76 @@
...
+    }
+};
+
+/**
+ *  @function compatibility
+ */
+const compatibility = function () {
+    const browser = bowser.getParser(window.navigator.userAgent);
+    const unsupportedBrowsers = Object.keys(UNSUPPORTED_BROWSERS);
+    for (let i = 0; i < unsupportedBrowsers.length; i++) {
```
minrwhite, line 63: You can actually use [`Array.some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) in IE>=9 [If you'd appreciate this looking a little cleaner?]

packages/services/f-compatibility/src/index.js
minrwhite: Surprisingly, [yes](https://stackoverflow.com/a/2356926)!

DamianMullins: Wow didn't know IE supported those keywords!

packages/services/f-compatibility/src/index.js
```diff
@@ -1,21 +1,76 @@
...
+import { UNSUPPORTED_BROWSERS, BROWSER_DATA } from './constants';
+
+const bowser = require('bowser/bundled');
+
+/* eslint-disable func-names */
+
 /**
- *  @class compatibility
- *  @type {Object}
+ *  @function attachEvents
```
DamianMullins: I don't think we need the `@function` JSDoc declarations as these functions are all using the `function` keyword so should already be picked up as such.

Maybe a short description for each one would suffice?

packages/services/f-compatibility/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-compatibility",
   "description": "Fozzie Compatibility - Service to check browser compatibilities and warn users with incompatible browsers",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "maxBundleSize": "100kB",
```
ashleynolan: Quick Q – what's the bundle size of this component? We should definitely update the maBundleSize here, as hopefully, it's not anywhere near 100kB 😆

simonsandersje: Updated to 10KB

packages/services/f-compatibility/src/index.js
simonsandersje: JSDocs updated

packages/services/f-compatibility/src/index.js
simonsandersje: The service is not just about IE11 - we'll be adding other browsers. So for future extensibility, I'd prefer to leave as is.

minrwhite (approved):

FayeCarter (approved):

PR #1396, opened 2021-10-27
f-feature-management@v0.5.0 - refactor for first integration

jnewc:
A few changes for the Feature Management service:
 * Allow injection of axios client
 * Delegate tracking to integrating client instead of calling data layer directly (allows tracking w/ SSR)
 * Separate config file loading from initialisation, and return config to service consumers to allow more flexible caching
 * Test fixes / updates / cleanup

---


packages/services/f-feature-management/src/configStore.js
```diff
@@ -8,18 +9,25 @@ let interval = null;
...
  * @param {string} configAsJson JSON containing the feature configuration.
  * @param {function} onUpdated Optional callback function that gets called when new feature config gets loaded
+ * @return {object} The config as JSON
  */
 function init (configAsJson, onUpdated) {
     if (configAsJson) {
-        const newConfig = JSON.parse(configAsJson);
+        const newConfig = typeof (configAsJson) === 'string' ? JSON.parse(configAsJson) : configAsJson;
+
+        logger.logInfo(`Received feature config created at ${configCreatedAt}`);
```
xander-marjoram: Will this log `null` for the initial config load?

xander-marjoram: Wondering if this log makes sense inside the `if` block below

packages/services/f-feature-management/src/configStore.js
```diff
@@ -30,28 +38,32 @@ function init (configAsJson, onUpdated) {
...
 
 /**
  * Loads config from CDN and starts polling if requested.
  * @param {object} cdnSettings Contains scope, environment, key and optionally host override and pollInterval (defaults to 30s).
  * @param {function} onUpdated Optional callback function called when new config is loaded.
+ * @param {object} httpClient Optional http client to use for making CDN request.
+ * @returns The raw JSON config (for caching purposes)
  */
-async function loadFromCdn (cdnSettings, onUpdated) {
+async function loadFromCdn (cdnSettings, onUpdated, httpClient = axios) {
```
xander-marjoram: Is it worth passing in an [_instance_](https://github.com/axios/axios#creating-an-instance) of axios rather than the whole library?

packages/services/f-feature-management/src/configStore.js
jnewc: This was actually a duplicate of the log further down - removed it 👍

xander-marjoram (dismissed):

packages/services/f-feature-management/package.json
```diff
@@ -41,7 +41,7 @@
     "eslint-plugin-json-schema-validator": "1.0.48",
     "eslint-plugin-jsonc": "1.3.1",
     "jest": "25.4.0",
-    "jest-fetch-mock": "3.0.3",
+    "jest-mock-axios": "^4.4.1",
```
minrwhite: It's standard within this repo to pin to a specific version if possible:
```suggestion
    "jest-mock-axios": "4.4.1",
```

packages/services/f-feature-management/README.md
```diff
@@ -63,5 +69,10 @@ const myFeatureValue = featureManagement.getValue('my-feature-value');
...
 |`cdn.pollInterval`|integer|No (defaults to 30000)|Poll interval in ms.|
 
+### Optional Axios Client:
+
+If an http client is provided, it will be used in place of an internal (axios) client.
+
 ## Integration with vue
-The intention is to make a wrapping library that integrates with Vue and makes these configuration values reactive.
+
+See [f-feature-management-vue](../f-feature-management-vue).
```
minrwhite: Might be better to make this an absolute reference; otherwise this could fail to resolve when the README is surfaced in places other than an instance of the repository.

packages/services/f-feature-management/src/FeatureManager.js
```diff
@@ -49,11 +52,16 @@ class FeatureManager {
 
             const context = getContext();
 
+            if (!context) {
+                logger.logError('Unable to provide feature values as no context was required. Check your settings for the plugin');
```
minrwhite: This might be my misunderstanding of how this is supposed to work, but 'required' seems the wrong word here? Some suggestions that make more sense to my mind:
```suggestion
                logger.logError('Unable to provide feature values as no context was provided. Check your settings for the plugin');
```
```suggestion
                logger.logError('Unable to provide feature values as no context was given. Check your settings for the plugin');
```
```suggestion
                logger.logError('Unable to provide feature values as no context was obtained. Check your settings for the plugin');
```

packages/services/f-feature-management/test/configStore.test.js
```diff
@@ -1,151 +1,157 @@
...
 
-    //Assert
-    expect(fetch).toHaveBeenCalledWith(`https://test.com/config/v1/${cdnSettings.scope}/${cdnSettings.environment}-${cdnSettings.key}`);
-  });
+    afterEach(() => {
+        mockAxios.reset();
+    });
 
-  it('should honour different pollInterval', async () => {
+    it('should make an initial call to fetch', async () => {
```
minrwhite:
```suggestion
    it('should make an initial call to httpClient', async () => {
```

packages/services/f-feature-management/test/configStore.test.js
```diff
@@ -1,151 +1,157 @@
...
-    expect(getFeature('key2')).toBeTruthy();    
-    expect(getFeature('key2').testVal).toBe('val1');
-  });
+    it('should honour different pollInterval', async () => {
+        // Arrange
+        const newSettings = {
+            ...cdnSettings,
+            host: 'https://test.com',
+            pollInterval: 60000
+        };
```
minrwhite: This is shared from lines 77-81 - worth extracting to a shared `alternateSettings` const, or making these specific to each test?

packages/services/f-feature-management/test/configStore.test.js
```diff
@@ -1,151 +1,157 @@
...
-  it('should honour different host', async () => {
-    
-    //Arrange
-    const newSettings = { ...cdnSettings, host: 'https://test.com', pollInterval: 60000 };
+    return loadFromCdnPromise;
+}
 
-    //Act
-    await loadFromCdn(newSettings);
+describe('When calling loadFromCdn', () => {
```
minrwhite, line 35: I think it'd be worth adding a new test to check a passed httpClient is interacted with appropriately, given that this changeset adds that to the interface.

packages/services/f-feature-management/src/FeatureManager.js
jnewc: I think I was going for "required context was not provided" and somehow ended up with this grammatical mess - I like your suggestion of "provided" so let's go with that.

packages/services/f-feature-management/test/configStore.test.js
jnewc: Extraction makes sense 👍

packages/services/f-feature-management/test/configStore.test.js
jnewc: How do you mean? The mock client is passed on L28

minrwhite (dismissed):

xander-marjoram (approved):

packages/services/f-feature-management/test/configStore.test.js
minrwhite: Ah that was my misunderstanding of how this test was written; all good!

minrwhite (approved):

PR #1398, opened 2021-10-27
f-spinner@0.2.0 - Add spinner functionality

FayeCarter:
### Added
- Event handling to start/stop spinner.
- Tests to cover change
ezgif com-gif-maker


packages/components/atoms/f-spinner/src/components/Spinner.vue
```diff
@@ -1,32 +1,46 @@
...
+        <span :class="{ ['is-hidden']: shouldShowSpinner }">
+            <slot />
+        </span>
     </div>
 </template>
 
 <script>
 
 export default {
     name: 'VSpinner',
```
tommcclean: Guessing this should be `FSpinner` (Name property)

packages/components/atoms/f-spinner/src/components/Spinner.vue
```diff
@@ -1,32 +1,46 @@
...
-    props: {
+
+    data () {
+        return {
+            shouldShowSpinner: true
+        };
+    },
+
+    mounted () {
+        this.$on('stopSpinner', () => {
```
tommcclean: I would simply edit the template to handle this better, you could use a method call instead of direct assignment; but I probably wouldn't :)

EX: Raise the event in the child; rather than raising it on the parent
```
<slot 
  @start-spinner="shouldShowSpinner = true"
  @stop-spinner="shouldShowSpinner = false" />
```

Then the child raises an event on itself; not on $parent `this.$emit('stop-spinner')`

packages/components/atoms/f-spinner/stories/TestComponent.vue
```diff
@@ -0,0 +1,17 @@
...
+    </p>
+</template>
+
+<script>
+export default {
+    name: 'TestComponent',
+
+    mounted () {
+        setTimeout(() => {
+            this.$parent.$emit('stopSpinner');
```
tommcclean: In tandem with other suggestion
`this.$emit('stop-spinner')`

packages/components/atoms/f-spinner/src/components/Spinner.vue
FayeCarter: So I tried that originally but sadly it turns out you can't emit from a slot 🤦  https://github.com/vuejs/vue/issues/4332

tommcclean (dismissed):
Do update the readme with usage also and perhaps describe why we are emitting in this style so people don't keep trying to refactor it!

Maybe worth just dropping a comment in the mounted to say why briefly

BanksySan (dismissed):

CHANGELOG.md
```diff
@@ -3,6 +3,15 @@
...
 
+v5.1.0
+ ------------------------------
+ *October 28, 2021*
+
+### Added
+- `fozzie` package version bump.
+
+
+
```
tommcclean: NIT: nbsp

tommcclean (dismissed):

packages/components/atoms/f-spinner/src/components/Spinner.vue
```diff
@@ -1,32 +1,49 @@
...
+        return {
+            shouldShowSpinner: true
+        };
+    },
+
+    mounted () {
+        /**
+         * events are handled in the mounted as components can't handle `this.$emit` from a slot.
+         * https://github.com/vuejs/vue/issues/4332
+         *
```
BanksySan, line 27: Superfluous blank line in comment.

BanksySan (approved):

JoshuaNg2332 (approved):

packages/components/atoms/f-spinner/src/components/Spinner.vue
```diff
@@ -1,32 +1,49 @@
...
     <div
-        :class="$style['c-spinner']"
-        data-test-id="spinner">
-        I am a VSpinner Component (GB)
+        data-test-id="spinner-component">
+        <span
+            v-if="shouldShowSpinner"
+            :class=" $style['c-spinner']" />
+
+        <span :class="{ ['is-hidden']: shouldShowSpinner }">
```
ashleynolan, line 8: Sorry for the delay on my comment – just wanted to check what sort of tags you're wrapping here?

The reason I ask is that `span` tends to be for text-type inline content, whereas `div`is for more block level elements, so if this is wrapping a big block on content, I'd switch this wrapper tag to be a `div`.

PR #1415, opened 2021-10-29
global - adding customized local indentname to solve our css bleed issues

matthewhardern:
I have been experiencing issues with our css bleeds and after speaking with @dandel10n I had an idea as to why this was causing us issues. 

The problem lies in how css modules is setting class names, the hash used in the class name is consistent across builds of a component, so when two version exist together i.e registration page pulls latest f-button but checkout pulls an older version the css class names are identical when css modules spits them out. 

This would normally not be an issue for general use cases with css modules, but with our use case this causes an issue where the two versions of styles compete. To solve this (with some inspiration from @benwhite-deltas I have changed the outputted class names to include the package json version number. this will enable to version of button to be used within coreweb or storybook at the same time. 

I will open as draft for now as this will involve updating all components to fix this. if we agree that the webpack change that applies to all would not need changing on a case by case basis then I will create a global config file to avoid repetition.

### DIFF
--- a/packages/components/atoms/f-button/vue.config.js
+++ b/packages/components/atoms/f-button/vue.config.js
@@ -1,10 +1,22 @@
 const path = require('path');
+const packageJson = require('./package.json');
 
 const rootDir = path.join(__dirname, '..', '..');
 const sassOptions = require('../../../../config/sassOptions')(rootDir);
 
 // vue.config.js
 module.exports = {
+    css: {
+        requireModuleExtension: false,
+        loaderOptions: {
+            css: {
+                modules: {
+                    localIdentName: `[name]__${packageJson.version}--[hash:base64:7]`
+                },
+                localsConvention: 'camelCaseOnly'
+            }
+        }
+    },
     chainWebpack: config => {
         config.module
             .rule('scss-importer')
--- a/packages/components/atoms/f-card/vue.config.js
+++ b/packages/components/atoms/f-card/vue.config.js
@@ -1,10 +1,22 @@
 const path = require('path');
+const packageJson = require('./package.json');
 
 const rootDir = path.join(__dirname, '..', '..');
 const sassOptions = require('../../../../config/sassOptions')(rootDir);
 
 // vue.config.js
 module.exports = {
+    css: {
+        requireModuleExtension: false,
+        loaderOptions: {
+            css: {
+                modules: {
+                    localIdentName: `[name]__${packageJson.version}--[hash:base64:7]`
+                },
+                localsConvention: 'camelCaseOnly'
+            }
+        }
+    },
     chainWebpack: config => {
         config.module
             .rule('scss-importer')
--- a/packages/components/atoms/f-error-message/vue.config.js
+++ b/packages/components/atoms/f-error-message/vue.config.js
@@ -1,10 +1,22 @@
 const path = require('path');
+const packageJson = require('./package.json');
 
 const rootDir = path.join(__dirname, '..', '..');
 const sassOptions = require('../../../../config/sassOptions')(rootDir);
 
 // vue.config.js
 module.exports = {
+    css: {
+        requireModuleExtension: false,
+        loaderOptions: {
+            css: {
+                modules: {
+                    localIdentName: `[name]__${packageJson.version}--[hash:base64:7]`
+                },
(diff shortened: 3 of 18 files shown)


DamianMullins:
One potential downside to this is that the CSS bundle sizes for consuming applications could increase if the packages referenced use different versions downstream. 

I'd suggest that when updating a package, such as `f-button`, that people ensure that all other components using it are also updated to the latest version. That would also mean that consuming applications may have to update multiple packages versions.

Maybe the changes you're suggesting in this PR coupled with guidelines around keeping dependant components up to date would be a good middle ground? That way, hopefully, we can help to ensure that the CSS bundle sizes are kept to a minimum.

matthewhardern:
> One potential downside to this is that the CSS bundle sizes for consuming applications could increase if the packages referenced use different versions downstream.
> 
> I'd suggest that when updating a package, such as `f-button`, that people ensure that all other components using it are also updated to the latest version. That would also mean that consuming applications may have to update multiple packages versions.
> 
> Maybe the changes you're suggesting in this PR coupled with guidelines around keeping dependant components up to date would be a good middle ground? That way, hopefully, we can help to ensure that the CSS bundle sizes are kept to a minimum.

I agree to your statement, but if we are going to have multiple teams usings this library then and it contains pages, then I don't think we can reasonably assume that every team's updating schedule is aligned. Therefore we will need a solution that covers for this. 

I however agree to guidelines around components being constantly updated, we would need to cover that as we don't want teams sitting on old versions of components.

SachaZvetelman:
Nice! Just a small note, make sure you update the generator component so that newly created components don't introduce the issue again.

matthewhardern:
> Nice! Just a small note, make sure you update the generator component so that newly created components don't introduce the issue again.

I will make that a change as part of this PR.

ashleynolan:
@DamianMullins @matthewhardern I don't think it'll affect overall downstream bundle sizes – we only see this issue in development as part of storybook, due to all components running through the same webpack config at that point.

When the component gets published out to NPM, the hash of different versions of a component will always be different at that stage (again, from what I've seen).

So I actually think the downstream bundle sizes will be the same as they are now (which still need to be addressed, but I don't think this change will make things any worse)

DamianMullins:
@ashleynolan I've checked this in menu in the past as I was concerned about bundle sizes increasing and only one version of the f-button styles is added to the production stylesheets.

I could be mistaken, but I've checked this in the menu branch where I've added a version of `f-button` (v3.2.0) that isn't used in any of the other components (such as f-megamodal that uses v3.0.2), and the CSS still only contains a single set of CSS.

Give me a shout and I can send you a link for the compiled CSS.

zlatinivanov-je:
Is there something that can be done with stricter component versioning e.g. if there is a visual/functional change to go to a major version up. Possibly also the use of `peerDependencies`... and other considerations as mentioned above

matthewhardern:
@ashleynolan @DamianMullins I can confirm that the hash does not change across build / versions currently

dandel10n:
> I will open as draft for now as this will involve updating all components to fix this. if we agree that the webpack change that applies to all would not need changing on a case by case basis then I will create a global config file to avoid repetition.

@matthewhardern I don't see the issue with having this config change on a root level. Could you make it global if there are no other objections? I can help if you are busy. 

I agree with the point that we need to keep the dependant components versions up-to-date but it is a whole other big topic for the discussion and can be done away from this PR, right?

ashleynolan:
Just an update on this as @dandel10n and myself have been digging into CSS Modules yesterday to see what was going on. It turns out that the hash generated is a feature of CSS modules, rather than a bug.

The way CSS Modules works, is that it hashes based on the filepath and the selector name, as its goal is to stop selectors clashing between files – so between button.vue and searchbox.vue for instance. That's why between versions of packages where the CSS has changed, the hash stays the same, as the filepath and selector are the same (so that's expected behaviour).

The problem we've hit is where we pull in button twice (for example, in checkout if we pull in f-alert and f-modal, they both have the same selectors as they generate the hash from the same filepath).

We're looking at the best option atm, but we're hoping a combination of peerDependencies (as @zlatinivanov-je suggested) and perhaps adding a hash based on the content of the CSS itself in page components to enable caching across pages without clashing selectors – I'll keep you up-to-date with progress though.

PR #1420, opened 2021-11-01
f-button - added component/a11y/visual tests for icon buttons #trivial

LTurns:
### Added

- Component, visual and accessibility tests for icon-buttons
- Updated `IconButton.stories` to show all types and sizes of icons - similar to `Button.stories` - https://github.com/justeat/fozzie-components/pull/1343
- Created separate demo file to still show icons in isolation

Updated stories.js file -
image

Added grey background option for displaying icons which aren't as visible in light & dark mode
image

Added a separate controls option for viewing the names of each icon
image


packages/components/atoms/f-button/stories/Button.stories.js
```diff
@@ -45,12 +45,7 @@ export const ButtonComponent = (args, { argTypes }) => ({
 });
 
 ButtonComponent.argTypes = {
-    ...SharedButtonArgTypes,
-    actionType: {
```
dandel10n: As I mentioned in `sharedButtonArgTypes.js` better leave it here

packages/components/atoms/f-button/stories/sharedButtonArgTypes.js
```diff
@@ -24,5 +24,10 @@ export default {
         control: { type: 'text' },
         description: 'Should be presented for the button to become a link <a> element',
         defaultValue: ''
+    },
+    actionType: {
```
dandel10n: I haven't added action type to sharedButtonArgTypes as it is not used for link buttons. So better to add it separately as it has been done for regular button story.

packages/components/atoms/f-button/stories/IconButton.stories.js
```diff
@@ -1,88 +1,90 @@
...
+                class="u-spacingBottom--large"
+                v-for="(icon, index) in list"
+                :key="index"
+                :buttonType="icon.type"
+                :buttonSize="icon.size"
+                :actionType="actionType"
+                :href="href"
+                :isIcon="true"
+                :isLoading="isLoading">
+                <CrossIcon />
```
dandel10n, line 44: It would be useful to have visually hidden button text after `<CrossIcon />` as well
```
<span class="is-visuallyHidden">
    {{ icon.text }}
</span>
```

packages/components/atoms/f-button/stories/IconButton.stories.js
```diff
@@ -1,88 +1,90 @@
...
+    decorators: [withA11y]
 };
 
-export const IconButtonComponent = () => ({
+export const IconButtonComponent = (args, { argTypes }) => ({
     components: {
         FButton,
         CrossIcon,
         PlusIcon,
         MoreVerticalIcon
```
dandel10n, line 28: Because now you use only a `CrossIcon` in the template you can remove `PlusIcon` and `moreVerticalIcon` from the import.

dandel10n:
One more thing I noticed is that adding a background for the icon button story will properly show all the buttons:

Like:
Screenshot 2021-11-02 at 11 14 45

instead of:
Screenshot 2021-11-02 at 11 15 07

ashleynolan:
@dandel10n related to your last comment – it can be easier to show some components on a slight background. We do this for the card component [like this in the storybook config](https://github.com/justeat/fozzie-components/blob/master/packages/components/atoms/f-card/stories/card.stories.js#L4-L16), so may be something you could do here for icon buttons?

packages/components/atoms/f-button/stories/Button.stories.js
```diff
@@ -5,7 +5,17 @@ import FButton from '../src/components/Button.vue';
...
 export default {
     title: 'Components/Atoms/f-button',
-    decorators: [withA11y]
+    decorators: [withA11y],
+    parameters: {
+        // background spans across all stories in f-button
+        backgrounds: {
+            values: [
+                { name: 'Grey', value: '#f5f5f5' },
+                { name: 'Dark', value: '#333' }
```
xander-marjoram: How are these used?

packages/components/atoms/f-button/stories/IconButton.stories.js
```diff
@@ -1,88 +1,127 @@
...
+        control: { type: 'select', options: ['button', 'submit', 'reset'] },
+        description: 'Choose the action type of the button',
+        defaultValue: 'button'
+    }
+};
+
+IconButtonComponent.args = {
+    iconButtonLists: [
+        [
+            { type: 'primary', size: 'large', text: 'Large Primary Icon Button' },
```
xander-marjoram, line 105: LGTM for now, but I'm wondering if it might be better in future to have this structure built programmatically, like nested `for each type in VALID_BUTTON_TYPES.iconButton` `for each size in VALID_BUTTON_SIZES`, etc, then if you want to add a new size you can just update the sizes array rather than adding a new object for each type.

packages/components/atoms/f-button/stories/IconButton.stories.js
```diff
@@ -1,88 +1,127 @@
...
+        [
+            { type: 'ghost', size: 'large', text: 'Large Ghost Icon Button' },
+            { type: 'ghost', size: 'medium', text: 'Medium Ghost Icon Button' },
+            { type: 'ghost', size: 'small', text: 'Small Ghost Icon Button' },
+            { type: 'ghost', size: 'xsmall', text: 'Xsmall Ghost Icon Button' }
+        ],
+        [
+            { type: 'ghostTertiary', size: 'large', text: 'Large Ghost Tertiary Icon Button' },
+            { type: 'ghostTertiary', size: 'medium', text: 'Medium GhostTertiary Icon Button' },
+            { type: 'ghostTertiary', size: 'small', text: 'Small GhostTertiary Icon Button' },
```
xander-marjoram, line 125: Ultra nit, feel free to update another time
```suggestion
            { type: 'ghostTertiary', size: 'small', text: 'Small Ghost Tertiary Icon Button' },
```

packages/components/atoms/f-button/test-utils/component-objects/f-button--link.component.js
```diff
@@ -1,6 +1,6 @@
 const Page = require('@justeat/f-wdio-utils/src/page.object');
 
-module.exports = class Buttons extends Page {
+module.exports = class LinkButtons extends Page {
```
xander-marjoram: Should the name be a plural?

packages/components/atoms/f-button/test/visual/f-button.visual.desktop.spec.js
```diff
@@ -141,4 +142,31 @@ describe('f-button Desktop visual tests', () => {
             });
         });
     });
+
+    describe('primary, secondary, inverse, ghostInverse, ghostTertiary, and ghost icon buttons in all 4 sizes', () => {
```
xander-marjoram: Maybe a slightly more futureproof name
```suggestion
    describe('all valid iconButton types in all sizes', () => {
```

packages/components/atoms/f-button/stories/IconButton.stories.js
```diff
@@ -1,88 +1,137 @@
...
+    decorators: [withA11y],
+    parameters: {
+        backgrounds: {
+            values: [
+                { name: 'White', value: '#fff' },
+                { name: 'Dark', value: '#333' },
+                { name: 'Blue', value: '#bfe6ff' }
+            ],
+            default: 'White'
+        }
```
LTurns, line 19: @xander-marjoram I've just used this to add a different background colour option so it's easier to see the icons - opted for the grey/black originally but as @dandel10n suggested the blue makes them pop more 😄 I've added the white and dark still and kept white as default - otherwise, all the stories in f-button would have the blue background

packages/components/atoms/f-button/stories/IconButton.stories.js
LTurns: ah yep agree that would definitely be better and more future proof

packages/components/atoms/f-button/test-utils/component-objects/f-button--link.component.js
LTurns: I was wondering this too but these classes go to the story files with multiple buttons on the page, so I think it's okay to leave as is

LTurns: changed my mind and renamed these actually since they're always imported as a singular in the tests 👍

xander-marjoram (dismissed):

packages/components/atoms/f-button/stories/IconButton.stories.js
```diff
@@ -1,88 +1,144 @@
...
 export default {
     title: 'Components/Atoms/f-button',
-    decorators: [withKnobs, withA11y]
+    decorators: [withA11y],
+    parameters: {
+        backgrounds: {
+            values: [
+                { name: 'White', value: '#fff' },
+                { name: 'Dark', value: '#333' },
+                { name: 'Blue', value: '#bfe6ff' }
```
ashleynolan: I would probably make one of the options one of the background colours we use for components/containers in PIE.

So atm, we have:

```scss
$color-background-default: $color-grey-10;
$color-background-dark: $color-grey-70;
$color-container-default: $color-white;
$color-container-subtle: $color-grey-20;
$color-container-strong: $color-grey-30;
$color-container-dark: $color-grey-60;
```

And the greys for those map to:

```scss
$color-grey-10: #fcfcfb;
$color-grey-20: #f5f3f1;
$color-grey-30: #efedea;
$color-grey-60: #2b3639;
$color-grey-70: #242e30;
```

So I'd probably choose one of those, potentially instead of the Dark option if you wanted to keep the blue one.

LTurns: Ah thanks Ash, that's handy - I've gone for the middle one, seems to be the best for showing the light and dark icons. Added a screenshot above, let me know what you think 👍

xander-marjoram (approved):

packages/components/atoms/f-button/test-utils/component-objects/f-button.component.js
```diff
@@ -1,6 +1,6 @@
 const Page = require('@justeat/f-wdio-utils/src/page.object');
 
-module.exports = class Buttons extends Page {
+module.exports = class Button extends Page {
```
siggerzz, line 3: Should this be renamed to be more specific to the type of button?

LTurns: could rename it to `ActionButton` potentially, but in storybook it's just called `Button` and the others are `Icon Button` and `Link Button` so it could probably stay as for consistency 😄

LTurns: could change to `ActionButton` potentially, but it's referenced as `Button` in storybook and the others are `Icon Button` and `Link Button` so we could also probably leave it for consistency

siggerzz (approved):

PR #1423, opened 2021-11-01
f-header@7.1.0 - tab focus and loop

AndyMayJE:
### Changed

- Mobile menu tab functionality to ensure hidden content is not focussed when tabbing and focus loops

https://user-images.githubusercontent.com/20644043/139708038-59b3ec28-55be-4080-bd6b-0d731534f3ea.mov

.


packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
```diff
@@ -75,12 +75,33 @@ export default {
         isOpen: {
             type: Boolean,
             default: false
+        },
+        isBelowMid: {
```
ashleynolan, line 79: Maybe not one for this PR, but for a future EE ticket, would be to make this property part of the state, so that you don't need to pass it down through props. Would likely be a little tidier that way potentially (but not needed as part of this PR)

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -505,28 +518,93 @@ export default {
...
+                } else if (!e.shiftKey && isLastElement) {
+                    // If focused on the last item and user presses tab, go to the first focusable item
+                    this.firstFocusableEl.focus();
+
+                    e.preventDefault();
+                }
+            }
+        },
+
+        tabLoop () {
```
ashleynolan, line 560: Can you add a JSDoc comment for this function?

packages/components/organisms/f-header/src/components/UserNavigationPanel.vue
```diff
@@ -61,6 +61,17 @@ export default {
         returnLogoutUrl: {
             type: String,
             default: ''
+        },
+        countrySelectorIsClosedOnMobileView: {
```
ashleynolan: If this is a Boolean, I'd try and start it with `is` or `has`. Does this just define if the Country selector is closed, or if it should be closed on mobile (it's a little ambiguous atm)

```suggestion
        isCountrySelectorClosedOnMobileView: {
```

AndyMayJE: `countrySelectorIsClosedOnMobileView` was an existing computed prop so I have renamed as well.

zlatinivanov-je (approved):
My personal preference is to manage focus events via DOM Semantics rather than JavaScript. I see you have good tests coverage and video looks good. Happy to approve

simonsandersje (approved):

PR #1424, opened 2021-11-01
f-restaurant-card@v0.7.0 - badges

jamieomaguire:
### Added
- RestaurantBadge and RestaurantBadges components
- Basic positioning styles for badges within restaurant image

Tested on Chrome & IOS safari

## Notes:

- I would prefer to use `<ul>` and `<li>` tags for the badges, however currently some of our microsites have some pretty heavy overrides pulled in from Fozzie that make this difficult. I think it would be suboptimal to write an override for an override. The future plan is to bring said microsites up to date so that they no longer pull in the overrides. Then we can use some more semantic markup for the badges

- One problem with passing down colours is it makes it harder to directly use pie design tokens

- For now this will be a restaurant card subcomponent, however it could quite easily be extracted to it's own Atom within fozzie-components in the future

### On Searchweb:

Screenshot 2021-11-02 at 11 00 29

Screenshot 2021-11-02 at 10 59 20


packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantBadges/RestaurantBadge.vue
```diff
@@ -0,0 +1,56 @@
...
+export default {
+    name: 'RestaurantBadge',
+    props: {
+        text: {
+            type: String,
+            required: true
+        },
+        textColour: {
+            type: String,
+            default: '#242e30'
```
1itvinka: I don't have a good solution in mind yet, but I don't really like how it's hardcoded. If fozzie tokens change - we'll need to manually change all such values, and it's likely that we'll forget to do it.

jamieomaguire: I agree - I can't see another way at the moment but open to ideas

jamieomaguire: my initial solution hard coded some css modifier classes, so a consumer could pass in a colour scheme token like 'warm' or 'dark' with pie design tokens being leveraged. But then we have to add N classes for N colour and background colour combos. So I'm not sure which is worse.

1itvinka (dismissed):

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantBadges/RestaurantBadge.vue
1itvinka: yeah, may be fozzie variables on the project level. Don't know..

jamieomaguire: @ashleynolan do you have any thoughts on how I could handle this? I think it's reasonable to assume we will get X badge colour requests in the future given we currently already need to support this kind of change. We could always just add another modifier class each time (could end up not being many) but it feels pretty inflexible

jamieomaguire: yeah if we could define a set of global JS vars that could work :P

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -49,7 +55,14 @@
             <component
                 :is="errorBoundary"
                 tier="3">
-                <slot name="new-label" />
+                <!-- TODO - we want to translate this within the component using i18n.
```
minrwhite, line 58: To confirm - is this work listed as `TODO` captured in an internal ticket?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -49,7 +55,14 @@
...
+                For now we'll just need to pass down a translated string from the consuming site -->
+                <restaurant-badge
+                    v-if="newBadgeText"
+                    :is-large="true"
+                    :text="newBadgeText"
+                    :background-colour="'#e5faef'"
+                    :text-colour="'#017A39'" />
```
minrwhite: Consistency in casing please:
```suggestion
                    :background-colour="'#e5faef'"
                    :text-colour="'#017a39'" />
```

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantBadges/RestaurantBadge.vue
```diff
@@ -0,0 +1,56 @@
+<template>
+    <span
+        :class="[
+            [$style['c-restaurantBadge']], {
+                [$style['.c-restaurantBadge--isLarge']]: isLarge
+            }]"
```
minrwhite: I don't think the first style needs the `[]` adding as it's part of the list. However to make it look consistent you could just revert to object syntax and pin the first style to `true` to make it visually easier to parse (IMO):
```suggestion
        :class="{
            [$style['c-restaurantBadge']]: true,
            [$style['.c-restaurantBadge--isLarge']]: isLarge
        }"
```

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantBadges/RestaurantBadge.vue
minrwhite: Does:
```js
import PieTokens from '@justeat/pie-design-tokens/dist/tokens.json';
const textColour = PieTokens.theme.jet.color.alias.default['content-default'];
```
not work, for example?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: Yeah completely, the plan is to add i18n as a stretch goal after initial delivery due to f-globalisation work needed first (which I think Core web are owning)

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantBadges/RestaurantBadge.vue
jamieomaguire: just spotted a typo in my classname! Will do this locally and push!

dawidchar (approved):

minrwhite (approved):

DamianMullins:
@jamieomaguire Related to this, the badges look like they are tag components in PIE, right? I've recently added all of the styles into the Fozzie SCSS and will be looking at creating the Vue component versions of these shortly. Not sure if that has any bearing on your changes here or not but thought I'd mention.

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -49,7 +55,14 @@
             <component
                 :is="errorBoundary"
                 tier="3">
-                <slot name="new-label" />
+                <!-- TODO - we want to translate this within the component using i18n.
+                For now we'll just need to pass down a translated string from the consuming site -->
+                <restaurant-badge
```
ashleynolan, line 60: One thing worth checking with design – I think these are named `Tags` in PIE ([link to the design is here](https://www.figma.com/file/R2rBfzJP0hG0MZorq6FLZ1/%E2%9A%A1%EF%B8%8F%F0%9F%A7%8A--PIE-2.-0-Global---Light-%E2%9A%A1%EF%B8%8F%F0%9F%A7%8A?node-id=742%3A328)). If so, it may be worth updating one way or the other (so either getting them to rename theirs to `badges` or us renaming ours here to reference `Tags `.

Not an urgent one, but would be good to update before we use the badge naming everywhere and it becomes harder to change 😄

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantBadges/RestaurantBadge.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+    <span
+        :class="{
+            [$style['c-restaurantBadge']]: true,
+            [$style['c-restaurantBadge--isLarge']]: isLarge
+        }"
+        :style="`color:${textColour}; background-color:${backgroundColour};`"
```
ashleynolan, line 7: I can see why you're doing it this way (as it does make the component super flexible), but would there not be a finite number of colour combinations we should adhere to as part of our design system?

If there is, then I would definitely go with using modifiers, as the only problem with making the component this flexible is that it means that colour combinations that haven't been approved by the design team can be used by consumers.

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: Ah I see - yes I'm happy to rename to tags :D will do in the next PR! cheers

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantBadges/RestaurantBadge.vue
jamieomaguire: Yeah I do agree. I'm happy to go with that approach. @1itvinka you may want to do similar for the icon text you're working on.

1itvinka: 👍

PR #1436, opened 2021-11-03
f-error-boundary@v0.1.0 - Initial generated component.

DamianMullins:
Component, documentation etc to follow.

### Added
- Initial generated component


yarn.lock
```diff
@@ -25135,6 +25250,18 @@ vite@2.4.4:
   optionalDependencies:
     fsevents "~2.3.2"
 
+vite@2.6.9:
```
jamieomaguire: I didn't realise we could use vite with fozzie components yet!

jamieomaguire:
is this going to behave similar to the version we use on SW?

jamieomaguire (dismissed):

packages/components/atoms/f-error-boundary/src/index.js
```diff
@@ -0,0 +1,14 @@
+
```
ashleynolan: Potentially stupid q – is this a component, or is it more of a service? (just as I kind of expect atoms to have a `.vue` style SFC)

yarn.lock
ashleynolan: We can use it for JS services – for anything using SFC structure, we're going to look at moving to Vite shortly when we update to Vue 3 (as we need to look at what things we rely on that may need to be handled by Vite)

DamianMullins: We'll be able to vite for this component as it has no dependencies on SCSS. I'll post a draft PR for discussion after this one.

packages/components/atoms/f-error-boundary/src/index.js
DamianMullins: It'll be a renderless component (in a Vue file), I kept this file here for the initial commit but it'll be removed in the next PR.

DamianMullins:
@jamieomaguire very similar, yes. Hoping to offer a few different ways of reporting the errors whilst also being configurable.

packages/components/atoms/f-error-boundary/src/index.js
DamianMullins: @ashleynolan I've updated the PR so that it now references a Vue component.

ashleynolan (approved):

jamieomaguire (approved):

PR #1438, opened 2021-11-03
f-restaurant-card@v0.8.0 - ETA, distance, address

1itvinka:
### Added
- DeliveryTimeMeta and IconText components, which we use for ETA, distance, and address data

### Fixed
- Styles of cuisines component

---
### Notes:
- We'll update the icons to reflect the ICING Stage 3 design later, when `f-vue-icon` package is updated.
---

### In Storybook:
- If **ETA** is available:
Screenshot 2021-11-03 at 17 44 56
- Else if **distance** is available:
Screenshot 2021-11-03 at 17 45 41
- Else if **address** is available:
Screenshot 2021-11-03 at 17 45 56
- Tile View:
Screenshot 2021-11-03 at 17 46 38

### On Searchweb:
Screenshot 2021-11-03 at 18 01 44
Screenshot 2021-11-03 at 18 01 56
Screenshot 2021-11-03 at 18 02 08

### DIFF
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
@@ -87,8 +87,7 @@
                 <delivery-time-meta
                     v-if="displayDeliveryTimeMeta"
                     v-bind="deliveryTimeData"
-                    data-test-id="restaurant-delivery-time-meta"
-                    :is-list-item="isListItem" />
+                    data-test-id="restaurant-delivery-time-meta" />
             </component>
             <!-- END ERROR BOUNDARY -->
 
@@ -238,6 +237,11 @@ export default {
                 this.deliveryTimeData.distance ||
                 this.deliveryTimeData.address;
         }
+    },
+    provide () {
+        return {
+            isListItem: this.isListItem
+        };
     }
 };
 </script>
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
@@ -147,7 +147,7 @@ import RestaurantDish from './subcomponents/RestaurantDish.vue';
 import RestaurantCuisines from './subcomponents/RestaurantCuisines.vue';
 import RestaurantBadges from './subcomponents/RestaurantBadges/RestaurantBadges.vue';
 import RestaurantBadge from './subcomponents/RestaurantBadges/RestaurantBadge.vue';
-import DeliveryTimeMeta from './subcomponents/DeliveryTimeMeta.vue';
+import DeliveryTimeMeta from './subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue';
 
 const {
     'support-positive': newBadgeTextColour,
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
@@ -83,6 +83,16 @@
                 :is="errorBoundary"
                 tier="3">
                 <slot name="meta-items" />
+
+                <delivery-time-meta
+                    v-if="deliveryTimeData.eta || deliveryTimeData.distance || deliveryTimeData.address"
+                    data-test-id="restaurant-delivery-time-meta"
+                    :address="deliveryTimeData.address"
+                    :distance="deliveryTimeData.distance"
+                    :distance-for-screen-readers="deliveryTimeData.distanceForScreenReaders"
+                    :eta="deliveryTimeData.eta"
+                    :eta-for-screen-readers="deliveryTimeData.etaForScreenReaders"
+                    :is-list-item="isListItem" />
             </component>
             <!-- END ERROR BOUNDARY -->
 
@@ -137,6 +147,7 @@ import RestaurantDish from './subcomponents/RestaurantDish.vue';
 import RestaurantCuisines from './subcomponents/RestaurantCuisines.vue';
 import RestaurantBadges from './subcomponents/RestaurantBadges/RestaurantBadges.vue';
 import RestaurantBadge from './subcomponents/RestaurantBadges/RestaurantBadge.vue';
+import DeliveryTimeMeta from './subcomponents/DeliveryTimeMeta.vue';
 
 const {
     'support-positive': newBadgeTextColour,
@@ -160,7 +171,8 @@ export default {
         RestaurantDish,
         RestaurantCuisines,
         RestaurantBadges,
-        RestaurantBadge
+        RestaurantBadge,
+        DeliveryTimeMeta
     },
     mixins: [ErrorBoundaryMixin],
     // NOTE: These are merely some placeholder props and not indicative of the props we will end up using
@@ -213,6 +225,10 @@ export default {
         newBadgeText: {
             type: String,
             default: null
+        },
+        deliveryTimeData: {
+            type: Object,
+            default: () => ({})
         }
     },
     data () {
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
@@ -85,7 +85,7 @@
                 <slot name="meta-items" />
 
                 <delivery-time-meta
-                    v-if="deliveryTimeData.eta || deliveryTimeData.distance || deliveryTimeData.address"
+                    v-if="displayDeliveryTimeMeta"
                     v-bind="deliveryTimeData"
                     data-test-id="restaurant-delivery-time-meta"
                     :is-list-item="isListItem" />
@@ -231,6 +231,13 @@ export default {
         return {
             subcomponentColourSchemes
         };
+    },
+    computed: {
+        displayDeliveryTimeMeta () {
+            return this.deliveryTimeData.eta ||
+                this.deliveryTimeData.distance ||
+                this.deliveryTimeData.address;
+        }
     }
 };
 </script>
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
@@ -86,12 +86,8 @@
 
                 <delivery-time-meta
                     v-if="deliveryTimeData.eta || deliveryTimeData.distance || deliveryTimeData.address"
+                    v-bind="deliveryTimeData"
                     data-test-id="restaurant-delivery-time-meta"
-                    :address="deliveryTimeData.address"
-                    :distance="deliveryTimeData.distance"
-                    :distance-for-screen-readers="deliveryTimeData.distanceForScreenReaders"
-                    :eta="deliveryTimeData.eta"
-                    :eta-for-screen-readers="deliveryTimeData.etaForScreenReaders"
                     :is-list-item="isListItem" />
             </component>
             <!-- END ERROR BOUNDARY -->
--- a/packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
+++ b/packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
@@ -167,4 +167,17 @@ describe('RestaurantCard.v1', () => {
             expect(wrapper.find('[data-test-id="restaurant-inner-content-badges"]').exists()).toBe(false);
         });
     });
+
+    describe('Delivery Time Meta', () => {
+        it('should not be shown if there is no eta, distance or address data', () => {
+            // arrange
+            const propsData = {};
+
+            // act
+            const wrapper = mount(RestaurantCardV1, { propsData });
+
+            // assert
+            expect(wrapper.find('[data-test-id="restaurant-delivery-time-meta"]').exists()).toBe(false);
+        });
+    });
 });
--- a/packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
+++ b/packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
@@ -179,5 +179,20 @@ describe('RestaurantCard.v1', () => {
             // assert
             expect(wrapper.find('[data-test-id="restaurant-delivery-time-meta"]').exists()).toBe(false);
         });
+
+        it.each(['eta', 'distance', 'address'])('should not be shown if there is %p data', dataKey => {
+            // arrange
+            const propsData = {
+                deliveryTimeData: {
+                    [dataKey]: 'Test'
+                }
+            };
+
+            // act
+            const wrapper = mount(RestaurantCardV1, { propsData });
+
+            // assert
+            expect(wrapper.find('[data-test-id="restaurant-delivery-time-meta"]').exists()).toBe(true);
+        });
     });
 });
--- a/packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
+++ b/packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
@@ -180,7 +180,7 @@ describe('RestaurantCard.v1', () => {
             expect(wrapper.find('[data-test-id="restaurant-delivery-time-meta"]').exists()).toBe(false);
         });
 
-        it.each(['eta', 'distance', 'address'])('should not be shown if there is %p data', dataKey => {
+        it.each(['eta', 'distance', 'address'])('should be shown if there is %p data', dataKey => {
             // arrange
             const propsData = {
                 deliveryTimeData: {
--- a/packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
@@ -3,7 +3,6 @@
         <icon-text
             v-if="eta"
             :text="eta"
-            :accessible-text="etaForScreenReaders"
             :is-bold="true"
             :color="$options.PieTokens.theme.jet.color.global.green"
             :hide-icon-in-tile-view="true"
@@ -14,7 +13,6 @@
         <icon-text
             v-else-if="distance"
             :text="distance"
-            :accessible-text="distanceForScreenReaders"
             :is-bold="true"
             :hide-icon-in-tile-view="true"
             :data-test-id="'distance-icon-text'">
@@ -24,7 +22,6 @@
         <icon-text
             v-else-if="address"
             :text="address"
-            :accessible-text="addressForScreenReaders"
             :data-test-id="'address-icon-text'">
             <map-pin-icon />
         </icon-text>
@@ -49,23 +46,11 @@ export default {
             type: String,
             default: null
         },
-        addressForScreenReaders: {
-            type: String,
-            default: null
-        },
         distance: {
             type: String,
             default: null
         },
-        distanceForScreenReaders: {
-            type: String,
-            default: null
-        },
         eta: {
-            type: String,
-            default: null
-        },
-        etaForScreenReaders: {
             type: String,
             default: null
         }
--- a/packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
@@ -34,7 +34,7 @@
 <script>
 import { MapPinIcon, ClockIcon } from '@justeat/f-vue-icons';
 import PIE_TOKENS from '@justeat/pie-design-tokens/dist/tokens.json';
-import IconText from './IconText.vue';
+import IconText from '../IconText.vue';
 
 export default {
     PieTokens: PIE_TOKENS,
--- a/packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
@@ -76,6 +76,3 @@ export default {
     }
 };
 </script>
-
-<style lang="scss" module>
-</style>
--- a/packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
@@ -4,7 +4,7 @@
             v-if="eta"
             :text="eta"
             :is-bold="true"
-            :color="$options.PieTokens.theme.jet.color.global.green"
+            :color="'green'"
             :hide-icon-in-tile-view="true"
             data-test-id="delivery-meta-eta">
             <clock-icon data-test-id="delivery-meta-eta-icon" />
@@ -30,11 +30,9 @@
 
 <script>
 import { MapPinIcon, ClockIcon } from '@justeat/f-vue-icons';
-import PIE_TOKENS from '@justeat/pie-design-tokens/dist/tokens.json';
 import IconText from '../IconText.vue';
 
 export default {
-    PieTokens: PIE_TOKENS,
     name: 'DeliveryTimeMeta',
     components: {
         IconText,
--- a/packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
@@ -6,8 +6,8 @@
             :is-bold="true"
             :color="$options.PieTokens.theme.jet.color.global.green"
             :hide-icon-in-tile-view="true"
-            data-test-id="eta-icon-text">
-            <clock-icon />
+            data-test-id="delivery-meta-eta">
+            <clock-icon data-test-id="delivery-meta-eta-icon" />
         </icon-text>
 
         <icon-text
@@ -15,15 +15,15 @@
             :text="distance"
             :is-bold="true"
             :hide-icon-in-tile-view="true"
-            data-test-id="distance-icon-text">
-            <map-pin-icon />
+            data-test-id="delivery-meta-distance">
+            <map-pin-icon data-test-id="delivery-meta-distance-icon" />
         </icon-text>
 
         <icon-text
             v-else-if="address"
             :text="address"
-            data-test-id="address-icon-text">
-            <map-pin-icon />
+            data-test-id="delivery-meta-address">
+            <map-pin-icon data-test-id="delivery-meta-address-icon" />
         </icon-text>
     </div>
 </template>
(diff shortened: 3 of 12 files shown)


jamieomaguire:
Very nitpicky but in storybook could you please add an example of a longer address that _might_ span multiple lines?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -83,6 +83,16 @@
...
+
+                <delivery-time-meta
+                    v-if="deliveryTimeData.eta || deliveryTimeData.distance || deliveryTimeData.address"
+                    data-test-id="restaurant-delivery-time-meta"
+                    :address="deliveryTimeData.address"
+                    :distance="deliveryTimeData.distance"
+                    :distance-for-screen-readers="deliveryTimeData.distanceForScreenReaders"
+                    :eta="deliveryTimeData.eta"
+                    :eta-for-screen-readers="deliveryTimeData.etaForScreenReaders"
+                    :is-list-item="isListItem" />
```
jamieomaguire:
```suggestion
                <delivery-time-meta
                    v-if="deliveryTimeData.eta || deliveryTimeData.distance || deliveryTimeData.address"
                    v-bind="deliveryTimeData"
                    data-test-id="restaurant-delivery-time-meta"
                    :is-list-item="isListItem" />
```

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -83,6 +83,16 @@
                 :is="errorBoundary"
                 tier="3">
                 <slot name="meta-items" />
+
+                <delivery-time-meta
+                    v-if="deliveryTimeData.eta || deliveryTimeData.distance || deliveryTimeData.address"
```
jamieomaguire: could extract to a computed property named something like "displayDeliveryTimeMeta"

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
```diff
@@ -0,0 +1,81 @@
...
+    </div>
+</template>
+
+<script>
+import { MapPinIcon, ClockIcon } from '@justeat/f-vue-icons';
+import PIE_TOKENS from '@justeat/pie-design-tokens/dist/tokens.json';
+import IconText from '../IconText.vue';
+
+export default {
+    PieTokens: PIE_TOKENS,
```
jamieomaguire: What's the benefit of doing this as opposed to exposing the pie tokens as part of the data object?

 I'm not sure if this approach offers any benefit and it seems a bit non-standard? Really interested to hear your thoughts!

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
```diff
@@ -0,0 +1,81 @@
...
+        },
+        isListItem: {
+            type: Boolean,
+            default: false
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
```
jamieomaguire: if we don't need it - delete!

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
1itvinka: love it!!!! completely forgot to do it. thanks!

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
1itvinka: yeah, I think that exposing the tokens as a part of the data object might be more readable, so I'm happy to refactor. This approach looks a bit neater, but I guess it's just a personal preference. @dawidchar what do you think?

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/_tests/DeliveryTimeMeta.test.js
```diff
@@ -0,0 +1,78 @@
...
+            eta: '20 - 25 min'
+        };
+
+        // act
+        const wrapper = shallowMount(sut, { propsData });
+
+        // assert
+        expect(wrapper.exists()).toBe(true);
+    });
+    it('should show eta text when data is present', () => {
```
jamieomaguire, line 18: we currently have a unique data-test-id for eta and another for distance and address. I think we should preserve this. These tests are quite coupled to the Icon Text implementation at the moment. 

I think we should keep unique test IDs and check for the presence of those instead of having the same test ID for all icon text instances (a testing nightmare in the future) and avoid inspecting text output .

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -0,0 +1,110 @@
...
+                :style="`color:${color};`"
+                data-test-id="icon-text-visible-text">
+                {{ text }}
+            </span>
+        </p>
+        <span
+            class="is-visuallyHidden"
+            data-test-id="icon-text-accessible-text">
+            {{ accessibleText || text }}
+        </span>
```
jamieomaguire: I don't think we currently do this on SearchWeb, and we're probably not going to start generating new copy as part of this work - we'd need to get new string values for all tenants which is out of scope. So we can probably leave this alone for now as it isn't something we've had as a ticket from accessibility testing. 

If it's something we want to add in future, then we probably want to do that as a separate piece of work - we'll need to work with copywriters etc.

jamieomaguire: Screenshot 2021-11-04 at 09 07 07

for context - the current implementation

jamieomaguire: I'd also be keen to know if this could be achieved via aria attributes as opposed to adding more DOM nodes to the page

packages/components/molecules/f-restaurant-card/stories/RestaurantCard.stories.js
```diff
@@ -46,7 +46,14 @@ RestaurantCardComponent.args = {
         cuisines: ['Mexican', 'Burgers', 'Chinese'],
         imageBadges: [{ text: 'Promoted', textColour: badgeColourSchemes.promoted.text, backgroundColour: badgeColourSchemes.promoted.background }, { text: 'StampCards', textColour: badgeColourSchemes.stampcards.text, backgroundColour: badgeColourSchemes.stampcards.background }],
         contentBadges: [{ text: 'BTA Winner' }, { text: 'Michelin Star' }, { text: 'Tried & Tasted' }, { text: 'New Ownership' }, { text: 'Delivered by Menulog' }, { text: 'A very very very super long unrealistic but necessary to test badge that hopefully never happens' }],
-        newBadgeText: 'NEW'
+        newBadgeText: 'NEW',
```
DamianMullins, line 49: MInor thing; I'd use title-case here and then use CSS `text-transform: uppercase` to convert the casing.

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -0,0 +1,110 @@
...
+export default {
+    name: 'IconText',
+    props: {
+        text: {
+            type: String,
+            required: true
+        },
+        accessibleText: {
+            type: String,
+            default: null
```
jamieomaguire: Also unsure if this is worthwhile if it just duplicates the regular text prop when no accessible string is provided

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -0,0 +1,110 @@
...
+        },
+        color: {
+            type: String,
+            default: null
```
jamieomaguire, line 43: So this is going to set `fill` and `color` to null when none is provided? I think it would be worth adding a fallback here :)

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+    <div>
+        <p
+            :class="{
+                [$style['c-restaurantCard-iconText']]: true,
+                [$style['c-restaurantCard-iconText--bold']]: isBold
+            }"
+            aria-hidden="true"
+            :data-test-id="dataTestId">
```
jamieomaguire: I can't see this being utilised in the DeliveryTimeMeta component - but like what I mentioned on the test file comment - would be worth passing something unique down here for eta vs distance/address and testing for that (sorry if you are but I can't see it!)

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -0,0 +1,110 @@
+<template>
+    <div>
+        <p
+            :class="{
+                [$style['c-restaurantCard-iconText']]: true,
+                [$style['c-restaurantCard-iconText--bold']]: isBold
+            }"
+            aria-hidden="true"
```
jamieomaguire: if this removes the entire component and it's children from the accessibility tree then I can't see how the accessible text stuff would get picked up by a screen reader. Perhaps I've misunderstood the purpose of this role but I'm unsure if we want it here. Can you let me know your thoughts please? :)

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
1itvinka: We do this on HomeWeb, and I don't want to make the home page less accessible as a result of this work. For SearchWeb we can use the regular text for now, which is reflected in the code: If there's no accessible text - we use the regular text for screen readers. It's a huge difference for screen reader users to hear "1 mile" vs "1 mile from your location".

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -0,0 +1,110 @@
...
+    flex: 1;
+}
+
+.c-restaurantCard-iconText-icon {
+    width: spacing(x2);
+    height: spacing(x2);
+    margin-right: spacing(x0.5);
+}
+
+.c-restaurantCard-iconText-icon--hideOnMidBellow {
```
jamieomaguire:
```suggestion
.c-restaurantCard-iconText-icon--hideOnSmallScreens {
```

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -0,0 +1,110 @@
...
+                [$style['c-restaurantCard-iconText']]: true,
+                [$style['c-restaurantCard-iconText--bold']]: isBold
+            }"
+            aria-hidden="true"
+            :data-test-id="dataTestId">
+            <span
+                v-if="showIcon"
+                :class="{
+                    [$style['c-restaurantCard-iconText-icon']]: true,
+                    [$style['c-restaurantCard-iconText-icon--hideOnMidBellow']]: hideIconInTileView
```
jamieomaguire: small typo on the world 'below'

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -0,0 +1,110 @@
...
+    display: flex;
+    align-items: center;
+}
+
+.c-restaurantCard-iconText--bold {
+    font-weight: $font-weight-bold;
+}
+
+.c-restaurantCard-iconText-content {
+    display: block;
```
jamieomaguire: I think flex items are displayed as blocks by default but ignore if needed!

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
dawidchar: I believe Vue simply doesn't apply the style when the value is null, so it falls back to the "global" colour

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantCuisines.vue
```diff
@@ -38,8 +38,8 @@ export default {
 
 .c-restaurantCard-cuisines-item {
     position: relative;
-    padding-right: spacing(x1.25);
-    margin-right: spacing(x0.6);
+    padding-right: spacing();
+    margin-right: spacing(x0.5);
```
jamieomaguire, line 42: 🔥🔥

packages/components/molecules/f-restaurant-card/src/components/subcomponents/_tests/IconText.test.js
```diff
@@ -0,0 +1,88 @@
...
+
+        // act
+        const wrapper = shallowMount(sut, { propsData });
+
+        const accessibleText = wrapper.find('[data-test-id="icon-text-accessible-text"]').text();
+
+        // assert
+        expect(accessibleText).toBe('foo');
+    });
+
```
jamieomaguire: again, I question the purpose of duplicating the default text if no accessible string provided - if anything that might make the screen reader repeat the same thing twice for each restaurant. As I said in my other comment this might need to be something we add in future due to translations etc.

packages/components/molecules/f-restaurant-card/src/components/subcomponents/_tests/IconText.test.js
```diff
@@ -0,0 +1,88 @@
...
+    it('when icon is present and hideInTileView is true and in tile mode, hides Icon', () => {
```
jamieomaguire: This is quite a complicated test case (in it's wording). Might be worth creating a describe block for the icon rendering, then a case when tile mode and hide are true etc

packages/components/molecules/f-restaurant-card/stories/RestaurantCard.stories.js
```diff
@@ -46,7 +46,14 @@ RestaurantCardComponent.args = {
...
+        deliveryTimeData: {
+            address: '29 - 32 The Oval, London',
```
jamieomaguire: can we use a longer address here just so it's clear what the behaviour is when the copy is longer than we'd prefer?

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
DamianMullins: Convention used in Fozzie is to use "showAboveMid" etc. "mid" being the name of the breakpoint where you'd like the behaviour to intersect.

jamieomaguire: ah I see - go with that then! Just fix that small typo :)

packages/components/molecules/f-restaurant-card/stories/RestaurantCard.stories.js
jamieomaguire: We'd probably need a class modifier triggered via props in that case - the badge was part of a previous PR of mine so happy to fix forward but keen to hear how you'd handle that flexibly @DamianMullins - just an uppercase prop etc? 'NEW' is the only example currently where badge text is uppercase

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
jamieomaguire: So then I'd suggest 2 things:

- only render this span if there is accessible text; there's no point in duplicating the same text 
- create an internal ticket so that we don't forget that we now need these translations available in searchweb

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
jamieomaguire: awesome!

packages/components/molecules/f-restaurant-card/stories/RestaurantCard.stories.js
1itvinka: Screenshot 2021-11-04 at 10 21 43

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
1itvinka: > only render this span if there is accessible text; there's no point in duplicating the same text

if there's no accessible text, we should at least use regular text. Otherwise, this information will be missed completely from the screen readers. Or we need some logic inside `aria-hidden="true"` which will only make the component less readable. Happy to hop on a call and discuss.

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
1itvinka: this block of code is visually hidden, but it's picked up by the screen readers
```
       <span
            class="is-visuallyHidden"
            data-test-id="icon-text-accessible-text">
            {{ accessibleText || text }}
        </span>
```

packages/components/molecules/f-restaurant-card/src/components/subcomponents/_tests/IconText.test.js
1itvinka: it won't repeat it twice. The block that has `aria-hidden="true"` will be ignored by screen readers. Only the accessible text (or the fallback to regular text if the accessible text is not provided) will be visible to screen readers.

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
dawidchar: We could potentially use Vue to add/remove the aria-hidden in that case 🤔

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
dawidchar: I wrote some tests this morning that utilise that, merging them now 😄

dawidchar: Alternatively, I can apply the test id to the parent icon-text element and not pass it down as a prop, what do you think?

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
dawidchar: Because the pie tokes don't need to be reactive, this is the preferred approach as it will import them statically instead of reactively

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/_tests/DeliveryTimeMeta.test.js
dawidchar: Happy to refactor, it was done because the text content included visible and accessible text and so assertions were harder. But if we move the accessible text to a aria tag, we can avoid having to do this

packages/components/molecules/f-restaurant-card/src/components/subcomponents/_tests/IconText.test.js
dawidchar: Fixed 😄

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
jamieomaguire: tbh I quite like the idea of just applying it from the parent component as opposed to a prop - the prop is super useful if we need granular control over any nested test IDs - so happy to go with your preference on this

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -0,0 +1,101 @@
+<template>
+    <div>
```
dawidchar: Should we remove this root div as the root element can now be the `p` tag

1itvinka: good spot!

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue
1itvinka: refactored it to use modifiers as discussed here https://github.com/justeat/fozzie-components/pull/1424#discussion_r742745599

packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
```diff
@@ -167,4 +167,32 @@ describe('RestaurantCard.v1', () => {
...
+            const propsData = {};
+
+            // act
+            const wrapper = mount(RestaurantCardV1, { propsData });
+
+            // assert
+            expect(wrapper.find('[data-test-id="restaurant-delivery-time-meta"]').exists()).toBe(false);
+        });
+
+        it.each(['eta', 'distance', 'address'])('should be shown if there is %p data', dataKey => {
```
jamieomaguire, line 183: the test is fine - the title is a little misleading though as it's actually just rendering the entire deliverytimemeta if 1 of the 3 is present

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/_tests/DeliveryTimeMeta.test.js
jamieomaguire: unresolved because I think this is still a legitimate issue with the test approach - I think we should give eta, delivery and address their own unique test IDs and simply look for those. Rather than giving them all the same ID and checking text output

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/_tests/DeliveryTimeMeta.test.js
```diff
@@ -0,0 +1,193 @@
...
+
+            // act
+            const wrapper = mount(sut, { propsData });
+
+            const etaIconExists = wrapper.findAll('[data-test-id="eta-icon-text"] [data-test-id="icon-text-icon"]').exists();
+            const etaTextExists = wrapper.findAll('[data-test-id="eta-icon-text"] [data-test-id="icon-text-visible-text"]').exists();
+
+            // assert
+            expect(etaIconExists).toBe(false);
+            expect(etaTextExists).toBe(true);
```
jamieomaguire: I'm in two minds about this. Its very tightly coupled to the implementation of icon text. You could always use shallowMount here - that will render out something like `<icon-text-stub show-icon="false" />` - you could then test that the correct attribute is added instead of relying on the icon text implementation (which could change at any time)

Try using shallowMount and logging out `wrapper.html()` and hopefully you'll see what I mean?

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/_tests/DeliveryTimeMeta.test.js
```diff
@@ -0,0 +1,193 @@
...
+            // act
+            const wrapper = mount(sut, { propsData });
+
+            const distanceIconExists = wrapper.findAll('[data-test-id="distance-icon-text"] [data-test-id="icon-text-icon"]').exists();
+            const distanceTextExists = wrapper.findAll('[data-test-id="distance-icon-text"] [data-test-id="icon-text-visible-text"]').exists();
+
+            // assert
+            expect(distanceIconExists).toBe(false);
+            expect(distanceTextExists).toBe(true);
+        });
```
jamieomaguire: same as the above comment - I think we should just test that we pass down the expected props and then the icon-text component will check that the icon does or does not render based on the props

packages/components/molecules/f-restaurant-card/src/components/subcomponents/_tests/IconText.test.js
```diff
@@ -0,0 +1,69 @@
...
+        });
+
+        describe('List view', () => {
+            beforeAll(() => {
+                config.provide = { isListItem: true };
+            });
+
+            afterAll(() => {
+                config.provide = {};
+            });
```
jamieomaguire, line 48: I think we might prefer beforeEach and afterEach here as this sounds like it would preserve state between future tests added (might be wrong here)

dawidchar: This sets up the component in the list item view, which is what all the tests in the describe block are testing in, no other state would be preserved between tests

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/_tests/DeliveryTimeMeta.test.js
dawidchar: I see what you're saying, I think I have a nice solution for this, I'll see if it works 🤔

packages/components/molecules/f-restaurant-card/stories/RestaurantCard.stories.js
jamieomaguire: can fix in next PR

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
1itvinka: ticket to address the accessibility of the component https://justeattakeaway.atlassian.net/browse/SEARCH-1996

packages/components/molecules/f-restaurant-card/stories/RestaurantCard.stories.js
DamianMullins: Might be best to refer to PIE people, from what I've seen the casing has all been title-case. Otherwise feel free to approach however you like, ignore me if you think this is the best way 😃

ashleynolan: I would probably reach out to design on this and see which badges should be which casing (they may have a set of guidelines to follow for this)

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/_tests/DeliveryTimeMeta.test.js
```diff
@@ -0,0 +1,185 @@
...
+        };
+
+        // act
+        const wrapper = mount(sut, { propsData });
+
+        const visibleText = wrapper.text();
+        const addressExists = wrapper.find('[data-test-id="delivery-meta-address"]').exists();
+
+        // assert
+        expect(visibleText).toBe('baz');
```
jamieomaguire, line 68: we should probably test only the desired variant is rendered and the other 2 are not :)

jamieomaguire (approved):

dawidchar (approved):

packages/components/molecules/f-restaurant-card/src/components/subcomponents/DeliveryTimeMeta/_tests/DeliveryTimeMeta.test.js
1itvinka: will be fixed forward in the next PR

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -0,0 +1,104 @@
...
+        isListItem: {
+            default: false
+        }
+    },
+    props: {
+        text: {
+            type: String,
+            required: true
+        },
+        color: {
```
ashleynolan, line 41: One suggestion for a future improvement to this – I'd probably have some validation over what colours this can be. That way there's a known set of colours that the icon can be set as (because otherwise, passing in blue atm, it won't actually make it blue, it'll just add a class that doesn't do anything I think?)

Sorry for the delayed response – just been trying to get to inbox zero due to the OneApp stuff that was going on recently!

PR #1445, opened 2021-11-04
f-wdio-utils@v0.4.0 – Adding `Pages` and `Templates` to the storybook url builder

ashleynolan:
### Added
- `Pages` and `Templates` added to the storybook Url builder.


ashleynolan:
Merging this so I can move pages and templates while people are offline tonight. Would be good if @JoshuaNg2332, @siggerzz & @LTurns can retrospectively review this in the morning in case they see any issues that I may have missed.

PR #1448, opened 2021-11-05
f-error-boundary@v0.2.0 - Implement component logic.

DamianMullins:
### Added
- Implemented logic in component to handle capturing errors and providing information back to the parent component.
- logger service to generate the logging payload.
- Unit tests for logger service.
- Storybook examples with demo components.
- Readme description.

### Changed
- Moved component into the `components` directory.
- Used updated Vite config definition.

### Removed
- Dead status badges on the readme.


packages/components/atoms/f-error-boundary/src/components/tests/ErrorBoundary.test.js
```diff
@@ -0,0 +1,140 @@
...
+    computed,
+    scopedSlots
+} = {}) => shallowMount(ErrorBoundary, {
+    localVue,
+    computed,
+    propsData,
+    scopedSlots
+});
+
+describe('ModalDeliveryFeesInfo', () => {
```
simonsandersje: Should this be something like `describe('ErrorBoundary', () => {` ?

DamianMullins: It should be, yes. Fixed now 👍🏻

simonsandersje (dismissed):

packages/components/atoms/f-error-boundary/src/components/ErrorBoundary.vue
```diff
@@ -0,0 +1,73 @@
...
+            this.hasError = true;
+            this.error = error;
+            this.vm = vm;
+            this.info = info;
+
+            this.$emit('on-error', {
+                error,
+                vm,
+                info,
+                tier: this.tier,
```
minrwhite, line 52: Do we need tier to be represented here, when we're also including it as part of the `loggerPayload` object?

packages/components/atoms/f-error-boundary/src/components/tests/ErrorBoundary.test.js
```diff
@@ -0,0 +1,140 @@
...
+                expect(wrapper.html()).toContain(`error: ${error}`);
+                expect(wrapper.html()).toContain(`vm: ${vm}`);
+                expect(wrapper.html()).toContain(`info: ${info}`);
+                expect(wrapper.html()).toContain(`tier: ${tier}`);
+                expect(wrapper.html()).toContain(`loggerPayload: ${loggerPayload}`);
+            });
+        });
+
+        describe('when `hideOnError` is true and `hasError` is true', () => {
+            it('should return slot props when values are set', async () => {
```
minrwhite:
```suggestion
            it('should not return slot props when values are set', async () => {
```

packages/components/atoms/f-error-boundary/src/components/tests/ErrorBoundary.test.js
```diff
@@ -0,0 +1,140 @@
...
+                <p>hasError: {{ props.hasError }}</p>
+                <p>error: {{ props.error }}</p>
+                <p>vm: {{ props.vm }}</p>
+                <p>info: {{ props.info }}</p>
+                <p>tier: {{ props.tier }}</p>
+                <p>loggerPayload: {{ props.loggerPayload }}</p>
+            </div>`
+        };
+
+        describe('when `hideOnError` is false and `hasError` is false', () => {
```
minrwhite: Worth expanding this to be an `.each()` over `false`&`false`/`true`&`false`/`false`&`true` condition sets?

packages/components/atoms/f-error-boundary/src/components/ErrorBoundary.vue
DamianMullins: I included it in case people want to use it for any other purposes, outside of logging.

packages/components/atoms/f-error-boundary/src/components/tests/ErrorBoundary.test.js
DamianMullins: Done 👍🏻

packages/components/atoms/f-error-boundary/src/components/tests/ErrorBoundary.test.js
```diff
@@ -33,16 +33,23 @@ describe('ErrorBoundary', () => {
...
         };
 
-        describe('when `hideOnError` is false and `hasError` is false', () => {
-            it('should return slot props when values are set', async () => {
+        describe.each([
+            [false, false],
+            [false, true],
+            [true, false],
+            [true, true]
+        ])('when `hideOnError` is %s and `hasError` is %s', (
```
minrwhite, line 41: 🎩💯

minrwhite (approved):

jamieomaguire (approved):

PR #1449, opened 2021-11-05
f-globalisation@v1.1.0 - Merge messages rather than replacing them

tommcclean:
> Not ready for review; would like to verify this solves the problem reported

**Problem**
- Currently each component using `f-globalisation` replaces the messages set for the tenant on load
- This works okay in most cases as the internationalisation is scoped to the component.
- In .NET/Vue microsites this scoping seems to work differently - thus replacing all localisation messages in some use cases

**Solution**
- Instead of replacing localisation messages, append the new ones on top of the old ones
- Prioritise the new messages over any old messages with the same key


DamianMullins (dismissed):

DamianMullins:
@tommcclean Apologies on jumping the gun for approvals, I didn't see the draft status!

tommcclean:
> @tommcclean Apologies on jumping the gun for approvals, I didn't see the draft status!

No worries; im on holiday next week, so will probably come back to it then; but wanted it available incase this was causing an urgent problem for somebody

PR #1452, opened 2021-11-08
f-link@2.3.0

AndyMayJE:
### Added

- Support for custom link CSS class via `link-class` attribute. Also ensures other `$attrs` are bound to the component, not the wrapper.


packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -114,6 +113,18 @@ export default {
...
+
+    methods: {
+        /**
+         * Set data.linkClass attribute and bind all other $attrs to the component.
+         * linkClass added manually to nested component as inheritAttrs: false does not affect class bindings.
+         */
+        bindAttrs () {
+            const { 'link-class': linkClass, ...atributes } = this.$attrs;
+            this.linkClass = linkClass;
+            return atributes;
```
xander-marjoram: Small typo.
Also, does `$attrs` not have `linkClass` available as a camelCase property?
```suggestion
            const { 'link-class': linkClass, ...attributes } = this.$attrs;
            this.linkClass = linkClass;
            return attributes;
```

xander-marjoram: Also, TIL about using the spread operator in assignment. Maybe calling it something like `otherAttributes` might make its use slightly clearer?

packages/components/atoms/f-link/README.md
```diff
@@ -78,6 +78,7 @@ The props that can be defined are as follows (if any):
 | `isFullWidth` | `Boolean` | `false` | Link set as full width. |
 | `noLineBreak` | `Boolean` | `false` | Removes white space. |
 | `isDistinct` | `Boolean` | `false` | Changes default link colour (dark grey) to blue. |
+| `link-class` | `String` | `n/a` (this is an expected attribute rather than a required prop) |  Allows parent component to add a CSS class to the `<a>` or `<router-link>`  |
```
xander-marjoram: Sounds like this is actually an optional attribute, rather than expected attribute or an optional/required prop. If it's not included, then no class is added, right?

```suggestion
| `link-class` | `String` | `n/a` (this is an optional attribute rather than a required prop) |  Allows parent component to add a CSS class to the `<a>` or `<router-link>`  |
```

xander-marjoram (commented):
A few small comments/questions

packages/components/atoms/f-link/README.md
AndyMayJE: Nice `otherAttributes` is much better

packages/components/atoms/f-link/src/components/Link.vue
AndyMayJE: No, it is available in `$attrs` exactly as used on the parent element.

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -3,6 +3,7 @@
         <component
             :is="linkType"
             :class="[
+                linkClass,
```
ashleynolan, line 6: Just wanted to double-check (I'm sure I asked this before) – what's the main difference between having this prop and just passing through a class on the component (so like `<f-link class="blah">`? I'm guessing just passing in the class you don't get the same kind of access to it as a prop gives you?

AndyMayJE: @ashleynolan The class attribute will always get added to the parent span rather than the `a` or `router-link` component as `inheritAttrs: false` option does not affect style and class bindings. So we get:

`<span class="blah"><router-link ... /></span>`

Rather than:

`<span><router-link class="blah" ... /></span>`

ashleynolan: Makes sense – I didn't realise classes didn't transfer down as well. Good to know 👍

ashleynolan (approved):

packages/components/atoms/f-link/src/components/Link.vue
```diff
@@ -33,17 +35,14 @@ let uid = 0;
...
 
+    inheritAttrs: false,
+
     props: {
         isExternalSite: {
             type: Boolean,
             default: false
         },
 
-        isRouterLink: {
```
zlatinivanov-je, line 42: What replaces `isRouterLink`? We have a case where that needs to be set.

zlatinivanov-je (commented):
Happy to approve, just to clarify setting router type links

packages/components/atoms/f-link/src/components/Link.vue
xander-marjoram: @zlatinivanov-je If you pass in the `to` attribute instead of `href`, it will behave as a router-link.

AndyMayJE: @zlatinivanov-je This logic was actually changed in 2.2.0 but the prop was leftover and is no longer used anywhere. The `linkType` computed property determines if to render `<a>` or `<router-link>` based on the `to`/`href` attributes.

https://github.com/justeat/fozzie-components/blob/master/packages/components/atoms/f-link/src/components/Link.vue#L108

zlatinivanov-je (approved):

xander-marjoram (approved):

PR #1456, opened 2021-11-09
fozzie-components@v5.3.0 – #globalconfig Node engine updated to v12 and lerna config update

ashleynolan:
### Added
- `templates` and `pages` component directories added to lerna config.

### Changed
- Updated engine from Node v10 > v12 on all packages (as we only test in Node 12+).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v5.3.0
+------------------------------
+*November 9, 2021*
+
+### Added
+- `templates` and `pages` component directories added to lerna config.
+
+### Changed
+- Updated engine from Node v10 > v12 on all packages (as we only test in Node 12+).
+
+
 v5.2.0
 ------------------------------
 *November 5, 2021*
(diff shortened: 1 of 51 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
 
+v5.3.0
+------------------------------
+*November 9, 2021*
+
+### Added
+- `templates` and `pages` component directories added to lerna config.
+
+### Changed
+- Updated engine from Node v10 > v12 on all packages (as we only test in Node 12+).
```
DamianMullins, line 15: Do you think this change should mean a major version bump in each package? If anyone is still using node 10 there aren't going to be able to use the packages anymore.

ashleynolan: The only reason I haven’t made this a major version is because we moved the repo over to Node 12 a while back - so the components may not work in Node 10 already.

As far as I understand, components _could_ still work on Node 10, but as our CI doesn’t test this and hasn’t for a while, I figured it best to update the package info.

DamianMullins:
@ashleynolan Yeah makes sense, thought I'd mention as yarn and npm won't perform the install if you're using an older version of node but you can override that with a command line flag (`--ignore-engines`).

CHANGELOG.md
DamianMullins: @ashleynolan Yeah makes sense, thought I'd mention as yarn and npm won't perform the install if you're using an older version of node but you can override that with a command line flag (--ignore-engines).

DamianMullins (approved):

PR #1460, opened 2021-11-09
f-card-with-content@0.2.0 - Add component functionality

xander-marjoram:
### Added
- Heading
- Description
- Primary and secondary buttons
  - `href`s are passed in via prop like `primary-button-href`.
  - Emit events like `primary-button-click`.
- Named slot for icon.
- Storybook story.

---

20211111T114247

---

### Changed
- Updated documentation for visual regression test setup.

---

## UI Review Checks

- [x] ~Unit~ visual tests have been [created|updated]

## Browsers Tested

- [x] Chrome (latest)


packages/components/molecules/f-card-with-content/CHANGELOG.md
```diff
@@ -4,6 +4,19 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.2.0
+------------------------------
+*November 10, 2021*
```
ashleynolan: Good forward planning 😆

packages/components/molecules/f-card-with-content/README.md
```diff
@@ -69,15 +73,31 @@ There may be props that allow you to customise its functionality.
 
 The props that can be defined are as follows (if any):
 
-| Prop  | Type  | Default | Description |
-| ----- | ----- | ------- | ----------- |
+| Prop                  | Type     | Default | Description                                                   |
+| --------------------- | -------- | ------- | ------------------------------------------------------------- |
+| `cardHeading`         | `String` | `''`    | If given, will render an `h1` tag within the card.            |
```
ashleynolan: Might not be needed for Contact Prefs, but may be good to have the heading type definable (so letting peoples choose `h2`/`h3` for instance would likely be useful. Could be a separate ticket potentially to do this – whatever you think makes the most sense.

packages/components/molecules/f-card-with-content/src/components/CardWithContent.vue
```diff
@@ -1,32 +1,107 @@
...
+            {{ cardHeading }}
+        </h1>
+
+        <p
+            v-if="cardDescription"
+            :class="$style['c-cardWithContent-description']">
+            {{ cardDescription }}
+        </p>
+
+        <f-button
```
ashleynolan, line 25: Could these need to link anywhere potentially?

If so, should there be a way of defining an `href` for these links/buttons? (or were you thinking it would get handled slightly differently through the emitters?)

packages/components/molecules/f-card-with-content/src/components/CardWithContent.vue
```diff
@@ -1,32 +1,107 @@
...
-    margin: auto;
-    border: 1px solid red;
-    font-family: $font-family-base;
-    @include font-size(heading-m);
+    flex-direction: column;
+    align-items: center;
+    text-align: center;
+}
+
+.c-cardWithContent-heading {
```
ashleynolan, line 92: Not sure if this will be the case here, but usually headings are slightly different sizes on narrow compared with wide viewport widths. So may need to define that here too (the font-map usually has two values, so if this is the wide font-size, you can do `@include font-size(heading-s, 'true', 'narrow');` (or something v similar to that to output the narrow size)

ashleynolan (commented):
Few comments/q's

packages/components/molecules/f-card-with-content/src/components/CardWithContent.vue
xander-marjoram: Yes, I did originally have the `href`s, but looking at how it works in f-checkout they [already have an event handler](https://github.com/justeat/fozzie-components/blob/master/packages/components/pages/f-checkout/src/components/Error.vue#L30) so thought it would be easier to stick to that pattern, and it's more flexible too.

packages/components/molecules/f-card-with-content/src/components/CardWithContent.vue
xander-marjoram: These styles were lifted straight out of f-checkout, which already works pretty well. The fact the heading is already small means it doesn't look strange when the viewport gets narrow.

packages/components/molecules/f-card-with-content/README.md
xander-marjoram: Yep, I considered it but I think it makes sense to add this when it's required.

matthewhardern:
This feels incredibly close to what f-media-element does??

matthewhardern (dismissed):
I am pretty sure we could accomplish this with combination of f-card and f-media-element, correct me if I am wrong though, but I am against adding components that do a specific thing when we could combine others to do that task. Or atleast modify it to do that task.

xander-marjoram:
@matthewhardern The idea of this component was to create a specific reusable format of `f-card` which we intend to use quite widely for error pages, but we wanted to keep to a more generic name so that it could be used for other purposes too, e.g., `f-takeawaypay-activation`, `f-loyalty`, etc.

matthewhardern:
> @matthewhardern The idea of this component was to create a specific reusable format of `f-card` which we intend to use quite widely for error pages, but we wanted to keep to a more generic name so that it could be used for other purposes too, e.g., `f-takeawaypay-activation`, `f-loyalty`, etc.

This is kind of my point we use f-media-element atm to serve the error / no results found parts of f-loyalty its already doing the above for that page.

DamianMullins:
Is there any description of the use case for this component? I'm a little confused about the differences between this and `f-card` and when you should use each?

xander-marjoram:
@DamianMullins I did add a description to the readme but happy to expand on it if you would like.

packages/components/molecules/f-card-with-content/README.md
ashleynolan: Can you make a separate ticket for it in EE?

packages/components/molecules/f-card-with-content/src/components/CardWithContent.vue
ashleynolan: I would check how accessible this is potentially - seems a bit strange to me to use events to redirect if it’s going to a url (if it’s not a form submission or something like that). 

Also, does that mean the buttons would do nothing if there was any sort of issue with the JS?

packages/components/molecules/f-card-with-content/src/components/CardWithContent.vue
ashleynolan: The narrow/wide headings are what’s set in PIE, so they should be used wherever those headings are used (as I understand it). Don’t want to tie this ticket to it, but this should be fixed so if make a ticket to check this with design and fix it if it is an issue

packages/components/molecules/f-card-with-content/src/components/CardWithContent.vue
```diff
@@ -1,32 +1,107 @@
...
+        <div
+            v-if="$slots.icon"
+            :class="$style['c-cardWithContent-icon']">
+            <slot name="icon" />
+        </div>
+
+        <h1
+            v-if="cardHeading"
+            :class="$style['c-cardWithContent-heading']">
+            {{ cardHeading }}
```
DamianMullins, line 16: What if you need to add markup to a heading, for example, I've needed to add `span` tags to heading a fair bit recently to allow for a mix of styles or hidden a11y content. It's not really possible with `prop`s, but you can do it with `slot`s. This is probably more evident with the cardDescription` below — you're limited to passing a single string of text that will be rendered into a single paragraph.

Pretty sure I've had this conversation elsewhere at some point but I'd always use slots to add content as they are much more flexible. In fact, I see you're using one above for the icon.

xander-marjoram: Yeah I see the need for that, and this can always be changed later as the use cases change, this definitely isn't set in stone, it does just enough for our use case at the moment.

The difficulty has been how far do we go with swapping everything out with slots? If we replace everything with a slot then we're effectively back at f-card, and we'd need to duplicate the common styling wherever this type of card is used.

packages/components/molecules/f-card-with-content/src/components/CardWithContent.vue
xander-marjoram: I agree and my preference would have been to use links. Checkout's error page button currently does some extra stuff as well, like removing a menu-related cookie before handling the redirect.

I understand that links would be the preference for semantic HTML, but "Go back button" should still be easily interpreted by screen reader users?

DamianMullins: You can use a link and still perform some operations using a `click` handler I believe?

packages/components/molecules/f-card-with-content/README.md
xander-marjoram: Ticket made 👍🏻

packages/components/molecules/f-card-with-content/src/components/CardWithContent.vue
xander-marjoram: Have updated to match the designs (which incidentally also needed to be updated 😄)

packages/components/molecules/f-card-with-content/src/components/CardWithContent.vue
xander-marjoram: Have updated to use an object which contains the text, handles `href` and `to`, and I've left in the event emitters

ashleynolan (approved):

oliversweb (approved):

PR #1463, opened 2021-11-10
f-checkout@3.6.0 - move all form logic out of `Checkout.vue`

FayeCarter:
### Changed
- `CheckoutForm` component.
- moved all form logic out of `Checkout.vue`.


packages/components/pages/f-checkout/src/components/CheckoutForm.vue
```diff
@@ -0,0 +1,326 @@
+<template>
+    <form
```
FayeCarter, line 2: Most of this has just been moved, it's not new

charliearlie (approved):

BanksySan (approved):

PR #1404, opened 2021-10-27
f-checkout@3.0.0 - Remove loading state

FayeCarter:
### Changed
- Remove spinner.

PR #1474, opened 2021-11-17
f-checkout@3.6.2 - Revert v3.0.0

FayeCarter:
### Fixed
- Spinner and loading state added back in.

### DIFF
--- a/packages/components/pages/f-checkout/src/components/Checkout.vue
+++ b/packages/components/pages/f-checkout/src/components/Checkout.vue
@@ -223,7 +223,6 @@ export default {
             shouldShowSpinner: false,
             isLoading: false,
             errorFormType: null,
-            isFormSubmitting: false,
             availableFulfilmentTimesKey: 0,
             checkoutAnalyticsService: new CheckoutAnalyticsService(this)
         };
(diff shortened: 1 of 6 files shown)


packages/components/pages/f-checkout/src/components/Checkout.vue
```diff
@@ -798,6 +817,28 @@ export default {
...
+
+        /**
+         * Sets the submitting state of the Checkout form. When true a spinner is displayed on the submit button
+         * This is done to allow us to test the setting of this value and ensure it is called with the correct value in the correct order.
+         * @param  {boolean} isFormSubmitting  - whether the form should be in a submitting state or not.
+         */
+        setSubmittingState (isFormSubmitting) {
+            this.isFormSubmitting = isFormSubmitting;
+        },
+
```
minrwhite: This function doesn't match up with #1404 and doesn't seem to be referenced anywhere else in the PR - I'm assuming this was left in from another changeset?

minrwhite (commented):
One comment, but everything else looks good

packages/components/pages/f-checkout/src/components/Checkout.vue
FayeCarter: Ah yes, this should be in a different component now. Thanks! I'll remove it

FayeCarter: Removed Thanks!

minrwhite (approved):

BanksySan (approved):

PR #1475, opened 2021-11-17
f-checkout@v3.7.0 - Remove spinner from checkout form.

FayeCarter:
### Removed
- Spinner from checkout form.

### DIFF
--- a/packages/components/pages/f-checkout/src/components/Checkout.vue
+++ b/packages/components/pages/f-checkout/src/components/Checkout.vue
@@ -223,7 +223,6 @@ export default {
             shouldShowSpinner: false,
             isLoading: false,
             errorFormType: null,
-            isFormSubmitting: false,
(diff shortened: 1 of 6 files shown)


BanksySan (dismissed):

BanksySan (dismissed):

packages/components/pages/f-checkout/src/components/Checkout.vue
```diff
@@ -414,9 +404,6 @@ export default {
          */
         async initialise () {
             this.setExperimentValues(this.experiments);
-            this.isLoading = true;
```
charliearlie, line 417: Does this not need to stay to avoid the bug we had? It's set to false in resetLoadingState.

BanksySan (approved):

charliearlie (approved):

PR #1479, opened 2021-11-18
f-account-info@v0.2.0 - Initial UI

tommcclean:
Not a complete component; just a development - adding rough UI from designs.

I may split to sub-internal-components later

v0.2.0
*November 19, 2021*

### Added
- Initial UI elements - similar to design
- Initial en-GB translation messages

image

### DIFF
--- a/packages/components/pages/f-account-info/src/components/AccountInfo.vue
+++ b/packages/components/pages/f-account-info/src/components/AccountInfo.vue
@@ -103,9 +103,16 @@
 import { VueGlobalisationMixin } from '@justeat/f-globalisation';
 
 import Card from '@justeat/f-card';
+import '@justeat/f-card/dist/f-card.css';
+
 import FormField from '@justeat/f-form-field';
+import '@justeat/f-form-field/dist/f-form-field.css';
+
 import FLink from '@justeat/f-link';
+import '@justeat/f-link/dist/f-link.css';
+
 import FButton from '@justeat/f-button';
+import '@justeat/f-button/dist/f-button.css';
 
 import tenantConfigs from '../tenants';
 
@@ -153,11 +160,6 @@ export default {
 </script>
 
 <style lang="scss" module>
-@import '@justeat/f-card/dist/f-card.css';
-@import '@justeat/f-form-field/dist/f-form-field.css';
-@import '@justeat/f-link/dist/f-link.css';
-@import '@justeat/f-button/dist/f-button.css';
-
 .c-accountInfo-customerCareLink {
     display: block;
     margin-bottom: spacing(x4);
--- a/packages/components/pages/f-account-info/src/components/AccountInfo.vue
+++ b/packages/components/pages/f-account-info/src/components/AccountInfo.vue
@@ -1,49 +1,178 @@
 <template>
-    <div
-        :class="$style['c-accountInfo']"
-        data-test-id="accountInfo">
-        {{ copy.text }}
-    </div>
+    <card
+        :card-heading="$t('accountDetails')"
+        data-test-id="account-info-card"
+        has-outline
+        is-page-content-wrapper>
+        <h2
+            class="u-spacingBottom--large">
+            {{ $t('yourDetails') }}
+        </h2>
+
+        <form
+            method="post"
+            @submit.prevent="onFormSubmit">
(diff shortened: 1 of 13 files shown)


FayeCarter (dismissed):

packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -1,49 +1,175 @@
...
+                :label-text="$t('fields.lastNameLabel')"
+                :placeholder="$t('fields.lastNamePlaceholder')"
+                :value="fields.lastName" />
+
+            <h2
+                class="u-spacingBottom--large">
+                {{ $t('deliveryAddress') }}
+            </h2>
+
+            <form-field
```
ashleynolan, line 45: Doesn't need to be done as part of this PR, but I'd check with @Preet-Sandhu / @ross-gratton if these form-fields could benefit from being grouped into fieldsets, from an accessibility standpoint.

packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -1,49 +1,175 @@
...
 
 <script>
-import { globalisationServices } from '@justeat/f-services';
-import tenantConfigs from '../tenants';
+import { VueGlobalisationMixin } from '@justeat/f-globalisation';
 
-// eslint-disable-next-line no-unused-vars
-import AccountInfoServiceApi from '../services/AccountInfoServiceApi';
+import Card from '@justeat/f-card';
+import '@justeat/f-card/dist/f-card.css';
```
ashleynolan, line 106: Better to load in the CSS as part of the style block

ashleynolan (commented):
Couple of small comments

packages/components/pages/f-account-info/src/components/AccountInfo.vue
tommcclean: Worth updating the readme's for the atoms - I followed these

I would put in the style block in coreweb, but didn't in fozzie as I didn't expect it would handle it

ashleynolan: Actually, I think you're right. It doesn't actually matter in the component, it's in CoreWeb it matters as that's where it gets extracted. So you're fine to leave it here in the component itself

tommcclean: I have moved it to styles and it still works; so probably worth us updating the readme for the atoms to get consistency as we can

packages/components/pages/f-account-info/src/components/AccountInfo.vue
tommcclean: Ill ask and pick it up in my next PR if needed

packages/components/pages/f-account-info/src/components/AccountInfo.vue
tommcclean: FYI: works locally when in styles; but not circleCI

image

FayeCarter (approved):

ashleynolan (approved):

packages/components/pages/f-account-info/src/tenants/en-GB.js
```diff
@@ -1,4 +1,32 @@
+const messages = {
+    accountDetails: 'Account Details',
+    yourDetails: 'Your details',
+    contactCustomerCareTeam: "If you'd like to change your email address, get in touch with our customer care team.",
```
xander-marjoram, line 4: Nit: single quotes (feel free to address in your next PR)

packages/components/pages/f-account-info/src/tenants/en-GB.js
```diff
@@ -1,4 +1,32 @@
...
+    deliveryAddress: 'Delivery Address',
+    fields: {
+        emailAddressLabel: 'Email address',
+        emailAddressPlaceholder: 'Your email address...',
+        firstNameLabel: 'First name*',
```
xander-marjoram, line 9: In later implementations, we should probably add the asterisks based on whether the property is required or not

xander-marjoram (approved):

dandel10n (approved):

PR #1480, opened 2021-11-19
f-feature-management@0.7.0 - Improved error handling

jnewc:
Minor improvements to error handling in f-feature-management vue service

### DIFF
--- a/packages/services/f-feature-management/package.json
+++ b/packages/services/f-feature-management/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@justeat/f-feature-management",
-  "version": "0.6.0",
+  "version": "0.7.0",
   "description": "Service for querying feature values provided by Feature Management",
   "main": "dist/f-feature-management.umd.js",
   "module": "dist/f-feature-management.es.js",
--- a/packages/services/f-feature-management/src/lib/trackExperiment.js
+++ b/packages/services/f-feature-management/src/lib/trackExperiment.js
@@ -22,7 +22,7 @@ function trackExperiment (experiment, onTrack) {
     try {
         onTrack(key, variant);
     } catch (error) {
-        logger.logError('An exception occurred while trying to send an experiment event', error);
+        logger.logError('An exception occurred while trying to send an experiment event', { error });
     }
 }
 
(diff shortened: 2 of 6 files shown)


packages/services/f-feature-management/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@justeat/f-feature-management",
-  "version": "0.6.0",
+  "version": "0.7.0",
```
xander-marjoram, line 3: Version doesn't match the changelog / PR title 👀

packages/services/f-feature-management/src/lib/trackExperiment.js
```diff
@@ -22,7 +22,7 @@ function trackExperiment (experiment, onTrack) {
     try {
         onTrack(key, variant);
     } catch (error) {
-        logger.logError('An exception occurred while trying to send an experiment event', error);
+        logger.logError('An exception occurred while trying to send an experiment event', { error });
```
xander-marjoram, line 25: Looks to me like `logError` [only takes one parameter](https://github.com/justeat/fozzie-components/blob/master/packages/services/f-feature-management/src/lib/logger.js#L5)?

jnewc: That's the default internal logger for non-Vue usages. See `createLoggerWrapper` for the one used when using the vue service (`FeaturesService`)

xander-marjoram (dismissed):

packages/services/f-feature-management/src/vue/services/features.service.js
```diff
@@ -77,20 +77,24 @@ export default class FeaturesService {
...
 
     /**
      * Loads the latest features configuration from the CDN
      */
     async update () {
-        const newConfigJson = await this.featureManagement.loadFromCdn();
-        this.store.dispatch(storeAction, newConfigJson);
+        try {
+            const newConfigJson = await this.featureManagement.loadFromCdn();
+            this.store.dispatch(storeAction, newConfigJson);
```
minrwhite: As [`dispatch()` returns a `Promise`](https://vuex.vuejs.org/api/#dispatch), it might be worth `await`ing this if we want to catch potential problems with the action as well.

minrwhite (approved):

xander-marjoram (approved):

PR #1489, opened 2021-11-23
f-card@3.2.0 - Add prop cardSizeCustom to allow two different widths.

kevinrodrigues:
We need two sized cards for the accounts pages, rather than setting the sizes on multiple components (f-contact-preferences, f-previous-orders etc) we will use this prop to change the width as it is rendered instead of using css styles in multiple places.

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/atoms/f-card/src/components/Card.vue
+++ b/packages/components/atoms/f-card/src/components/Card.vue
@@ -4,7 +4,8 @@
         :class="[
             $style['c-card'], {
                 [$style['c-card--outline']]: hasOutline,
-                [$style['c-card--pageContentWrapper']]: isPageContentWrapper
+                [$style['c-card--pageContentWrapper']]: isPageContentWrapper,
+                [$style[`c-card--cardSizeCustom--${cardSizeCustom}`]]: cardSizeCustom !== ''
             }]">
         <div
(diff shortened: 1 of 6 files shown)


siggerzz (approved):
Component + Storybook story look fine 👍🏼 

Don't know enough about Jest to comment

simonsandersje (approved):

packages/components/atoms/f-card/src/components/Card.vue
```diff
@@ -142,4 +148,13 @@ $card--pageContentWrapper-width           : 472px; // so that it falls on our 8p
     .c-card-heading--rightAligned {
         text-align: right;
     }
+
+    // The two card sizes used on the accounts page. e.g Previous orders & contact preferences cards.
+    .c-card--cardSizeCustom--large {
```
ashleynolan, line 153: Small naming thing – would just make this `c-card--sizeLarge` and `c-card--sizeMedium` as it's just a modifier on the base card class.

kevinrodrigues: Fixed. https://github.com/justeat/fozzie-components/pull/1503

PR #1504, opened 2021-11-25
f-contact-preferences@v0.9.0 - Register Module and Initialise Page based on Authorisation

oliversweb:
### Added
- Expose new isAuthFinished `required` property

### Changed
- Registered the Store Module on Create()
- Initialise Page based on isAuthFinished


packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -142,8 +147,25 @@ export default {
         ])
     },
 
+    watch: {
+        immediate: true,
```
SachaZvetelman: Does this work correctly? I thought `immediate` had to be inside the bit you're watching, and not outside.

packages/components/pages/f-contact-preferences/src/components/_tests/ContactPreferences.test.js
```diff
@@ -143,6 +166,34 @@ describe('ContactPreferences Component', () => {
...
+
+        it('should not call initialise method if the authorisation has not completed', () => {
+            // Arrange
+            sutProps = { ...sutProps, isAuthFinished: false };
+
+            // Act
+            wrapper = mountContactPreferences();
+
+            // Assert
+            expect(storeActions.loadPreferences).not.toHaveBeenCalled();
```
SachaZvetelman: The test's name refers to the initialise method, but the assert refers to something else. It's a bit confusing 😅

packages/components/pages/f-contact-preferences/src/components/_tests/ContactPreferences.test.js
```diff
@@ -143,6 +166,34 @@ describe('ContactPreferences Component', () => {
...
+        });
+
+        it('should only call initialise method once the authorisation has completed', async () => {
```
SachaZvetelman, line 189: If I understand correctly, I think in this test what you're trying to check is that when your watcher sees that the authorisation is complete, the initialise method is called.

To me, this test would look something like this (kinda pseudo-code):

```
// Arrange
wrapper = mountContactPreferences();

// Act
wrapper.vm.$options.watch.isAuthFinished.call(wrapper.vm, true);

// Assert
expect(initialiseSpy).toHaveBeenCalled();
```
More info: https://github.com/vuejs/vue-test-utils/issues/331#issuecomment-382037200

packages/components/pages/f-contact-preferences/stories/ContactPreferences.stories.js
```diff
@@ -72,11 +68,12 @@ export const ContactPreferencesComponent = () => ({
...
     template: '<contact-preferences ' +
     ':authToken="authToken" ' +
+    ':isAuthFinished="true" ' +
     ':locale="locale" ' +
-    ':smart-gateway-base-url="smartGatewayBaseUrl" ' +
+    'smart-gateway-base-url="" ' +
     ':prepareStory="prepareStory" ' +
     // eslint-disable-next-line no-template-curly-in-string
-    ' :key="`${authToken},${locale},${smartGatewayBaseUrl}`" />'
+    ' :key="`${authToken},${locale}`" />'
```
SachaZvetelman:
```suggestion
    ':key="`${authToken},${locale}`" />'
```

packages/components/pages/f-contact-preferences/src/components/_tests/ContactPreferences.test.js
oliversweb: @SachaZvetelman I want to mount the component with the `isAuthFinished` = false then check that the `initialise()` has not been called then set the prop and recheck that the `initialise()` has now been called. If I don't check it wasn't called on the `mount` I cannot prove the check; after setting the prop to `true`, was not for the previous `mount` action.

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
oliversweb: There is an alternative way ... I will try that 👍

packages/components/pages/f-contact-preferences/src/components/_tests/ContactPreferences.test.js
oliversweb: It was simpler to just check if a method inside the `Initialise()` had been called but I can set up a new spy 👍

packages/components/pages/f-contact-preferences/stories/ContactPreferences.stories.js
oliversweb: I think it was always like that do didn't want to break anything but can remove 👍

packages/components/pages/f-contact-preferences/src/components/_tests/ContactPreferences.test.js
SachaZvetelman: If you comment out the watcher, does this test still pass? I'm not sure. If it fails when you do that, I'm ok to approve. Could you please check that?

oliversweb: @SachaZvetelman I comment out the watcher and it failed

SachaZvetelman (approved):

JoshuaNg2332 (approved):

PR #1505, opened 2021-11-25
f-account-info@0.5.0 - Stop f-spinner once loaded

oliversweb:
### Added
- Send and event to stop the f-spinner invoked by the parent component


packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -174,7 +177,7 @@ export default {
             } catch (error) {
                 // TODO - to be added with next ticket
             } finally {
-                // TODO - Stop auth spinner to be added with next ticket
+                this.$parent.$emit(STOP_LOADING_SPINNER_EVENT);
```
kevinrodrigues: Tests?

packages/components/pages/f-account-info/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
 
+v0.5.0
+------------------------------
+*November 25, 2021*
+
+### Added
+- Send and event to stop the f-spinner invoked by the parent component
+
+
+
```
kevinrodrigues, line 26: Super minor but 3 line breaks (should be two).

packages/components/pages/f-account-info/src/constants.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export const STOP_LOADING_SPINNER_EVENT = 'stop-spinner';
```
kevinrodrigues: Should this be placed in an events file / folder rather than constants so we have all events sitting in one area of the app.

oliversweb: There will only be about x5 constant in the whole of the component and will be very similar to [this](https://github.com/justeat/fozzie-components/blob/master/packages/components/pages/f-contact-preferences/src/constants.js) so might be overkill.

packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -153,12 +162,22 @@ export default {
         };
     },
 
-    async mounted () {
-        await this.initialise();
+    watch: {
+        isAuthFinished () {
```
AndyMayJE, line 167: Consider using the immediate: true option to trigger the callback immediately with the current value of the expression: https://vuejs.org/v2/api/#vm-watch

AndyMayJE (dismissed):

packages/components/pages/f-account-info/src/constants.js
ashleynolan: A super small thing – I'd likely name this `SPINNER_EVENT_STOP_LOADING` or `EVENT_SPINNER_STOP_LOADING`, just as that way other potential events can have the same prefix `EVENT_blah_blah_blah` at the start of the constant name.

ashleynolan (commented):
Looks good – just one super small comment

packages/components/pages/f-account-info/src/components/_tests/AccountInfo.test.js
```diff
@@ -15,16 +18,68 @@ const i18n = {
...
+        sutProps = {
+            authToken: token,
+            isAuthFinished: true,
+            smartGatewayBaseUrl: baseUrl
+        };
+    });
+
+    describe('when mounting the component', () => {
+        it('should be defined', () => {
+            // Arrange
```
JoshuaNg2332: Can probably remove this.

ashleynolan (approved):

AndyMayJE (approved):

PR #1519, opened 2021-12-01
f-contact-preferences@0.10.0 - Update to use f-form-field

AndyMayJE:
### Changed

- Generic checkboxes to use `f-form-field`

image

### DIFF
--- a/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
+++ b/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
@@ -1,6 +1,7 @@
 <template>
     <div
-        data-test-id="contact-preferences">
+        data-test-id="contact-preferences"
+        :class="$style['c-contactPreferences']">
         <card-component
             v-if="!shouldShowErrorPage"
             :card-heading="$t('heading')"
@@ -278,6 +279,7 @@ export default {
 </style>
 
 <style lang="scss" module>
+
 .c-contactPreferences-fieldset {
     display: flex;
     flex-flow: column;
@@ -288,6 +290,16 @@ export default {
 
 .c-contactPreferences-subtitle {
     @include font-size(heading-s);
+}
+
+.c-contactPreferences {
+    .c-contactPreferences-formField {
+        margin-top: 0;
+    }
+    .c-contactPreferences-formField+.c-contactPreferences-formField {
+        margin-top: 0;
+        margin-bottom: spacing(x2);
+    }
 }
 
 .c-contactPreferences-btn {
--- a/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
+++ b/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
@@ -273,7 +273,7 @@ export default {
 
 <style lang="scss">
 .c-formField-label,
-.c-formField-label-description {
+.c-formField-label-description  {
     margin-bottom: 0;
 }
 </style>
@@ -296,7 +296,7 @@ export default {
     .c-contactPreferences-formField {
         margin-top: 0;
     }
-    & + .c-contactPreferences-formField {
+    .c-contactPreferences-formField + .c-contactPreferences-formField {
         margin-top: 0;
         margin-bottom: spacing(x2);
     }
--- a/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
+++ b/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
@@ -295,10 +295,9 @@ export default {
 .c-contactPreferences {
     .c-contactPreferences-formField {
         margin-top: 0;
-    }
-    .c-contactPreferences-formField + .c-contactPreferences-formField {
-        margin-top: 0;
-        margin-bottom: spacing(x2);
+        & + & {
+            margin-bottom: spacing(x2);
+        }
     }
 }
 
--- a/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
+++ b/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
@@ -18,6 +18,7 @@
                     <fieldset
                         :class="$style['c-contactPreferences-fieldset']">
                         <f-form-field
+                            :class="$style['c-contactPreferences-form-field']"
                             input-type="checkbox"
                             :label-text=" $t(`${key}.email`)"
                             :label-description="getEmailDescription(key)"
@@ -28,6 +29,7 @@
                         />
 
                         <f-form-field
+                            :class="$style['c-contactPreferences-form-field']"
                             input-type="checkbox"
                             :label-text=" $t(`${key}.sms`)"
                             :label-description="getSmsDescription(key)"
@@ -267,6 +269,15 @@ export default {
 </script>
 
 <style lang="scss" module>
+
+.c-contactPreferences-form-field {
+    margin-top: 0 !important;
+    & label,
+    & label span {
+        margin-bottom: 0 !important;
+    }
+}
+
 .c-contactPreferences-fieldset {
     display: flex;
     flex-flow: column;
--- a/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
+++ b/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
@@ -174,7 +174,7 @@ export default {
         getEmailDescription (key) {
             if (!this.$te(`${key}.emailDescription`)) return undefined;
 
-            return this.$t(`${key}.emailDescription`);
+            return `(${this.$t(`${key}.emailDescription`)})`;
         },
 
         /**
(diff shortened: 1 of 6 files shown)


packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -17,40 +17,29 @@
...
-                                    [$style['c-contactPreferences-labelText--disabled']]: !isSmsEnabled
-                                }">
-                                {{ $t(`${key}.sms`) }}
-                                <template v-if="$te(`${key}.smsDescription`)">
-                                    <br>({{ $t(`${key}.smsDescription`) }})
-                                </template>
-                            </span>
-                        </label>
+                        <f-form-field
+                            :class="$style['c-contactPreferences-form-field']"
```
ashleynolan: To match the naming CSS naming convention, this should be `formField` here (and below where it's used too)

```suggestion
                            :class="$style['c-contactPreferences-formField']"
```

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -17,40 +17,29 @@
...
+                            :label-description="getEmailDescription(key)"
+                            :disabled="!isEmailEnabled"
+                            :data-test-id="`contact-preferences-${key}-email-checkbox`"
+                            :name="`contact-preferences-${key}-email`"
+                            :checked="emailValue"
+                            @input="(checked) => editPreferenceValue(key, Object.keys({ emailValue })[0], checked)"
+                        />
+
+                        <f-form-field
+                            :class="$style['c-contactPreferences-form-field']"
```
ashleynolan: As above:

```suggestion
                            :class="$style['c-contactPreferences-formField']"
```

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -177,6 +169,28 @@ export default {
             'editPreference'
         ]),
 
+
+
+        /**
+        * Returns translation string if it exists
+        * @param {string} key - The key of the preference that needs changing
+        */
+        getEmailDescription (key) {
```
ashleynolan, line 179: Just to clarify for my understanding more than anything – I'm guessing this is here to help with error handling, as passing the direct value if it's not defined causes issues?

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -257,6 +271,15 @@ export default {
 </script>
 
 <style lang="scss" module>
+
+.c-contactPreferences-form-field {
```
ashleynolan:
```suggestion
.c-contactPreferences-formField {
```

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -257,6 +271,15 @@ export default {
...
+    margin-top: 0 !important;
```
ashleynolan: Would be good to understand why you need the `!important` here and below – does the styling come down from the component otherwise and that takes precedence? If so, which rule is causing the issue, as usually this indicates a problem in the way the component defines its styling.

ashleynolan (changes requested):

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
AndyMayJE: `!important` is used by `f-form-field` in several places:

image

The scoping on `f-form-field` also makes it hard for the parent component to override styles defined as scoped classes.

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
AndyMayJE: Yes, if we don't perform the check a string is returned  = `news.smsDescription`. We are also wrapping the string in brackets ([existing functionality](https://github.com/justeat/fozzie-components/blob/master/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue#L33)) so the check ensures brackets are correctly added.

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -177,6 +169,28 @@ export default {
...
         ]),
 
+
+
+        /**
+        * Returns translation string if it exists
+        * @param {string} key - The key of the preference that needs changing
+        */
+        getEmailDescription (key) {
+            if (!this.$te(`${key}.emailDescription`)) return undefined;
```
xander-marjoram: I don't have a big problem with this, but I do think it is generally a little tidier to have a consistent return type 
```suggestion
            if (!this.$te(`${key}.emailDescription`)) return '';
```

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
xander-marjoram: Would be good if we could find a way to avoid using `!important` inside the form field component.

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -17,40 +17,29 @@
...
+                            @input="(checked) => editPreferenceValue(key, Object.keys({ emailValue })[0], checked)"
+                        />
+
+                        <f-form-field
+                            :class="$style['c-contactPreferences-formField']"
+                            input-type="checkbox"
+                            :label-text=" $t(`${key}.sms`)"
+                            :label-description="getSmsDescription(key)"
+                            :disabled="!isSmsEnabled"
+                            :data-test-id="`contact-preferences-${key}-sms-checkbox`"
```
xander-marjoram, line 39: Do these `data-test-id`s need updating? They don't seem to match up with [these](https://github.com/justeat/fozzie-components/pull/1519/files#diff-f77ef9c2d4e7399a7501bedce43bd813aa369fcb6349df4e9d0448e3bb248421).

AndyMayJE: No, the selector is having to use the wrapper div returned by `f-form-field` for `click()`. It fails when using the test id of the element as the styling hides the actual input field (opacity: 0).

image

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
AndyMayJE: I agree. 

We could remove the `u-spacing....` classes from `f-form-field` and use a class without `!important` This would allow us to style from the parent component without the `!important`?  I am hesitant to remove the `!important` from the base styles not knowing the impact this might have.

ashleynolan: So the reason the important is on these, is that utility/trump classes like that shouldn't be overwritten (otherwise, there's no point in it being a trump class).

I'd suggest removing these trump classes from f-form-field label (as you also suggested), as if they aren't always the value we want to use, they shouldn't be applied and will cause issues like this when trying to override them.

AndyMayJE: @ashleynolan  @xander-marjoram  would you mind taking a look at - https://github.com/justeat/fozzie-components/pull/1531

AndyMayJE: `f-form-field` has been updated and the use of `!important` removed from this pr :)

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -265,14 +288,20 @@ export default {
...
 }
 
+.c-contactPreferences {
+    .c-contactPreferences-formField {
+        margin-top: 0;
+    }
+    .c-contactPreferences-formField+.c-contactPreferences-formField {
+        margin-top: 0;
+        margin-bottom: spacing(x2);
+    }
```
ashleynolan:
```suggestion
    .c-contactPreferences-formField {
        margin-top: 0;

        & + .c-contactPreferences-formField {
            margin-top: 0;
            margin-bottom: spacing(x2);
        }
    }
```

packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -265,14 +288,20 @@ export default {
...
 
-.c-contactPreferences-labelText--disabled {
-    color: $color-content-disabled;
-}
-
 .c-contactPreferences-subtitle {
     @include font-size(heading-s);
 }
 
+.c-contactPreferences {
```
ashleynolan, line 295: Do you need to wrap these selectors in the outer `c-contactPreferences`?

AndyMayJE: Yes it needs the higher specificity to override the style on `f-form-field`

ashleynolan (dismissed):

xander-marjoram (approved):

ashleynolan (approved):

PR #1523, opened 2021-12-02
f-header@7.3.0 - Add support for custom navigation links

xander-marjoram:
### Added
- Support for custom links in the nav via `customNavLinks` prop.
- Relevant knobs to component story.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/organisms/f-header/src/components/Header.vue
+++ b/packages/components/organisms/f-header/src/components/Header.vue
@@ -24,6 +24,7 @@
 
             <navigation
                 :copy="copy"
+                :custom-nav-links="customNavLinks"
                 :show-delivery-enquiry="showDeliveryEnquiryWithContent"
                 :show-offers-link="showOffersLinkWithContent"
                 :show-help-link="showHelpLink"
@@ -60,6 +61,11 @@ export default {
         locale: {
             type: String,
             default: ''
+        },
+
+        customNavLinks: {
+            type: Array,
+            default: () => []
         },
 
         errorLog: {
--- a/packages/components/organisms/f-header/src/components/Navigation.vue
+++ b/packages/components/organisms/f-header/src/components/Navigation.vue
@@ -348,7 +348,8 @@ export default {
 
         customNavLinks: {
             type: Array,
-            default: () => []
+            default: () => [],
+            validator: links => links.every(link => link.text && link.url)
         }
     },
 
--- a/packages/components/organisms/f-header/src/components/Navigation.vue
+++ b/packages/components/organisms/f-header/src/components/Navigation.vue
@@ -82,6 +82,19 @@
                 :class="$style['c-nav-list']"
                 data-test-id="nav-list">
                 <li
+                    v-for="(customNavLink, index) in customNavLinks"
+                    :key="`custom-nav-link-${index}`"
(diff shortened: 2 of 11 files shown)


packages/components/organisms/f-header/src/components/Header.vue
```diff
@@ -62,6 +63,11 @@ export default {
             default: ''
         },
 
+        customNavLinks: {
```
ashleynolan, line 66: Should there be any validation for the values passed in via the array? (the main thing I'm thinking would be, what happens if someone specifies `text` but not `url`, or vice versa?).

packages/components/organisms/f-header/stories/header.stories.js
```diff
@@ -49,6 +49,12 @@ export const HeaderComponent = () => ({
         },
         showSkipLink: {
             default: boolean('Show skip link', true)
+        },
+        showCustomNavLinks: {
```
ashleynolan: Not sure if it's a bit confusing to have this control in storybook when it's not an actual prop? If you think it's worthwhile having here, I'd probably leave a comment just to say that the knob is only there for storybook to aid in showing and hiding an example of the links, as I'm guessing that's why it's there?

xander-marjoram: Have updated the story to use controls rather than knobs, which makes (or will make) it easier to pass complex values and allow for proper descriptions.

packages/components/organisms/f-header/src/components/Header.vue
xander-marjoram: This is possible and I'll look into this on Monday, hopefully an issue that easy to test wouldn't make it out of local development 😅

ashleynolan: I'd like to think so too, but always worth considering what happens if people misuse the contract that's provided – kind of acts as a form of documentation for anyone using the prop in the wrong way then.

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -331,6 +344,12 @@ export default {
         showCountrySelector: {
             type: Boolean,
             default: false
+        },
+
+        customNavLinks: {
+            type: Array,
+            default: () => [],
+            validator: links => links.every(link => link.text && link.url)
```
zlatinivanov-je, line 352: Would an Object containing GTM data produce a validation warning here? Should it be accounted for?

xander-marjoram: @zlatinivanov-je The `gtm` object isn't currently required, so I haven't included it in the validation, what kind of validation did you have in mind?

zlatinivanov-je: Looking at the code and trying it my self did not come with a good solution for what I was proposing. 
Resolving the conversation. 
Just wanted to try and flag if something is being missed.

zlatinivanov-je (approved):

DamianMullins (approved):

ashleynolan (approved):

PR #1525, opened 2021-12-06
generator-component@2.5.0 - Added visual regression option and changed file names

LTurns:
### Added
- Prompt option for visual regression
- Spec templates for visual regression

### Changed
-names in `test-utils` and `test` from `componentName` to `component-name`
-`f-wdio-utils` to latest version

### Removed
- Unused functions from within the component object - `open()` and `load()`

### DIFF
--- a/packages/tools/generator-component/generators/app/templates/__package__.json
+++ b/packages/tools/generator-component/generators/app/templates/__package__.json
@@ -35,12 +35,12 @@
     "build": "vite build",
     "lint": "eslint \"!(dist)/**/*.{js,vue}\""<% } %>,
     "lint:fix": "yarn lint --fix"<% if(config.isComponent) { %>,
-    "lint:style": "vue-cli-service lint:style",
-    "test:coverage": "jest test:unit --coverage",
-    "test": "vue-cli-service test:unit"<% if(config.needsComponentTests) { %>,
+    "lint:style": "vue-cli-service lint:style"<% } else { %>,
+    "test:coverage": "jest test:unit --coverage"<% } %>,
+    "test": "vue-cli-service test:unit"<% if(config.isComponent) { %><% if(config.needsComponentTests) { %>,
     "test-component:chrome": "cross-env-shell TEST_TYPE=component wdio ../../../../wdio-chrome.conf.js"<% } %>,
     "test-a11y:chrome": "cross-env-shell TEST_TYPE=a11y wdio ../../../../wdio-chrome.conf.js"<% if(config.needsVisualTests) { %>,
-    "test:visual": "cross-env-shell PERCY_TOKEN=${PERCY_TOKEN_F_<%= name.class.toUpperCase() %>} TEST_TYPE=visual percy exec -- wdio ../../../../wdio-chrome.conf.js"<% } %><% } %>
+    "test:visual": "cross-env-shell PERCY_TOKEN=${PERCY_TOKEN_F_<%= name.default.toUpperCase().replace('-', '_') %>} TEST_TYPE=visual percy exec -- wdio ../../../../wdio-chrome.conf.js"<% } %><% } %>
   },
   "browserslist": [
     "extends @justeat/browserslist-config-fozzie"
--- a/packages/tools/generator-component/generators/app/templates/__package__.json
+++ b/packages/tools/generator-component/generators/app/templates/__package__.json
@@ -52,7 +52,7 @@
     "@justeat/browserslist-config-fozzie": ">=1.2.0"
   },
   "devDependencies": {
-<% if(config.isComponent) { %>    "@justeat/f-wdio-utils": "0.4.0",
+<% if(config.isComponent) { %>    "@justeat/f-wdio-utils": "0.5.0",
     "node-sass-magic-importer": "5.3.2",
     "@samhammer/vue-cli-plugin-stylelint": "2.1.0",
     "@vue/cli-plugin-babel": "4.5.15",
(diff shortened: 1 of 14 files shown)


packages/tools/generator-component/generators/app/templates/__package__.json
```diff
@@ -6,9 +6,9 @@
   "module": "dist/f-<%= name.default %>.es.js"<% } %><% if(config.needsBundlewatch) { %>,
   "maxBundleSize": "<%= bundlewatchMaxSize %>kB"<% } %>,
   "files": [
-    "dist"<% if(config.needsComponentTests) { %>,
-    "test-utils"<% } %><% if(config.needsTestingApiMocks) { %>,
-    "src/services"<% } %>
+    "dist",
+    "test-utils",
+    "src/services"
```
oliversweb: I don't think this needs to be exported, does it?

oliversweb:
Shall we create desktop/mobile visual tests as part of the generator?

packages/tools/generator-component/generators/app/templates/__package__.json
ashleynolan: I agree – I think this only needs to be exported in certain circumstances?

packages/tools/generator-component/generators/app/templates/__package__.json
```diff
@@ -38,8 +38,9 @@
     "lint:style": "vue-cli-service lint:style"<% } else { %>,
     "test:coverage": "jest test:unit --coverage"<% } %>,
     "test": "vue-cli-service test:unit"<% if(config.needsComponentTests) { %>,
-    "test-component:chrome": "cross-env-shell TEST_TYPE=component wdio ../../../../wdio-chrome.conf.js",
-    "test-a11y:chrome": "cross-env-shell TEST_TYPE=a11y wdio ../../../../wdio-chrome.conf.js"<% } %>
+    "test-component:chrome": "cross-env-shell TEST_TYPE=component wdio ../../../../wdio-chrome.conf.js"<% } %>,
+    "test-a11y:chrome": "cross-env-shell TEST_TYPE=a11y wdio ../../../../wdio-chrome.conf.js"<% if(config.needsVisualTests) { %>,
```
ashleynolan, line 42: I'm guessing this will include this line even if the component is a service potentially? I think it probably needs to be wrapped in the above `needsComponentTests` if statement?

packages/tools/generator-component/generators/app/templates/__package__.json
LTurns: ah yep you're right the `src/services` doesn't - will revert that now. The `test-utils` technically would be needed even if there were no component tests as its used for our accessibility tests. But actually I don't think there'll be a situation where we'll start testing a component in CoreWeb that doesn't have component tests - so happy to revert this too! 🤦

packages/tools/generator-component/generators/app/templates/__package__.json
LTurns: I'm not sure - does this line solve that? https://github.com/justeat/fozzie-components/pull/1525/files#diff-22e5b67bfabcb4c42acfca9714ab1d95874f60b9bc33a20726363824807fb021R41 
I was wondering about putting it inside `needsComponentTests` but kept them separate in case there's a scenario where we'll need visual tests but not component tests. Happy to change though

ashleynolan: I think the `test-a11y:chrome` script isn't wrapped in any if statement atm though, so it'll always be included in the package.json atm?

packages/tools/generator-component/generators/app/templates/__package__.json
LTurns: > Shall we create desktop/mobile visual tests as part of the generator?

Added the starter template for these 😄  https://github.com/justeat/fozzie-components/pull/1525/files#diff-34c172c4d98130674099f6c035932983bb7c21369c31f26325836091386a6de8

packages/tools/generator-component/generators/app/templates/__package__.json
LTurns: oh I see, thank you! The accessibility tests are configured to always be generated even if there are no component tests - so it's probably good that the script isn't in the `needsComponentTests` anymore 😄  - I could wrap it around `isComponent` instead so it doesn't get added to services?

ashleynolan: Yeh, that'd be good. Just need to make sure it doesn't output for services, so as long as that case is covered then I'll be happy 👍

LTurns: awesome, thank you - just fixed 👍

oliversweb (dismissed):

packages/tools/generator-component/generators/app/templates/test-utils/component-objects/f-skeleton.component.js
```diff
@@ -1,25 +1,15 @@
...
-    }
-
-    open (url) {
-        super.open(url);
-    }
-
+    
     waitForComponent () {
         super.waitForComponent(this.component);
     }
```
oliversweb, line 15: If you are bumping wdio to v0.5.0 then you might want this little extra sugar ...
```suggestion
    waitForComponent (timeoutMs) {
        super.waitForComponent(this.component, timeoutMs);
    }
```

oliversweb (approved):

JoshuaNg2332 (approved):

PR #1526, opened 2021-12-06
f-account-info@v0.8.0 - Add the Vuex store (get data only)

oliversweb:
### Added
- Vuex store - get, map & store consumer details/address data
---
This is just the getting of the data .... the saving of the data will follow once we have the spec.


packages/components/pages/f-account-info/src/services/mapping/consumer.mapper.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Validates the parameter passed is an Array then finds the default consumers address
+ * then maps the raw api response to the view model.
+ * @param {object} addresses - The consumers addresses (i.e. response.data).
+ * @returns {object} A view model that can be used to presents the component.
+ */
+export const mapToConsumerAddress = addresses => {
+    let address;
+    if (Array.isArray(addresses)) {
+        address = addresses.find(e => e.IsDefault);
```
jamieomaguire, line 28: Do you need a chained `?.` in the event of e being null?

packages/components/pages/f-account-info/src/store/accountInfo.module.js
```diff
@@ -0,0 +1,48 @@
...
+
+            const responses = await Promise.all([getDetailsTask, getAddressTask]);
+
+            const details = mapToConsumerDetails(responses[0]?.data);
+            const address = mapToConsumerAddress(responses[1]?.data);
+
+            commit(UPDATE_CONSUMER_DETAILS, { details, address });
+        },
+
+        async editConsumerDetails ({ commit, state }, { field, value }) {
```
jamieomaguire: is async needed here if nothing is being awaited?

oliversweb: Good spot 👍

packages/components/pages/f-account-info/src/services/mapping/consumer.mapper.js
oliversweb: Added test to prove `e` is not evaluated in the case of no elements 👍

jamieomaguire (approved):

AndyMayJE (approved):

PR #1530, opened 2021-12-07
f-restaurant-card@v0.11.0 - add local legend logo

dawidchar:
v0.10.0
*December 07, 2021*

### Added
- Local Legend Logo

---
## Screenshots

image

### DIFF
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
@@ -57,7 +57,7 @@
                 <div
                     v-if="isPremier"
                     :class="[$style['c-restaurantCard-localLegend']]">
-                    <legend-icon />
+                    <legend-icon data-test-id="local-legend-icon" />
                 </div>
             </component>
             <!-- END ERROR BOUNDARY -->
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
@@ -37,13 +37,6 @@
                 {{ name }}
             </h3>
 
-            <!-- Local Legend -->
-            <div
-                v-if="isPremier"
-                :class="[$style['c-restaurantCard-localLegend']]">
-                <legend-icon />
-            </div>
-
             <!-- Cuisines -->
             <!-- START ERROR BOUNDARY -->
             <component
@@ -56,6 +49,18 @@
             </component>
             <!-- END ERROR BOUNDARY -->
 
+            <!-- Local Legend label -->
+            <!-- START ERROR BOUNDARY -->
+            <component
+                :is="errorBoundary"
+                tier="3">
+                <div
+                    v-if="isPremier"
+                    :class="[$style['c-restaurantCard-localLegend']]">
+                    <legend-icon />
+                </div>
+            </component>
+            <!-- END ERROR BOUNDARY -->
 
             <!-- New label -->
             <!-- START ERROR BOUNDARY -->
@@ -97,15 +102,6 @@
                     v-if="displayDeliveryTimeMeta"
                     v-bind="deliveryTimeData"
                     data-test-id="restaurant-delivery-time-meta" />
-            </component>
-            <!-- END ERROR BOUNDARY -->
-
-            <!-- Local Legend label -->
-            <!-- START ERROR BOUNDARY -->
-            <component
-                :is="errorBoundary"
-                tier="3">
-                <slot name="local-legend" />
             </component>
             <!-- END ERROR BOUNDARY -->
 
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
@@ -37,6 +37,13 @@
                 {{ name }}
             </h3>
 
+            <!-- Local Legend -->
+            <div
+                v-if="isPremier"
+                :class="[$style['c-restaurantCard-localLegend']]">
+                <legend-icon />
+            </div>
+
             <!-- Cuisines -->
             <!-- START ERROR BOUNDARY -->
             <component
@@ -144,7 +151,7 @@
 </template>
 
 <script>
-import { OfferIcon } from '@justeat/f-vue-icons';
+import { OfferIcon, LegendIcon } from '@justeat/f-vue-icons';
 import ErrorBoundaryMixin from '../assets/vue/mixins/errorBoundary.mixin';
 import RestaurantImage from './subcomponents/RestaurantImage/RestaurantImage.vue';
 import RestaurantLogo from './subcomponents/RestaurantLogo.vue';
@@ -168,7 +175,8 @@ export default {
         RestaurantRating,
         DeliveryTimeMeta,
         IconText,
-        OfferIcon
+        OfferIcon,
+        LegendIcon
     },
     mixins: [ErrorBoundaryMixin],
     // NOTE: These are merely some placeholder props and not indicative of the props we will end up using
@@ -229,6 +237,10 @@ export default {
         offer: {
             type: String,
             default: null
+        },
+        isPremier: {
+            type: Boolean,
+            default: false
         }
     },
     computed: {
@@ -319,5 +331,10 @@ export default {
     @include media('>mid') {
         bottom: spacing(x1.5);
     }
+}
+
+.c-restaurantCard-localLegend {
+    width: 45px;
+    height: 21px;
 }
 </style>
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
@@ -52,8 +52,8 @@
             <!-- Local Legend Logo -->
             <div
                 v-if="isPremier"
-                :class="[$style['c-restaurantCard-localLegend']]">
-                <legend-icon data-test-id="local-legend-icon" />
+                :class="[$style['c-restaurantCard-premier']]">
+                <legend-icon data-test-id="premier-icon" />
             </div>
 
             <!-- New label -->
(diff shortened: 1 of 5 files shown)


packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -49,6 +49,18 @@
...
+            <!-- START ERROR BOUNDARY -->
+            <component
+                :is="errorBoundary"
+                tier="3">
+                <div
+                    v-if="isPremier"
+                    :class="[$style['c-restaurantCard-localLegend']]">
+                    <legend-icon data-test-id="local-legend-icon" />
+                </div>
+            </component>
```
jamieomaguire: I hate to say this, but the error boundary seems redundant here - if this was going to break, it would be during the process of building the component props object to pass down from the consumer (search, home etc) - which I hope we would handle with try/catches or some other form of graceful error handling. 

I think it might be worth dropping the error boundary from this component (can do the same for similar in future PRs to 'fix forward', reserving it for more complex components that have some internal logic that could be prone to breaking. 

What do you think?

packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
```diff
@@ -236,4 +236,56 @@ describe('RestaurantCard.v1', () => {
             expect(wrapper.find('[data-test-id="restaurant-delivery-time-meta"]').exists()).toBe(true);
         });
     });
+
+    describe('Local Legend', () => {
+        it('should load an svg', () => {
```
jamieomaguire: I'm not completely sold on the benefits of this particular test - I think the three below are likely more than adequate

packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
```diff
@@ -236,4 +236,56 @@ describe('RestaurantCard.v1', () => {
...
+            };
+
+            // act
+            const wrapper = mount(RestaurantCardV1, { propsData });
+
+            // assert
+            expect(wrapper.find('[data-test-id="local-legend-icon"]').exists()).toBe(false);
+        });
+
+        it('should not show Local Legends when premier data is missing', () => {
```
jamieomaguire: could potentially it.each this and test with `null`, `undefined` and omitting the property as you've done below

jamieomaguire:
Please can you include some screen shots, and add an example into storybook? That makes it easier for other reviewers to see what you've added.

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -301,4 +310,9 @@ export default {
         bottom: spacing(x1.5);
     }
 }
+
+.c-restaurantCard-localLegend {
+    width: 45px;
+    height: 21px;
+}
```
jamieomaguire: It's difficult to review this without seeing some screenshots. Does this perfectly preserve the ratio?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -49,6 +49,18 @@
             </component>
             <!-- END ERROR BOUNDARY -->
 
+            <!-- Local Legend label -->
+            <!-- START ERROR BOUNDARY -->
+            <component
+                :is="errorBoundary"
+                tier="3">
+                <div
```
jamieomaguire: I'm not convinced we need the div. Have you identified the issue of why this is needed?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
dawidchar: Totally Agree, I wasn't going to add it in, just saw other components low risk use it, so decided to just leave it in

jamieomaguire: Feel free to question/challenge existing implementations - no need to copy if you think there's an issue with it! :)

packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
dawidchar: This ensures an SVG is loaded, if the icons package for example was not working, this test would fail, but also a simple manual test should catch this too, so happy to remove it

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
dawidchar: Yes, this preserves the ratio, I've added a screenshot

packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
jamieomaguire: I'm not sure if that's worth testing tbh. That would require a package update which I hope would be properly tested both manually and verified by visual tests - once we set those up. 

Not sure what would happen to the presence of the test-id if the library failed. Not sure we need to be too worried though.

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -49,6 +49,18 @@
...
+                    v-if="isPremier"
+                    :class="[$style['c-restaurantCard-localLegend']]">
+                    <legend-icon data-test-id="local-legend-icon" />
```
jamieomaguire: can we rename this to something like `premier-restaurant-icon`? In case the local legend thing changes. We probably shouldn't reference local legends directly and just refer to it as a premier restaurant.

jamieomaguire: same for css classes

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
dawidchar: The SVG seems to have an invisible _bottom margin_, so I added a div with a set height to remove it

In the screenshots, I removed the set height and let it auto scale and you can see whitespace under the SVG
image
image

Here it is with the set height SVG
image

packages/components/molecules/f-restaurant-card/src/components/_tests/RestaurantCard.v1.test.js
dawidchar: Awesome, happy to remove it

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
dawidchar: It might be that the viewBox of the SVG is larger than the SVG causing this issue

ashleynolan (dismissed):
Needs a version bump to the package.json, but otherwise looks good

siggerzz:
Unrelated to this ticket, but just a general observation. Looks like this component hasn't got many tests associated with it yet. Would it be possible to add some browser based tests for this component? 

You can see an overview of the tests we expect here - https://vue.pie.design/?path=/story/documentation-standards-testing-overview--page

In terms of effort and value, I think it'd be good to create some visual regression tests at a minimum, as they're pretty easy to implement. You can find a guide here - https://vue.pie.design/?path=/story/documentation-standards-testing-visual-regression--page

It obviously doesn't have to be done as part of this PR, but would be good to do ASAP, since this component is used in SearchWeb.

We're happy to assist / pair with this if needed 👍🏼

jamieomaguire:
> Unrelated to this ticket, but just a general observation. Looks like this component hasn't got many tests associated with it yet. Would it be possible to add some browser based tests for this component?
> 
> You can see an overview of the tests we expect here - https://vue.pie.design/?path=/story/documentation-standards-testing-overview--page
> 
> In terms of effort and value, I think it'd be good to create some visual regression tests at a minimum, as they're pretty easy to implement. You can find a guide here - https://vue.pie.design/?path=/story/documentation-standards-testing-visual-regression--page
> 
> It obviously doesn't have to be done as part of this PR, but would be good to do ASAP, since this component is used in SearchWeb.
> 
> We're happy to assist / pair with this if needed 👍🏼

Thanks @siggerzz - we will absolutely be implementing browser, visual and accessibility tests before this component is put to use. We've put off adding them just yet because this component is in a very constant state of change both with markup, design and some of the functionality. 

Is that an acceptable approach? 

Just to be clear, I have added this component to SearchWeb behind a toggle and it is absolutely not planned to go into use until the component phase 1 is completed and tests mentioned above are added. So there isn't a risk of this affecting production.

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -49,6 +49,11 @@
             </component>
             <!-- END ERROR BOUNDARY -->
 
+            <!-- Local Legend Logo -->
```
jamieomaguire, line 52: NIT: maybe change to isPremier logo?

jamieomaguire (approved):

jamieomaguire:
@siggerzz also, would be fantastic to pair on some of these tests when we write them - thank you for the offer!

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
dawidchar: Will fix forward 😄

sonnyprince:
Unsure if this will come in the future but do we need to amend the alignment? - was just looking at figma:
Screenshot 2021-12-08 at 13 26 09

dawidchar:
> Unsure if this will come in the future but do we need to amend the alignment? - was just looking at figma: Screenshot 2021-12-08 at 13 26 09

Yes potentially @sonnyprince, the alignment of the metadata will be part of another ticket. Currently, this is just to bring the component into the restaurant card and have it working correctly.

sonnyprince (approved):

PR #1534, opened 2021-12-07
f-wdio-utils@v0.5.0 - Expose the timeout option on `waitForComponent()` method

oliversweb:
### Changed
- `waitForComponent` - Expose the timeout option on the underlying `waitForExist` method.
---
spec - https://webdriver.io/docs/api/element/waitForExist/

### DIFF
--- a/packages/services/f-wdio-utils/src/page.object.js
+++ b/packages/services/f-wdio-utils/src/page.object.js
@@ -1,6 +1,8 @@
 const { buildUrl } = require('./storybook-extensions');
 
 class Page {
+    #defaultWaitTimeout = 500;
+
     constructor (componentType, componentName) {
         this.title = 'Component URLS';
         this.componentType = componentType;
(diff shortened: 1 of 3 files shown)


packages/services/f-wdio-utils/src/page.object.js
```diff
@@ -19,8 +21,8 @@ class Page {
         return this;
     }
 
-    waitForComponent (component) {
-        component.waitForExist();
+    waitForComponent (component, timeoutMs = this.#defaultWaitTimeout) {
```
DamianMullins, line 24: Is there any benefit of setting this on a private field rather than inline?

DamianMullins (approved):

packages/services/f-wdio-utils/src/page.object.js
oliversweb: x2 reasons really, one so I can give the value a name and secondly so I can get rid of the annoying eslint warning :)

LTurns (approved):

PR #1535, opened 2021-12-07
f-account-info@0.10.0 - Add defence to form submission.

kevinrodrigues:
## UI Review Checks

## Browsers Tested

- [ ] Chrome (latest)
- [ ] Internet Explorer 11


packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -300,6 +316,18 @@ export default {
...
+
+        /**
+         * Send through the field & value that has been edited.
+         *
+         * @TODO store in vuex store once that is wired up.
+         *
+         * @param field
+         * @param value
+         */
+        editConsumerDetails (field, value) {
```
zlatinivanov-je, line 321: Will it make sense to listen to the `form` `change` event rather than each field?

zlatinivanov-je (commented):
One comment to consider. Let me know what you think

packages/components/pages/f-account-info/src/components/AccountInfo.vue
kevinrodrigues: Valid point. Will give this a try 👍

kevinrodrigues: Have fixed this now.

packages/components/pages/f-account-info/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.8.0
+------------------------------
+*December 07, 2021*
+
+### Added
+- `isFormDirty` checks to avoid submitting the form when the data hasn't changed.
```
tommcclean: Totally uneducated question on this one; but I noticed when working on validation, there is a this.$v.isDirty or something. Could we use that for the same purpose rather than introducing new properties?

kevinrodrigues: Will take a look to see if that will work here 👍

ashleynolan: Yeh, if this is using vuelidate, it will have a dirty state set on it ([the docs on that are here](https://vuelidate-next.netlify.app/guide.html#the-dirty-state))

simonsandersje (dismissed):

packages/components/pages/f-account-info/CHANGELOG.md
kevinrodrigues: @ashleynolan @tommcclean Have looked at this and as far as I can see the `$v.dirty` state is dependant on user interaction, i.e  used to keep fields from showing as invalid, before the user has interacted with them. What we want to do is track the value change on the form rather than just the interaction i.e clicking in and out of the input field (this toggles the $v.dirty value from false > true from what i can see). I think the name `isFormDirty` might be better named to something like `hasFormUpdate` or something to avoid confusion with the vuelidate name. Let me know if I have missed something on the `$v.dirty` docs tho, i could be wrong.

packages/components/pages/f-account-info/src/components/AccountInfo.vue
kevinrodrigues: Have decided to revert this one as i think it's a little clearer when the method is called on each field, with it at the form level it's a little more obscure as to the field name and target value we're passing. For readability will keep as is if that is ok.

oliversweb (dismissed):

DamianMullins (dismissed):

simonsandersje (dismissed):

oliversweb (dismissed):

DamianMullins (dismissed):

oliversweb (approved):

simonsandersje (approved):

PR #1537, opened 2021-12-08
f-account-info@v0.9.0 - Wired up the store to the template

oliversweb:
### Changed
- Wired up the new Vuex Store that exposes the State `consumer` to the Template plus wired up the load action
---
Wiring up the edit and save actions to follow on separate prs

### DIFF
--- a/packages/components/pages/f-account-info/package.json
+++ b/packages/components/pages/f-account-info/package.json
@@ -43,7 +43,8 @@
     "@justeat/f-globalisation": "1.0.0"
   },
   "peerDependencies": {
-    "@justeat/browserslist-config-fozzie": ">=1.2.0"
+    "@justeat/browserslist-config-fozzie": ">=1.2.0",
+    "vuex": "3.5.1"
   },
   "devDependencies": {
     "@justeat/f-wdio-utils": "0.4.0",
--- a/packages/components/pages/f-account-info/package.json
+++ b/packages/components/pages/f-account-info/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-account-info",
   "description": "Fozzie Account Info - The account information page",
-  "version": "0.8.0",
+  "version": "0.9.0",
   "main": "dist/f-account-info.umd.min.js",
   "maxBundleSize": "100kB",
   "files": [
--- a/packages/components/pages/f-account-info/package.json
+++ b/packages/components/pages/f-account-info/package.json
@@ -3,7 +3,7 @@
   "description": "Fozzie Account Info - The account information page",
   "version": "0.9.0",
   "main": "dist/f-account-info.umd.min.js",
-  "maxBundleSize": "100kB",
+  "maxBundleSize": "200kB",
   "files": [
     "dist",
     "test-utils"
--- a/packages/components/pages/f-account-info/package.json
+++ b/packages/components/pages/f-account-info/package.json
@@ -47,7 +47,7 @@
     "vuex": "3.5.1"
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


packages/components/pages/f-account-info/package.json
```diff
@@ -43,10 +43,11 @@
     "@justeat/f-globalisation": "1.0.0"
   },
   "peerDependencies": {
-    "@justeat/browserslist-config-fozzie": ">=1.2.0"
+    "@justeat/browserslist-config-fozzie": ">=1.2.0",
+    "vuex": "3.5.1"
```
DamianMullins: Peer deps should be set as a range, typically setting a minimum known version and allowing any version higher to allow for more flexibility in consuming applications.

```suggestion
    "vuex": ">=3.5.1"
```

packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -230,38 +221,44 @@ export default {
...
             }
         }
     },
 
-    mounted () {
+    created () {
+        if (!this.$store.hasModule('fAccountInfoModule')) {
+            this.$store.registerModule('fAccountInfoModule', fAccountInfoModule);
+        }
+    },
```
SachaZvetelman, line 245: Don't we normally do this in the consuming application? At least we are doing that for Checkout.

packages/components/pages/f-account-info/src/components/_tests/AccountInfo.test.js
```diff
@@ -18,48 +36,106 @@ const createStore = ({ state = {}, actions = {} } = {}) => new Vuex.Store({
...
+    storeOverride = null,
+    initialiseOverride = null
+} = {}) => {
+    const store = storeOverride || createStore({ state, actions });
+    store.registerModule = registerStoreModuleSpy;
+    store.hasModule = hasModuleSpy;
+
+    if (initialiseOverride) {
+        AccountInfo.methods.initialise = initialiseOverride;
+    }
```
SachaZvetelman: For this kind of pattern in JavaScript, I personally prefer this one-liner:
`AccountInfo.methods.initialise = initialiseOverride || AccountInfo.methods.initialise;`

packages/components/pages/f-account-info/src/components/AccountInfo.vue
oliversweb: Yes, this is now the preferred way ... (_before it was thought before that we couldn't do this and used to do this in the parent (coreweb)_)

SachaZvetelman: Sure, as long as we always have a store, then I definitely prefer this approach 👍 
Just for my understanding, how are we ensuring we have a store when this component it used?

packages/components/pages/f-account-info/package.json
```diff
@@ -43,10 +43,11 @@
     "@justeat/f-globalisation": "1.0.0"
   },
   "peerDependencies": {
-    "@justeat/browserslist-config-fozzie": ">=1.2.0"
+    "@justeat/browserslist-config-fozzie": ">=1.2.0",
+    "vuex": ">=3.5.1"
```
DamianMullins, line 47: I think you [might also need this change](https://github.com/justeat/fozzie-components/pull/1540/files#diff-8d69a0c8f25e9aa68bf113b42dee7ed1f1dacb2c71b1a292101d02863e631608) to further prevent the packages from being bundled up when built?

SachaZvetelman (approved):

DamianMullins (approved):

PR #1539, opened 2021-12-09
f-header@8.0.0 - Fix invisible icons in mobile nav for highlight theme, and add `tallBelowMid` prop

xander-marjoram:
### Changed
- Fix some display issues for highlight theme
  - Make icons visible in mobile nav.
  - Reduce header height for mobile viewports.

### Added
- Supporting visual regression tests.

---

## UI Review Checks

- [x] Tests have been [created|updated]

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/organisms/f-header/CHANGELOG.md
+++ b/packages/components/organisms/f-header/CHANGELOG.md
@@ -4,6 +4,19 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v7.4.0
+------------------------------
+*December 9, 2021*
+
+### Changed
+- Fix some display issues for highlight theme
+  - Make icons visible in mobile nav.
+  - Reduce header height for mobile viewports.
+
+### Added
+- Supporting visual regression tests.
+
+
 v7.3.0
 ------------------------------
 *December 6, 2021*
--- a/packages/components/organisms/f-header/CHANGELOG.md
+++ b/packages/components/organisms/f-header/CHANGELOG.md
@@ -9,9 +9,7 @@ v7.4.0
 *December 9, 2021*
 
 ### Changed
-- Fix some display issues for highlight theme
-  - Make icons visible in mobile nav.
-  - Reduce header height for mobile viewports.
+- Make icons visible in mobile nav for highlight theme.
 
 ### Added
 - Supporting visual regression tests.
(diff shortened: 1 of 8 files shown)


packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -13,7 +13,7 @@
                 'is-hidden--noJS',
                 $style['c-nav-trigger'],
                 $style['c-nav-toggle'],
-                { [$style['c-nav-toggle--altColour']]: isAltColour },
+                { [$style['c-nav-toggle--altColour']]: isAltColour || (headerBackgroundTheme === 'highlight' && navIsOpen) },
```
xander-marjoram, line 16: I thought it was better to add the additional logic here rather than extending `isAltColour` to support the unique case of:
1. Icon is the nav toggle
2. The theme is `highlight`
3. The mobile nav is open

packages/components/organisms/f-header/CHANGELOG.md
```diff
@@ -4,6 +4,19 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
 
 
+v7.4.0
+------------------------------
+*December 9, 2021*
+
+### Changed
+- Fix some display issues for highlight theme
+  - Make icons visible in mobile nav.
+  - Reduce header height for mobile viewports.
```
ashleynolan: I think I know why you've adjusted the height (as it looked broken and too big) but I think this is used (maybe on search?) so that an extra input text field is covered inside the header potentially (if you look on SERP live it shows this on mobile)

It may be worth providing a prop that extends the height for this use case, as atm I guess it's just being made the default as the highlight header is only ever used on SERP.

xander-marjoram: Added a new prop to make the header "tall below mid", and added a visual test to cover that.

simonsandersje (approved):

packages/components/organisms/f-header/src/components/Header.vue
```diff
@@ -248,7 +254,16 @@ html:global(.is-navInView) {
 
     .c-header--highlightBg {
         background-color: $color-support-brand-01;
-        min-height: 88px;
+
+        @include media('>mid') {
+            min-height: 88px;
```
ashleynolan, line 259: Does this rule still need to be there now with the new prop? (should the header just be the same height, even when highlighted?)

ashleynolan (commented):
Looks good – just one small comment

packages/components/organisms/f-header/src/components/Header.vue
xander-marjoram: Yep, the highlight theme header is currently taller than the default themed header.

ashleynolan (approved):

PR #1546, opened 2021-12-10
f-header@9.0.0 - add atom deps to peer deps

dandel10n:
### Added
- **Breaking Change**: Added `f-button` and `f-popover` dependencies to peer dependencies. Now `f-button` and `f-popover` should be included as a dependency of the consuming component or application.

### Removed
- **Breaking Change**: Removed `f-button` and `f-popover` styles import from the component. Make sure to import `f-button` and `f-popover` styles in your application.


DamianMullins (dismissed):

ashleynolan:
Seems like the visual regression tests have broken now – do you need to add the styles into those tests?

ashleynolan (approved):

kevinrodrigues (approved):

PR #1548, opened 2021-12-10
f-account-info@v0.11.0 - Wired up the store action `editConsumerDetails` to the Template

oliversweb:
### Changed
- Wired up the Vuex Store `editConsumerDetails` action to the Template

### DIFF
--- a/packages/components/pages/f-account-info/src/components/AccountInfo.vue
+++ b/packages/components/pages/f-account-info/src/components/AccountInfo.vue
@@ -15,7 +15,7 @@
             method="post"
             @submit.prevent="onFormSubmit">
             <form-field
-                v-model="consumer.firstName"
+                :value="consumer.firstName"
                 data-test-id="account-info-consumer-firstName"
                 maxlength="50"
                 :label-text="$t('consumer.firstNameLabel')"
@@ -35,7 +35,7 @@
             </form-field>
 
             <form-field
-                v-model="consumer.lastName"
+                :value="consumer.lastName"
                 data-test-id="account-info-consumer-lastName"
                 maxlength="50"
                 :label-text="$t('consumer.lastNameLabel')"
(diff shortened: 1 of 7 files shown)


simonsandersje (dismissed):

packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -310,15 +320,12 @@ export default {
...
-         * @TODO store in vuex store once that is wired up.
-         *
-         * @param field
-         * @param value
-         */
-        // eslint-disable-next-line no-unused-vars
-        editConsumerDetails (field, value) {
+        * A generic method that updates the State (e.g. 'consumer.<field> = value')
+        * @param {string} field - The field of the consumer that needs changing
+        * @param {string} value - The new value the consumer field
```
kevinrodrigues: `The new value of the consumer field`?

packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -310,15 +320,12 @@ export default {
...
+        */
+        editConsumer (field, value) {
```
kevinrodrigues: Maybe we can call this `onEditConsumerDetails` or `onEditConsumer`?

packages/components/pages/f-account-info/src/components/_tests/AccountInfo.test.js
```diff
@@ -154,20 +154,39 @@ describe('AccountInfo', () => {
         });
     });
 
-    describe('`methods`', () => {
```
kevinrodrigues, line 157: Personally i think this describe makes it easier to keep things (method tests in this case) together, otherwise it gets hard to find places to add new tests.

kevinrodrigues (dismissed):

kevinrodrigues (approved):

simonsandersje (approved):

PR #1551, opened 2021-12-13
f-header@9.1.0 - jet logo

dandel10n:
### Added
- Jet logo support via `useJetLogo` prop.

### Updated
- f-vue-icons to the latest (3.3.0).

### Removed
- Arrow button size override from mobile country selector as with the latest version of f-vue-icons arrow doesn't have a transparent spacing around it.

### DIFF
--- a/packages/components/organisms/f-header/README.md
+++ b/packages/components/organisms/f-header/README.md
@@ -92,7 +92,7 @@ The props that can be defined are as follows:
 | userInfoUrl               | `String`      | `/api/account/details` | URL to call to retrieve the userInfo (when `userInfoProp` isn't set). |
 | showCountrySelector       | `Boolean`     | `false` | Defines whether the country selector should be shown in the navigation. |
 | showSkipLink              | `Boolean`     | `true`  | Set to false if you need to remove skip-to-main-content link from the header. |
-| useJetLogo                | `Boolean`     | `false`  | Set to true if you want to show Jet logo in the header. |
+| shouldUseJetLogo                | `Boolean`     | `false`  | Set to true if you want to show Jet logo in the header. |
 **Important:** if you're adding a new property to show/hide something on the navigation bar, you probably want to check the `hasNavigationLinks` computed property, since you might have to update it.
 
 ### CSS styles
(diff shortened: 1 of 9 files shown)


packages/components/organisms/f-header/src/components/Header.vue
```diff
@@ -137,6 +137,11 @@ export default {
         tallBelowMid: {
             type: Boolean,
             default: false
+        },
+
+        useJetLogo: {
```
kevinrodrigues: More of a suggestion but would `shouldUseJetLogo` work better as a name?

packages/components/organisms/f-header/src/components/_tests/Logo.test.js
```diff
@@ -47,6 +47,23 @@ describe('Logo', () => {
         expect(logo).toBeDefined();
     });
 
+    it('should render Jet logo if jet theme passed', () => {
+        // Arrange
+        const propsData = {
+            theme: 'jet',
```
kevinrodrigues, line 53: Instead of setting the theme i think this test should be passing the prop value `useJetLogo` as `true` and then the theme (computed) would automatically allow this test to pass.

dandel10n: Because this is the test for logo component it doesn't know about `shouldUseJetLogo` prop as it a header component's prop. I added another test to the header.test.js to cover that prop.

packages/components/organisms/f-header/src/components/Header.vue
dandel10n: Thanks, changed

packages/components/organisms/f-header/README.md
```diff
@@ -85,7 +85,7 @@ The props that can be defined are as follows:
 | userInfoUrl               | `String`      | `/api/account/details` | URL to call to retrieve the userInfo (when `userInfoProp` isn't set). |
 | showCountrySelector       | `Boolean`     | `false` | Defines whether the country selector should be shown in the navigation. |
 | showSkipLink              | `Boolean`     | `true`  | Set to false if you need to remove skip-to-main-content link from the header. |
-
+| shouldUseJetLogo                | `Boolean`     | `false`  | Set to true if you want to show Jet logo in the header. |
```
matthewhardern, line 95: For backwards compatibility would it be better to be default to true?

dandel10n: The Jet logo will be used mostly for Ops team's projects. For customer facing sites the logos are still the same (JE amd Menulog) and their choice is covered by our globalisation service.

kevinrodrigues (dismissed):

DamianMullins (approved):

xander-marjoram (approved):

packages/components/organisms/f-header/README.md
matthewhardern: cool

PR #1554, opened 2021-12-14
f-form@0.2.0 - Add basic components

FayeCarter:
### Added
- Add basic form components.

image

### DIFF
--- a/packages/components/organisms/f-form/README.md
+++ b/packages/components/organisms/f-form/README.md
@@ -32,6 +32,13 @@ npm install @justeat/f-form
 ```
 
 
+The package also has dependencies that need to be installed by consuming components/applications:
+
+| Dependency     | Command to install                   | Styles to include                                       |
+| -----          | -----                                | -----                                                   |
+| f-button       | `yarn add @justeat/f-button`         | `import '@justeat/f-button/dist/f-button.css';`         |
+| f-form-field   | `yarn add @justeat/f-form-field`     | `import '@justeat/f-form-field/dist/f-form-field.css';` |
+
 
 ### Vue Applications
 
--- a/packages/components/organisms/f-form/README.md
+++ b/packages/components/organisms/f-form/README.md
@@ -4,7 +4,7 @@
 
 <img width="125" alt="Fozzie Bear" src="../../../../bear.png" />
 
-Generic form component for basic forms. Can be used to collect data from text inputs.
+Generic form component for creating basic forms. Can be used to collect data from text inputs.
 
 </div>
 
@@ -38,12 +38,12 @@ npm install @justeat/f-form
 You can import it in your Vue SFC like this (please note that styles have to be imported separately):
 
 ```js
-import FForm from '@justeat/f-form';
+import VForm from '@justeat/f-form';
 import '@justeat/f-form/dist/f-form.css';
 
 export default {
     components: {
-        FForm
+        VForm
     }
 }
 ```
@@ -65,13 +65,13 @@ export default {
 
 ### Props
 
-To add fields to the form and text to the form button a `formData` Object prop should be passed in. The `formData` Object should contain.
+To add fields to the form, and text to the form button, a `formData` Object prop should be passed in. The `formData` Object should contain:
  * **formFields** - ` Array of objects. Object should include
     * **name** - The field name that data should be stored under
-    * **value** - The value of the field
+    * **value** - The default value of the field
     * **translations** - an object containing:
-        * **label** - the field label
- *  **buttonText** - The text displayed on the button.
+        * **label** - the field's displayed text label
+ *  **buttonText** - The text displayed on the form's action button
 
 > **Example `formData`**
 ```
--- a/packages/components/organisms/f-form/README.md
+++ b/packages/components/organisms/f-form/README.md
@@ -75,7 +75,7 @@ export default {
 To add fields to the form, and text to the form button, a `formData` Object prop should be passed in. The `formData` Object should contain:
  * **formFields** - ` Array of objects. Object should include
     * **name** - The field name that data should be stored under
-    * **value** - The default value of the field
+    * **value** - The value of the field. When not set, this defaults to ''.
     * **translations** - an object containing:
         * **label** - the field's displayed text label
  *  **buttonText** - The text displayed on the form's action button
--- a/packages/components/organisms/f-form/README.md
+++ b/packages/components/organisms/f-form/README.md
@@ -69,7 +69,7 @@ To add fields to the form and text to the form button a `formData` Object prop s
  * **formFields** - ` Array of objects. Object should include
     * **name** - The field name that data should be stored under
     * **value** - The value of the field
-    * **translations** - an object contating
+    * **translations** - an object containing:
         * **label** - the field label
  *  **buttonText** - The text displayed on the button.
 
--- a/packages/components/organisms/f-form/README.md
+++ b/packages/components/organisms/f-form/README.md
@@ -4,7 +4,7 @@
 
 <img width="125" alt="Fozzie Bear" src="../../../../bear.png" />
 
-fozzie form
+Generic form component for basic forms. Can be used to collect data from text inputs.
 
 </div>
 
@@ -38,12 +38,12 @@ npm install @justeat/f-form
 You can import it in your Vue SFC like this (please note that styles have to be imported separately):
 
 ```js
-import VForm from '@justeat/f-form';
+import FForm from '@justeat/f-form';
 import '@justeat/f-form/dist/f-form.css';
 
 export default {
     components: {
-        VForm
+        FForm
     }
 }
 ```
@@ -65,22 +65,26 @@ export default {
 
 ### Props
 
-There may be props that allow you to customise its functionality.
-
-The props that can be defined are as follows (if any):
+To add fields to the form and text to the form button a `formData` Object prop should be passed in. The `formData` Object should contain.
+ * **formFields** - ` Array of objects. Object should include
+    * **name** - The field name that data should be stored under
+    * **value** - The value of the field
+    * **translations** - an object contating
+        * **label** - the field label
+ *  **buttonText** - The text displayed on the button.
 
 > **Example `formData`**
 ```
 formData: {
-    formFields: {
-        firstName: {
+    formFields: [
+        {
             name: 'firstName',
             value: '',
             translations: {
                 label: 'First Name'
             }
         },
-        lastName: {
+        {
             name: 'lastName',
             value: '',
             translations: {
@@ -88,15 +92,15 @@ formData: {
             }
         }
         ...
-    },
+    ],
     buttonText: 'Continue'
 }
 ```
 
 | Prop  | Type  | Default | Description |
 | ----- | ----- | ------- | ----------- |
-| `formData` | Object | - | Object containing button text and form field data. Should be structured as above. |
-| `isFormSubmitting` | Boolean | - | Allows the parent component to set a loading spinner on the submit button while any asynchronous calls are carried out by the parent |
+| `formData` | Object | **Required Prop** | Object containing button text and form field data. Should be structured as above. |
+| `isFormSubmitting` | Boolean | false | Allows the parent component to set a loading spinner on the submit button while any asynchronous calls are carried out by the parent |
 
 ### Events
 
--- a/packages/components/organisms/f-form/README.md
+++ b/packages/components/organisms/f-form/README.md
@@ -69,8 +69,34 @@ There may be props that allow you to customise its functionality.
 
 The props that can be defined are as follows (if any):
 
+> **Example `formData`**
+```
+formData: {
+    formFields: {
+        firstName: {
+            name: 'firstName',
+            value: '',
+            translations: {
+                label: 'First Name'
+            }
+        },
+        lastName: {
+            name: 'lastName',
+            value: '',
+            translations: {
+                label: 'Last Name'
+            }
+        }
+        ...
+    },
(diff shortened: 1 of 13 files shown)


packages/components/organisms/f-form/src/components/Form.vue
```diff
@@ -1,46 +1,103 @@
 <template>
-    <div
+    <form
```
ashleynolan, line 2: Let me know when you're free as would be good to go over this component – we played around with the idea of having a generic form component when we built registration, but for one reason or another decided against it. WOuld be good to understand the reasons for building one now, as guessing it's for use in checkout?

ashleynolan (changes requested):

packages/components/organisms/f-form/src/components/_tests/Form.test.js
```diff
@@ -1,10 +1,92 @@
 import { shallowMount } from '@vue/test-utils';
-import VForm from '../Form.vue';
+import FForm from '../Form.vue';
+
+const formData = {
+    formFields: {
```
xander-marjoram: Did you try using an array instead of an object? I think it would make a bit more sense to me; the ordering is a bit more explicit and you wouldn't need to name each object on top of its `name` property.

FayeCarter: Oh I hadn't thought of that, I'll give it a go, thanks!

packages/components/organisms/f-form/README.md
```diff
@@ -65,19 +65,51 @@ export default {
 
 ### Props
 
-There may be props that allow you to customise its functionality.
+To add fields to the form and text to the form button a `formData` Object prop should be passed in. The `formData` Object should contain.
+ * **formFields** - ` Array of objects. Object should include
+    * **name** - The field name that data should be stored under
+    * **value** - The value of the field
+    * **translations** - an object contating
```
simonsandersje: Nit: Text needs a small typo correction.

FayeCarter: damn 🤦 
Thanks I'll change it now and install a better spell checker for vs code 😆

packages/components/organisms/f-form/README.md
```diff
@@ -4,7 +4,7 @@
 
 <img width="125" alt="Fozzie Bear" src="../../../../bear.png" />
 
-fozzie form
+Generic form component for basic forms. Can be used to collect data from text inputs.
```
ashleynolan:
```suggestion
Generic form component for creating basic forms. Can be used to collect data from text inputs.
```

packages/components/organisms/f-form/src/components/_tests/Form.test.js
```diff
@@ -1,10 +1,89 @@
...
+            it('should emit `updated` event with fieldName and field value', () => {
+                // Arrange
+                const wrapper = shallowMount(FForm, { propsData });
+                const payload = { fieldName: 'firstName', value: 'Joe' };
+
+                // Act
+                wrapper.vm.updateField(payload);
+
+                // Assert
+                expect(wrapper.emitted('updated').length).toBe(1);
```
kevinrodrigues: I think you can squash this into one expect which covers both:
``` javascript
// Arrange
const spy = jest.spyOn(wrapper.vm, '$emit');

// Assert
expect(spy).toHaveBeenCalledWith('updated', payload);
```

FayeCarter: Thanks!

kevinrodrigues (dismissed):

packages/components/organisms/f-form/README.md
```diff
@@ -38,12 +38,12 @@ npm install @justeat/f-form
 You can import it in your Vue SFC like this (please note that styles have to be imported separately):
 
 ```js
-import VForm from '@justeat/f-form';
+import FForm from '@justeat/f-form';
```
ashleynolan: I would probably keep this as VForm – we don't have a hard rule over whether to use F{component_name} or V{component_name}, and FForm does look a bit like a typo (to me anyway 😄 )

packages/components/organisms/f-form/README.md
```diff
@@ -38,12 +38,12 @@ npm install @justeat/f-form
...
 import '@justeat/f-form/dist/f-form.css';
 
 export default {
     components: {
-        VForm
+        FForm
```
ashleynolan: Same as above

packages/components/organisms/f-form/README.md
```diff
@@ -65,19 +65,51 @@ export default {
 
 ### Props
 
-There may be props that allow you to customise its functionality.
+To add fields to the form and text to the form button a `formData` Object prop should be passed in. The `formData` Object should contain.
```
ashleynolan:
```suggestion
To add fields to the form, and text to the form button, a `formData` Object prop should be passed in. The `formData` Object should contain:
```

packages/components/organisms/f-form/README.md
```diff
@@ -65,19 +65,51 @@ export default {
...
+ * **formFields** - ` Array of objects. Object should include
+    * **name** - The field name that data should be stored under
+    * **value** - The value of the field
+    * **translations** - an object containing:
+        * **label** - the field label
```
ashleynolan:
```suggestion
        * **label** - the field's displayed text label
```

packages/components/organisms/f-form/README.md
```diff
@@ -65,19 +65,51 @@ export default {
...
+ *  **buttonText** - The text displayed on the button.
```
ashleynolan:
```suggestion
 *  **buttonText** - The text displayed on the form's action button
```

packages/components/organisms/f-form/README.md
```diff
@@ -65,19 +65,51 @@ export default {
 
 ### Props
 
-There may be props that allow you to customise its functionality.
+To add fields to the form and text to the form button a `formData` Object prop should be passed in. The `formData` Object should contain.
+ * **formFields** - ` Array of objects. Object should include
+    * **name** - The field name that data should be stored under
+    * **value** - The value of the field
```
ashleynolan:
```suggestion
    * **value** - The default value of the field
```

Also, is it a requirement that this is set if there is no default value?

packages/components/organisms/f-form/package.json
```diff
@@ -40,7 +40,9 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-services": "1.7.0"
+    "@justeat/f-services": "1.7.0",
+    "@justeat/f-button": "3.0.2",
+    "@justeat/f-form-field": "4.0.1"
```
ashleynolan: We've started to make any subcomponents like `f-button` and `f-form-field` peerDependencies which then get pulled in at the page component level (so in `f-checkout` then for instance). Essentially f-button and f-form-field would be included as a `peerDependency` (so they don't get built as part of the f-form component) and also included as a `devDependency` (so that the tests can still run with no problems).

You can see how to do this by looking at a PR [like this one](https://github.com/justeat/fozzie-components/pull/1546/files) (as you'll need to add a tiny bit of config as well so that the peerDependency is kept external when the component is built). Give me a msg if you have any questions around this though and can pair on it.

packages/components/organisms/f-form/src/components/Form.vue
```diff
@@ -1,46 +1,104 @@
...
+            :is-loading="isFormSubmitting">
+            {{ formData.buttonText }}
+        </f-button>
+    </form>
 </template>
 
+
 <script>
+import FButton from '@justeat/f-button';
+import '@justeat/f-button/dist/f-button.css';
```
ashleynolan: Just a note that when you move these to be `peerDependencies`, you won't need to include the CSS here, as it'll get pulled in as part of the page level component (i.e. Checkout) instead. You'll need the styles included in any storybook files you create for testing purposes though.

packages/components/organisms/f-form/src/components/Form.vue
```diff
@@ -1,46 +1,104 @@
...
-.c-form {
-    display: flex;
-    justify-content: center;
-    min-height: 80vh;
-    width: 80vw;
-    margin: auto;
-    border: 1px solid red;
-    font-family: $font-family-base;
-    @include font-size(heading-m);
+.c-form-form {
```
ashleynolan: Does this class exist? I can only see `c-form` in the markup?

ashleynolan (changes requested):
Few comments, but looks good overall 👍

packages/components/organisms/f-form/src/components/Form.vue
FayeCarter: 🤦

packages/components/organisms/f-form/src/components/Form.vue
FayeCarter: Awesome, Thanks

packages/components/organisms/f-form/package.json
```diff
@@ -43,10 +43,14 @@
     "@justeat/f-services": "1.7.0"
   },
   "peerDependencies": {
-    "@justeat/browserslist-config-fozzie": ">=1.2.0"
+    "@justeat/browserslist-config-fozzie": ">=1.2.0",
+    "@justeat/f-button": "3.x.x",
```
xander-marjoram: What's the need to have button and form-field as both dev and peer dependencies?

FayeCarter: This is something Ash recommended here https://github.com/justeat/fozzie-components/pull/1554#discussion_r779624384.
It's in peer so they don't get built as part of the f-form component
Then it's in dev so that the tests can still run with no problems

xander-marjoram (dismissed):

packages/components/organisms/f-form/package.json
```diff
@@ -43,10 +43,14 @@
...
+    "@justeat/f-form-field": "4.x.x"
```
ashleynolan:
```suggestion
    "@justeat/f-button": "3.x",
    "@justeat/f-form-field": "4.x"
```

packages/components/organisms/f-form/src/components/Form.vue
```diff
@@ -1,46 +1,102 @@
...
+        },
+
+        onFormSubmit () {
+            this.$emit('form-submitting');
+        },
+
+        fieldProps (field) {
+            return {
+                name: field.name,
+                value: field.value,
```
ashleynolan: You could make this default to empty and then if the `field.value` is set then use that value

```suggestion
                value: field.value || '',
```

That would save having to pass in an empty string for field objects with no default value.

packages/components/organisms/f-form/src/components/_tests/Form.test.js
```diff
@@ -1,10 +1,89 @@
 import { shallowMount } from '@vue/test-utils';
-import VForm from '../Form.vue';
+import FForm from '../Form.vue';
+
+const formData = {
```
ashleynolan, line 5: It would be good to have some tests around the expected result of a consumer not setting the formData as expected. For example, what happens if they don't pass in a default `value`, `name` or a `label`? (Same for `buttonText`?)

packages/components/organisms/f-form/src/components/Form.vue
```diff
@@ -1,46 +1,102 @@
...
+
+        <f-button
+            button-type="primary"
+            button-size="large"
+            :class="$style['c-form-submitButton']"
+            is-full-width
+            action-type="submit"
+            data-test-id="submit-button"
+            :is-loading="isFormSubmitting">
+            {{ formData.buttonText }}
```
ashleynolan: I'd probably have a sensible default `buttonText` in case someone doesn't pass one in for some reason (for example, `Submit`).

packages/components/organisms/f-form/src/components/Form.vue
```diff
@@ -1,46 +1,102 @@
...
+        onFormSubmit () {
+            this.$emit('form-submitting');
+        },
+
+        fieldProps (field) {
+            return {
+                name: field.name,
+                value: field.value,
+                'label-text': field.translations?.label,
+                'input-type': field.inputType,
```
ashleynolan: Can't remember from when you ran through it, but does this get passed in through the `formData` object?

FayeCarter: It will do in a later PR 😬. I've removed it for now

packages/components/organisms/f-form/src/components/_tests/Form.test.js
FayeCarter: I've added a `formDataValidator` method now https://github.com/justeat/fozzie-components/pull/1554/files#diff-6129c1b3d5cfb74495e312f10baf699cd88ab3fa678749e2edde92146fab55eaR103 (and tests)

packages/components/organisms/f-form/README.md
```diff
@@ -65,19 +72,51 @@ export default {
 
 ### Props
 
-There may be props that allow you to customise its functionality.
+To add fields to the form, and text to the form button, a `formData` Object prop should be passed in. The `formData` Object should contain:
+ * **formFields** - ` Array of objects. Object should include
+    * **name** - The field name that data should be stored under
+    * **value** - The default value of the field
```
ashleynolan:
```suggestion
    * **value** - The value of the field. When not set, this defaults to ''.
```

ashleynolan (commented):
One small comment, otherwise LGTM

ashleynolan (approved):

charliearlie (approved):

PR #1556, opened 2021-12-14
f-form@v0.3.0 - add field validations

FayeCarter:
### Added
- Add field validations.
image


matthewhardern:
please rebase from master so we can add this into the approval process thanks

packages/components/organisms/f-form/README.md
```diff
@@ -96,7 +110,8 @@ formData: {
 | Prop  | Type  | Default | Description |
 | ----- | ----- | ------- | ----------- |
 | `formData` | Object | - | Object containing button text and form field data. Should be structured as above. |
-| `isFormSubmitting` | Boolean | - | Allows the parent component to set a loading spinner on the submit button while any asynchronous calls are carried out by the parent |
+| `locale` | String | `en-gb` | Sets the translation file to use. |
```
xander-marjoram: Nit
```suggestion
| `locale` | String | `en-GB` | Sets the translation file to use. |
```

packages/components/organisms/f-form/src/components/_tests/Form.test.js
```diff
@@ -43,21 +18,66 @@ describe('Form', () => {
...
+
+    it('should have one form with method "post"', () => {
+        // Arrange
+        const wrapper = shallowMount(FForm, { propsData });
+
+        // Act
+        const forms = wrapper.findAll('form');
+
+        // Assert
+        expect(forms.length).toBe(1);
```
xander-marjoram: It it likely that there will be two or more forms? Not sure it's worth checking for that

packages/components/organisms/f-form/src/components/Form.vue
```diff
@@ -67,33 +105,112 @@ export default {
...
+
+        invalidFieldsSummary () {
+            const invalidFieldCount = this.$v.formFields.$dirty
+                && validations.getFormValidationState(this.$v.formFields).invalidFields.length;
+
+            if (!invalidFieldCount) return null;
+
+            return invalidFieldCount === 1
+                ? this.copy.errorMessages.singleFieldError
+                : this.copy.errorMessages.multipleFieldErrors.start + invalidFieldCount + this.copy.errorMessages.multipleFieldErrors.end;
```
xander-marjoram: Instead of enforcing this kind of sentence structure could you try something like this, where you have a template and the number gets passed in?
https://kazupon.github.io/vue-i18n/guide/formatting.html#support-ruby-on-rails-i18n-format

packages/components/organisms/f-form/src/components/_tests/Form.test.js
```diff
@@ -76,17 +96,205 @@ describe('Form', () => {
...
+
+            beforeEach(() => {
+                touchSpy = jest.spyOn($v, '$touch');
+            });
+
+            afterEach(() => {
+                jest.clearAllMocks();
+            });
+
+            it('should call `.touch()` ', () => {
```
xander-marjoram, line 251: Do we need to specifically know that it has called `touch`, or is it better to check that the fields are now dirty?

packages/components/organisms/f-form/src/components/_tests/Form.test.js
```diff
@@ -76,17 +96,205 @@ describe('Form', () => {
...
+                // Assert
+                expect(wrapper.vm.isFormValid()).toBeTruthy();
+            });
+
+            it('should return `false` if `$v` is invalid', () => {
+                // Arrange
+                $v.$invalid = true;
+
+                // Act
+                const wrapper = shallowMount(FForm, { propsData, mocks: { $v } });
```
xander-marjoram: Could setting `$invalid` like this potentially carry state across tests? I wonder if something like this might be a bit more robust:

(Feel free to trim it down a bit visually if you think it takes up too many lines)

```suggestion
                // Arrange & Act
                const wrapper = shallowMount(FForm, {
                    propsData,
                    mocks: {
                        $v: {
                            ...$v,
                            $invalid: true
                        }
                    }
                });
```

packages/components/organisms/f-form/src/components/_tests/Form.test.js
```diff
@@ -76,17 +96,205 @@ describe('Form', () => {
...
+            });
+        });
+
+        describe('onFormSubmit ::', () => {
+            let isFormValidSpy;
+            let scrollToFirstInlineErrorSpy;
+            let wrapper;
+
+            beforeEach(() => {
+                isFormValidSpy = jest.spyOn(FForm.methods, 'isFormValid');
```
xander-marjoram: Instead of spying on the method, I think it would be a little cleaner to mock the underlying value of `$v.$invalid`.

Spies are perfect when you want to test that something is being called, but also maintain the original functionality.
As you're changing the functionality as part of the test (by mocking the return value), it's probably not an ideal usage.

packages/components/organisms/f-form/src/components/_tests/Form.test.js
```diff
@@ -76,17 +96,205 @@ describe('Form', () => {
...
+                // Arrange
+                const isValidPhoneNumberSpy = jest.spyOn(validations, 'isValidPhoneNumber');
+
+                const wrapper = shallowMount(FForm, { propsData });
+
+                // Act
+                wrapper.vm.isValidPhoneNumber();
+
+                // Assert
+                expect(isValidPhoneNumberSpy).toHaveBeenCalledWith(formData.formFields.mobileNumber.value, 'en-GB');
```
xander-marjoram: For example, this is a good use of a spy.
It's an external method so you can't just stub it with `jest.fn` (unless you use a module mock which can get a bit hairy), and you're also using an expectation like `toHaveBeenCalledWith`.

packages/components/organisms/f-form/src/components/_tests/Form.test.js
```diff
@@ -76,17 +96,205 @@ describe('Form', () => {
...
+                // Arrange & Act
+                const wrapper = shallowMount(FForm, { propsData });
+
+                // Assert
+                expect(wrapper.vm.hasValidationMessages('lastName')).toBeFalsy();
+            });
+        });
+
+        describe('hasInvalidErrorMessage ::', () => {
+            it('should return true if passed field has `invalidValidationMessages`', () => {
```
xander-marjoram: Is `invalidValidationMessages` a variable somewhere?

packages/components/organisms/f-form/src/components/_tests/FormField.test.js
```diff
@@ -0,0 +1,459 @@
...
+
+    describe('methods ::', () => {
+        afterEach(() => {
+            jest.clearAllMocks();
+        });
+
+        describe('updateField ::', () => {
+            it('should emit `updated` event with fieldName and field value', () => {
+                // Arrange
+                const wrapper = shallowMount(FormField, { propsData, provide: { $v } });
```
xander-marjoram: I think in all other places, provide is a function, does this matter? Also could be worth pulling the whole thing out into a variable to save a few lines, etc.

packages/components/organisms/f-form/src/components/_tests/FormField.test.js
```diff
@@ -0,0 +1,459 @@
...
+                    })
+                });
+
+                // Assert
+                expect(wrapper.vm.errorType).toEqual(expected);
+            });
+        });
+
+        describe('ariaLabel ::', () => {
+            it('should return the mobile number with spaces after every character', () => {
```
xander-marjoram: Do we need to handle any bad values?

packages/components/organisms/f-form/src/components/_tests/FormField.test.js
```diff
@@ -0,0 +1,459 @@
...
+                wrapper.vm.updateField(payload);
+
+                // Assert
+                expect(wrapper.emitted('updated').length).toBe(1);
+                expect(wrapper.emitted('updated')[0][0]).toEqual(payload);
+            });
+        });
+
+        describe('formFieldBlur ::', () => {
+            it('should call `touch` for validation', () => {
```
xander-marjoram, line 458: Again, it's probably worth testing the result (form field is now dirty) rather than the implementation (`touch` method has been called).

packages/components/organisms/f-form/src/tenants/en-GB.js
```diff
@@ -1,4 +1,10 @@
 export default {
     locale: 'en-GB',
-    text: 'I am a VForm Component (GB)'
+    errorMessages: {
+        singleFieldError: 'There is 1 error in the form',
```
xander-marjoram, line 4: Maybe worth looking into pluralisation too: https://kazupon.github.io/vue-i18n/guide/pluralization.html#pluralization

packages/components/organisms/f-form/src/components/_tests/Form.test.js
xander-marjoram: Having thought about this some more, I'm not sure how easy this will be whilst mocking the `$v` object without just testing your mock. I think what you've got now is fine

packages/components/organisms/f-form/src/components/_tests/FormField.test.js
xander-marjoram: See other comment on other thread

packages/components/organisms/f-form/README.md
```diff
@@ -73,17 +73,17 @@ export default {
...
- * **formFields** - ` Array of objects. Object should include
+ * **formFields** - Array of objects. Object should include
     * **name** - The field name that data should be stored under
-    * **value** - The value of the field
+    * **value** - TThe value of the field. When not set, this defaults to ''.
     * **translations** - an object containing:
-        * **label** - the field label
-        * **validationMessages** - an object that can containin two types of validation messages:
+        * **label** - the field's displayed text label
+        * **validationMessages** - an object that can contain two types of validation messages:
```
zlatinivanov-je, line 95: This is interesting, however it seems to be limiting to `required` and `invalid` without relation to the validation rule(s). We build something similar recently and had to build a field validity object, also observe validation order... 
It can easily get complicated. Happy to chat to you about that if that is the on the cards here.

FayeCarter: Yeah would be good to see what you've already built to see if there's anything that could be improved. 
At the moment it works out the validation message here https://github.com/justeat/fozzie-components/pull/1556/files#diff-fdc8192e0a3e68bb066e4093bc5964e0edec3662062fca586688c40fc5a9ca34R121 wheich is the same as is currently used in checkout

PR #1560, opened 2021-12-15
fozzie-components@v5.7.0 – Adding storybook:serve-visualregression #globalconfig

ashleynolan:
### Added
- Added storybook serve npm script to only spin up storybook for packages that have visual regression tests.


siggerzz (approved):

siggerzz:
After discussion, this approach won't work as component / a11y tests will fail.

PR #1566, opened 2021-12-17
f-image-tile@0.2.0  - Add toggle functionality

AlixRichards:
---

### Added
- ImageTile component structure and toggling behaviour
- ImageTile stories

This PR adds the base markup and toggling functionality to the component.
Minimal styling has been added  - this will happen in subsequent PRs

Screenshots below show the two stories added - the component and component in a container.

---

## UI Review Checks

- [X] README and/or UI Documentation has been [created]
- [X] Unit tests have been [created]

## Browsers Tested

- [X] Chrome (latest)

Screenshot 2021-12-17 at 10 27 49

Screenshot 2021-12-17 at 10 28 00


packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -12,20 +51,108 @@ export default {
...
+
+.c-imageTile-label {
+    display: flex;
+    flex-flow: column wrap;
+}
+
+.c-imageTile-label--selected {
+  border: 1px solid green;
+}
+
```
AlixRichards, line 149: ^ I'm using this as a visual guide whilst developing and before the components actual styles are added. This will be removed but is needed for now

packages/components/atoms/f-image-tile/stories/TestContainer.vue
```diff
@@ -0,0 +1,52 @@
...
+};
+</script>
+
+<style lang="scss" module>
+    .c-testContainer {
+        width: 200px;
+        margin: auto;
+        background-color: pink;
+    }
+</style>
```
AlixRichards, line 51: These are temporary styles that I'm using for development but will be removed when the components actual styling is added in later PRs

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -1,8 +1,43 @@
...
-        data-test-id="imageTile">
-        I am a ImageTile Component (GB)
+        data-test-id="image-tile-component">
+        <a
+            :class="[
+                $style['c-imageTile-link'], {
+                    [$style['c-imageTile-link--toggle']]: !isLink
+                }]"
+            :href="isLink ? href : '#'"
+            :aria-hidden="isLink ? false : true"
```
1itvinka:
```suggestion
            :aria-hidden="!isLink"
```

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -1,8 +1,43 @@
...
+            :tabindex="isLink ? false : -1"
```
1itvinka, line 12: according to documentation the tabindex attribute, if specified, must have a value that is a valid integer https://html.spec.whatwg.org/multipage/interaction.html#attr-tabindex. Should we use "0" instead of "false"?

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -1,8 +1,43 @@
...
+                    [$style['c-imageTile-label--selected']]: isFilterSelected
+                }]"
+            :for="`imageTileToggle-${tileId}`"
+            data-test-id="image-tile-label">
+            <img
+                :class="$style['c-imageTile-image']"
+                :src="imgSrc"
+                data-test-id="image-tile-image"
+                alt="">
+            <span :aria-hidden="isLink ? true : false">
```
1itvinka:
```suggestion
            <span :aria-hidden="isLink">
```
or
```suggestion
            <span :aria-hidden="!!isLink">
```

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -12,20 +47,108 @@ export default {
...
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+}
+
+.c-imageTile-link--toggle {
+  display: block;
+  position: static;
+  pointer-events: none;
```
1itvinka: sorry for nitpicking. It would be great to have 4 spaces at the beginning of each line here.

AlixRichards: Good spot 👍

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: Updated to use this 👍

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: Updated to `:aria-hidden="isLink"` 👍

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: When `isLink` is true I don't want the `tabindex` property to be output.
When `isLink` is false I want `tabindex` to be `-1`

This is working for me but agree if we wanted to specify a value for tabindex when `isLink` is true `0` would be a more appropriate value

simonsandersje: A value of -1 will remove the item from the normal tabbing sequence: https://dequeuniversity.com/rules/axe/3.0/tabindex
So this is the correct behaviour.
Also, an integer is a positive or negative whole number, including 0.

simonsandersje (dismissed):

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -12,20 +47,108 @@ export default {
     name: 'ImageTile',
     components: {},
     props: {
+        href: {
+            type: String,
+            default: 'https://www.google.com'
```
1itvinka:
```suggestion
            default: null
```

as far as I understand on searchweb we don't have links for image tiles. In this case it's better to have null as a default value.

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -12,20 +47,108 @@ export default {
...
+        },
+        tileId: {
+            type: String,
+            default: 'Chicken'
```
1itvinka:
```suggestion
            default: null
```

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -12,20 +47,108 @@ export default {
...
+        },
+        isSelected: {
+            type: Boolean,
+            default: false
+        },
+        isLink: {
+            type: Boolean,
+            default: false
+        },
```
1itvinka, line 65: I would remove this prop and check if there's `href` value to check if it's a link. Don't have a strong opinion on this though.

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -12,20 +47,108 @@ export default {
...
+        displayText: {
+            type: String,
+            default: 'Chicken'
```
1itvinka:
```suggestion
            default: null
```

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -12,20 +47,108 @@ export default {
...
+        },
+        imgSrc: {
+            type: String,
+            default: 'https://via.placeholder.com/150'
```
1itvinka:
```suggestion
            default: null
```

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -1,8 +1,43 @@
...
         :class="$style['c-imageTile']"
-        data-test-id="imageTile">
-        I am a ImageTile Component (GB)
+        data-test-id="image-tile-component">
+        <a
+            :class="[
+                $style['c-imageTile-link'], {
+                    [$style['c-imageTile-link--toggle']]: !isLink
+                }]"
+            :href="isLink ? href : '#'"
```
dandel10n, line 10: In case when there is no href provided should the component change link to a span or just not have an `a` tag wrapper at all?

packages/components/atoms/f-image-tile/src/components/_tests/ImageTile.test.js
```diff
@@ -7,4 +7,164 @@ describe('ImageTile', () => {
...
+
+                    const image = wrapper.find('[data-test-id="image-tile-image"]');
+
+                    // Assert
+                    expect(image.attributes('src')).toContain(propsData.imgSrc);
+                });
+            });
+
+            describe('isSelected :: ', () => {
+                it('should make `isFilterSelected` true when set to true', () => {
```
dandel10n: You can use [it.each](https://github.com/justeat/fozzie-components/blob/7c23f7f154ccd83bf9dd7e63e1ca1ed99798dea3/packages/components/pages/f-checkout/src/components/_tests/AgeVerification.test.js#L109-L121) for these type of checks. We have many examples in the repo, like [here](https://github.com/justeat/fozzie-components/blob/7c23f7f154ccd83bf9dd7e63e1ca1ed99798dea3/packages/components/pages/f-checkout/src/components/_tests/AgeVerification.test.js#L109-L121)

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: I like the separation of having the `isLink` prop. Agree that I could achieve the functionality by checking the `href` value but for me having `isLink` makes it clear the component can have different behaviours.

packages/components/atoms/f-image-tile/src/components/_tests/ImageTile.test.js
AlixRichards: Awesome, I'll update this 👍

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -1,8 +1,43 @@
...
+            data-test-id="image-tile-input"
+            @change="toggleFilter">
+        <label
+            :class="[
+                $style['c-imageTile-label'], {
+                    [$style['c-imageTile-label--selected']]: isFilterSelected
+                }]"
+            :for="`imageTileToggle-${tileId}`"
+            data-test-id="image-tile-label">
+            <img
```
jamieomaguire, line 32: is it worth adding `role="presentation"` as the image doesn't provide any additional information to the display text?

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -12,20 +47,108 @@ export default {
...
+         * Toggles the isFilterSelected data property
+         * when the checkbox is toggled.
+         *
+         */
+        toggleFilter () {
+            this.isFilterSelected = !this.isFilterSelected;
+
+            this.$nextTick(() => {
+                this.$emit('toggle', this.filterEmitObject(this.tileId));
+            });
```
jamieomaguire: this is nice - hopefully it circumvents any blocking side effects of the event being emitted! 🤞

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -12,20 +47,108 @@ export default {
...
+        },
+        /**
+         * Creates the object that is $emitted
+         * when the checkbox is toggled
+         *
+         * @param {tileId} string
+         */
+        filterEmitObject (tileId) {
```
jamieomaguire: super NIT: Could we rename this function something like `createEmitObject`?

we should probably avoid any references to the word `filter` to keep it super presentational and less domain aware

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -12,20 +47,108 @@ export default {
...
+            default: null
+        },
+        imgSrc: {
+            type: String,
+            default: null
+        }
+    },
+    data () {
+        return {
+            isFilterSelected: false
```
jamieomaguire: super NIT: mentioned in another comment - perhaps it's worth avoiding references to the word `filter` just to keep any domain references out of this component's concerns

packages/components/atoms/f-image-tile/src/components/_tests/ImageTile.test.js
```diff
@@ -7,4 +7,148 @@ describe('ImageTile', () => {
         const wrapper = shallowMount(ImageTile, { propsData });
         expect(wrapper.exists()).toBe(true);
     });
+
+    describe('props :: ', () => {
```
jamieomaguire, line 11: I haven't noticed this `::` convention before

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: I created a POC and wiki doc which you can find [here] (https://justeattakeaway.atlassian.net/wiki/spaces/FDG/pages/2981953822/Search-2136+Image+Tile+-+Phase+1+MVP+-+Spike+HTML+and+behaviour+implementation) where I explore a options for the component markup 

I considered conditional markup changes but my preference is to have same HTML whatever the state of the component. For now I'd like to keep the markup as is but I have further tickets in the epic that focus on these areas in more detail so I will keep your comment in mind and am open to changing this in the future if it seems more appropriate/better solution 👍

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -12,20 +47,108 @@ export default {
...
+    },
+    watch: {
+        isSelected (newValue) {
+            if (newValue !== this.isFilterSelected) {
+                this.isFilterSelected = newValue;
+            }
+        }
+    },
+    mounted () {
+        this.isFilterSelected = this.isSelected;
```
jamieomaguire: This lifecycle method is client only, I believe. Therefore it'll cause issues during SSR on any consuming apps I believe. Might be worth looking into `created` or others. [Reference](https://ssr.vuejs.org/guide/universal.html#component-lifecycle-hooks)

jamieomaguire: In fact, I think the `watch` method may take care of this for you - worth checking!

jamieomaguire: I think:

```
watch: {
  isSelected: {
    immediate: true,
    handler (newValue) {
      if (newValue !== this.isFilterSelected) {
        this.isFilterSelected = newValue;
      }
    }
  }
}
```

Might work

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -1,8 +1,43 @@
...
+                }]"
+            :href="isLink ? href : '#'"
+            :aria-hidden="!isLink"
+            :tabindex="isLink ? false : -1"
+            data-test-id="image-tile-link">
+            <span class="is-visuallyHidden">
+                {{ displayText }}
+            </span>
+        </a>
+        <input
```
jamieomaguire, line 18: my comment seems to have disappeared, but I asked before if this checkbox is going to be announced by screenreaders even when the tile is being used soley as a link? Might be worth using dynamic markup if so - what do you think?

Could always throw a conditional aria-hidden on it if not

packages/components/atoms/f-image-tile/src/components/_tests/ImageTile.test.js
```diff
@@ -7,4 +7,148 @@ describe('ImageTile', () => {
...
+                    const image = wrapper.find('[data-test-id="image-tile-image"]');
+
+                    // Assert
+                    expect(image.attributes('src')).toContain(propsData.imgSrc);
+                });
+            });
+
+            describe('isSelected :: ', () => {
+                const isSelectedTrue = true;
+                const isSelectedFalse = false;
```
jamieomaguire, line 85: NIT: I think the code is self-explanatory enough just passing in booleans to `it.each` and naming the `expected` parameter as you have. If it were me I'd probably skip adding the two variables above

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: I'll add a note to the image ticket 👍

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: Yes! Let me change this 👍

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: Yep, good spot 👍

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: Cool, let me add this to the relevant ticket and I can explore that further there 👍

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: I created a POC and wiki doc which you can find [here] (https://justeattakeaway.atlassian.net/wiki/spaces/FDG/pages/2981953822/Search-2136+Image+Tile+-+Phase+1+MVP+-+Spike+HTML+and+behaviour+implementation) where I explore a options for the component markup

I will bear this in mind within the screen-reader experience ticket and am open to amending if it seems the best solution but during my testing of the POC I don't remember having this issue. Markup may be slightly different above so will test thoroughly within that ticket 👍

jamieomaguire: Sure thing - if it seems fine on screen readers then the markup is A-Ok!

packages/components/atoms/f-image-tile/src/components/_tests/ImageTile.test.js
AlixRichards: I really like the explicitness of the variables, however I see your point. I'm going to leave as is for now but I am open to potentially revisiting this when I add further tests 👍

jamieomaguire (approved):

simonsandersje (approved):

dandel10n (approved):

1itvinka (approved):

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
ashleynolan: I tend to agree with @1itvinka – the only issue with having a prop for whether it's a link as well as an `href` prop is that you're allowing more room for error for consumers of the component. For someone new to the component, if I add a link to the `href` prop, I'd expect the component to act like a link (I wouldn't expect it to not do that without also setting the `isLink` prop).

For me, `isLink` should be a computed prop which does a check for whether href is set, unless you ever want there to be a use-case where the `href` can be set to a value and `isLink` also set to false.

PR #1568, opened 2021-12-17
f-account-info@0.14.0 - Adding analytics.

SachaZvetelman:
v0.14.0
*December 17, 2021*

### Added
- Analytics requirements when form is submitted.


packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -17,7 +17,7 @@
                 method="post"
                 @submit.prevent="onFormSubmit">
                 <form-field
-                    :value="consumer.firstName"
+                    v-model="consumer.firstName"
```
SachaZvetelman: These changes were done because the inputs were not passing validations at any point, so they were broken. @kevinrodrigues will probably take a look at this in detail next week, and if not, we'll check it in the new year 😄

tommcclean: What do you mean not "passing validations"?

SachaZvetelman: Basically, you'd put a valid value in any of the fields, and you'd get the red message under the field saying it was an invalid value.

tommcclean (dismissed):

packages/components/pages/f-account-info/src/services/analytics.js
```diff
@@ -0,0 +1,26 @@
+export default class AccountInfoAnalyticsService {
+    constructor ({ $gtm }) {
+        this.$gtm = $gtm;
```
oliversweb: Maybe make this private ... `this.#$gtm`

packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -321,6 +326,8 @@ export default {
                 return;
             }
 
+            this.accountInfoAnalyticsService.trackFormSubmission(this.hasAddressBeenUpdated);
```
oliversweb: Maybe move this to line 340 so we only emit the event on a successful change

packages/components/pages/f-account-info/src/services/providers/_tests/__snapshots__/analytics.test.js.snap
```diff
@@ -0,0 +1,55 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
oliversweb: Is this in the correct folder?

packages/components/pages/f-account-info/src/services/providers/_tests/analytics.test.js
```diff
@@ -0,0 +1,42 @@
+import AccountInfoAnalyticsService from '../../analytics';
```
oliversweb: again ... should this be in the `service/_tests` folder only

packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -246,9 +249,11 @@ export default {
                 cookies: this.$cookies,
                 baseUrl: this.smartGatewayBaseUrl
             }),
+            accountInfoAnalyticsService: new AccountInfoAnalyticsService(this),
```
oliversweb: Maybe just pass in `this.$gtm` rather than the whole context

packages/components/pages/f-account-info/src/components/AccountInfo.vue
xander-marjoram: I suppose one advantage of having it here is that if the submission breaks it may be detectable by the increase in button clicks

packages/components/pages/f-account-info/src/components/AccountInfo.vue
oliversweb: I changed the `:v-model` to `:value` because you should not bind a vuex store state model to a bi-direction binder, it will complain because it will try and update the state directly rather than going via a mutation.

packages/components/pages/f-account-info/src/components/_tests/AccountInfo.test.js
```diff
@@ -251,6 +290,36 @@ describe('AccountInfo', () => {
                     // Assert
                     expect(wrapper.vm.hasFormUpdate).toEqual(true);
                 });
+
+                describe('form field is part of the address', () => {
+                    it('should set `hasAddressBeenUpdated` to `true`', async () => {
+                        // Arrange
+                        wrapper = await mountAccountInfo();
+
+                        const element = wrapper.find('[data-test-id="account-info-consumer-line1"]');
```
xander-marjoram: Should we have tests for the other address lines too?

packages/components/pages/f-account-info/src/components/AccountInfo.vue
oliversweb: We can check if the submission is broken by the logs

packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -331,11 +336,13 @@ export default {
                 // TODO - to be added with next ticket
                 this.$log.info('Consumer details saved successfully', ['account-pages', 'account-info']);
                 this.hasFormUpdate = false;
+                this.accountInfoAnalyticsService.trackFormSubmission(this.hasAddressBeenUpdated);
             } catch (error) {
                 this.$log.error('Error saving consumer details', error, ['account-pages', 'account-info']);
                 this.handleErrorState(new AccountInfoError(error.message, error?.response?.status));
             } finally {
                 this.setSubmittingState(false);
+                this.hasAddressBeenUpdated = false;
```
oliversweb: Maybe move this under 339 now otherwise it may reset in the case of a failure

tommcclean (approved):

oliversweb (approved):

packages/components/pages/f-account-info/src/components/_tests/AccountInfo.test.js
```diff
@@ -251,6 +290,46 @@ describe('AccountInfo', () => {
...
+                        ['lastName'],
+                        ['phoneNumber']
+                    ])('should not set `hasAddressBeenUpdated` to `true` when `field` is %s', async field => {
+                        // Arrange
+                        wrapper = await mountAccountInfo();
+
+                        const element = wrapper.find(`[data-test-id="account-info-consumer-${field}"]`);
+
+                        // Act
+                        element.vm.$emit('input', 'Harry Potter');
```
xander-marjoram, line 327: Might be better to use a valid phone number?
```suggestion
                    it.each([
                        ['firstName', 'Harry'],
                        ['lastName', 'Potter'],
                        ['phoneNumber', '07890123456']
                    ])('should not set `hasAddressBeenUpdated` to `true` when `field` is %s', async (field, inputValue) => {
                        // Arrange
                        wrapper = await mountAccountInfo();

                        const element = wrapper.find(`[data-test-id="account-info-consumer-${field}"]`);

                        // Act
                        element.vm.$emit('input', inputValue);
```

PR #1569, opened 2021-12-17
f-restaurant-card@v0.11.1 - Add f-vue-icons as dependency

jamieomaguire:
v0.11.1
*December 17, 2021*
### Added
- @justeat/f-vue-icons added as dev dependency

### DIFF
--- a/packages/components/molecules/f-restaurant-card/CHANGELOG.md
+++ b/packages/components/molecules/f-restaurant-card/CHANGELOG.md
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.11.1
+------------------------------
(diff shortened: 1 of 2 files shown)


1itvinka (dismissed):

oliversweb (dismissed):

1itvinka (approved):

siggerzz (commented):
Just needs a yarn.lock update and then I'm happy to merge 👍🏼

jamieomaguire:
> Just needs a yarn.lock update and then I'm happy to merge 👍🏼

@ashleynolan - updating yarn at the root level doesn't appear to update the lock file. Is there something else I should be doing? As I understand it, there is not an individual lock file for each component package

siggerzz (approved):

ashleynolan:
@jamieomaguire I don't think this would have needed a `yarn.lock` update, as the package you added is already being used, so likely wouldn't have added anything to the lock file. I think it would only have amended it if you were adding a different version of the same package.

PR #1570, opened 2021-12-20
f-header@9.2.0 - remove ml theme style overrides

dandel10n:
### Removed
- 'ml' theme override for Logo.
- width css definition from logo icon to make sure that icons for different tenants (je and ml) are rendered correctly with same height.
- nb-No and da-DK from copy, storybook story and tests.

### Changed
- Fixed Logo tests.

### Updated
- fozzie package to the latest (v6.2.0).


packages/components/organisms/f-header/src/components/_tests/Logo.test.js
```diff
@@ -4,7 +4,7 @@ import Logo from '../Logo.vue';
 describe('Logo', () => {
     it('should be defined', () => {
         const propsData = {
-            theme: 'je',
+            theme: 'jet',
```
DamianMullins: Do we still need this prop?

dandel10n: We need this prop for setting the right logo, but actually I noticed that some tests should fail with changing `je` to `jet` but they don't :) I will dig into this :+1:

DamianMullins: Ah right, cool. OK, I'll keep an eye on any changes to the PR.

dandel10n: Ok, I think now the changes make more sense :)

packages/components/organisms/f-header/src/components/_tests/Logo.test.js
```diff
@@ -17,51 +16,48 @@ describe('Logo', () => {
         // Arrange
         const propsData = {
             theme: 'ml',
-            headerBackgroundTheme: 'transparent',
```
AlixRichards, line 20: Do these both still need the `theme` to be set?

AlixRichards (approved):

DamianMullins (approved):

PR #1581, opened 2022-01-11
#trivial - component tests for account-info

LTurns:
### Added
- component tests in `f-account-info.component.spec`
- test ids in `DeleteAccount.vue` and `EmailAddressField.vue`
- `vuelidate` and `f-services` package in `package.json`  - to resolve current issue on storybook prod - https://vue.pie.design/?path=/story/components-pages--account-info-component

### Changed
- data-test-ids in `AccountInfo.vue`  - the `formfield` component needs a `name` element to override the `data-test-id` it already has - https://github.com/justeat/fozzie-components/blob/master/packages/components/atoms/f-form-field/src/components/FormField.vue 
- test filenames from `accountInfo` to `account-info`

### DIFF
--- a/packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
+++ b/packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
@@ -1,5 +1,5 @@
 exports.COMPONENT = '[data-test-id="account-info"]';
-exports.CHANGE_EMAIL_ADDRESS_LINK = '[name="account-info-change-email-link"]';
+exports.CHANGE_EMAIL_ADDRESS_LINK = '[name="account-info-change-emailAddress-link"]';
 exports.SAVE_CHANGES_BUTTON = '[data-test-id="account-info-save-changes-button"]';
 exports.CHANGE_PASSWORD_BUTTON = '[data-test-id="account-info-change-password-button"]';
 exports.DELETE_ACCOUNT_LINK = '[name="account-info-delete-account-link"]';
--- a/packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
+++ b/packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
@@ -1,8 +1,8 @@
 exports.COMPONENT = '[data-test-id="account-info"]';
-exports.CHANGE_EMAIL_ADDRESS_LINK = '[id="change-email-address-link"]';
-exports.SAVE_CHANGES = '[data-test-id="account-info-save-changes-button"]';
-exports.CHANGE_PASSWORD = '[data-test-id="account-info-change-password-button"]';
-exports.DELETE_ACCOUNT = '[id="delete-account"]';
+exports.CHANGE_EMAIL_ADDRESS_LINK = '[name="account-info-change-email-link"]';
+exports.SAVE_CHANGES_BUTTON = '[data-test-id="account-info-save-changes-button"]';
+exports.CHANGE_PASSWORD_BUTTON = '[data-test-id="account-info-change-password-button"]';
+exports.DELETE_ACCOUNT_LINK = '[name="account-info-delete-account-link"]';
 
 
 exports.FIELDS = {
--- a/packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
+++ b/packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
@@ -17,7 +17,7 @@ exports.FIELDS = {
         invalidError: '[data-test-id="consumer-lastName-invalid-error"]'
     },
     emailAddress: {
-        input: '[data-test-id="formfield-account-info-consumer-emailAddress"]'
+        input: '[data-test-id="formfield-account-info-consumer-emailAddress-input"]'
     },
     phoneNumber: {
         input: '[data-test-id="formfield-account-info-consumer-phoneNumber-input"]',
--- a/packages/components/pages/f-account-info/test-utils/component-objects/f-account-info.component.js
+++ b/packages/components/pages/f-account-info/test-utils/component-objects/f-account-info.component.js
@@ -14,10 +14,26 @@ module.exports = class AccountInfo extends Page {
     }
 
     get component () { return $(COMPONENT); }
-    get changeEmailAddressLink () { return $(CHANGE_EMAIL_ADDRESS_LINK); }
-    get saveChangesButton () { return $(SAVE_CHANGES_BUTTON); }
-    get changePasswordButton () { return $(CHANGE_PASSWORD_BUTTON); }
-    get deleteAccountLink () { return $(DELETE_ACCOUNT_LINK); }
+
+    LinksAndButtons = {
+        changeEmailAddressLink: {
+            get cta () { return $(CHANGE_EMAIL_ADDRESS_LINK); }
+        },
+        saveChangesButton: {
+            get cta () { return $(SAVE_CHANGES_BUTTON); }
+        },
+        changePasswordButton: {
+            get cta () { return $(CHANGE_PASSWORD_BUTTON); }
+        },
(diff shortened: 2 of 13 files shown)


packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
```diff
@@ -0,0 +1,46 @@
+exports.COMPONENT = '[data-test-id="account-info"]';
+exports.CHANGE_EMAIL_ADDRESS_LINK = '[name="account-info-change-emailAddress-link"]';
```
siggerzz, line 2: My only worry with relying on the `name` property is that if it changes in the future, it's not immediately clear that it will cause the tests to break.

IMO we should be consistent in using `data-test-id`. If we're using `name` elsewhere in the repo, we should look to change these back in future PR's 👍🏼

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
```diff
@@ -0,0 +1,46 @@
...
+exports.SAVE_CHANGES_BUTTON = '[data-test-id="account-info-save-changes-button"]';
+exports.CHANGE_PASSWORD_BUTTON = '[data-test-id="account-info-change-password-button"]';
+exports.DELETE_ACCOUNT_LINK = '[name="account-info-delete-account-link"]';
```
siggerzz, line 5: ditto

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
LTurns: yep I agree actually, probably worth making a ticket to revert `f-form-field` so it no longer has this function and we can simply add the `data-test-id` in - I've made a note to add this

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
```diff
@@ -0,0 +1,46 @@
...
+        emptyError: '[data-test-id="consumer-firstName-empty-error"]',
+        invalidError: '[data-test-id="consumer-firstName-invalid-error"]'
+    },
+    lastName: {
+        input: '[data-test-id="formfield-account-info-consumer-lastName-input"]',
+        emptyError: '[data-test-id="consumer-lastName-empty-error"]',
+        invalidError: '[data-test-id="consumer-lastName-invalid-error"]'
+    },
+    emailAddress: {
+        input: '[data-test-id="formfield-account-info-consumer-emailAddress"]'
```
siggerzz: Have we catered for invalid / empty emails in the component?

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
```diff
@@ -0,0 +1,46 @@
...
+    city: {
+        input: '[data-test-id="formfield-account-info-consumer-locality-input"]',
+        emptyError: '[data-test-id="consumer-city-empty-error"]'
+    },
+    postcode: {
+        input: '[data-test-id="formfield-account-info-consumer-postcode-input"]',
+        emptyError: '[data-test-id="consumer-postcode-empty-error"]',
+        invalidError: '[data-test-id="consumer-postcode-invalid-error"]'
+    }
+};
```
siggerzz, line 46: Ditto for these other fields. It seems to me that we should have 'invalid' validation states for the address fields too.

Maybe not something for this PR, but worth raising with @oliversweb perhaps?

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info.component.js
```diff
@@ -0,0 +1,134 @@
...
+        return this.component.isDisplayed();
+    }
+
+    /**
+    * @description
+    * Select all the text in the field and then performs a backspace to clear the field
+    *
+    * @param {String} fieldName The name of the field input it is clearing
+    */
+    clearBlurField (fieldName) {
```
siggerzz, line 91: Nit: Lets move this to `f-wdio-utils` soon. We're using it a lot in other component objects.

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info.component.js
```diff
@@ -0,0 +1,134 @@
...
+    * @description
+    * Inputs customer details into the account-info component.
+    *
+    * @param {Object} customerInput customer input details
+    */
+    populateAccountForm (field, customerInput) {
+        this.fields[field].input.setValue(customerInput[field].input);
+    }
+
+    clickOutOfInputField () {
```
siggerzz, line 110: Purpose of this? Is this because errors only appear on blur of fields?

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
LTurns: It's not possible to have an invalid/empty email in this component - the input box is completely disabled with the user's email inside -
image

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info.component.js
```diff
@@ -0,0 +1,134 @@
...
+
+    changePasswordButtonCanBeClicked () {
+        return this.changePasswordButton.isClickable();
+    }
+
+    deleteAccountLinkCanBeClicked () {
+        return this.deleteAccountLink.isClickable();
+    }
+
+    isDisabled (field) {
```
siggerzz, line 126: Should this be `return !this.fields[field].input.isEnabled();`?

packages/components/pages/f-account-info/test/component/f-account-info.component.spec.js
```diff
@@ -0,0 +1,96 @@
...
+        // Assert
+        expect(accountInfo.isComponentDisplayed()).toBe(true);
+    });
+
+    it('should show that the email formfield is disabled', () => {
+        // Assert
+        expect(accountInfo.isDisabled('emailAddress')).toBe(true);
+    });
+
+    it('should check if all call to actions are clickable', () => {
```
siggerzz: Might be worth changing this to a generic `canBeClicked(buttonName)` function and using `mocha-each` to for `forEach` this test.

Currently, I feel if an assert fails, it won't execute future asserts. This can be avoided with the above approach.

siggerzz (changes requested):
Looking good! Just a few minor queries + potential suggestions.

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info.component.js
LTurns: Was thinking this too 👏 we might be able to make a general `populateForm` function too, instead of repeating for every component ^ have made a note to add a ticket

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info.component.js
LTurns: yep it is indeed :)

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info.component.js
LTurns: ah yes thank you!

JoshuaNg2332 (approved):

packages/components/pages/f-account-info/test-utils/component-objects/f-account-info-selectors.js
LTurns: spoken to @tommcclean about this, we don't require 'invalid' validation states for the address fields - they're just required fields needed to send to the API. However, we need the postcode to be in the correct format (use regex for this), hence why this has the extra validation 👍

siggerzz (approved):

PR #1584, opened 2022-01-13
f-checkout@3.13.3 - Add test id and focus

FayeCarter:
### Added
- Test id for guest header.
- Focus for first input in fieldset.

### DIFF
--- a/packages/components/pages/f-checkout/src/components/Address.vue
+++ b/packages/components/pages/f-checkout/src/components/Address.vue
@@ -2,7 +2,8 @@
     <div data-test-id='address-component'>
         <fieldset :class="$style['c-address-group']">
             <legend
-                :class="$style['c-address-label']">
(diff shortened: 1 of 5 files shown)


charliearlie (approved):

packages/components/pages/f-checkout/src/components/Address.vue
```diff
@@ -54,6 +55,12 @@ export default {
             fieldType: 'address',
             fieldLength: '50'
         };
+    },
+
+    methods: {
+        addFocus () {
+            document.getElementById('formField-line-1').focus();
```
BanksySan, line 62: I think this method should either be in the checkout-form-field or we should expose the name of the field and reference it by that as we have a bit of ikky coupling as it is.

BanksySan (changes requested):

BanksySan (approved):

PR #1591, opened 2022-01-14
f-checkout@3.14.0 - Fix getAddressClosestToPostcode logic

charliearlie:
The logic inside `getAddressClosestToPostcode` would format the postcode which exists in the `je-location` cookie by removing the whitespace, and then would compare it to a postcode returned in the list of addresses from the API. These postcodes returned from the API often had spaces so there would rarely ever be an address found in this line 
```
let address = addresses.find(a => a.ZipCode === formattedPostcode);
```

Now there will be. 🙂

### DIFF
--- a/packages/components/pages/f-checkout/src/services/addressService.js
+++ b/packages/components/pages/f-checkout/src/services/addressService.js
@@ -44,14 +44,16 @@ function getDefaultAddress (addresses) {
     return !addresses ? null : addresses.find(a => a && a.IsDefault);
 }
 
+const formatPostcode = postcode => postcode.replace(/ /g, '').replace(/-/g, '');
+
 function getAddressClosestToPostcode (postcode, addresses) {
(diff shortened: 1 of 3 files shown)


packages/components/pages/f-checkout/src/services/addressService.js
```diff
@@ -44,14 +44,16 @@ function getDefaultAddress (addresses) {
     return !addresses ? null : addresses.find(a => a && a.IsDefault);
 }
 
+const formatPostcode = postcode => postcode.replace(/ /g, '').replace(/-/g, '');
```
BanksySan: You probably was to use the whitespace character class instead of a space.

BanksySan (changes requested):

BanksySan (approved):

FayeCarter (approved):

PR #1593, opened 2022-01-17
f-account-info - small changes to enable contract tests #trivial

LTurns:
Small PR to enable contract tests - 

### Added
-  `src/services` and `src/constants.js` to `dist` in package.json

### DIFF
--- a/packages/components/pages/f-account-info/src/services/providers/Consumer.api.js
+++ b/packages/components/pages/f-account-info/src/services/providers/Consumer.api.js
@@ -7,10 +7,14 @@ import {
 } from '../../constants';
 
 export default class ConsumerApi {
+    #httpClient;
+    #cookies;
+    #baseUrl;
+
     constructor ({ httpClient, cookies, baseUrl } = {}) {
-        this.httpClient = httpClient;
-        this.cookies = cookies;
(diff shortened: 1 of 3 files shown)


siggerzz (dismissed):

packages/components/pages/f-account-info/src/services/providers/Consumer.api.js
```diff
@@ -7,14 +7,10 @@ import {
...
 export default class ConsumerApi {
-    #httpClient;
-    #cookies;
-    #baseUrl;
-
     constructor ({ httpClient, cookies, baseUrl } = {}) {
-        this.#httpClient = httpClient;
-        this.#cookies = cookies;
-        this.#baseUrl = baseUrl;
+        this.httpClient = httpClient;
```
oliversweb: Have you made these variable public for any reason?

LTurns: yep unfortunately when testing with `yalc` I kept getting an error because they're private properties - `Cannot read property 'get' of undefined`

oliversweb (approved):

JoshuaNg2332 (approved):

PR #1594, opened 2022-01-17
f-header@9.3.0 - Fix keyboard navigation and accessibility of user menu

xander-marjoram:
### Added
- Tests for `UserNavigationPanel`'s computed property `tabIndex`.
- `aria-label` to user account submenu button.

### Fixed
- User menu was not visibly toggling with keyboard navigation.

### Changed
- User account submenu link to button for increased accessibility.
- Screen readers will now announce the country selector title, e.g., "Select your country".
- Reorganised and rewritten some of the Navigation unit tests.

---

## UI Review Checks

## Browsers Tested
- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -151,14 +151,18 @@
                     data-test-id="user-info-icon"
                     v-on="isBelowMid ? null : { mouseover: openUserMenu, mouseleave: closeUserMenu }"
                     @keyup.esc="closeUserMenu">
-                    <a
+                    <button
```
dandel10n, line 154: Could you use ghost f-button in this case? There is `hasIcon=leading` prop so it should look like you need, I think

packages/components/organisms/f-header/src/components/UserNavigationPanel.vue
```diff
@@ -62,15 +66,19 @@ export default {
             type: String,
             default: ''
         },
-        isCountrySelectorClosedOnMobileView: {
+        isCountrySelectorOpen: {
             type: Boolean,
             default: true
```
dandel10n: Should the default value become false if the prop now states "open"?

packages/components/organisms/f-header/src/components/Navigation.vue
xander-marjoram: It's a good idea but there are quite a lot of differences with the styling at the minute, so I would need to put in lots of overrides which does defeat the point slightly. Perhaps once we have new header designs it might be more fitting?

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -525,18 +533,25 @@ export default {
                 this.firstFocusableEl = firstFocusableEl;
                 this.lastFocusableEl = lastFocusableEl;
 
+                // Add loop-handling event listener
                 nav.addEventListener('keydown', this.keyActions);
             });
         },
 
+        toggleUserMenu () {
+            if (this.userMenuIsOpen) {
```
FayeCarter: Could this be `this.userMenuIsOpen ? this.closeUserMenu() : this.openUserMenu();`

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -85,13 +85,13 @@ export default {
...
     methods: {
-        closeCountrySelector () {
+        closeCountrySelector (withFocus = true) {
             this.isCountrySelectorOpen = false;
             this.$emit('close-country-selector');
-            const {
-                countrySelectorToggle
-            } = this.$refs;
-            countrySelectorToggle.focus();
+            const { countrySelectorToggle } = this.$refs;
```
FayeCarter: should this be inside the if statement?

packages/components/organisms/f-header/src/components/Navigation.vue
xander-marjoram: eslint isn't a massive fan of it, saying it's an "unused expression", so I've had to stick a `return` on the front which works just fine

simonsandersje (approved):

dandel10n (approved):

PR #1598, opened 2022-01-17
f-braze-adapter@v4.1.0  - Location filter

matthewhardern:
This PR adds a location filter to the braze consumer under filters. It filters by users current location against any cards that contain a `location` attribute. I have added tests to support this case.


packages/services/f-braze-adapter/src/services/utils/transformCardData.js
```diff
@@ -49,6 +49,7 @@ const transformCardData = card => {
         restaurant_logo_url: restaurantLogoUrl,
         icon_1: icon = restaurantLogoUrl,
         location,
+        locations,
```
minrwhite: Based on the ticket, we're not expecting `locations` as a key within the extras, instead `location` can potentially have more than one location in, in a comma-separated fashion.

packages/services/f-braze-adapter/src/services/BrazeConsumer.js
```diff
@@ -108,7 +110,8 @@ class BrazeConsumer {
             sortByCardOrder,
             cards => filterByEnabledCardTypes(cards, this.enabledCardTypes),
             cards => filterByBrands(cards, this.brands),
-            filterByCurrentlyActive
+            filterByCurrentlyActive,
+            cards => locationFilter(cards, currentLocation)
```
minrwhite, line 114: ~We don't want 'For You' cards filtered out when they have a location set, but when there is no currently searched for location. This can be revised later on (i.e. location dropdown solution) but for now please allow this filter to be added for homepage (and therefore HCC) only.~ As discussed with Ivan, we can ensure that this filter applies with respect to the templatable properties required by the url string, and filter based on that rather than the presence or otherwise of the location by itself.

matthewhardern: I have now added in a check for the URL as discussed

packages/services/f-braze-adapter/src/services/utils/locationFilter.js
```diff
@@ -19,6 +30,10 @@ const locationFilter = (cards, { location }) => cards.filter(card => {
         if (!location) {
             return false;
         }
+        // if we don't have a valid lat long throw it out
+        if ((needsLongitude && needsLatitude) && !hasValidLocation(latitude, longitude)) {
```
minrwhite: I don't believe it's guaranteed that a lat or lon will be required to also have a location, in which case this should also be outside of the `if (needsLocation) {` block. Also, this should be an OR condition rather than AND on the (admittedly odd, but possible) chance that a string has included a latitude placeholder without a longitude placeholder, or vice-versa.

minrwhite (commented):
One comment but otherwise looks good.

packages/services/f-braze-adapter/src/services/utils/locationFilter.js
matthewhardern: On the first point it may not require a location, but do we want to filter out any that do not have a location KVP? My understanding was we would filter out any that have invalid lat, long if they have a location? Second point good catch will make that change.

minrwhite: Yes, good point - but we shouldn't make that dependent on also having a location placeholder as well. I.e. we might have a url `www.example.com/$LAT$/$LON$` that doesn't have a reference to the location, but we should allow that to show (and the placeholders populated accordingly) if the `location` KVP matches to the last stored location.

matthewhardern: made the suggested changes based on our previous conversations.

packages/services/f-braze-adapter/src/services/utils/locationFilter.js
```diff
@@ -0,0 +1,61 @@
...
+    const needsLocation = card?.url?.includes('$LOCATION$');
+    const needsLatitude = card?.url?.includes('$LAT$');
+    const needsLongitude = card?.url?.includes('$LON$');
+    const hasPlaceHolder = needsLongitude || needsLatitude || needsLocation;
+
+    // will be undefined if no locations KVP exists
+    const locationsList = card?.location?.split(',');
+    // return true if NO KVP and if user has no location and hasPlaceholder is false or user has location and hasPlaceholder is true
+    if (locationsList === undefined) {
+        return !!((needsLocation && location) || (!needsLocation && hasValidLocation(latitude, longitude)) || (!hasPlaceHolder));
```
minrwhite: For readability (and to preempt `hasValidLocation()` evaluation if we don't need to) I'd suggest:
```suggestion
        return !!((!hasPlaceHolder)
            || (needsLocation && location)
            || (!needsLocation && hasValidLocation(latitude, longitude)));
```

packages/services/f-braze-adapter/src/services/utils/_tests/locationFilter.test.js
```diff
@@ -43,135 +49,55 @@ const MOCK_CARDS_WITHOUT_LOCATION_KVP = [
...
+    describe('card has NO Location KVP', () => {
+        it.each([
+            [4, { location: MOCK_CURRENT_LOCATION.location, latitude: MOCK_CURRENT_LOCATION.latitude, longitude: MOCK_CURRENT_LOCATION.longitude }],
+            [3, { location: null, latitude: MOCK_CURRENT_LOCATION.latitude, longitude: MOCK_CURRENT_LOCATION.longitude }],
+            [3, { location: MOCK_CURRENT_LOCATION.location, latitude: null, longitude: null }],
+            [3, { location: MOCK_CURRENT_LOCATION.location, latitude: '0.000', longitude: '0.000' }],
+            [3, { location: MOCK_CURRENT_LOCATION.location, latitude: null, longitude: MOCK_CURRENT_LOCATION.longitude }],
+            [3, { location: MOCK_CURRENT_LOCATION.location, latitude: MOCK_CURRENT_LOCATION.latitude, longitude: null }],
+            [1, { location: null, latitude: null, longitude: null }]
+        ])('should return %s cards when the users location is %o', (numberOfCards, locationObject) => {
```
minrwhite: For this and the similar section above, I think for understandability's sake it would be good to have a simple description string describing each case, and to help identifying the test case in question:
```suggestion
            [4, 'has all location attributes', { location: MOCK_CURRENT_LOCATION.location, latitude: MOCK_CURRENT_LOCATION.latitude, longitude: MOCK_CURRENT_LOCATION.longitude }],
            [3, 'only has geocoords', { location: null, latitude: MOCK_CURRENT_LOCATION.latitude, longitude: MOCK_CURRENT_LOCATION.longitude }],
            [3, 'only has location', { location: MOCK_CURRENT_LOCATION.location, latitude: null, longitude: null }],
            [3, 'has invalid geocoords', { location: MOCK_CURRENT_LOCATION.location, latitude: '0.000', longitude: '0.000' }],
            [3, 'is missing latitude', { location: MOCK_CURRENT_LOCATION.location, latitude: null, longitude: MOCK_CURRENT_LOCATION.longitude }],
            [3, 'is missing longitude', { location: MOCK_CURRENT_LOCATION.location, latitude: MOCK_CURRENT_LOCATION.latitude, longitude: null }],
            [1, 'has no valid location attributes', { location: null, latitude: null, longitude: null }]
        ])('should return %s cards when the users location %s', (numberOfCards, description, locationObject) => {
```

minrwhite (approved):

simonsandersje (approved):

PR #1602, opened 2022-01-18
f-checkout@3.15.0 - Re-call note config if note not accepted

charliearlie:


packages/components/pages/f-checkout/src/components/Checkout.vue
```diff
@@ -5,7 +5,8 @@
             v-if="checkoutErrorMessage"
             ref="errorMessage"
             v-bind="errorProps.props"
-            @created="handleDialogCreation">
+            @created="handleDialogCreation"
+            @note-not-accepted="handleNoteNotAccepted">
```
BanksySan, line 9: Could we have a positive boolean, i.e. `noteAccepted` instead of the negative?

BanksySan (changes requested):

packages/components/pages/f-checkout/src/components/Checkout.vue
charliearlie: I'm not sure what you mean? The event is when the note isn't accepted so we need to handle when the note isn't accepted?

BanksySan: Of course.  It's an event, sorry, ignore me.

BanksySan (approved):

FayeCarter (approved):

PR #1603, opened 2022-01-18
fozzie-components@v5.9.0 - Automatic package publishing to npm

siggerzz:
- Add new `publish-to-npm` workflow in `.circleci/config.yml` to automatically publish packages to npm as part of `master` builds
- Update Storybook / Contributing guides around npm publishing best practice.
- Update Percy CLI dependency.
- Add `"tag" : "beta` property to `@justeat/f-searchbox` `package.json` to ensure the version is tagged correctly once published.
- Make `@justeat/storybook` a private package so it's not published to npm

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -486,6 +486,17 @@ jobs:
       - slack_notify_fail
       - slack_notify_success
 
+  publish-to-npm:
+    executor: node
+    steps:
+      - attach_workspace:
+          at: .
+      - run:
+          name: Publish unreleased package versions to npm
+          command: npx lerna publish from-package --ignore-scripts --yes
+      - slack_notify_fail
+      - slack_notify_success
+
 workflows:
   version: 2
 
@@ -532,6 +543,16 @@ workflows:
               ignore: ['gh-pages']
           requires:
             - unit-tests-<< matrix.component_type >>
+      - publish-to-npm:
+          context: web-core
+          filters:
+            branches:
+              only: master
+          requires:
+            - browser-tests-atoms
+            - browser-tests-molecules
+            - browser-tests-organisms
+            - browser-tests-pages
 
   deploy-storybook:
     jobs:
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,26 +73,8 @@ If you are making changes that do not need a version bump in any part of the rep
 
 ## Publishing via npm
 
-To publish a package to **npm**, please first ensure that you have registered an account on [npmjs.com](https://www.npmjs.com/).
-
-Internal contributors (JET Employees) should request to be added to the Just Eat npm organisation by asking in the #guild-frontend Slack channel and providing their npm username. This will then allow you to publish the packages from our mono-repo.
-
-To be able to publish you will need to run `npm login` first. You will also need to add your github access token to `.npmrc` file for components to be published to our Storybook. More info about that can be found [here](https://docs.github.com/en/packages/guides/configuring-npm-for-use-with-github-packages).
-
-External contributors should raise a pull request and the release will be handled for you once approved.
-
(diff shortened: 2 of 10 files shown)


packages/components/molecules/f-searchbox/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "@justeat/f-searchbox",
   "description": "Fozzie Searchbox – Just Eat Takeaway Global Searchbox",
+  "tag": "beta",
```
siggerzz, line 4: By adding this property to beta packages, we ensure that the package will be tagged correctly on the registry, and won't override the `latest` release

packages/storybook/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@justeat/storybook",
+  "private":"true",
```
siggerzz, line 3: Making this private as it's for monorepo use only, so there's no need to publish it!

.circleci/config.yml
```diff
@@ -532,6 +543,16 @@ workflows:
               ignore: ['gh-pages']
           requires:
             - unit-tests-<< matrix.component_type >>
+      - publish-to-npm:
```
siggerzz, line 546: Ensure packages are only published on master once all builds are successful.

simonsandersje (approved):

JoshuaNg2332 (approved):

.circleci/config.yml
ashleynolan: Out of interest, what happens if the build does fail? Does it just mean we'd have to run the publish manually or will it publish next time there's a green build?

CONTRIBUTING.md
```diff
@@ -73,26 +73,8 @@ If you are making changes that do not need a version bump in any part of the rep
...
-2. Update the components version in its `package.json` to be `3.0.0-beta.0`. Subsequent v3 beta releases will increment this integer suffix `-beta.x`
-3. Once a PR has been merged into the v3 release branch, the package should be published using the following command: `npm publish --tag beta`.
-   This publishes the package using the tag `beta` ensuring that it doesn't get tagged as the `latest` version of the package.
-4. When the package is ready for full release as v3, the `package.json` of the component should be updated to `3.0.0` in its `package.json` and a PR from the v3 release branch should be made back into the `main`/`master` branch. When this has been merged, the component can be published as normal using `npm publish`
+When working with beta package versions, you must ensure the `tag: "beta"` property is present in the package's `package.json` file **before** merging to master. Failure to do this will result in the `latest` version being overwritten on the registry.
```
ashleynolan, line 80: Worth having a chat about this flow. I'd recommend keeping beta releases manual and in a separate branch, mainly as there may also be ongoing dev happening on master concurrently which will need versioned releases.

ashleynolan: Also as a sidenote – I think searchbox is probably the edge case for how we usually develop beta releases (in that the original package is in another repo) so having the option to publish beta's like you've mentioned is still good, but not the typical use-case I would say.

stories/getting-started/contributing-guide.stories.mdx
```diff
@@ -49,31 +52,11 @@ If you have any questions about how to implement any of these testing standards,
...
-2. Update the components version in its `package.json` to be `3.0.0-beta.0`. Subsequent v3 beta releases will increment this integer suffix `-beta.x`
-3. Once a PR has been merged into the v3 release branch, the package should be published using the following command: `npm publish --tag beta`.
-   This publishes the package using the tag `beta` ensuring that it doesn't get tagged as the `latest` version of the package.
-4. When the package is ready for full release as v3, the `package.json` of the component should be updated to `3.0.0` in its `package.json` and a PR from the v3 release branch should be made back into the `main`/`master` branch. When this has been merged, the component can be published as normal using `npm publish`
-
+When working with beta package versions, you must ensure the `tag: "beta"` property is present in the package's `package.json` file **before** merging to master. Failure to do this will result in the `latest` version being overwritten on the registry.
```
ashleynolan, line 59: Same here as above

PR #1604, opened 2022-01-19
f-image-tile@0.3.0 - Implement image

AlixRichards:
---

### Added
- ImageTile image and fallback image
- Supporting tests and stories

---

For this ticket the image and fallback image have been implemented, the text is not yet styled and will be picked up in its own ticket. In the first instance this component is being built to replace an existing SearchWeb component however in the future in could be used in other contexts. I've decided to allow the component to go full width and be constrained by its container. Below are screenshots of the 1) the component on its own 2) within a container with fallback image

## UI Review Checks
Screenshot 2022-01-19 at 09 07 23

Screenshot 2022-01-19 at 09 07 03

- [X] Unit tests have been [created]

## Browsers Tested

- [X] Chrome (latest)
- [X] Firefox (latest)
- [X] Safari (latest)

This ticket only covers adding the image as part of the Fozzie component. It doesn't cover its integration into searchWeb. However I have included some WIP images of what this currently looks like when added to SearchWeb replacing the form__toggle component for context.

Additional screenshots
Screenshot 2022-01-18 at 15 05 27
Screenshot 2022-01-18 at 15 06 06


packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -147,8 +186,25 @@ export default {
   border: 1px solid green;
 }
```
AlixRichards, line 182: I'm using the above style for development but will be removed in a future PR when the components different states are styled

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -147,8 +186,25 @@ export default {
...
 
+.c-imageTile-imageContainer {
+  display: block;
+  background-color: rgba($image-tile-background-color, $image-tile-background-opacity);
+  border-radius: $radius-rounded-b;
+  background-image: var(--bg-image);
```
jamieomaguire, line 188: I absolutely love the use of css variables here and might retro fit it into the restaurant card! 

I suppose now we don't support IE 11 we're good to use them? Just want to confirm because it's so nice!

packages/components/atoms/f-image-tile/stories/Demo.stories.js
```diff
@@ -32,12 +36,17 @@ DemoImageTileComponent.argTypes = {
...
     },
     cuisineId: {
         control: { type: 'text' },
         description: 'Cuisine id',
         defaultValue: 'Chicken'
+    },
+    fallbackImage: {
+        control: { type: 'text' },
+        description: 'Fallback image link',
+        defaultValue: 'https://d30v2pzvrfyzpo.cloudfront.net/a/sw/img/wallpaper.png'
```
jamieomaguire: something useful that was recommended to me for f-restaurant-card is to self-host an image and import into the file, rather than rely on an external source.

In case that sounds like something you'd be interested in I'll link the implementation [here](https://github.com/justeat/fozzie-components/blob/41c6f5f62d1b6b4e69daafd35f61bcd2caf69951/packages/components/molecules/f-restaurant-card/stories/RestaurantCard.stories.js#L8) and [here](https://github.com/justeat/fozzie-components/blob/41c6f5f62d1b6b4e69daafd35f61bcd2caf69951/packages/components/molecules/f-restaurant-card/stories/RestaurantCard.stories.js#L31) :)

jamieomaguire (dismissed):

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: Totally. I checked [caniuse](https://caniuse.com/css-variables) and support outside IE11 looks pretty good. Where it isn't supported this will fallback to just an orange background. But potentially I could improve this by using @supports.

packages/components/atoms/f-image-tile/stories/Demo.stories.js
AlixRichards: Ahh perfect, will take a look and implement thank you!

AlixRichards: Awesome, I've updated now 🎇

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: I'm going to leave this as is for now - but create another task/ticket to see if I can improve the experience for places where css variables aren't supported 👍

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -69,14 +75,43 @@ export default {
...
+        },
+        /**
+         * Returns a css variable from the fallback image prop
+         *
+         */
+        cssVars () {
+            const cssVariable = JSON.stringify(this.fallbackImage);
+
+            return {
+                '--bg-image': `url(${cssVariable})`
```
sonnyprince, line 106: the use of this is awesome! top notch stuff ✂️ 🧗

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -147,8 +186,25 @@ export default {
...
+}
+
 .c-imageTile-image {
-    display: block;
+    border-radius: $radius-rounded-b;
+    position: absolute;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
```
sonnyprince, line 197: if this is more for centralised position ignore me - but might be able to drop the setting of the bottom and right values (or top and left)

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -69,14 +75,43 @@ export default {
...
             isToggleSelected: false
         };
     },
+    computed: {
+        /**
+         * Returns a boolean which adds or removes
+         * the 'presentation' role from the image
+         *
+         */
+        addPresentationRole () {
```
1itvinka:
```suggestion
        isPresentationRole () {
```
The name with the action in it is a bit misleading for me. I would name booleans with "is" prefix. It's a bit more clear that this function doesn't add anything, but it just returns a boolean value.

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -69,14 +75,43 @@ export default {
...
+            default: ''
         }
     },
     data () {
         return {
             isToggleSelected: false
         };
     },
+    computed: {
+        /**
```
1itvinka: I would remove the comment because we can rewrite the code so that it's self-explanatory. Unnecessary comments can be dangerous in the long run.

jamieomaguire (dismissed):

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: Yeah for sure! We can remove the bottom and right 👍

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
AlixRichards: 🥳

jamieomaguire (approved):

1itvinka (approved):

PR #1606, opened 2022-01-19
f-mega-modal@4.0.1 - added aria-label to component

LTurns:
### Added
- `aria-labelledBy` to MegaModal.vue to remove the below violations in storybook

aria-labelledBy reads out the title of the modal (references the id) as soon as modal opens. Yalced to f-checkout and it removes the violations there. Works in f-cookie-banner pending a tiny aria-tag change.

image

### DIFF
--- a/packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
+++ b/packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
@@ -19,7 +19,7 @@
                 [$style['is-positioned-bottom']]: isPositionedBottom
             }]"
             role="dialog"
-            aria-labelledBy="modal-title">
+            :aria-labelledby="descriptionId">
             <div
                 ref="megaModalDocument"
                 :class="['c-megaModal-document', {
@@ -51,7 +51,7 @@
                 <component
                     :is="titleHtmlTag"
                     v-if="title"
-                    id="modal-title"
+                    :id="descriptionId"
                     :class="['c-megaModal-title', $style['c-megaModal-title']]"
                     data-test-id="mega-modal-title">
                     {{ title }}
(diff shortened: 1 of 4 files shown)


packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -50,6 +51,7 @@
                 <component
                     :is="titleHtmlTag"
                     v-if="title"
+                    id="modal-title"
```
DamianMullins: I know it shouldn't happen, but I've seen it before, sometimes two modals can be displayed at once which would mean there could be duplicate ids in the page markup that I think would break this?

LTurns: ah good point, thanks @DamianMullins! Just thinking about how else to go about it and @FayeCarter just sent me this which could be a good work around - https://github.com/justeat/fozzie-components/blob/master/packages/components/atoms/f-link/src/components/Link.vue#L79 👏

DamianMullins: Sounds good to me 👍🏻

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -136,18 +140,32 @@ export default {
             type: String,
             default: ''
         },
-
+        ariaTitle: {
```
DamianMullins: Can't see this used anywhere?

LTurns: ah yep no longer needed

FayeCarter (approved):

DamianMullins (approved):

PR #1617, opened 2022-01-24
f-alert@v4.1.0 - Applied Icing Designs

oliversweb:
### Changed
- Aligned the title with the icons
- Set the distance between the title and description & margins to meet the Icing designs.
---
image


SachaZvetelman:
I think the indentation was there on purpose, but @ashleynolan can explain more about that, hopefully 👍 
Other than that, it looks good to me.

ashleynolan (commented):
Looks good to me. If we don't have visual regression checks for alert, should probably make a test excellence ticket to create some (to make sure spacing like this only changes intentionally, not it's been _Iced_ 😆 )

ashleynolan (approved):

SachaZvetelman (approved):

PR #1622, opened 2022-01-26
f-checkout@v3.16.0 - added `f-wdio-utils` functions to tests

LTurns:
### Changed
- Updated `component`, `visual` and `a11y` tests with functions from `f-wdio-utils`

`setFieldValue` function can now be called directly from this package - see PR for more info:
https://github.com/justeat/fozzie-components/pull/1618


packages/components/pages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -237,111 +238,12 @@ module.exports = class Checkout extends Page {
         return this.errorMessageDupOrderGoToHistory.isDisplayed();
     }
 
-    /**
```
siggerzz, line 240: Very nice! 😱

packages/components/pages/f-checkout/test/accessibility/axe-accessibility.spec.js
```diff
@@ -170,11 +171,11 @@ describe('Accessibility tests', () => {
...
             mobileNumber: '+447111111111',
-            line1: '1 Bristol Road',
-            locality: 'Bristol',
-            postcode: 'BS1 1AA'
+            addressLine1: '1 Bristol Road',
+            addressLocality: 'Bristol',
+            addressPostcode: 'BS1 1AA'
         };
-        checkout.populateCheckoutForm(checkoutInfo, customerInfo);
+        Object.keys(customerInfo).forEach(field => checkout.setFieldValue(field, customerInfo[field]));
```
SachaZvetelman: Nit: please add an empty line before and after this line (and the same applies to the other files 🙂)

SachaZvetelman: Besides that, given you have this everywhere, it may be worth extracting it to its own function. Not sure. I'll leave it up to you.

packages/components/pages/f-checkout/test/component/f-checkout.component.spec.js
```diff
@@ -5,6 +5,8 @@ const Checkout = require('../../test-utils/component-objects/f-checkout.componen
 let checkout;
 let checkoutInfo;
 
+const orderTime = 'Wednesday 01:45';
```
SachaZvetelman, line 8: Unless I'm wrong, this is only used in one test at the moment, so we can move this `const` to the scope of that test rather than having it as a global variable.

packages/components/pages/f-checkout/test/visual/f-checkout-authenticated.visual.desktop.spec.js
```diff
@@ -80,13 +80,11 @@ describe('f-checkout - Collection - Authenticated - Desktop Visual Tests', () =>
 
     it('should display the illegal mobile number error message', () => {
         // Arrange
-        const customerInfo = {
-            mobileNumber: '123'
-        };
+        const input = '123';
```
SachaZvetelman: I preferred the `mobileNumber` name, since it's a bit more descriptive of the data it holds.

packages/components/pages/f-checkout/test/accessibility/axe-accessibility.spec.js
LTurns: ah yep good point, will try that now

packages/components/pages/f-checkout/test/component/f-checkout.component.spec.js
LTurns: ooo used twice actually, so will probably keep it global 😄

packages/components/pages/f-checkout/test/visual/f-checkout-authenticated.visual.desktop.spec.js
LTurns: sounds good, will change now 👍

JoshuaNg2332 (dismissed):

packages/components/pages/f-checkout/test-utils/component-objects/f-checkout.component.js
```diff
@@ -135,6 +135,14 @@ module.exports = class Checkout extends Page {
             get error () { return $(FIELDS.addressAdministrativeArea.error); }
         }
     }
+
+    get customerInput () { return this.customerInfo || 'customer info has not been set'; }
+
+    set expectedCustomerInfo (info) {
+        this.customerInfo = info;
+    }
```
SachaZvetelman: There are a few things here that are confusing to me:

1. You have a `getter` called `customerInput` but it returns an `object` (most of the times 😬). In the lines above, when we're talking about inputs, we are referring to an input field (e.g. `text`).
2. Your getter returns an `object` (if you have one), otherwise a `String`. Normally you should try to return the same type in a function/method/property. So, instead, I'd return an empty `object`. The problem, otherwise, is that the consumers of this `getter` need to worry about handling different types of results. Makes sense? 😄
3. You can (and normally should, I believe) use the same name for the `getter` and the `setter`.

JoshuaNg2332 (approved):

FayeCarter (approved):

oliversweb (approved):

PR #1624, opened 2022-01-27
f-contact-preferences@0.14.0 - Amending styles to match the designs.

SachaZvetelman:
v0.14.0
*January 26, 2022*

### Changed
- Styles to match the designs.

_Note: the main heading is smaller than requested. This is known and it's because fozzie's heading sizes are not exactly matching the designs provided. There's an ongoing conversation about this with the design team and should be amended in fozzie shortly._

**Images:**
image

image

### DIFF
--- a/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
+++ b/packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
@@ -324,11 +324,6 @@ export default {
     margin: spacing(x2) 0 spacing(x4);
 }
 
-.c-formField-label,
-.c-formField-label-description {
-    margin-bottom: 0;
-}
-
 .c-contactPreferences-subtitle {
     @include font-size(heading-s);
 }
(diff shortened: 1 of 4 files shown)


packages/components/pages/f-contact-preferences/src/components/ContactPreferences.vue
```diff
@@ -313,21 +315,18 @@ export default {
...
 .c-contactPreferences-fieldset {
     display: flex;
     flex-flow: column;
     border: none;
     padding: 0;
-    margin: spacing(x2) 0;
+    margin: spacing(x2) 0 spacing(x4);
+}
+
+.c-formField-label,
```
ashleynolan: Won't moving these into the `module` block mean that it'll get scoped and not apply to the `formField` component?

SachaZvetelman: Good spot. It's actually no longer needed, though. Removing it :)

oliversweb (approved):

ashleynolan (approved):

PR #1629, opened 2022-01-28
f-image-tile@v0.5.0 - Keyboard navigation

AlixRichards:
---

### Added
- Focus styles for keyboard navigation

---

This PR adds focus styles and tabindex for keyboard navigation

## UI Review Checks
Screenshot 2022-01-28 at 12 07 24
Screenshot 2022-01-28 at 12 06 44

## Browsers Tested

- [X] Chrome (latest)
- [X] Firefox
- [X] Safari

### DIFF
--- a/packages/components/atoms/f-image-tile/src/components/ImageTile.vue
+++ b/packages/components/atoms/f-image-tile/src/components/ImageTile.vue
@@ -202,7 +202,7 @@ $image-tile-text-transform: translate3d(5px, 0, 0);
     pointer-events: none;
 
     &:focus {
-        outline: none;
+        outline: none; // Focus styles not needed in toggle state.
     }
 }
 
@@ -211,7 +211,7 @@ $image-tile-text-transform: translate3d(5px, 0, 0);
     flex-flow: column wrap;
 
     &:focus {
-        outline: none;
+        outline: none; // Prevents Safari doubling focus styles.
     }
 }
 
--- a/packages/components/atoms/f-image-tile/src/components/ImageTile.vue
+++ b/packages/components/atoms/f-image-tile/src/components/ImageTile.vue
@@ -24,11 +24,13 @@
             class="is-visuallyHidden"
             :class="$style['c-imageTile-checkbox']"
             data-test-id="image-tile-input"
+            :tabindex="!isLink ? 0 : false"
             @change="toggleFilter">
         <label
             :class="$style['c-imageTile-label']"
             :for="`imageTileToggle-${tileId}`"
-            data-test-id="image-tile-label">
+            data-test-id="image-tile-label"
+            :tabindex="!isLink ? -1 : false">
             <span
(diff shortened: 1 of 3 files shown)


packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -179,17 +190,29 @@ $image-tile-text-transform: translate3d(5px, 0, 0);
...
 
 .c-imageTile-link--toggle {
     display: block;
     position: static;
     pointer-events: none;
+
+    &:focus {
+        outline: none;
+    }
 }
```
AlixRichards, line 207: This component has two states - link and toggle. When in toggle state I am removing focus styles for the anchor. Focus styles are added for the toggle state, they are just not needed here.

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -179,17 +190,29 @@ $image-tile-text-transform: translate3d(5px, 0, 0);
...
 
 .c-imageTile-label {
     display: flex;
     flex-flow: column wrap;
+
+    &:focus {
+        outline: none;
+    }
 }
```
AlixRichards, line 216: Removing outline from the label to prevent Safari doubling focus styles.

sonnyprince (dismissed):

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
1itvinka: it might be worth adding this comment to code, because it's not obvious why this code was added.

1itvinka: and this information might be useful if someone is refactoring this component

AlixRichards: Sure! I can add comments to both ✨

1itvinka: thanks!

1itvinka (approved):

packages/components/atoms/f-image-tile/src/components/ImageTile.vue
```diff
@@ -24,11 +24,13 @@
             class="is-visuallyHidden"
             :class="$style['c-imageTile-checkbox']"
             data-test-id="image-tile-input"
+            :tabindex="!isLink ? 0 : false"
```
kevinrodrigues, line 27: I thought tabindex can only have integers as values?

AlixRichards: Yes that's correct 👍 
With the above if `isLink` is true the `tabindex` attribute won't be output.

dandel10n (approved):

PR #1634, opened 2022-02-01
f-form-field@4.7.0 - postcss inline svg

dandel10n:
### Added
- `postcss-assets` plugin to enable inline assets injection for css background images.

There is one issue still left to figure out: I had to copy assets to storybook to make it work there. I couldn't find the right way for storybook to use the assets from component's folders. The decision to make is either to accept this as there shouldn't be many assets like this or open another issue to find the solution as this doesn't block postcss-assets plugin to be enabled.


packages/components/atoms/f-form-field/src/assets/img/icons/check--grey.svg
```diff
@@ -0,0 +1,3 @@
+<svg fill="none" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 8">
```
ashleynolan, line 1: Not necessarily for this PR, but as an improvement – would there be a way to add the file to f-icons and pull it in from the `node_modules` folder? All the icons in f-icons should be available as part of the dist folder in that package.

packages/storybook/src/assets/img/icons/check--grey.svg
```diff
@@ -0,0 +1,3 @@
+<svg fill="none" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 10 8">
```
ashleynolan, line 1: As mentioned above – I think if you can add the icon to f-icons (or if there's n icon in there already that matches this one) then that would save you needing to duplicate the icon to storybook

ashleynolan (dismissed):
Looks good – one improvement to hook into f-icons, but that can be a separate ticket

packages/components/atoms/f-form-field/src/assets/img/icons/check--grey.svg
dandel10n: I wasn't sure about adding whole f-icon package for 2 small icons. And for most of icons we can change colour via css and in this case I added 2 same tick icons with different fill colour because it is used as a css background image.

ashleynolan: The whole icon-set shouldn't get bundled – only the ones you actually bundle inline. 

You might be able to use CSS filters to change the colour of solid icons like this too – imagine it would be ok for grey and white like this.

SachaZvetelman (dismissed):
LGTM, and fixes an annoying issue we had 🥳

SachaZvetelman (approved):

packages/components/atoms/f-form-field/src/assets/img/icons/check--grey.svg
dandel10n: Nice, I will create a separate ticket for this :+1:

ashleynolan (approved):

PR #1639, opened 2022-02-01
fozzie-components@v6.1.0 – Updating to ESLint v8 #globalconfig

ashleynolan:
### Changed
- Updated repo to use ESLint v8
- wdio plugin added to ESLint

### Removed
- Reliance on vue-cli for JS linting as meant we were tied to the version it supported (which is currently v6).

---

## UI Review Checks

- [x] Unit tests have been |updated


packages/services/f-test-data-creator/src/test-data-generator.js
```diff
@@ -1,14 +1,13 @@
...
 
 module.exports = class TestDataGenerator {
-    get firstName () { return 'John'; }
+    static get firstName () { return 'John'; }
 
-    get lastName () { return 'Testington'; }
+    static get lastName () { return 'Testington'; }
 
-    get emailAddress () { return this._emailAddress; }
+    static get emailAddress () { return this._emailAddress; }
```
DamianMullins: Does it matter that the getter is static and the setter isn't?

DamianMullins: Also does referencing `this`cause any issues?

packages/services/f-globalisation/src/mixins/globalisation.mixin.vue
```diff
@@ -1,5 +1,6 @@
 <script>
 export default {
+    name: 'GlobalisationMixin',
```
DamianMullins, line 3: Is this a new rule that means we have to add a name to mixins?

packages/components/atoms/f-image-tile/stories/TestContainer.vue
```diff
@@ -16,7 +16,7 @@
 import ImageTile from '../src/components/ImageTile.vue';
 
 export default {
-    name: 'Container',
+    name: 'TestContainer',
```
DamianMullins, line 19: The `name` property is already generated automatically, do we need a rule to enforce adding the property manually?

packages/components/organisms/f-footer/test-utils/component-objects/f-footer.component.js
```diff
@@ -14,36 +14,36 @@ module.exports = class Footer extends Page {
...
+    static get countryLink () { return this.countryValue != null ? this.countryValue : 'Please set a country value'; }
 
     set expectedDownloadIcon (icon) {
-        this.downloadIconValue = this.downloadIcons.filter(element => element.getAttribute('data-test-id').includes(icon))[0];
+        [this.downloadIconValue] = this.downloadIcons.filter(element => element.getAttribute('data-test-id').includes(icon));
     }
 
     set expectedSocialIcon (icon) {
-        this.socialIconValue = this.socialIcons.filter(element => element.getAttribute('data-test-id').includes(icon))[0];
+        [this.socialIconValue] = this.socialIcons.filter(element => element.getAttribute('data-test-id').includes(icon));
```
DamianMullins, line 42: I did not know you could assign instance properties inside a destructure like this, nice.

packages/components/pages/f-offers/stories/Offers.stories.js
```diff
@@ -60,7 +60,7 @@ export const VOffersComponent = (args, { argTypes }) => ({
     }),
 
     watch: {
-        offersState: function (val) {
+        offersState: function (val) { // eslint-disable-line func-names
```
DamianMullins, line 63:
```suggestion
        offersState (val) {
```

Does the function shorthand syntax work here?

packages/services/f-braze-adapter/src/services/_tests/BrazeConsumer.test.js
```diff
@@ -71,18 +71,18 @@ describe('BrazeConsumer', () => {
...
     // Arrange
-    function instantiateBrazeConsumer () {
-        const testOptions = {
-            ...mockConsumerOptions,
-            [parameter]: value[Object.keys(value)[0]]
-        };
-        // eslint-disable-next-line no-unused-vars
-        const consumer = new BrazeConsumer(testOptions);
-    }
+        function instantiateBrazeConsumer () {
```
DamianMullins, line 74: Minor thing but the comment above looks like it should also be aligned with the code?

packages/services/f-test-data-creator/src/test-data-generator.js
ashleynolan: I think this one can be removed (I did a big find and replace, so any getter's with references to this. shouldn't need static added to them).

packages/services/f-globalisation/src/mixins/globalisation.mixin.vue
ashleynolan: Nope – it's basically my quick fix for the lining complaint of having one-word component names. I didn't want to update all the files and references (as there are quite a lot of errors across the repo like this) so I just added names into the exports.

packages/components/atoms/f-image-tile/stories/TestContainer.vue
ashleynolan: Same as above – the linting started picking up on all the one-word components (like this one). This was the safest way I could think of to fix them until people go in and actually update their components to be two word descriptors as is recommended in the vue docs

packages/components/pages/f-offers/stories/Offers.stories.js
ashleynolan: I was a bit nervous changing code too much (mainly as I don't want to break anything across the repo!).

I'll flag this and try it in a separate PR if that sounds ok? (as I'm having problems getting the tests to run on Circle atm with our memory limits)

DamianMullins: Sure, sounds good to me.

packages/services/f-test-data-creator/src/test-data-generator.js
ashleynolan: Just as a side note – I had to get rid of all the static references as it killed all the tests 🥴

PR #1640, opened 2022-02-02
f-restaurant-card@v0.16.0 - Preorder & Collection

jamieomaguire:
v0.16.0
*Februrary 02, 2022*

### Added
- New `RestaurantAvailability` component for displaying pre-order and collection info
- Supporting unit tests
- New blue colour variant to `IconText` class modifiers
- Storybook file for the `RestaurantAvailability` component
### Changed
- Updated f-vue-icons to `3.5.0`
- Use new clock-small icon on ETA
- Directly target the SVG path elements in `IconText` colour modifiers to avoid default fills present in some icons
- Simplified restaurant Storybook selector

Screenshot 2022-02-02 at 13 52 20
Screenshot 2022-02-02 at 13 51 25

### DIFF
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
@@ -77,6 +77,15 @@
                     v-bind="rating" />
             </component>
 
+            <component
+                :is="errorBoundary"
+                v-if="availability"
+                :tier="3">
+                <restaurant-availability
+                    v-bind="availability"
+                    data-test-id="restaurant-availability" />
+            </component>
+
             <!-- Offline Icon -->
 
             <!-- Meta Items List -->
@@ -147,6 +156,7 @@ import RestaurantRating from './subcomponents/RestaurantRating/RestaurantRating.
 import DeliveryTimeMeta from './subcomponents/DeliveryTimeMeta/DeliveryTimeMeta.vue';
 import IconText from './subcomponents/IconText.vue';
 import RestaurantFees from './subcomponents/RestaurantFees/RestaurantFees.vue';
+import RestaurantAvailability from './subcomponents/RestaurantAvailability/RestaurantAvailability.vue';
 import RenderlessSlotWrapper from './RenderlessSlotWrapper';
 
 export default {
@@ -164,6 +174,7 @@ export default {
         OfferIcon,
         LegendIcon,
         RestaurantFees,
+        RestaurantAvailability,
         RenderlessSlotWrapper
     },
     mixins: [ErrorBoundaryMixin],
@@ -237,6 +248,10 @@ export default {
         fees: {
             type: Object,
             default: () => {}
+        },
+        availability: {
+            type: Object,
+            default: null
         }
     },
     computed: {
--- a/packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
@@ -86,6 +86,10 @@ export default {
     color: $color-green;
 }
 
+.c-restaurantCard-iconText-content--blue {
+    color: $color-blue;
+}
+
 .c-restaurantCard-iconText-icon {
     width: spacing(x2);
     height: spacing(x2);
@@ -99,6 +103,14 @@ export default {
 }
 
 .c-restaurantCard-iconText-icon--green {
-    fill: $color-green;
+    path {
+        fill: $color-green;
+    }
+}
+
+.c-restaurantCard-iconText-icon--blue {
+    path {
+        fill: $color-blue;
+    }
 }
 </style>
--- a/packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
@@ -82,12 +82,12 @@ export default {
     line-height: spacing(x2);
 }
 
-.c-restaurantCard-iconText-content--green {
-    color: $color-green;
+.c-restaurantCard-iconText-content--colorSupportPositive {
+    color: $color-support-positive;
 }
 
-.c-restaurantCard-iconText-content--blue {
-    color: $color-blue;
+.c-restaurantCard-iconText-content--colorSupportInfo {
+    color: $color-support-info;
 }
 
 .c-restaurantCard-iconText-icon {
@@ -102,15 +102,15 @@ export default {
     }
 }
 
-.c-restaurantCard-iconText-icon--green {
+.c-restaurantCard-iconText-icon--colorSupportPositive {
     path {
-        fill: $color-green;
+        fill: $color-support-positive;
     }
 }
 
-.c-restaurantCard-iconText-icon--blue {
+.c-restaurantCard-iconText-icon--colorSupportInfo {
     path {
-        fill: $color-blue;
+        fill: $color-support-info;
     }
 }
 </style>
--- a/packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
@@ -2,13 +2,13 @@
     <p
         :class="{
             [$style['c-restaurantCard-iconText']]: true,
+            [$style[`c-restaurantCard-iconText--${color}`]]: true,
             [$style['c-restaurantCard-iconText--bold']]: isBold
         }">
         <span
             v-if="showIcon"
             :class="{
                 [$style['c-restaurantCard-iconText-icon']]: true,
-                [$style[`c-restaurantCard-iconText-icon--${color}`]]: true,
                 [$style['c-restaurantCard-iconText-icon--hideOnMidBelow']]: hideIconInTileView
             }"
             data-test-id="icon-text-icon">
@@ -16,8 +16,7 @@
         </span>
         <span
             :class="{
-                [$style['c-restaurantCard-iconText-content']]: true,
-                [$style[`c-restaurantCard-iconText-content--${color}`]]: true
+                [$style['c-restaurantCard-iconText-content']]: true
             }"
             data-test-id="icon-text-visible-text">
             {{ text }}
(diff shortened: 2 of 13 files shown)


yarn.lock
```diff
@@ -2454,6 +2454,13 @@
   dependencies:
     babel-helper-vue-jsx-merge-props "2.0.3"
 
+"@justeat/f-vue-icons@3.4.0":
```
jamieomaguire, line 2457: Not quite sure why it isn't at 3.5.0... looking into this

jamieomaguire: @ashleynolan @dandel10n I've tried a number of times to reinstall yarn but it never appears to update the lock file to `3.5.0`. Is this a concern?

ashleynolan: I think this change likely comes from another package still using this older version, rather than you moving to v3.5.

So I wouldn't worry about it – yarn/lock works in mysterious ways 😆

jamieomaguire: Thank you!

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantAvailability/_tests/RestaurantAvailability.test.js
```diff
@@ -0,0 +1,96 @@
+import { mount } from '@vue/test-utils';
+import { ClockSmallIcon, WalkingSmallIcon } from '@justeat/f-vue-icons';
+import availabilityTypes from '../availabilityTypes';
+import sut from '../RestaurantAvailability.vue';
```
1itvinka, line 4: is it a widely used term? I wonder if just `component` or something like this will be more descriptive?

1itvinka (dismissed):

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantAvailability/_tests/RestaurantAvailability.test.js
jamieomaguire: It stands for `system under test`. It's a bit of an old habit from previous projects when writing test code. I believe it's relatively well known but perhaps not so much in the world of front-end? I am more than happy to rename it to something more widely understood

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
```diff
@@ -237,6 +248,10 @@ export default {
         fees: {
             type: Object,
             default: () => {}
+        },
+        availability: {
+            type: Object,
+            default: null
         }
```
AlixRichards, line 255: fees and availability are both objects but have two different defaults - could availability's default also be default: () => {} or default: () => ({})?

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -86,6 +86,10 @@ export default {
     color: $color-green;
 }
 
+.c-restaurantCard-iconText-content--blue {
+    color: $color-blue;
+}
+
```
AlixRichards: `$color-blue` is a global pie token, catching up with Ash last week he highlighted that we should avoid referencing the global token directly but instead use the alias tokens. Maybe this could be updated to `$color-support-info`

Attached is the pie token link if that's useful - https://unpkg.com/browse/@justeat/pie-design-tokens@1.4.0/dist/jet.scss

Also I would revisit the naming of this `c-restaurantCard-iconText-content--blue` to be more descriptive of what the different state is rather than what it looks like to make it more future proof.  So maybe that's `content--info`* or something on those lines

* looks like this is could be preorder so maybe `content--preorder` is appropriate

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -99,6 +103,14 @@ export default {
...
+    path {
+        fill: $color-green;
+    }
+}
+
+.c-restaurantCard-iconText-icon--blue {
+    path {
+        fill: $color-blue;
+    }
 }
```
AlixRichards, line 107: Same with these color variables and naming conventions

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
jamieomaguire: I'd prefer the default to be null so we can simply not render the component if the entire data object is missing. I'm doing to default most of them to null in a future ticket!

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantAvailability/RestaurantAvailability.vue
```diff
@@ -0,0 +1,106 @@
...
+            validator: isValidAvailabilityType
+        },
+        // The translated version of the availability type for the current tenant i.e. 'Pre-order', 'Preordina'
+        availabilityTranslatedName: {
+            type: String,
+            default: null
+        },
+        // The accompanying message for the availability type. i.e Opening times, delivering from...
+        detailsMessage: {
+            type: String,
```
AlixRichards, line 62: No need to change this but `availabilityMessage` unifies it with the other props :)

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
jamieomaguire: good point - updated!

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -82,8 +82,12 @@ export default {
     line-height: spacing(x2);
 }
 
-.c-restaurantCard-iconText-content--green {
-    color: $color-green;
+.c-restaurantCard-iconText-content--colorSupportPositive {
```
sonnyprince: should / could this class be `c-restaurantCard-iconText--colorSupportPositive`? 
`content` of `iconText` would just be text?
If not do we have an opportunity to reduce dom elms?

sonnyprince: (also same for below)

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
```diff
@@ -98,7 +102,15 @@ export default {
...
 
-.c-restaurantCard-iconText-icon--green {
-    fill: $color-green;
+.c-restaurantCard-iconText-icon--colorSupportPositive {
+    path {
+        fill: $color-support-positive;
+    }
+}
+
+.c-restaurantCard-iconText-icon--colorSupportInfo {
```
sonnyprince: As this is duped for both the icon and the icon text I would simplfy the class and use on both. 

eg. `c-restaurantCard-iconText--colorSupportInfo`.

or atleast combine the classnames do not as much dupe? but this might not be needed due to my first comment

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantAvailability/RestaurantAvailability.vue
```diff
@@ -0,0 +1,106 @@
...
+    &:after {
+        content: '\2022'; // round bullet character
+        color: $color-content-subdued;
+        line-height: 1;
+        margin: 0 spacing(x0.5);
+    }
+}
+
+.c-restaurantCard-availability-detailsMessage {
+    margin-top: 0;
```
sonnyprince: where does the margin come from that needs resetting? 
Should  we be using a modifier to apply to those that need vs applying to all then removing?

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantAvailability/RestaurantAvailability.vue
jamieomaguire: updated for consistency!

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantAvailability/_tests/RestaurantAvailability.test.js
sonnyprince: >is it a widely used term? I wonder if just component or something like this will be more descriptive?

SUT did always confuse me too and lowecase sut feels more like a name that an acronym - maybe, for readability, we should create a ticket to rename all to component (or something similar)

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantAvailability/RestaurantAvailability.vue
jamieomaguire: it was from the `<p>` being used for it - I just switched to a `<span>` and removed the margin override :)

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
jamieomaguire: yeah I like this! just updated

packages/components/molecules/f-restaurant-card/src/components/subcomponents/IconText.vue
jamieomaguire: I've combined the classes and reduced the dupe!

sonnyprince (approved):

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.v1.vue
AlixRichards: Ok cool 👍

AlixRichards (approved):

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantAvailability/_tests/RestaurantAvailability.test.js
jamieomaguire: happy to do that! will fix forward :)

PR #1656, opened 2022-02-07
f-restaurant-card@v0.17.1 - add CSS module tests to Restaurant Images

dawidchar:
Add Unit tests to test the CSS classes (CSS Modules) within the Restaurant Image components - Restaurant image & logo


packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantImage/_tests/RestaurantImage.test.js
```diff
@@ -1,6 +1,13 @@
...
+import { mount, config } from '@vue/test-utils';
 import RestaurantImage from '../RestaurantImage.vue';
 
+// Allows for the testing of CSS module classes
+config.mocks.$style = new Proxy({}, {
+    get (_, name) {
+        return name;
+    }
+});
+
```
jamieomaguire: this seems useful enough to either abstract into a utility for the entire repo, or there may be something already being used. @ashleynolan do you happen to know if testing CSS Module classes has been figured out on any existing components?

I think others would benefit from something like this if not.

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantLogo/_tests/RestaurantLogo.test.js
```diff
@@ -0,0 +1,57 @@
+import { mount, config } from '@vue/test-utils';
+import RestaurantLogo from '../RestaurantLogo.vue';
+
+// Allows for the testing of CSS module classes
+config.mocks.$style = new Proxy({}, {
+    get (_, name) {
+        return name;
+    }
+});
```
jamieomaguire: at the very least I'd write this once for the component repository and import it where needed. Rather than keep rewriting it

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantImage/_tests/RestaurantImage.test.js
jamieomaguire: I'd also probably elaborate a little in the comment as to what this is actually doing as it isn't immediately obvious :)

dawidchar: Makes sense, happy to do that 😄

dawidchar: I've found this 
https://github.com/justeat/fozzie-components/blob/caf032fb10f3dc5475c505cb812799e6b80c82b8/packages/components/molecules/f-media-element/src/components/_tests/MediaElement.test.js#L103-L105

dawidchar: I'm assuming that by making a global change, I'd have to do a version bump and update the change log for the root project.

ashleynolan: Only just seen this (I think it must have got lost in my emails when I was on holiday at the time).

I think this could potentially be useful – may be worth a chat on Slack about it (not sure if it being closed means it's not relevant anymore though?)

jamieomaguire: I think this is still relevant @ashleynolan :)

ashleynolan: What's the best way to pick this back up? Shall we have a hangout to discuss it when you're free later in the week?

PR #1661, opened 2022-02-08
fozzie-components@v7.0.0 – Switch to dart-sass #globalconfig

kevinrodrigues:
### Changed
- **Breaking Change** – repo now uses `dart-sass` instead of `node-sass` for styles.
  There is still an outstanding issue in `vue-jest` and it's support of dart-sass which means for components
  wanting to compile styles in jest (to test classnames for instance), `node-sass` still needs to be included
  as part of that packages `devDepenedencies`. For all other packages, `experimentalCSSCompile: false` has been
  set to turn off jest compiling styles, otherwise the tests won't run as expected.
- Eyeglass has been removed in favour of using `sass-loader`'s node_modules path resolution, as this comes out of the box.

### Removed
- Any references to fozzie `$theme` variables in `common.scss` files, as no longer needed.
- Removed webpack tilda `~` for fozzie scss refrences, as not needed – sass-loader does this automatically now.


packages/components/atoms/f-image-tile/jest.config.js
```diff
@@ -28,7 +28,8 @@ module.exports = {
 
     globals: {
         'vue-jest': {
-            hideStyleWarn: true // We hide style warnings given the first time we run the tests it complains about some styles. The second time the tests are run, the warning disappears. https://github.com/vuejs/vue-jest/issues/178#issuecomment-529175129
+            hideStyleWarn: true, // We hide style warnings given the first time we run the tests it complains about some styles. The second time the tests are run, the warning disappears. https://github.com/vuejs/vue-jest/issues/178#issuecomment-529175129
+            experimentalCSSCompile: false
```
kevinrodrigues, line 32: Need to remove this before merging ideally but without this in place `vue-jest` complains. Bit more info [here](https://github.com/vuejs/vue-jest/issues/137):

ashleynolan: After spending a few days looking at this, I think we'll need to keep these in so that we can update to dart-sass. It doesn't cause an issue for the majority of packages – any that it does cause an issue in can temporarily add node-sass to their devDependencies in order to test style specific features.

package.json
```diff
@@ -57,7 +57,7 @@
     "@babel/preset-env": "7.14.9",
     "@justeat/browserslist-config-fozzie": "1.2.0",
     "@justeat/eslint-config-fozzie": "5.1.0",
-    "@justeat/fozzie": "7.3.0",
+    "@justeat/fozzie": "8.0.0-beta.1",
```
ashleynolan: Will update this to v8 when [this PR gets merged](https://github.com/justeat/fozzie/pull/366)

packages/tools/generator-component/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/generator-component",
   "description": "Generator Component – A generator for Fozzie components",
-  "version": "2.6.0",
+  "version": "2.8.0",
```
dandel10n, line 4: Why 2.8.0 and not 2.7.0?

dandel10n (commented):
One tiny question about version bump, other than that looks good to me

oliversweb (approved):

packages/components/templates/f-template-subnav/jest.config.js
```diff
@@ -28,7 +28,8 @@ module.exports = {
 
     globals: {
         'vue-jest': {
-            hideStyleWarn: true // We hide style warnings given the first time we run the tests it complains about some styles. The second time the tests are run, the warning disappears. https://github.com/vuejs/vue-jest/issues/178#issuecomment-529175129
+            hideStyleWarn: true, // We hide style warnings given the first time we run the tests it complains about some styles. The second time the tests are run, the warning disappears. https://github.com/vuejs/vue-jest/issues/178#issuecomment-529175129
+            experimentalCSSCompile: false // hoping this will be a temporary fix, as tests fail when updating to dart-sass currently with vue-cli
```
DamianMullins, line 32: Do you need to update the changelogs for these types of changes?

packages/tools/generator-component/package.json
ashleynolan: There was a 2.7.0 before, but I forgot to actually bump the version number last time, so I just skipped the version

DamianMullins (approved):

kevinrodrigues (commented):
Thanks alot for picking this up whilst I was away @ashleynolan 🎖️

PR #1671, opened 2022-02-14
f-restaurant-card@v0.20.0 - remove wrapper component and increase image size

jamieomaguire:
v0.20.0
*Februrary 15, 2022*

### Added
- Click handler logic unit test
- A new constants JS file to house the custom event key (and any more in the future)

### Changed
- Deleted the `RestaurantCard` wrapper component previously wrapping the `RestaurantCard.v1` component
- Renamed `RestaurantCard.v1` component to `RestaurantCard`
- Updated storybook with flattend props object
- Updated click handler logic to emit an object containing the restaurant ID
- Background image now takes up a maximum of 30% width, up from 20%

Screenshot 2022-02-14 at 14 45 33


packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
```diff
@@ -1,45 +1,374 @@
...
-    </component>
+    <a
+        :href="url"
+        :class="[
+            $style['c-restaurantCard'], {
+                [$style['c-restaurantCard--listItem']]: isListItem,
+                [$style['c-restaurantCard--hasImg']]: !!imgUrl
+            }]"
+        data-test-id="restaurantCard-component"
+        @click="handleClick">
```
AlixRichards, line 10: Although there is a bit of a mixture, convention in fozzie looks to mostly be kebab case for test ids so might want to change this to `data-test-id="restaurant-card-component"`

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
```diff
@@ -1,45 +1,374 @@
...
+
+        </restaurant-image>
+
+        <!-- primary content -->
+        <div :class="$style['c-restaurantCard-content']">
+            <!-- Restaurant Name -->
+            <h3
+                :class="$style['c-restaurantCard-name']"
+                data-test-id="restaurant_name"
+                data-search-name>
```
AlixRichards, line 40: Might want to update this one too `data-test-id="restaurant-card-name"`

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
```diff
@@ -1,45 +1,374 @@
...
+
+            <!-- Meta Items List -->
+            <component
+                :is="errorBoundary"
+                v-if="displayDeliveryTimeMeta"
+                :tier="3">
+                <delivery-time-meta
+                    v-bind="deliveryTimeData"
+                    data-test-id="restaurant-delivery-time-meta" />
+            </component>
```
AlixRichards, line 96: Might want to change this to `data-test-id="delivery-time-meta"` aso it matches the component name and follows your other component test ids

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
```diff
@@ -1,45 +1,374 @@
...
+                    :class="$style['c-restaurantCard-tags']"
+                    test-id-position="inner-content"
+                    :tags="tags.contentTags" />
+            </component>
+
+            <!-- Offers -->
+            <icon-text
+                v-if="hasOffer"
+                data-test-id="restaurant-offer"
+                :text="offer"
```
AlixRichards, line 123: Same here

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
```diff
@@ -1,45 +1,374 @@
...
+            return !!this.tags?.imageTags?.length;
+        },
+        hasOffer () {
+            return !!this.offer?.length;
+        },
+        displayDeliveryTimeMeta () {
+            return this.deliveryTimeData.eta ||
+                this.deliveryTimeData.distance ||
+                this.deliveryTimeData.address;
+        },
```
AlixRichards, line 282: Could it cause an error/problem if none of these existed?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
```diff
@@ -1,45 +1,374 @@
...
-            v-for="(_, slotName) in $slots"
-            v-slot:[slotName]>
-            <slot :name="slotName" />
-        </template>
-    </component>
+    <a
+        :href="url"
+        :class="[
+            $style['c-restaurantCard'], {
+                [$style['c-restaurantCard--listItem']]: isListItem,
```
AlixRichards, line 6: I wonder if maybe this class should be `c-restaurantCard--item` or another name? For me there's a bit of a dissonance in using list item as a name when the element is `<a>` rather than a `<li>`

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
jamieomaguire: Absolutely - the next PR is to sync the test IDs between searchweb and fozzie so I'll be updating all of them in that change :)

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
jamieomaguire: it wouldn't error - although the logic could be tighter as it is treating the result as truthy/falsey. I'll update this to tighten it up!

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
jamieomaguire: yeah it's a tricky one - this is for when they are on desktop, but I'd like to avoid using the name 'desktop'. list item is certainly confusing, and I think 'item' might be equally ambiguous. I'll have a think on a better name for this and update

ashleynolan: What's the use-case for the modifier? Is it that it appears in the SERP listings?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
ashleynolan: Yeh, data attributes are case insensitive, so should always be lowercase (which is why it's the only place I think we don't use camelCase)

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
jamieomaguire: Screenshot 2022-02-15 at 11 57 12


@ashleynolan the `listItem` modifier is the term we dubbed for the desktop style listing you can see in the image attached

jamieomaguire: (the one on the right)

ashleynolan: Hmm, naming is hard 😆 

I think as long as it makes sense to you all, that's the main thing tbh. Could use something like `c-restaurantCard--stretchPanel` or something equally vague to describe it 🤷

jamieomaguire: Indeed it is 😆 ok cool - as @AlixRichards rightly pointed out, listItem is not the most understandable term. I'll see what she thinks of stretchPanel and will happily change if so!

AlixRichards: `stretchPanel` is potentially cool, I'm still unclear what the modifier does even seeing the image 🤔 
Can you remind me what this looks like on desktop without the modifier 🤞

jamieomaguire: @AlixRichards the modifier makes it look like the right hand image above. Otherwise it's just the left hand image but full width :)

dawidchar (dismissed):

dawidchar (dismissed):

dawidchar (approved):

AlixRichards (approved):

PR #1692, opened 2022-02-21
f-services@v1.14.0 - add Italian validations

FayeCarter:
### Added
- Italian postcode and phone validations


simonsandersje (approved):

packages/services/f-services/tests/validations.test.js
```diff
@@ -300,4 +319,26 @@ describe('isValidPhoneNumber', () => {
...
     });
+
+    it.each([
+        ['+39 06 01234667', true],
+        ['+39 06 0123', false],
+        ['+39 06 01234567890123456', false],
+        ['+39 30 123467', true],
+        ['+39 30 123', false],
+        ['+39 30 123456780123456', false],
+        ['01234567', true],
```
xander-marjoram, line 330: I'm certainly no expert in Italian phone numbers, but this one looks quite short. Where did you get the list of cases from?

FayeCarter: So I couldn't find a list of test phone numbers anywhere, If you happen to know where to find them that would be amazing.
I ended up using the consumer web regex as a guide, (which just says that it needs to be 6 digits minimum) and then the source of at truth, wikipedia 😬

xander-marjoram: As long as we've got analytics on this field I think it should be fine

charliearlie (approved):

PR #1694, opened 2022-02-22
f-restaurant-card@v0.24.0 - style data points layout

jamieomaguire:
v0.24.0
*Februrary 22, 2022*

### Changed
- Styled the positioning of restaurant card data points

## Browsers Tested

- [X] Chrome (latest)
- [ ] Internet Explorer 11
- [X] Mobile IPhone 13

### Desktop 
Screenshot 2022-02-23 at 13 49 59

Screenshot 2022-02-23 at 13 05 03

Screenshot 2022-02-23 at 13 10 36

Screenshot 2022-02-23 at 13 11 04

### Mobile
Screenshot 2022-02-23 at 13 11 37

Screenshot 2022-02-23 at 13 12 50

Screenshot 2022-02-23 at 13 13 24


AlixRichards:
Looks really good! I have a couple of minor points/checks:

1. Using `order` - wanted to double check that there are no negative effects for screen reader users
2. I think you could remove a lot of the css comments. They are super valuable for the clearfix but you might not need the others
3. `only-child` nice I have never used this before so I learnt something!
4. Looks great! Really like the use of grid
5. f-alert - is this used in the restaurant card? Checking why it gets updated here

jamieomaguire:
> Looks really good! I have a couple of minor points/checks:
> 
> 1. Using `order` - wanted to double check that there are no negative effects for screen reader users
> 2. I think you could remove a lot of the css comments. They are super valuable for the clearfix but you might not need the others
> 3. `only-child` nice I have never used this before so I learnt something!
> 4. Looks great! Really like the use of grid
> 5. f-alert - is this used in the restaurant card? Checking why it gets updated here

Thanks @AlixRichards !

All great points - I'll remove some of the extra CSS comments. 

Regarding `order` and screen-reader impact, I don't think it would negatively impact the screen-reader experience. I think either data point (offers and tags) can be announced in either order without degrading the experience. What we might do in the near future is add an `aria-label` to the outermost element that summarises the entire card's data, which will allow us to use proper grammar etc. I am more than happy to be told otherwise by anyone a bit more screen-reader savvy, however! 

f-alert - this seemed to update itself when I ran yarn on the root, strangely. I will revert this change as the update I originally made isn't in use anymore so it can go!

AlixRichards:
Awesome, thanks @jamieomaguire ✨

AlixRichards (approved):

1itvinka (approved):

PR #1695, opened 2022-02-22
f-takeawaypay-activation@v2.10.0 - fixing logging to be on the new logger instance

matthewhardern:
Tests in CoreWeb are failing due to the new `$log` being used and this package not being updated to support this new interface. This PR changes the code and tests to fix this.


packages/components/pages/f-takeawaypay-activation/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-takeawaypay-activation",
   "description": "Fozzie TakeawayPay Activation - Handles Takeaway Pay activation for users",
-  "version": "2.9.0",
+  "version": "2.10.0",
```
minrwhite, line 4: I think this should be a major version bump, as effectively the change in shape of the logger object this component consumes is a backwards compatibility break.

PR #1696, opened 2022-02-22
f-braze-adapter@5.0.0 - updated logWhileInitialising function

dandel10n:
### Changed
- `logWhileInitialising` function and test in line with new log interface.

### Removed
- local logging service.


matthewhardern (dismissed):

packages/services/f-braze-adapter/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-braze-adapter",
   "description": "Fozzie Metadata Service",
-  "version": "4.2.0",
+  "version": "4.3.0",
```
minrwhite: IMO, this should be a major version bump as it is a backwards compatibility break.

matthewhardern (approved):

minrwhite (approved):

PR #1720, opened 2022-03-04
f-account-info@v0.18.0 - Added the save method to the store

oliversweb:
### Added
- Added Save method to vuex store

### DIFF
--- a/packages/components/pages/f-account-info/src/store/accountInfo.module.js
+++ b/packages/components/pages/f-account-info/src/store/accountInfo.module.js
@@ -1,6 +1,7 @@
 import {
     mapToConsumerDetails,
-    mapToConsumerAddress
+    mapToConsumerAddress,
+    mapToConsumerUpdate
 } from '../services/mapping/consumer.mapper';
 import {
     UPDATE_CONSUMER_DETAILS,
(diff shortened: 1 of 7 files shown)


packages/components/pages/f-account-info/src/store/accountInfo.module.js
```diff
@@ -28,6 +29,12 @@ export default {
             commit(UPDATE_CONSUMER_DETAILS, { details, address });
         },
 
+        async saveConsumerDetails ({ state }, { api, authToken }) {
```
oliversweb, line 32: PR main change here

kevinrodrigues: I might have missed some context but does this belong here, it doesn't seem to really save anything to the store from what i can see and looks like just an api call which should probably go in a service instead, again I could have missed something obvious here.

oliversweb: The template fields are wired up to the store so as each field is changes it changes the store .... but the submit button then the saves the current store to the api

kevinrodrigues (dismissed):

JoshuaNg2332 (dismissed):

PR #1739, opened 2022-03-10
f-filter-pill@v0.3.0 - Style the component

1itvinka:
### Added
- Animation

### Changed
- Renamed SCSS variables

### Fixed
- Outline in Safari

---

How animation looks like:

https://user-images.githubusercontent.com/3179649/157729567-e9d31462-c138-40d3-b312-9ca7e50e2544.mov

Long text:

https://user-images.githubusercontent.com/3179649/157729572-41682d2d-6f1b-4ccc-939c-b6a71844d5c0.mov

---

## Browsers Tested

- [x] Chrome (latest)
- [x] Safari
- [x] Firefox

### DIFF
--- a/packages/components/atoms/f-filter-pill/CHANGELOG.md
+++ b/packages/components/atoms/f-filter-pill/CHANGELOG.md
@@ -3,6 +3,16 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.3.0
+------------------------------
+*March 10, 2022*
+
+### Added
+- Animation
+
+### Fixed
+- Outline in Safari
+
 v0.2.0
 ------------------------------
 *February 23, 2022*
--- a/packages/components/atoms/f-filter-pill/CHANGELOG.md
+++ b/packages/components/atoms/f-filter-pill/CHANGELOG.md
@@ -10,6 +10,9 @@ v0.3.0
 ### Added
 - Animation
 
+### Changed
+- Renamed SCSS variables
+
 ### Fixed
 - Outline in Safari
 
--- a/packages/components/atoms/f-filter-pill/CHANGELOG.md
+++ b/packages/components/atoms/f-filter-pill/CHANGELOG.md
@@ -8,7 +8,7 @@ v0.3.0
 *March 10, 2022*
 
 ### Added
-- Animation
+- Animation on hover/select/unselect
 
 ### Changed
 - Renamed SCSS variables
--- a/packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
+++ b/packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
@@ -105,8 +105,10 @@ export default {
 </script>
 
 <style lang="scss" module>
-$filter-pill-selected: $color-content-positive;
-$filter-pill-checkbox-width: 16px;
+$selected-color: $color-content-positive;
+$checkbox-width: 16px;
+$transition-duration: 0.2s;
+$ease: ease-in-out;
 
 @mixin ellipsis() {
     white-space: nowrap;
@@ -115,17 +117,18 @@ $filter-pill-checkbox-width: 16px;
 }
 
 .c-filterPill {
-    outline: 1px solid $color-grey-30;
+    box-shadow: 0 0 0 1px $color-grey-30;
     border-radius: $radius-rounded-e;
     background: $color-white;
+    transition: .1s;
 
     &:focus-within:not(.c-filterPill--disabled) {
-        outline: 2px solid $color-focus;
+        box-shadow: 0 0 0 2px $color-focus;
     }
 }
 
 .c-filterPill--selected {
-    outline-color: $filter-pill-selected;
+    box-shadow: 0 0 0 2px $color-focus;
 }
 
 .c-filterPill-label {
@@ -138,10 +141,11 @@ $filter-pill-checkbox-width: 16px;
 
     &:focus {
         outline: none; // Prevents Safari doubling focus styles.
+        box-shadow: none;
     }
 
     .c-filterPill--selected & {
-        color: $filter-pill-selected;
+        color: $selected-color;
     }
 
     .c-filterPill--disabled & {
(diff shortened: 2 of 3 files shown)


packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
```diff
@@ -115,17 +117,18 @@ $filter-pill-checkbox-width: 16px;
 }
 
 .c-filterPill {
-    outline: 1px solid $color-grey-30;
+    box-shadow: 0 0 0 1px $color-grey-30;
```
1itvinka: the `outline` was changed to `box-shadow` to get rounded corners on the outline in Safari. I'm super happy with this solution, but other solutions are more ugly. 😬

AlixRichards: Nice! I might need to update the imageTile to do the same ✨

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
```diff
@@ -105,8 +105,10 @@ export default {
 </script>
 
 <style lang="scss" module>
-$filter-pill-selected: $color-content-positive;
-$filter-pill-checkbox-width: 16px;
+$selected-color: $color-content-positive;
+$checkbox-width: 16px;
+$transition-duration: 0.2s;
+$ease: ease-in-out;
```
AlixRichards, line 112: I think its good to prefix these variables with the component name like you had before - maybe we can check the fozzie convention but potentially these could be: `$filter-pill-selected-color`, `$filter-pill-ease` ...

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
AlixRichards: I know this wasn't added in this PR but I missed it previously: `$color-grey-30` is a global color variable which we should avoid using directly. So potentially could update this to `$color-border-default` instead (https://unpkg.com/browse/@justeat/pie-design-tokens@1.4.0/dist/jet.scss)

I think there might be a couple of other globals like `$color-white` which might also want swapped out

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
```diff
@@ -154,17 +158,29 @@ $filter-pill-checkbox-width: 16px;
...
+    transform: translate3d(0, 0, 0) scale(0) rotate(-45deg);
+    will-change: transform, opacity, width, min-width, margin-right;
+
+    @media (prefers-reduced-motion: no-preference) {
+        transition: transform $transition-duration $ease,
+                    opacity $transition-duration $ease,
+                    width $transition-duration $ease,
+                    min-width $transition-duration $ease,
+                    margin-right $transition-duration $ease;
+    }
```
AlixRichards, line 171: I wonder if this can be tweaked a bit. `will-change` needs to be used sparingly as its possible it could cause a performance impact rather than improvement.  Also translate3d() is great because it uses the GPU. Could the margin-right be changed using this instead? 

Wondering if you could then transition only - transform, opacity & width and it look the same 🤞

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
1itvinka: thanks! Will fix it in this PR

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
1itvinka: yeah, I'm not sure if we have the convention. The variables won't leak into other components, so I don't really see much sense in making them longer and harder to read. Happy to change if we have such guidelines.

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
1itvinka: hm. good point! will play a bit more with it

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
AlixRichards: For sure, I've taken a look at the other atoms and a couple of the molecules in Fozzie and they generally follow this pattern but up to you whether you want to change or not 👍

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
```diff
@@ -154,17 +158,29 @@ $filter-pill-checkbox-width: 16px;
     align-self: center;
     opacity: 0;
     width: 0;
+    min-width: 0;
+    transform: translate3d(0, 0, 0) scale(0) rotate(-45deg);
+    will-change: transform, opacity, width, min-width, margin-right;
+
+    @media (prefers-reduced-motion: no-preference) {
```
jamieomaguire, line 164: 🔥

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
```diff
@@ -154,17 +158,29 @@ $filter-pill-checkbox-width: 16px;
     align-self: center;
     opacity: 0;
     width: 0;
+    min-width: 0;
+    transform: translate3d(0, 0, 0) scale(0) rotate(-45deg);
+    will-change: transform, opacity, width, min-width, margin-right;
```
jamieomaguire: This `will-change` property might not be necessary. It sounds as though it's only worth adding as a last resort for known performance issues, however I'm not an expert!

Basically it might be more resource intensive to use this property on lots of elements, by the sounds of it.

[Reference](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change)

packages/components/atoms/f-filter-pill/CHANGELOG.md
```diff
@@ -3,6 +3,19 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.3.0
+------------------------------
+*March 10, 2022*
+
+### Added
+- Animation
```
kevinrodrigues: Could you add a little info more here please.

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
1itvinka: Changed it to follow the pattern. I'm still not sure if consistency is more important than readability, but let's see how it goes, and adjust along the way.

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
1itvinka: Good point! I removed it for now. Let's see if we have any performance along the way.

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
1itvinka: I removed `will-change` for now. Let's see if we face any performance issues along the way and if we do we'll see how to refactor it. 

We need `min-width` because otherwise when the text is long, the icon will shrink. 

I don't see a clean readable solution on how to get rid of `margin-right`. If the icon was always visible we could use flex gap. The other solution could be to add `translate3d()` to the text element. But the code will be less maintainable, as the animation part will be spread across different elements. Happy to do it if we see that it's beneficial in terms of performance.

jamieomaguire (approved):

kevinrodrigues (approved):

PR #1740, opened 2022-03-11
f-header@v9.7.0 - update component to match latest design

LTurns:
### Changed
- component to match latest designs

| Current | New |
| --- | ----------- |
| image| image|
| - | <ul><li>Header container height updated to 80px</li><li>Logo colour changed to `$color-support-brand-01` </li> <li>Flag size increased to 21px</li> <li>Small padding and margin adjustments in country selector panel and horizontal nav</li></ul> |

| Current | New |
| --- | ----------- |
|| |
| - | <ul><li>Updated left side margin to `56px` with all text inline (including account info)</li><li>Moved position of border to under text only</li> <li>Icon centred between text and edge of screen</li> <li>border-top changed to `1px` (below logo)</li> <li>all font updated to 1rem</li></ul> |

| Current | New |
| --- | ----------- |
|| |
| - | <ul><li>Updated padding around `select your country`</li><li>Updated arrow & flag size</li> <li>icon and text mirrors margin and padding of previous nav (see above)</li></ul> |

As changes are minor, these will be checked with design via storybook after PR has been merged - any follow up alterations will be done in a separate PR


packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -8,30 +8,30 @@
 */
 
 $nav-text-size                     : 'body-l';
-$nav-text-size--narrow             : 'body-s';
+$nav-text-size--narrow             : 'body-1';
 $nav-text-weight                   : $font-weight-bold;
 $nav-text-color                    : $color-content-link;
 $nav-text-color--hover             : darken($color-content-link, $color-hover-01);
-$nav-text-color--narrow            : $color-content-subdued;
+$nav-text-color--narrow            : $color-interactive-primary;
```
ashleynolan: I think this is the same colour as `$color-content-link` (which is used for the `$nav-text-color`).

If so, you could get rid of the `$nav-text-color--narrow` variable and then remove line 205 (which sets the color) and update line 196 to just point to `$nav-text-color` now that both widths use the same grey.

packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -186,34 +188,62 @@ $nav-toggleIcon-space              : 5px;
...
 
-    @include media('>mid') {
+    @include media('<=mid') {
+        display: flex;
+        float: center;
+        margin-left: spacing(h);
+        padding: spacing(c) spacing(d) spacing(c) 0;
+        color: $nav-text-color;
         @include font-size($nav-text-size);
+        font-weight: 300;
```
ashleynolan: Weirdly, we don't have `300` in the design tokens, so if it's set to that in the designs, may need to ask them if it should be 400 instead (as that's the lightest version we have).

The current design tokens just have:

```scss
$font-weight-bold: 700;
$font-weight-extrabold: 800;
$font-weight-regular: 400;
```

Just as a general design rule – if a value like this isn't [in the design tokens](https://unpkg.com/@justeat/pie-design-tokens/dist/jet.scss), then it is either a design mistake or the design tokens need updating after speaking to the design team.

packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -186,34 +188,62 @@ $nav-toggleIcon-space              : 5px;
...
         font-weight: $nav-text-weight;
-        color: $nav-text-color;
-        border-bottom: none;
-        height: $header-height;
+        margin: 0;
+        padding: 0;
     }
 }
 
+// TODO: MAKE THIS NOT USE FLOATS
```
ashleynolan: Did you want to fix this before merging? Happy either way (can help you with it once I'm back, or just ask someone else on the team for help with it if you need it)

packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -8,30 +8,30 @@
 */
 
 $nav-text-size                     : 'body-l';
-$nav-text-size--narrow             : 'body-s';
+$nav-text-size--narrow             : 'body-l';
```
ashleynolan: If this is the same now on narrow and wide, you could just get rid of `$nav-text-size--narrow` and delete the extra reference in the selector then as well

packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -186,34 +188,62 @@ $nav-toggleIcon-space              : 5px;
...
+.c-nav-list-link--country-selector {
+    @include media('>mid') {
+        text-decoration: none;
+        display: flex;
+        float: center;
+        margin-left: spacing(h);
+        padding: spacing(c) spacing(d) spacing(c) 0;
+        color: $nav-text-color;
+        @include font-size($nav-text-size);
+        font-weight: 300;
```
ashleynolan: Same here as above – we don't have 300 set in the design tokens, so this may be an issue with the designs.

packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -133,22 +133,24 @@ $nav-toggleIcon-space              : 5px;
...
 .c-nav-list-icon--flag {
     height: $countrySelector-flag-height;
     width: $countrySelector-flag-width;
-    margin-right: spacing();
+    display: flex;
+    float: left;
+    margin: spacing(c) spacing(d);
+}
+
+.c-nav-list-icon--feature-flag {
```
ashleynolan: I'm not sure the name of this selector describes it that well atm (I'm not sure what the difference between `c-nav-list-icon--feature-flag` and `c-nav-list-icon--flag`).

Doees this just apply in certain circumstances?

packages/components/organisms/f-header/src/components/CountrySelector.vue
```diff
@@ -7,33 +7,30 @@
         ]"
         v-on="isBelowMid ? null : { mouseover: openCountrySelector, mouseleave: (() => closeCountrySelector(false)) }"
         @keyup.esc="closeCountrySelector">
+        <flag-icon
+            data-test-id="current-flag-icon"
+            :country-code="copy.currentCountryKey"
+            :class="[
+                { [$style['c-nav-list-icon--feature-flag']]: !isBelowMid },
+                { [$style['c-nav-list-icon--flag']]: isBelowMid }]" />
```
ashleynolan: Would be good to understand what you're doing with these classes. It feels looking at how it's applied that you could potentially use media queries to adjust the style between `isBelowMid` and `!isBelowMid` rather than applying difference classnames, but would be good to understand what different styling this is applying

packages/components/organisms/f-header/src/components/NavLink.vue
```diff
@@ -1,18 +1,24 @@
...
+            name="icon" />
+        <a
+            v-bind="$attrs"
+            :data-trak="JSON.stringify(dataTrak)"
+            :href="href"
+            :class="[
+                $style['c-nav-list-link'],
+                { [$style['c-nav-list-link--country-selector']]: isCountrySelectorPanel & !isBelowMid },
+                { [$style['c-nav-list-link--alt']]: isAltColour },
+                { [$style['c-nav-list-link-with-border']]: hasBorderBottom & isBelowMid },
```
ashleynolan: To fit the CSS naming scheme, this would be `c-nav-list-link-withBorder` (so camelCase).

packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
```diff
@@ -34,23 +37,26 @@
...
-                        <flag-icon
-                            :country-code="country.flagKey"
-                            :class="$style['c-nav-list-icon--flag']" />
-                        <span>
-                            {{ country.localisedName }}
-                        </span>
-                    </a>
+                        :has-border-bottom="false"
+                        :is-below-mid="isBelowMid"
+                        :is-country-selector-panel="true"
```
ashleynolan: Rather than use a prop for this (which adds a classname), could you not add a class to this component `c-countrySelector-country-link`, that you then add styles for here in this component? (I think that should work with CSS modules, but may need to try it)

packages/components/organisms/f-header/src/components/UserNavigationPanel.vue
```diff
@@ -89,31 +90,22 @@ export default {
...
-    height: auto;
-    @include font-size('body-s');
-    text-decoration: none;
-    border-bottom: 1px solid $color-border-default;
+    @include media('>mid') {
+        display: block;
+        padding: spacing(c) spacing(d);
+        margin: 0;
+        font-family: $font-family-base;
+        font-weight: 300;
```
ashleynolan: Same as above – 300 isn't a font-weight in the design tokens currently

ashleynolan (dismissed):

packages/components/organisms/f-header/src/assets/scss/navigation.scss
LTurns: ah I copied this note over with the code but didn't write it, happy to have a go at fixing though - think there are quite a few floats so could be worth doing separately once the PR's been merged 😄

packages/components/organisms/f-header/src/assets/scss/navigation.scss
LTurns: just checked with design and font-weight is at `400` actually so have removed all reference to `300` and changed to `$font-weight-regular` 👍

packages/components/organisms/f-header/src/assets/scss/navigation.scss
LTurns: the `c-nav-list--icon-feature-flag` references the current flag which opens the popover - agree it's not very clear though so I've changed it to `c-nav-icon-current-flag`

packages/components/organisms/f-header/src/components/CountrySelector.vue
LTurns: Agree, turns out there was no need to add the additional props - actually only ended up needed the `c-nav-icon-current-flag` class after a bit of refactoring 😄

packages/components/organisms/f-header/src/components/NavLink.vue
LTurns: fixed

packages/components/organisms/f-header/src/components/UserNavigationPanel.vue
LTurns: fixed 👍 have noticed that `font-weight: 300` also exists in f-checkout so can check with designs and add a small ticket to change that to `font-weight-regular`

packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
LTurns: Yes, that's fixed it thank you - I've created `c-countrySelector-country-link a` so it's declared here but only styles the link inside the `NavLink` component 😄

packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -186,34 +200,67 @@ $nav-toggleIcon-space              : 5px;
...
-        height: $header-height;
+        margin: 0;
+        padding: 0;
     }
 }
 
+.c-nav-list-item--horizontallyAlignedAboveMid {
+    @include media('>mid') {
+        float: left;
+        padding: 28px spacing(d);
```
1itvinka: not sure if we'd like to be consistent using spacing variables vs actual values?

packages/components/organisms/f-header/src/components/Header.vue
```diff
@@ -264,13 +264,13 @@ html:global(.is-navInView) {
         background-color: $color-support-brand-01;
 
         @include media('>mid') {
-            min-height: 88px;
+            min-height: 80px;
```
1itvinka: might be worth moving it to the variable

LTurns: thanks @1itvinka , using the `$header-height` variable here now instead 👍

packages/components/organisms/f-header/src/assets/scss/navigation.scss
LTurns: that makes sense - I've changed it to use actual values in this instance 👍

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -692,10 +684,13 @@ export default {
 }
 
 .c-nav-featureLink {
-    display: block;
     width: $nav-featureLinkIcon-width;
     height: $nav-featureLinkIcon-height;
 
+    & path {
+            fill: $nav-icon-color;
+        }
```
SachaZvetelman: Nit: indentation looks odd here.

SachaZvetelman (dismissed):

packages/components/organisms/f-header/src/assets/scss/navigation.scss
1itvinka: hm. I wonder if `padding: spacing(e) spacing(d);` will still look good?

packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -186,34 +202,68 @@ $nav-toggleIcon-space              : 5px;
...
-    padding: spacing(c) spacing(d);
-    margin: 0;
-    font-family: $nav-text-subFont;
-    color: $nav-text-color--narrow;
-    @include font-size($nav-text-size--narrow);
-    font-weight: 300;
+.c-nav-list-link {
     text-decoration: none;
-    border-bottom: 1px solid $color-border-default;
+    font-family: $nav-text-subFont;
```
dandel10n: I wonder if this font declaration was left after moving to using one default font? Could you please try to remove it and see if it is necessary here as if not we can remove the `$nav-text-subFont` var and this line as well

packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -186,34 +202,68 @@ $nav-toggleIcon-space              : 5px;
...
+    color: $nav-text-color;
+    font-weight: $font-weight-regular;
```
dandel10n: Same for font-weight, it should be regular by default so worth checking if we can just remove this line

packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
```diff
@@ -34,23 +37,23 @@
...
-                    <a
-                        :data-trak='`{
-                            "trakEvent": "click",
-                            "category": "engagement",
-                            "action": "header",
-                            "label": "${country.gtm}"
-                        }`'
+                    <nav-link
+                        :data-trak="{
+                            trakEvent: 'click',
```
dandel10n, line 42: I remember there were issues with this syntax before so could you please double check that tracking works as it should.

packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
```diff
@@ -114,9 +119,18 @@ export default {
...
+        padding-bottom: spacing(c);
+    }
+}
+
 .c-countrySelector {
     @include media('>mid') {
-        padding: spacing(e) 0;
+        padding-top: spacing(e);
+        padding-bottom: spacing(c);
+        width: 416px;
```
dandel10n: Normally we don't add exact width for containers like this one. It wasn't specified before so I am not sure if we need it here. Width of a grid of content + spacing around should be enough

packages/components/organisms/f-header/src/components/UserNavigationPanel.vue
```diff
@@ -89,31 +91,25 @@ export default {
...
-    text-decoration: none;
-    border-bottom: 1px solid $color-border-default;
-
-    &:hover,
-    &:focus {
-        font-weight: $font-weight-bold;
-        text-decoration: none;
-        color: $color-content-subdued;
-    }
+    font-weight: $font-weight-regular;
```
dandel10n: Same here, it should be regular by default so probable there is no need in this declaration

packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -186,34 +202,68 @@ $nav-toggleIcon-space              : 5px;
...
+
+.c-nav-list-item--horizontallyAlignedAboveMid {
+    @include media('>mid') {
+        float: left;
+        padding: 28px 16px;
+
+    }
+}
+
+.c-nav-list-link-withBorderTop a {
```
dandel10n: Here and on line 239 and 252 you target an `a` tag as a child component of wrapper with a classname. Can we add a class straing to the `a` tag in this case?

dandel10n (commented):
Looks good, I left few comments. There are a lot of small changes and code restructure so it is hard to spot everything just by looking at the code diff. I can run your branch locally tomorrow to have a better review or we can do it on a call together if it helps?

packages/components/organisms/f-header/src/assets/scss/navigation.scss
LTurns: ah yes, this one isn't needed anymore - have removed 🎉

packages/components/organisms/f-header/src/assets/scss/navigation.scss
LTurns: yep can also be removed

packages/components/organisms/f-header/src/components/UserNavigationPanel.vue
LTurns: has been removed too

packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
LTurns: have removed and altered spacing around instead

packages/components/organisms/f-header/src/assets/scss/navigation.scss
LTurns: The best work around I have for not adding the `a` tag to the class name is adding a prop into the child component - this seems to be the best way of specifying whether the nav links require a 'border-bottom` or not - for country selector they are not needed, but they are for the main menu in mobile mode - let me know what you think on this :)

packages/components/organisms/f-header/src/assets/scss/navigation.scss
LTurns: it looks _slightly_ too high with `spacing(e)` and too low with `spacing(f)` 🤦 The design is `28px` so it's central in the navbar - so may have to keep it as actual values for the meantime

packages/components/organisms/f-header/src/components/CountrySelectorPanel.vue
LTurns: we have reviewed this and can see that the `data-trak` here matches the `data-trak` in production, so changes can be left 👍

dandel10n (approved):

SachaZvetelman (approved):

PR #1707, opened 2022-02-25
f-development-context@1.2.0 - new log interface signature

dandel10n:
### Changed
- $logger interface signature to $log.

PR #1743, opened 2022-03-11
f-development-context@1.2.0 - new log interface signature - REVERT

oliversweb:
Reverts justeat/fozzie-components#1707

Will be corrected & re-applied in a new pr 👍

### DIFF
--- a/packages/storybook/package.json
+++ b/packages/storybook/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/storybook",
   "private":"true",
-  "version": "0.50.1",
+  "version": "0.50.0",
   "scripts": {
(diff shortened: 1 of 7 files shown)


packages/storybook/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/storybook",
   "private":"true",
-  "version": "0.50.1",
+  "version": "0.50.0",
```
kevinrodrigues, line 4: Should this not go to the next version as a fix?

oliversweb: I think we are undoing that revision

kevinrodrigues (approved):

PR #1745, opened 2022-03-14
#trivial - attempt to fix broken build.

kevinrodrigues:
Not sure if the syntax / path is correct here, also not sure how to test this locally.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ commands:
             }
 
             echo 'Checking changes'
-            changes=`git diff --name-only origin/master...$CIRCLE_BRANCH | { grep -Ev '^packages/|yarn.lock|bear.png|.editorconfig' || true; }`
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -32,7 +32,7 @@ commands:
             }
 
             echo 'Checking changes'
-            changes=`git diff --name-only origin/master...$CIRCLE_BRANCH | { grep -Ev '^packages/|yarn.lock|bear.png|.editorconfig' || true; }`
+            changes=`git diff --name-only origin/master...$CIRCLE_BRANCH | { grep -Ev '^packages/|yarn.lock|bear.png|.editorconfig|services/f-development-context' || true; }`
```
kevinrodrigues, line 35: path wise i think it would work without the `services` specified but unsure.

matthewhardern: Not sure if I am correct but I think this is saying if a change does not occur in that grep then apply global change, so I think you want to exclude the `services/f-development-context`. I might be wrong though

kevinrodrigues:
Closing this PR for now.

PR #1746, opened 2022-03-15
f-filter-pill@v0.4.0 - Add `a` tag to the component for the SEO reasons

1itvinka:
The presence of a URL is currently required by SearchWeb SEO to be consumable by web crawlers.

No visual changes.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [x] Safari
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
+++ b/packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
@@ -8,7 +8,7 @@
         data-test-id="filter-item">
         <a
             :class="$style['c-filterPill-link']"
-            :href="toggleValue"
+            :href="href"
             tabindex="-1"
             aria-hidden="true"
             data-test-id="filter-pill-link">
@@ -60,6 +60,10 @@ export default {
             default: null
         },
         toggleValue: {
+            type: String,
+            default: null
+        },
+        href: {
             type: String,
             default: null
         },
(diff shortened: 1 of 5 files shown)


packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
```diff
@@ -6,6 +6,16 @@
                 [$style['c-filterPill--disabled']]: isToggleDisabled
             }]"
         data-test-id="filter-item">
+        <a
+            :class="$style['c-filterPill-link']"
+            :href="href"
+            tabindex="-1"
+            aria-hidden="true"
```
jamieomaguire, line 13: Do we want this hidden if there's no-js on the page? That's when this should essentially be behaving like a link, I believe.

Might be worth doubling checking if @AlixRichards knows how this is handled in the image tile :)

jamieomaguire (approved):

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
1itvinka: I created a ticket to address this https://justeattakeaway.atlassian.net/browse/SEARCH-2233. Because currently, we don't display anything inside filter pill when js is disabled, I don't think users should be able to click the links. I think we'll have to move this ticket to the next phase though.

AlixRichards (approved):

PR #1747, opened 2022-03-15
f-content-cards@v8.0.0-alpha.1

matthewhardern:
Adding a new Image component that will be used as part of multiple components to build up new content cards going forward, this will allow us to build new cards more easily and will enable us to conform to Icing changes in an easier manner. 

## UI Review Checks

Screenshot 2022-03-15 at 10 54 03

## Browsers Tested

- [x] Chrome (latest)


packages/components/organisms/f-content-cards/src/components/templates/Image.vue
```diff
@@ -0,0 +1,57 @@
+<template>
+    <div :class="$style['c-content-cardImage--wrapper']">
+        <img
+            data-test-id="content-card-background-image"
+            :class="$style['c-content-cardImage']"
+            :src="backgroundImageUrl"
+            alt="hey">
```
dandel10n, line 7: Should this be cleaned up?

dandel10n (commented):
Looks good, left one tiny comment

kevinrodrigues (approved):

packages/components/organisms/f-content-cards/src/components/templates/Image.vue
matthewhardern: yes good catch will clean up as part of my next PR.

packages/components/organisms/f-content-cards/src/components/templates/Image.vue
```diff
@@ -0,0 +1,57 @@
...
+            data-test-id="content-card-restaurant-logo"
+            :class="$style['c-content-cardLogo']"
+            :src="restaurantLogoUrl"
+            alt="">
+    </div>
+</template>
+
+<script>
+export default {
+    name: 'ContentCardImage',
```
DamianMullins, line 18: Very minor thing, but is there any reason you didn't use this as the filename? That way you can avoid including this property.

DamianMullins (approved):

packages/components/organisms/f-content-cards/src/components/templates/Image.vue
matthewhardern: No specific reason, I might make the change of file name to make it more specific so it's obviously associated with content cards. Will update as part of the next PR.

PR #1752, opened 2022-03-16
f-filter-pill@v0.5.0 - Basic skeleton loading logic

jamieomaguire:
v0.5.0
*March 16, 2022*

### Added
- `isLoading` prop to trigger skeleton loading
### Changed
- Removed `isToggleDisabled` property from filter pill component and instead directly reference the `isDisabled` prop
- Shorten how we pass props into the Storybook component

## Browsers Tested

- [X] Chrome (latest)

Screenshot 2022-03-16 at 11 37 44


packages/components/atoms/f-filter-pill/CHANGELOG.md
```diff
@@ -3,6 +3,16 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v0.5.0
+------------------------------
+*March 16, 2022*
+
+### Added
+- `isLoading` prop to trigger skeleton loading
+### Changed
```
1itvinka, line 13:
```suggestion

### Changed
```
nit. It might be more readable to add an empty line between the blocks

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
```diff
@@ -3,46 +3,51 @@
...
-                data-test-id="filter-pill-text">
-                {{ displayText }}
-            </span>
-            <span
-                :class="$style['c-filterPill-number']"
-                data-test-id="filter-pill-number">
-                {{ displayNumber }}
-            </span>
-        </label>
+        <div v-if="isLoading">
```
1itvinka, line 9: it might be a part of the styling ticket, but it would be great to extract the loading thing into a separate component.

1itvinka (dismissed):

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
jamieomaguire: absolutely - I'd suggest we do that as part of the styling work if that's ok :)

packages/components/atoms/f-filter-pill/CHANGELOG.md
jamieomaguire: done!

1itvinka (approved):

PR #1754, opened 2022-03-16
f-filter-pill@v0.5.0 - Basic skeleton loading logic

jamieomaguire:
v0.5.0
*March 16, 2022*

### Added
- `isLoading` prop to trigger skeleton loading
### Changed
- Removed `isToggleDisabled` property from filter pill component and instead directly reference the `isDisabled` prop
- Shorten how we pass props into the Storybook component

## Browsers Tested

- [X] Chrome (latest)

Screenshot 2022-03-16 at 11 37 44


packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
```diff
@@ -143,7 +145,7 @@ $filter-pill-ease: ease-in-out;
 }
 
 .c-filterPill--selected {
-    box-shadow: 0 0 0 2px $color-focus;
+    box-shadow: 0 0 0 1px $filter-pill-selected-color;
 }
```
AlixRichards, line 149: Does this want to be reduced to 1px as well as the colour changing?

AlixRichards (approved):

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
jamieomaguire: I'm unsure! This was a change @1itvinka asked me to add in this PR - I will try to find out

jamieomaguire: @AlixRichards there will be another PR here shortly to deliver the styling - if we need to tweak to 2px I can add the change there if that's cool :)

packages/components/atoms/f-filter-pill/src/components/FilterPill.vue
```diff
@@ -3,46 +3,51 @@
         :class="[
             $style['c-filterPill'], {
                 [$style['c-filterPill--selected']]: isToggleSelected,
-                [$style['c-filterPill--disabled']]: isToggleDisabled
+                [$style['c-filterPill--disabled']]: isDisabled
```
sonnyprince, line 6: Do we want to change this as the opposite state is `isToggleSelected`?
should we align? or not change?

jamieomaguire: I think if there are more improvements to be made here then it would be best suited to a separate PR. As I understand, the `isDisabled` prop is to prevent the component from being interacted with/greyed out

sonnyprince (approved):

dawidchar (approved):

PR #1759, opened 2022-03-17
f-feature-management@v0.9.0 - Replace old logger interface with new one

tommcclean:
- We need to replace the old logger interface here so that we can delete it from the consuming website.
- Feature Management is one of the few places not yet using the new interface.
- We changed to a newer interface in November and are replacing the older one throughout.
- Reach out if you want more info as I can't post specifics in this repo.

### DIFF
--- a/packages/services/f-feature-management/src/vue/services/features.service.js
+++ b/packages/services/f-feature-management/src/vue/services/features.service.js
@@ -3,6 +3,7 @@ import featuresModule from '../store/features.module';
 
 const namespace = 'f-feature-management';
 const storeAction = `${namespace}/update`;
+const loggingTags = 'feature-management';
 
 function trackExperiment (key, variant, analytics) {
     analytics.pushEvent({
@@ -24,7 +25,7 @@ function contextGetter (store, cookies) {
     };
 }
 
-function createSettings (store, serviceStore, cookies, analytics, logger) {
+function createSettings (store, serviceStore, cookies, analytics, log) {
     const { featureManagement: envConfig } = store.state.configuration.settings;
 
     const settings = {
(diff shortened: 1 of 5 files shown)


packages/services/f-feature-management/src/vue/services/features.service.js
```diff
@@ -41,28 +42,19 @@ function createSettings (store, serviceStore, cookies, analytics, logger) {
...
     if (initialConfigAsJson && initialConfigAsJson.features) {
-        logger.logInfo('[FeaturesService] Initialising with stored config');
+        log.info('[FeaturesService] Initialising with stored config', loggingTags);
         settings.initialConfigAsJson = initialConfigAsJson;
     }
 
     return settings;
 }
 
-function createLoggerWrapper (store, logger) {
```
tommcclean, line 57: ℹ️ - Refactored this out, as the store is no longer required - I assume this is the reason for the wrapper.

packages/services/f-feature-management/src/vue/services/features.service.js
```diff
@@ -93,7 +85,7 @@ export default class FeaturesService {
             const newConfigJson = await this.featureManagement.loadFromCdn();
             await this.store.dispatch(storeAction, newConfigJson);
         } catch (error) {
-            this.logger.logError('Failed to update features config file', { error });
+            this.log.error('Failed to update features config file', error, loggingTags);
```
tommcclean, line 88: ℹ️ - Exceptions can now be explicitly provided and given to Sentry

packages/services/f-feature-management/src/vue/services/features.service.js
jnewc: Yes, this is ancient. Good call.

jnewc (approved):

Umar--Khan (approved):

PR #1771, opened 2022-03-23
f-content-cards@v8.0.0-alpha.2 - Adding Content card Body component

matthewhardern:
PR adds the body component of the new content card card templates, unit tests have been added. Also is a change to file names based on previous PR suggestions.

## UI Review Checks

Screenshot 2022-03-21 at 15 52 29

## Browsers Tested

- [x] Chrome (latest)

### DIFF
--- a/packages/components/organisms/f-content-cards/src/components/templates/ContentCardBody.vue
+++ b/packages/components/organisms/f-content-cards/src/components/templates/ContentCardBody.vue
@@ -101,7 +101,7 @@ export default {
 .c-content-card-bodyWrapper {
     z-index: 10;
     width: 100%;
-    @include media('<=narrow'){
+    @include media('<=narrow') {
         padding: spacing(d);
     }
 }
--- a/packages/components/organisms/f-content-cards/src/components/templates/ContentCardBody.vue
+++ b/packages/components/organisms/f-content-cards/src/components/templates/ContentCardBody.vue
@@ -1,0 +1,108 @@
+<template>
+    <div :class="$style['c-content-card-bodyWrapper']">
+        <div :class="$style['c-content-card-body']">
+            <slot name="title">
+                <h3
+                    v-if="title"
+                    data-test-id="content-card-title"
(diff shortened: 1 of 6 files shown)


tommcclean (approved):

packages/components/organisms/f-content-cards/src/components/templates/ContentCardBody.vue
```diff
@@ -0,0 +1,108 @@
...
+                    data-test-id="content-card-subtitle"
+                    :class="$style['c-content-card-subtitle']">
+                    {{ subtitle }}
+                </p>
+            </slot>
+            <slot name="description">
+                <template v-if="description">
+                    <p
+                        v-for="(text, index) in description"
+                        :key="index"
```
FayeCarter: Could this cause a problem with multiple '0' keys? Does it need to be `description-${index}` or similar?

matthewhardern: not sure what your asking here, could you explain further? The index is scoped under that v-for.

minrwhite: The key uniqueness problem is not a concern across components/component instances, but this could potentially be a problem if another for loop is introduced to the component at some other point. To guard against this, making the key `description-${index}` seems like a good precaution.

matthewhardern: > The key uniqueness problem is not a concern across components/component instances, but this could potentially be a problem if another for loop is introduced to the component at some other point. To guard against this, making the key `description-${index}` seems like a good precaution.

It would only be an issue if you were to place another loop within this loop, as its scoped to that v-for

matthewhardern: happy to modify if it makes people happy, it also makes sense from a readability point of view

FayeCarter (approved):

PR #1779, opened 2022-03-28
f-popover@2.3.0 - Removed f-popover tooltip arrow from popover.vue

JoshuaNg2332:
v2.3.0
*March 28, 2022*

### Removed
- Tooltip arrow from `Popover.vue` as per the request of design.


packages/components/atoms/f-popover/src/components/Popover.vue
```diff
@@ -26,18 +26,6 @@ $tooltip-width                 : 10px;
         border-radius: $radius-rounded-c;
         padding: 0 $popover-padding;
         width: auto;
```
zlatinivanov-je, line 27: Line 20 contains `$tooltip-width` that seems to be made redundant with the change

JoshuaNg2332: Ah missed that! Thanks. Will address now.

zlatinivanov-je (approved):

jamieomaguire (approved):

PR #1788, opened 2022-04-04
f-content-cards@v8.0.0-alpha.4 - Adding Voucher code component

matthewhardern:
PR adds in a slightly modified version of the original voucher card voucher code component. Unit tests have been added / updated to support this new component.

## UI Review Checks

Screenshot 2022-03-29 at 08 58 07

## Browsers Tested

- [x] Chrome (latest)

### DIFF
--- a/packages/components/organisms/f-content-cards/src/components/templates/ContentCardVoucherCode.vue
+++ b/packages/components/organisms/f-content-cards/src/components/templates/ContentCardVoucherCode.vue
@@ -17,13 +17,13 @@
                 :class="$style['c-contentCard-voucher-code-cooldown']"
                 role="status">
                 <transition
-                    :leave-to-class="$style['c-contentCard-voucher-copy-cooldownTick-leave-to']"
-                    :leave-active-class="$style['c-contentCard-voucher-copy-cooldownTick-leave-active']">
+                    :leave-to-class="$style['c-contentCard-voucher-copy-cooldownTickLeaveTo']"
+                    :leave-active-class="$style['c-contentCard-voucher-copy-cooldownTickLeaveActive']">
                     <tick-icon
                         v-if="inCooldown"
                         :class="[
                             $style.zeta,
-                            $style['c-contentCard-voucher-code-cooldown-tick']
+                            $style['c-contentCard-voucher-code-cooldownTick']
                         ]" />
                 </transition>
                 {{ voucherCodeCopiedLabel }}
@@ -99,7 +99,7 @@ export default {
             return [
                 this.$style['c-contentCard-voucher-copy'],
                 { [this.$style['c-contentCard-voucher-copy--copied']]: this.copyState === COPY_STATE_COOLDOWN },
-                { [this.$style['c-contentCard-voucher-copy--transition-out']]: this.copyState === COPY_STATE_TRANSITIONOUT }
+                { [this.$style['c-contentCard-voucher-copy--transitionOut']]: this.copyState === COPY_STATE_TRANSITIONOUT }
             ];
         },
 
@@ -175,19 +175,19 @@ export default {
     color: $color-content-default;
 }
 
-.c-contentCard-voucher-copy--transition-out {
+.c-contentCard-voucher-copy--transitionOut {
     transition: color 1s;
 }
 
-.c-contentCard-voucher-copy-cooldownTick-leave-to {
+.c-contentCard-voucher-copy-cooldownTickLeaveTo {
     opacity: 0;
 }
 
-.c-contentCard-voucher-copy-cooldownTick-leave-active {
+.c-contentCard-voucher-copy-cooldownTickLeaveActive {
     transition: opacity 1s;
 }
 
-.c-contentCard-voucher-code-cooldown-tick {
+.c-contentCard-voucher-code-cooldownTick {
     fill: $color-content-default;
     width: spacing(d);
     height: spacing(d);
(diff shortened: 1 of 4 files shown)


packages/components/organisms/f-content-cards/src/components/templates/ContentCardVoucherCode.vue
```diff
@@ -0,0 +1,201 @@
...
+        :class="$style['c-contentCard-voucher']"
+        :data-test-id="`contentCard-${cardType}`"
+        @click="copyVoucherCode">
+        <span
+            :class="$style['c-contentCard-voucher-code']"
+            :data-test-id="`contentCard-${cardType}-code`">
+            {{ code }}
+        </span>
+        <span
+            :class="voucherCopyClasses"
```
DamianMullins, line 13: Any reason you haven't used the dynamic class approach here? e.g.

```html
:class="[$style['c-contentCard-voucher-copy'], {
    this.copyState === COPY_STATE_COOLDOWN: $style['c-contentCard-voucher-copy-copied'],
    this.copyState === COPY_STATE_TRANSITIONOUT: $style['c-contentCard-voucher-copy-transition-out']
}]"
```

packages/components/organisms/f-content-cards/src/components/templates/ContentCardVoucherCode.vue
```diff
@@ -0,0 +1,201 @@
...
+
+.c-contentCard-voucher-code {
+    text-align: left;
+}
+
+.c-contentCard-voucher-copy {
+    text-align: right;
+}
+
+.c-contentCard-voucher-copy-copied {
```
DamianMullins: This looks like it should be a modifier class? i.e. `c-contentCard-voucher-copy--copied`

packages/components/organisms/f-content-cards/src/components/templates/ContentCardVoucherCode.vue
```diff
@@ -0,0 +1,201 @@
...
+    color: $color-content-default;
+}
+
+.c-contentCard-voucher-copy-transition-out {
```
DamianMullins: The dashes represent child elements in our CSS naming structure, it looks like this should, again, be a modifier. But also, something like `c-contentCard-voucher-copy--transitionOut`

Same applies to the classes below.

packages/components/organisms/f-content-cards/src/components/templates/ContentCardVoucherCode.vue
matthewhardern: Not sure, it was copied across from the previous component. I agree it would simplify the code and make it more readable. I will make the change though. Thanks

packages/components/organisms/f-content-cards/src/components/templates/ContentCardVoucherCode.vue
matthewhardern: agreed it should be --

packages/components/organisms/f-content-cards/src/components/templates/ContentCardVoucherCode.vue
matthewhardern: as above agreed

packages/components/organisms/f-content-cards/src/components/templates/ContentCardVoucherCode.vue
matthewhardern: @DamianMullins I think this is because with `$style` you can't write `{$style['example-class']: this.copyState === COPY_STATE_COOLDOWN}` which would be the shorthand instead of using the ternary expression

DamianMullins: Ah yeah, sorry, my code above was completely wrong! You can apply conditional CSS module classes like this (we do this [in the mega-modal component](https://github.com/justeat/fozzie-components/blob/679296a56d83e49484146f4d9063c6c3bea7ca48/packages/components/molecules/f-mega-modal/src/components/MegaModal.vue#L14-L19).

```
:class="[$style['c-contentCard-voucher-copy'], {
    [$style['c-contentCard-voucher-copy-copied']]: this.copyState === COPY_STATE_COOLDOWN,
    [$style['c-contentCard-voucher-copy-transition-out']]: this.copyState === COPY_STATE_TRANSITIONOUT
}]"
```

matthewhardern: nice thanks that looks quite neat and tidy

matthewhardern: I have made your change but on the computed property to allow the continued use of the constant

DamianMullins: Sure, looks good. I think I've hit this before and had to wrap the constant in a computed property but I think it's fine to not do that here 👍🏻

packages/components/organisms/f-content-cards/src/components/templates/ContentCardVoucherCode.vue
```diff
@@ -0,0 +1,197 @@
...
+
+.c-contentCard-voucher-copy {
+    text-align: right;
+}
+
+.c-contentCard-voucher-copy--copied {
+    color: $color-content-default;
+}
+
+.c-contentCard-voucher-copy--transition-out {
```
DamianMullins: Apologies, but the dash in `transition-out` should be removed (and in the classes below) — switch them to camel-case and these all look good, e.g. `c-contentCard-voucher-copy--transitionOut`, `c-contentCard-voucher-copy--cooldownTickLeaveTo` etc

matthewhardern: yep makes sense thanks

DamianMullins (approved):

PR #1790, opened 2022-04-05
f-loyalty@v1.7.2 - changes to add in some logging in for decoding jwt

matthewhardern:
Adding in a minor change to log some information out around decoding jwt when it fails with invalid token.

### DIFF
--- a/packages/components/pages/f-loyalty/src/components/Loyalty.vue
+++ b/packages/components/pages/f-loyalty/src/components/Loyalty.vue
@@ -124,6 +124,7 @@ export default {
         await this.init({
             brazeApiKey: this.brazeApiKey,
             authToken: this.authToken,
+            logger: this.$log,
             errorHandler: this.handleErrors
         });
     },
--- a/packages/components/pages/f-loyalty/src/components/Loyalty.vue
+++ b/packages/components/pages/f-loyalty/src/components/Loyalty.vue
@@ -123,7 +123,8 @@ export default {
     async mounted () {
         await this.init({
             brazeApiKey: this.brazeApiKey,
-            authToken: this.authToken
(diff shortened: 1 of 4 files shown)


packages/components/pages/f-loyalty/src/store/loyalty.module.js
```diff
@@ -10,10 +10,31 @@ import {
...
+        return tokenData?.global_user_id;
+    } catch (error) {
+        logger.error(
+            'Error Decoding JWT token',
+            error,
+            'loyalty',
+            {
+                ...{
+                    authToken
+                },
```
DamianMullins: Adding `authToken` directly here should achieve the same thing without the spread I believe? 

```suggestion
                authToken,
```

matthewhardern: thanks fixed

DamianMullins (dismissed):

packages/components/pages/f-loyalty/src/components/Loyalty.vue
```diff
@@ -123,7 +123,8 @@ export default {
     async mounted () {
         await this.init({
             brazeApiKey: this.brazeApiKey,
-            authToken: this.authToken
+            authToken: this.authToken,
+            log: this.$log
         });
```
minrwhite: Assuming the error passes through this on it's way up the stack when it is thrown, I'd prefer if the error is caught here and `emit()`ed as an event rather than left to be handled by the global error handler.

matthewhardern: so have the `f-loyalty` emit an onError event that coreweb can listen to?

PR #1793, opened 2022-04-07
f-loyalty@v1.7.2 - jwt error logging and change to Tabs import

matthewhardern:
Adding in a minor change to log some information out around decoding jwt when it fails with invalid token.

### DIFF
--- a/packages/components/pages/f-loyalty/src/store/loyalty.module.js
+++ b/packages/components/pages/f-loyalty/src/store/loyalty.module.js
@@ -10,10 +10,31 @@ import {
 /**
  * @function resolveGlobalUserID
  * @param  {string} authToken - auth token.
+ * @param logger
  */
-const resolveGlobalUserID = authToken => {
-    const tokenData = jwtDecode(authToken);
-    return tokenData?.global_user_id;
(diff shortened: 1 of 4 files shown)


DamianMullins (dismissed):

packages/components/pages/f-loyalty/src/store/loyalty.module.js
```diff
@@ -10,10 +10,33 @@ import {
...
  * @function resolveGlobalUserID
  * @param  {string} authToken - auth token.
+ * @param logger
+ * @param errorHandler
  */
-const resolveGlobalUserID = authToken => {
-    const tokenData = jwtDecode(authToken);
-    return tokenData?.global_user_id;
+const resolveGlobalUserID = (authToken, logger, errorHandler) => {
+    // eslint-disable-next-line no-useless-catch
```
minrwhite: Is this comment still required, now we're performing an action and not rethrowing?

matthewhardern: good point

minrwhite (approved):

DamianMullins (approved):

PR #1795, opened 2022-04-08
f-restaurant-card@v0.29.4 - Fix minor design issues

dawidchar:
Fixed minor design issues raised during mob testing.

### Fixed
- Centre Restaurant dish names when calories data is not present
- Image tag spacing to match logo spacing
- Prevent ratings count from wrapping

image
---

### Ratings Wrapping
This issue happens typically on restaurants with a lot of data and users on the lower size of the restaurant list view
#### Before
image

#### After
image

---
_(Fixed issues raised in ticket **SEARCH-2698**)_

### DIFF
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
@@ -426,7 +426,7 @@ export default {
 }
 
 .c-restaurantCard-imageTags {
-    bottom: spacing(c);
+    bottom: spacing(d);
     left: spacing(d);
     position: absolute;
     margin-bottom: 0;
--- a/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
@@ -426,7 +426,7 @@ export default {
 }
 
 .c-restaurantCard-imageTags {
-    bottom: spacing(b);
+    bottom: spacing(c);
     left: spacing(d);
     position: absolute;
     margin-bottom: 0;
--- a/packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantDishes/RestaurantDish.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantDishes/RestaurantDish.vue
@@ -9,6 +9,7 @@
                 {{ name }}
             </span>
             <span
+                v-if="calories || portion"
                 :class="[$style['c-restaurantCard-dish-nutritionalInfo']]">
                 <span
                     v-if="calories"
--- a/packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantRating/RestaurantRating.vue
+++ b/packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantRating/RestaurantRating.vue
@@ -170,6 +170,7 @@ export default {
     align-items: center;
     padding: 0;
     margin: 0;
+    white-space: nowrap;
 }
 
 .c-restaurantCard-rating-iconWrapper {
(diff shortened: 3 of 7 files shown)


packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
```diff
@@ -426,7 +426,7 @@ export default {
 }
 
 .c-restaurantCard-imageTags {
-    bottom: spacing(b);
+    bottom: spacing(c);
```
jamieomaguire: is this the same value as the logo?

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantDishes/RestaurantDish.vue
```diff
@@ -9,6 +9,7 @@
                 {{ name }}
             </span>
             <span
+                v-if="calories || portion"
```
jamieomaguire, line 12: does this need a unit test?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
dawidchar: Yes and no, so the logo has a spacing of **d** which is 16px, while **c** is 12px, however the restaurant tag has a hardcoded 4px bottom margin, therefore it ends up being 16px, the hardcoded 4px is for when the tags are in tile view and when they wrap so that they don't touch each other

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantDishes/RestaurantDish.vue
dawidchar: Probably not, as all the current unit tests are passing (which test to ensure portions and calories are displayed), and we'd simply test if a span does or does not exist

jamieomaguire: ah I just noticed this span is empty, what is it for?

jamieomaguire: ah yes - sorry I missed that

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantRating/RestaurantRating.vue
```diff
@@ -170,6 +170,7 @@ export default {
     align-items: center;
     padding: 0;
     margin: 0;
+    white-space: nowrap;
```
jamieomaguire, line 173: is this going to potentially cause any new issues when in tight space? Also would `flex-wrap: nowrap` be useful here?

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
jamieomaguire: It sounds like that 4px margin might need looking at. Are you sure that can't be moved somewhere else?

dawidchar: I did think that, the alternative would be to remove the 4px and add a flex gap to the parent

packages/components/molecules/f-restaurant-card/src/components/subcomponents/RestaurantRating/RestaurantRating.vue
dawidchar: The only issue would be that a there might be a very slight overlap, but this issue seems to only appear on the lowest size of the list view and only if there is a new tag or a local legends tag

dawidchar: `flex-wrap: nowrap;` does not appear to fix the issue

packages/components/molecules/f-restaurant-card/src/components/RestaurantCard.vue
dawidchar: Updated 😄

AlixRichards (approved):

jamieomaguire (approved):

PR #1796, opened 2022-04-11
f-content-cards@v8.0.0-alpha.5 - Promotion Card Two

matthewhardern:
PR adds promotion card two. Card is the latest Icing design for promotion card two. I have included unit tests for the card.

## UI 

Screenshot 2022-04-11 at 08 34 03

## Browsers Tested

- [x] Chrome (latest)

### DIFF
--- a/packages/components/organisms/f-content-cards/stories/new/PromotionCardTwo.stories.js
+++ b/packages/components/organisms/f-content-cards/stories/new/PromotionCardTwo.stories.js
@@ -10,32 +10,30 @@ export default {
 
 function cardGenerator (type, title, image, icon, description, cta) {
     return {
-        card: {
-            type,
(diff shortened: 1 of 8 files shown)


packages/components/organisms/f-content-cards/stories/new/PromotionCardTwo.stories.js
```diff
@@ -0,0 +1,107 @@
...
+    data () {
+        return cardGenerator(
+            'Promotion_Card_1',
+            this.title,
+            this.image,
+            this.icon,
+            this.description,
+            this.cta
+        );
+    },
```
minrwhite: Will the fact that the `data()` function is returning the output from the generator rather than having a `card` computed property mean that any update to any of the args in storybook won't live update the component?

matthewhardern: good point, will make sure that this works

minrwhite (approved):

FayeCarter (approved):

PR #1802, opened 2022-04-12
f-header@v9.13.0 - small CSS updates to match design

LTurns:
### Changed
- added font size and color to nav buttons
- refactored `UserNavigationPanel` to take in `NavLink` component


JoshuaNg2332 (dismissed):

FayeCarter (approved):

JoshuaNg2332 (approved):

packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -166,6 +166,10 @@ $nav-link-focus-color              : $color-focus;
             text-decoration: underline;
         }
     }
+
+    @include media('<=mid') {
+        color: $nav-text-color;
```
ashleynolan, line 171: What's the reason for only applying this colour on mobile? Does the colour change between mobile/desktop?

packages/components/organisms/f-header/src/assets/scss/navigation.scss
```diff
@@ -166,6 +166,10 @@ $nav-link-focus-color              : $color-focus;
...
+    }
```
ashleynolan, line 172: You could just miss out the media query for this and put it above the `@include media('>mid') {` line, so that it's effectively the default colour.

packages/components/organisms/f-header/src/components/NavLink.vue
```diff
@@ -159,10 +158,12 @@ export default {
 
     &:hover {
         background: $color-container-subtle;
+        border-radius: 0 !important;
```
ashleynolan, line 161: What other styles makes you need `!important` here?

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -763,47 +768,27 @@ export default {
     }
 }
 
-.c-nav-list-text {
```
ashleynolan, line 766: Now that this class has beeen deleted, you'll likely need to rename `c-nav-list-text-sub` as the naming won't make sense otherwise (as that indicates that `c-nav-list-text-sub` is a child of `c-nav-list-text`)

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -711,8 +710,11 @@ export default {
 }
 
 .c-nav-list-btn-text {
+    @include font-size($nav-text-size);
+    color: $nav-text-color;
+    font-weight: $font-weight-regular;
     @include media('>mid') {
-        margin-top: 2px;
+        font-weight: $nav-text-weight;
```
ashleynolan, line 717: Super NIT – can you add an extra line break between normal CSS rules and any nested classes or media queries etc.

It just makes it easier to scan the styles (and it's something we can't enforce in stylelint)

PR #1803, opened 2022-04-12
f-loyalty@1.7.3 - auth token check

matthewhardern:
PR fixes missing check for auth token before getting global user id


packages/components/pages/f-loyalty/src/store/loyalty.module.js
```diff
@@ -73,12 +51,14 @@ export default {
             commit(ACTION_SET_GLOBAL_USER_ID, resolveGlobalUserID(authToken, logger, errorHandler));
         },
 
-        [ACTION_INITIALISE_LOYALTY]: ({ commit, dispatch }, {
-            brazeApiKey, authToken, logger, errorHandler
+        [ACTION_INITIALISE_LOYALTY]: ({ commit, dispatch, getters}, {
```
minrwhite: This should fix lint checks:
```suggestion
        [ACTION_INITIALISE_LOYALTY]: ({ commit, dispatch, getters }, {
```

packages/components/pages/f-loyalty/src/components/Loyalty.vue
```diff
@@ -123,9 +123,7 @@ export default {
     async mounted () {
         await this.init({
             brazeApiKey: this.brazeApiKey,
-            authToken: this.authToken,
-            logger: this.$log,
-            errorHandler: this.handleErrors
```
minrwhite, line 128: Can we remove `handleErrors()` as well if we're going to remove all reference to it? Or is it worth wrapping the `await this.init()` call in a `try { ... } catch (e) { this.handleErrors(e); throw e; }` to help identify any other errors (i.e. events surfaced by the vue dev tools, even if they don't have a handler attached).

matthewhardern: just removing for now

FayeCarter (approved):

minrwhite (approved):

PR #1805, opened 2022-04-13
fozzie-components@v7.6.0 - Specify required Node major versions of 12 / 14 for all components #globalconfig

siggerzz:
This PR forces either Node 12 / 14 when using the monorepo. As we start to refactor components so that they're compatible with Node 16, we'll be able to modify the `engines` value in `package.json`


xander-marjoram (approved):
Nice. Thanks for updating the generator template too ⭐

ashleynolan (approved):

PR #1817, opened 2022-04-27
f-checkout@3.26.0 - Axios wrapper for API calls that require a retry

charliearlie:
One issue we've had in Checkout is API calls failing where a repeat of the same call could rectify that. I believe this has been trialled on Android and iOS with some success. 

I have added this wrapper on only 2 smaller endpoints for now to gauge the success and will add it to further calls in the future.


FayeCarter (dismissed):

FayeCarter (dismissed):

DamianMullins (commented):
Looks good, we use an npm package in menu to do something similar, along with exponential backoffs, called [`exponential-backoff`](https://github.com/coveooss/exponential-backoff) which allows us to also log when retries occur.

DamianMullins (commented):
Looks good, we use an npm package in menu to do something similar, along with exponential backoffs, called [`exponential-backoff`](https://github.com/coveooss/exponential-backoff) which allows us to also log when retries occur.

DamianMullins (dismissed):

charliearlie:
> Looks good, we use an npm package in menu to do something similar, along with exponential backoffs, called [`exponential-backoff`](https://github.com/coveooss/exponential-backoff) which allows us to also log when retries occur.

Definitely looks like a good option @DamianMullins - I might build on this and use that library because logging a retry is important. Thanks for that.

BanksySan (approved):

FayeCarter (approved):

PR #1818, opened 2022-04-28
f-checkout@v3.26.0 - update error messages

FayeCarter:
Previously voiceover would only read 'You have * errors in the form' on the first submit. Screen readers would not be updated that there are still errors in the form after submitting again. 

Now when the submit button is pressed focus is put on the first error message. Each field will read 'There are * errors in the form. Please enter your ...'

### Changed
 - Invalid field error summary to be read with error messages.

### Added
 - Focus to first inline error.

### DIFF
--- a/packages/components/pages/f-checkout/src/components/CheckoutForm.vue
+++ b/packages/components/pages/f-checkout/src/components/CheckoutForm.vue
@@ -5,8 +5,8 @@
         @submit.prevent="onFormSubmit">
         <section
             v-if="invalidFieldsSummary"
+            id="error-summary-container"
             class="is-visuallyHidden"
-            role="alert"
             data-test-id="error-summary-container">
             {{ invalidFieldsSummary }}
         </section>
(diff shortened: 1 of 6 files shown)


BanksySan:
We need to also test this on Windows NVDA and Android Talkback.

packages/components/pages/f-checkout/src/components/CheckoutForm.vue
```diff
@@ -171,9 +173,13 @@ export default {
          */
         scrollToFirstInlineError () {
             this.$nextTick(() => {
-                const firstInlineError = document.querySelector('[data-js-error-message]');
+                const firstInlineError = document.querySelector('[aria-invalid="true"]');
 
                 this.scrollToElement(firstInlineError, { offset: -100 });
+
+                firstInlineError?.focus({
```
BanksySan, line 180: This might not be a good idea.  Repositioning the cursor is helpful for a signed person, but for a blind person is frustrating.
It's a bit like moving somebody in a wheelchair without asking them if they want to be moved.

BanksySan (changes requested):

packages/components/pages/f-checkout/src/components/CheckoutForm.vue
FayeCarter: Ah ok cool. The reason I tried it is because the screen already moves up to be on that field and it reads that particular error so thought it would help 😂 
Happy to remove it though

PR #1825, opened 2022-05-03
f-checkout@3.27.0 - Lowercase note config keys to match dine in

charliearlie:
A not so nice PR I have had to make due to the keys in the `checkout-note-types` response being in camelcase when Dine in has always been `dinein` as a service type. It was too much of a hassle for the change to be made on the backend, so until it is finally changed, we have had to make this workaround. 

A ticket will be made in the Checkout backlog for the backend team to allow us to revert this change. 🙂


oliversweb (approved):

packages/components/pages/f-checkout/src/store/checkout.module.js
```diff
@@ -250,8 +250,12 @@ export default {
          */
         getNotesConfiguration: async ({ commit, state }, { url, timeout }) => {
             const { data } = await checkoutApi.getNoteConfiguration(url, timeout);
-            const isSplitNotesEnabled = !data?.customerNotes?.serviceTypes[state.serviceType].orderNoteAccepted;
-            commit(UPDATE_NOTES_CONFIGURATION, { ...data?.customerNotes?.serviceTypes, isSplitNotesEnabled });
+            if (data?.customerNotes.serviceTypes) {
+                // This has to be done because the dinein key is sent to us as 'dineIn` ....
+                const notesByServiceType = Object.fromEntries(Object.entries(data.customerNotes.serviceTypes).map(([k, v]) => [k.toLowerCase(), v]));
```
jamieomaguire, line 255: looks good - I was wondering if this would be easier to understand if the const was named `notesByServiceTypes` (plural rather than singular) as, if I understand correctly, this is creating an object of all the service types in lowercase?

Certainly not a big deal, approving regardless :)

jamieomaguire (approved):

packages/components/pages/f-checkout/src/store/checkout.module.js
charliearlie: Yeah, makes sense. I will change it in my next PR with a note as about to make another one related to something else. Hopefully this workaround is only temporary for a short period too!

PR #1854, opened 2022-05-11
f-offers@v1.12.0 - Adding new version of content cards

matthewhardern:
PR adds the updated Icing changes and the new content cards from beta1 of v8 content cards. 

## UI Review Checks

Screenshot 2022-05-11 at 15 14 20
Screenshot 2022-05-11 at 15 14 00
Screenshot 2022-05-11 at 15 13 44
Screenshot 2022-05-11 at 15 13 24

## Browsers Tested

- [x] Chrome (latest)


packages/components/pages/f-offers/src/components/Results.vue
```diff
@@ -1,3 +1,5 @@
+
+
```
minrwhite: Minor, but I beleive having some leading newlines is not as standard.

minrwhite (approved):

DamianMullins (approved):

PR #1867, opened 2022-05-16
f-checkout@3.28.0 - Add address support for IE

charliearlie:
Added address support for IE as well as using the new assistive text slot in `f-form-field` to add a link to find your Eircode


FayeCarter (dismissed):

FayeCarter (dismissed):

FayeCarter (dismissed):

FayeCarter (approved):

packages/components/pages/f-checkout/src/services/addressService/tenants/ie.js
```diff
@@ -0,0 +1,42 @@
+function isFullPostCode (postcode) {
+    return postcode?.length >= 5;
```
BanksySan, line 2: Should this be a regex match instead of a length check as `ab!_90` would pass this check?

packages/components/pages/f-checkout/src/services/addressService/tenants/ie.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+function toFormattedPostcode (postcode) {
+    let formattedPostcode = postcode.toString().replace(/\s/g, '').replace(/-/g, '');
```
BanksySan, line 6: Should the matching regex be `\s+`?

BanksySan (approved):

packages/components/pages/f-checkout/src/services/addressService/tenants/ie.js
charliearlie: These are copied from existing tenants. Both comments.

packages/components/pages/f-checkout/src/services/addressService/tenants/ie.js
charliearlie: As mentioned below, we use this logic for every tenant

PR #1877, opened 2022-05-24
f-account-info@v0.22.0 - Redirect on 403

oliversweb:
### Removed
- Temp logging to record if auth expired or not

### Added
- Prop for 'login path' which is now used for 403 responses when getting data.

### DIFF
--- a/packages/components/pages/f-account-info/src/components/AccountInfo.vue
+++ b/packages/components/pages/f-account-info/src/components/AccountInfo.vue
@@ -214,7 +214,6 @@
 </template>
 
 <script>
-import jwtDecode from 'jwt-decode';
 import { mapActions, mapState } from 'vuex';
 import { VueGlobalisationMixin } from '@justeat/f-globalisation';
 import FErrorMessage from '@justeat/f-error-message';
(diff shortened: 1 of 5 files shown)


packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -366,30 +369,25 @@ export default {
             try {
                 await this.loadConsumerDetails({ api: this.consumerApi, authToken: this.authToken });
                 this.$log.info('Consumer details fetched successfully', standardLogTags);
+                // debug
```
oliversweb, line 372: I've added some debug code to assist with testing in QA and will remove and do a second pr once complete.

siggerzz (approved):

packages/components/pages/f-account-info/src/components/AccountInfo.vue
BanksySan: Is this still to be removed?

oliversweb: Not yet

kevinrodrigues (approved):

PR #1908, opened 2022-06-08
f-form-field@4.13.0 - Add required prop

FayeCarter:
### Added
- `isRequired` prop to add `*` to form label and `aria-required` attribute to input.


siggerzz:
@FayeCarter Looks like all the visual regression needs re-approving because all the screenshots now have `required` set to true.

Could we set it to false for the existing tests, and add a new test specifically for required base state?

You should be able to copy the existing base state test and just change the prop value 👍🏼

FayeCarter:
> @FayeCarter Looks like all the visual regression needs re-approving because all the screenshots now have `required` set to true.
> 
> Could we set it to false for the existing tests, and add a new test specifically for required base state?
> 
> You should be able to copy the existing base state test and just change the prop value 👍🏼

I've kept the original tests as they were and updated the screenshot so that they show the default values and written a new test for setting `isRequired` to `false`

packages/components/atoms/f-form-field/test/visual/f-form-field.visual.spec.js
```diff
@@ -43,4 +43,18 @@ describe('f-form-field visual tests', () => {
             await browser.percyScreenshot('f-form-field - Errored State', 'desktop');
         });
     });
+
+    describe('none required state', () => {
+        it('should display all fields in an none required state', async () => {
+            // Arrange
+            formField = await new FormField()
```
oliversweb: Is this correct ... do we need to await when instantiating new object?

jamieomaguire (approved):

jamieomaguire (approved):

oliversweb (approved):

ashleynolan:
@FayeCarter Delayed review (as on holiday, sorry-not-sorry 😉 ). I've got a couple of questions relating to this, as I'm not sure this needs an `isRequired` prop potentially – I think this got discussed in refinement, but may have got missed in the acceptance criteria afterwards potentially.

Mainly, the points were:

- As `required` is a default HTML attribute, we could use that for the default state (to add aria attributes and add the asterisk)
- If we wanted to remove the asterisk for specific instances where we don't need it, that would be when we'd use a prop to hide that, so that it could be removed (but the aria attributes should stay on the element)

If you've got 5 mins tomorrow would be good to discuss – may need to add a new ticket to adjust the functionality slightly.

PR #1928, opened 2022-06-10
#trivial - Move two components to dart-sass.

kevinrodrigues:
### Removed
- `node-sass` from `f-tabs` & `f-navigation-links`.


packages/components/molecules/f-navigation-links/src/components/_tests/__snapshots__/NavigationLinks.test.js.snap
```diff
@@ -2,43 +2,43 @@
 
 exports[`NavigationLinks should display the list elements correctly for each \`link\` supplied 1`] = `
 <nav aria-label="navigation" data-test-id="navigation-links-component">
-  <ul class="c-navigationLinks">
```
kevinrodrigues, line 5: I've had to update the snapshot here because the `experimentalCSSCompile` flag won't allow compilation of `$style` from vue. Otherwise we would need to keep the `node-sass` dep inside this component. From what i can see there is no actual resolution for this.

packages/components/molecules/f-tabs/src/components/Tab.vue
```diff
@@ -63,8 +63,8 @@ export default {
 
         transitionName () {
             return this[TABS_COMPONENT].animationDirection === DIRECTION.LEFT
-                ? this.$style['fade-in-right']
```
kevinrodrigues, line 66: These are changed because of the same reason above.

packages/components/molecules/f-tabs/src/components/tests/Tab.test.js
```diff
@@ -67,7 +67,7 @@ describe('Tab.vue', () => {
 
         it('should show', () => {
             // Assert
-            const tab = activeWrapper.find('.c-tab');
```
kevinrodrigues, line 70: Ditto.

PR #1929, opened 2022-06-10
f-account-info@v0.24.0 - Dynamically generate return URL

tommcclean:
- Return URL is currently hardcoded in the `f-account-info` page component
- This PR changes this to use `this.$route` so that it is dynamic
- It also protects the generated return URL by encoding it.


kevinrodrigues (dismissed):

jamieomaguire (dismissed):

packages/components/pages/f-account-info/src/components/AccountInfo.vue
```diff
@@ -301,7 +301,13 @@ export default {
     computed: {
         ...mapState('fAccountInfoModule', [
             'consumer'
-        ])
+        ]),
+
+        changePasswordUrl () {
+            const returnUrlPath = encodeURIComponent(this.$route.fullPath);
+
+            return `/account/change-password?returnUrl=${returnUrlPath}`;
```
oliversweb, line 311: You need to pass this `/account/change-password` in as a prop as we don't want to assume all users of this component use this path. This was the purpose of the ticket.

tommcclean: Why would this ever be different?

oliversweb: imho - That's not for us to decide, we make generic components and shouldn't hard code paths.

tommcclean: I agree in the case of the return URL because you could host the account info in other pages / in a modal in the future.

I disagree in the sense that the change password button will always link to the change password page, worth a discussion because I think we should be careful to avoid adding props for things that will likely never change.

Pages are less generic than atoms, molecules and organisms, making everything about a page generic would require a lot of configuration which also complicates them.

xander-marjoram: I'm with @tommcclean on this one. I feel that this is the kind of logic we can keep out of CoreWeb, which only needs to know that it's hosting `/account/info`, not telling the account info page where to find `/account/change-password`.

oliversweb: Change password will be a new self-contained page (eventually), so account info will have no knowledge of it ... secondly it also suits the scenario of changing the coreweb config if we wanted to use some other change password page ... note change password has no seo benefit so can change without too much of an issue. Also we create these components in an Open Source environment (whether anyone else uses them or not) so to keep this ethos alive and kicking we should stick to the rules of keep things generic and building for everyone not just us.

tommcclean: Discussed in slack

jamieomaguire (approved):

kevinrodrigues (approved):

PR #1942, opened 2022-06-13
#trivial - Move components to dart-sass.

kevinrodrigues:
### Removed
- `node-sass` from `f-tabs` & `f-navigation-links`.


packages/components/molecules/f-navigation-links/src/components/_tests/__snapshots__/NavigationLinks.test.js.snap
```diff
@@ -2,43 +2,43 @@
 
 exports[`NavigationLinks should display the list elements correctly for each \`link\` supplied 1`] = `
 <nav aria-label="navigation" data-test-id="navigation-links-component">
-  <ul class="c-navigationLinks">
```
kevinrodrigues, line 5: I've had to update the snapshots because css compilation when turned off can't pick these up. The compilation has been turned off because `node-sass` has been removed.

packages/components/molecules/f-tabs/src/components/Tab.vue
```diff
@@ -63,8 +63,8 @@ export default {
 
         transitionName () {
             return this[TABS_COMPONENT].animationDirection === DIRECTION.LEFT
-                ? this.$style['fade-in-right']
```
kevinrodrigues, line 66: Similar thing to above comment, $style can not be picked up in jest when using `experimentalCSSCompile: false` so have updated the test.

jamieomaguire (approved):

FayeCarter (approved):

PR #1945, opened 2022-06-14
fozzie-components@7.12.0 - Remove CircleCI caching in favour of building dependencies in CI

siggerzz:
Currently, when a non-global build runs in CI, we relying on restoring the dependency `dist` in order to speed up CI time. However, due to flaky master builds, we can run into an issue when the cache isn't being updated, which can result in old dependencies being pulled down.

This PR addresses this by telling Turborepo to build any dependencies before building the changed package. This should:

- Improve Percy stability
- Speed up local build times as we're no longer running `--concurrency=1`


FayeCarter (dismissed):

jamieomaguire (dismissed):

siggerzz:
Closing in favour of https://github.com/justeat/fozzie-components/pull/1956/

PR #1956, opened 2022-06-16
fozzie-components@v7.14.0 - Ensure devDependency version range matches peerDependencies

siggerzz:
This PR looks to solve some of the dependency chains issues in the monorepo where yarn pulls down the incorrect version /  certain dependencies aren't being built when they should.

It also removes the CircleCI caching functionality, and ensure any changed package dependencies get built, resolving any potential out-of-date cache / Percy issues.


packages/components/molecules/f-card-with-content/package.json
```diff
@@ -49,9 +49,7 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "peerDependencies": {
-    "@justeat/browserslist-config-fozzie": ">=1.2.0",
-    "@justeat/f-button": "3.x",
-    "@justeat/f-card": "3.x"
+    "@justeat/browserslist-config-fozzie": ">=1.2.0"
   },
   "devDependencies": {
     "@justeat/f-button": "3.3.0",
```
siggerzz: As you can see, the `peerDependencies` are also declared as a `devDependency`.

Currently, the `devDependency` version is ignored if a newer component version is found when resolving the `peerDependency`


For example, despite the fact we specify `"@justeat/f-button": "3.3.0",` `3.3.1` will get pulled down because of the `"3.x",` in the `peerDependency`.

Whilst we could just specify the version we want in the `peerDependency`, I don't believe this is best practice. to have both dev / peer deps

siggerzz:
Visual reg changes are due to the `f-alert` version bump where Icons have been updated.

packages/components/molecules/f-card-with-content/package.json
DamianMullins: My understanding of mixing the two types of dependencies is that it allows consuming projects to pass a package they already have installed into multiple dependant packages.

So `peerDependencies` allows consuming projects to pass a package version into multiple packages. Whereas `devDependencies` is there purely for when running this package in isolation i.e. testing or possibly Storybook? 

By removing `peerDependencies` are we not preventing consumers of the package from using whichever version they like and, possibly, allowing for duplicate versions of packages to be used? i.e. they might already use f-button v3.3.1 in their project but this package will bundle and use v3.0.0 alongside it.

ashleynolan: Yeh, so to confirm, the peerrDependency is there to make consumers explicitly aware that they need to use that version of a package alongside this one (so here, that is that f-button, f-card etc need to be installed as dependencies to use f-card-with-content).

The devDependency is solely there for testing as part of the unit tests and for storybook.

A potential solution to this could be to make the devDependency match the range of the peerDependency, so that the tests just run with the latest published version of the component (as essentially, they should run against this anyway, as that's what the peerDependency is saying). The only issue we could have with this is if someone manages to push a component change that does break a dependent test, but I don't see that as a downside, as we would want to know about that asap anyway?

kevinrodrigues: I'm not sure how easy it will be but keeping both devDep & peerDep in sync seems like the logical resolution here. I'm happy to help out if you need.

.circleci/config.yml
```diff
@@ -64,119 +64,11 @@ commands:
           command: |
             if [ "${RUN_ALL}" == "true" ];
             then
-              yarn << parameters.command_name >> --concurrency=1
+              yarn << parameters.command_name >>
             else
-              yarn << parameters.command_name >> --filter=[master] --concurrency=1
+              yarn << parameters.command_name >> --filter=[master] --no-deps
```
siggerzz, line 69: `--no-deps` ensures that if a package is changed, only itself and it's dependencies are built, and **not** it's dependants.

Happy to change this if we want to be safe though?

kevinrodrigues (dismissed):

packages/components/molecules/f-card-with-content/package.json
siggerzz: All fixed!

packages/components/molecules/f-card-with-content/package.json
```diff
@@ -54,8 +54,8 @@
     "@justeat/f-card": "3.x"
   },
   "devDependencies": {
-    "@justeat/f-button": "4.0.0",
-    "@justeat/f-card": "4.0.0",
+    "@justeat/f-button": "3.x",
```
siggerzz: Going to `3.x` for devDeps because that's what the peerDeps specify. @ashleynolan Would you prefer I update the peerDeps to `4.x`? If so, shall I make this a major version bump instead?

packages/components/molecules/f-mega-modal/package.json
```diff
@@ -53,7 +53,7 @@
     "body-scroll-lock": "3.x"
   },
   "devDependencies": {
-    "@justeat/f-button": "4.0.0",
+    "@justeat/f-button": "3.x",
```
siggerzz: ditto

packages/components/molecules/f-card-with-content/package.json
ashleynolan: Yes  please 👍

packages/components/molecules/f-mega-modal/package.json
ashleynolan: Yep, sounds good

kevinrodrigues (dismissed):

.circleci/config.yml
xander-marjoram: I guess this is fine as long as the dependants refer to specific versions. Then they will only be built once you specifically request the latest version.

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+v7.14.0
+------------------------------
+*June 23, 2022*
+
+
+### Changed
+- `turbo.json` to specify that dependencies should be built along with the changed package. 
+### Removed
+- CircleCI Caching to prevent issues with outdated deps.
+
```
xander-marjoram: Nits inc. spacing
```suggestion
v7.14.0
------------------------------
*June 24, 2022*

### Changed
- `turbo.json` to specify that dependencies should be built along with the changed package. 

### Removed
- CircleCI Caching to prevent issues with outdated deps.


```

packages/components/molecules/f-alert/CHANGELOG.md
```diff
@@ -3,6 +3,13 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v6.0.0
+------------------------------
+*Jun 23, 2022*
```
xander-marjoram: Please can you change all the `Jun`s to `June`s? 😄

siggerzz: My bad! 🤦🏼

packages/components/templates/f-template-subnav/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
 
+v0.7.0
+------------------------------
+*June 23, 2022*
+### Changed
+- Bumped devDependencies to prevent issue with conflicting versions of dependancy peerDeps
+
+### Changed
+- Updated `f-navigation-links` to v1.3.0.
+
```
xander-marjoram: A few spacing nits, also I think you've duplicated a bit of the changelog from the previous entry, probably from a merge conflict

```suggestion
v0.7.0
------------------------------
*June 24, 2022*

### Changed
- Bumped devDependencies to prevent issue with conflicting versions of dependency peerDeps


```

turbo.json
```diff
@@ -4,7 +4,7 @@
   "pipeline": {
     "build": {
       "cache": false,
-      "dependsOn": []
+      "dependsOn": ["^build"]
```
xander-marjoram, line 7: What does this do?

siggerzz: Tells Turborepo that any dependencies (that exist in the monorepo) need to be built before attempting to build a package, since we're not using the cache anymore.

packages/components/molecules/f-mega-modal/package.json
```diff
@@ -49,11 +49,11 @@
   ],
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.1.1",
-    "@justeat/f-button": "3.x",
+    "@justeat/f-button": "4.x",
     "body-scroll-lock": "3.x"
```
kevinrodrigues, line 53: Do we need to make this change in the devDep too?

siggerzz: It was already at 4.x 👍

ashleynolan (approved):

kevinrodrigues (approved):

packages/components/molecules/f-mega-modal/package.json
siggerzz: As mentioned, this PR only focuses on JE packages.

PR #1970, opened 2022-06-20
f-media-element@v3.0.0 - update to newer sass syntax

jamieomaguire:
f-media-element: v3.0.0
-----------------------------
*June 20, 2022*

### Changed
- Update to `@use` and `@forward` SASS syntax

Storybook: v0.52.10
*June 20, 2022*
### Changed
- updated vue.config to include f-media-element in components using @use and @forward

### DIFF
--- a/packages/components/molecules/f-media-element/src/components/MediaElement.vue
+++ b/packages/components/molecules/f-media-element/src/components/MediaElement.vue
@@ -129,6 +129,8 @@ export default {
 </script>
 
 <style lang="scss" module>
+@use '@justeat/fozzie/src/scss/fozzie' as f;
+
 $font-sizes: (
     sm: (
         title: heading-s,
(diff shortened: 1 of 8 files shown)


FayeCarter (approved):

packages/components/molecules/f-media-element/src/components/MediaElement.vue
```diff
@@ -181,7 +183,7 @@ $mediaElement-content-margin: spacing(e);
 }
 
 
-@each $name, $value in $breakpoints {
+@each $name, $value in f.$breakpoints {
```
kevinrodrigues, line 186: I think this needs to be called `f.$fozzie-breakpoints` im not 100% sure tho.

jamieomaguire: good point - I'll look into this

kevinrodrigues (approved):

PR #1971, opened 2022-06-20
f-alert@v5.0.1 - Read alert messages to accessible users

tommcclean:
- Alert messages (success and danger) are now read out when displayed on screen.
- This only works if the alert is rendered using `v-show` in the parent component, as `aria-live` needs to be present on DOM load, `v-if` excludes the markup from the DOM

### DIFF
--- a/packages/components/molecules/f-alert/src/components/_tests/Alert.test.js
+++ b/packages/components/molecules/f-alert/src/components/_tests/Alert.test.js
@@ -1,7 +1,12 @@
-import { shallowMount } from '@vue/test-utils';
+import {
+    shallowMount
+} from '@vue/test-utils';
 import FAlert from '../Alert.vue';
 
-const defaultPropsData = { heading: 'Alert title', type: 'info' };
+const defaultPropsData = {
+    heading: 'Alert title',
+    type: 'info'
+};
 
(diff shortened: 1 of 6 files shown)


packages/components/molecules/f-alert/src/components/_tests/Alert.test.js
```diff
@@ -164,4 +164,42 @@ describe('Alert', () => {
             });
         });
     });
+
+    describe('ariaLiveAttribute', () => {
```
jamieomaguire, line 168: do we want to test for the presence of the actual attribute in the html as well?

tommcclean: I imagine if the attribute was missing both tests would fail

tommcclean: I did go down the component test route, but it didn't look obvious how to provide different props, ill have another look and see.

tommcclean: Added a test to ensure the attribute is present

jamieomaguire (approved):

DamianMullins (approved):

PR #1984, opened 2022-06-23
#trivial - Update f skeleton loader 2

kevinrodrigues:
### Changed
- *** Breaking change *** Update to `@use` and `@forward` SASS syntax
- Bump `sass` version.


FayeCarter (dismissed):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "root",
-  "version": "7.13.0",
```
jamieomaguire, line 3: Should this be a major version bump?

jamieomaguire (dismissed):

jamieomaguire (dismissed):

jamieomaguire (dismissed):

jamieomaguire (dismissed):

oliversweb (dismissed):

packages/components/molecules/f-skeleton-loader/src/components/skeletons/RestaurantCarousel.vue
```diff
@@ -56,13 +56,13 @@ $breakpoints: (medium: 520px, large: 960px);
     &:nth-child(1) {
         display: block;
     }
-    @include media('>medium') {
-        margin: 0 spacing(d) 0 0;
+    @include f.media('>520px') {
```
xander-marjoram: How come these have changed from breakpoints to pixels?

packages/components/molecules/f-skeleton-loader/src/components/skeletons/Heading.vue
```diff
@@ -11,9 +11,11 @@ export default {
 
 
 <style lang="scss" module>
+@use '@justeat/fozzie/src/scss/fozzie' as f;
+
 .c-heading {
-    height: $skeleton-loader-heading-height;
+    height: common.$skeleton-loader-heading-height;
```
xander-marjoram, line 17: How is the `common` module brought in?

jamieomaguire: @xander-marjoram it's imported via the vue.config.js file for each component. We added some docs here to explain:
https://github.com/justeat/fozzie-components#reusable-styling-with-commonscss

packages/components/molecules/f-skeleton-loader/src/components/skeletons/RestaurantCarousel.vue
kevinrodrigues: Good spot. I actually have no idea how these have changed, i'm certain I didn't make this change. Will revert.

packages/components/molecules/f-skeleton-loader/src/components/skeletons/Heading.vue
xander-marjoram: Thanks, makes sense once you've read the webpack config!

packages/components/molecules/f-skeleton-loader/src/components/skeletons/RestaurantCarousel.vue
kevinrodrigues: Right I remember now. Further up in this file there's new values overriding the `$breakpoints` from within `fozzie` it was causing errors so I've just passed them in directly instead here.

ashleynolan: I'd probably assign the breakpoint value to a variable, as then you can name it so it's clear it's a specific breakpoint for this component (could call it `$bp-carouselMid` or something)

packages/components/molecules/f-skeleton-loader/src/components/skeletons/RestaurantCarousel.vue
```diff
@@ -56,13 +56,13 @@ $breakpoints: (medium: 520px, large: 960px);
...
+        margin: 0 f.spacing(d) 0 0;
         &:nth-child(2) {
             display: block;
         }
     }
-    @include media('>large') {
+    @include f.media('>960px') {
```
ashleynolan: Same here as above (`$bp-carouselWide`?)

jamieomaguire (approved):

packages/components/molecules/f-skeleton-loader/src/components/skeletons/RestaurantCarousel.vue
kevinrodrigues: Fixed.

BanksySan (approved):
Should we depreciate the previous versions?
https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

kevinrodrigues:
> Should we depreciate the previous versions? https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

I'm not sure we usually follow a process of deprecating a single package or lower when bumping to breaking changes? I think we could if we wanted to but there's probably no real value in doing so. We're covering how to migrate over in-order to use these new versions in some documentation tickets so that should help.

ashleynolan:
You only need to deprecate a package if it's been moved to another repo, or if it's been published by mistake. If it's just a new major version bump, it'll just take over the latest tag in the NPM registry.

PR #1989, opened 2022-06-23
#trivial - update common.scss imports

jamieomaguire:
Uses the default namespace for `common.scss` file across components that are using `@use` and `@forward` Sass syntax.

### DIFF
--- a/packages/components/atoms/f-button/src/components/Button.vue
+++ b/packages/components/atoms/f-button/src/components/Button.vue
@@ -357,7 +357,7 @@ $btn-icon-sizeXSmall-buttonSize        : 32px;
 
 .o-btn--primary,
 .o-btn--icon.o-btn--primary {
-    @include background-color($btn-primary-bgColor);
+    @include common.background-color($btn-primary-bgColor);
 
     &,
     &:link,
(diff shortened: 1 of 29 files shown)


FayeCarter (approved):

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -357,7 +357,7 @@ $btn-icon-sizeXSmall-buttonSize        : 32px;
 
 .o-btn--primary,
 .o-btn--icon.o-btn--primary {
-    @include background-color($btn-primary-bgColor);
+    @include common.background-color($btn-primary-bgColor);
```
kevinrodrigues, line 360: I think we should document this approach somewhere. Maybe the readme within this repo. Just so it's super clear that `common` is referring to the `@use` from the vue config. That way if anyone comes back to these components and say adds a mixin or file within `common.scss` they know how to reference it. It's probably obvious how this is working but worth stating it I think.

jamieomaguire: great point - as agreed I'll raise a separate PR to document this

kevinrodrigues (approved):

PR #1991, opened 2022-06-23
#trivial - update readme on common.scss usage

jamieomaguire:
Add new section to readme that helps understand how to use the `common.scss` file in components


kevinrodrigues (dismissed):
🎖️

PR #1998, opened 2022-06-27
fozzie-components@v7.14.1 - Fix Bundlewatch to only run against changed packages and it's dependants.

siggerzz:
Currently bundlewatch fails when diffing for non-global changes due to use no longer using the CircleCI cache.

This PR should fix Bundlewatch to only run against changed packages and its dependants.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -309,6 +309,7 @@ jobs:
     steps:
       - attach_workspace:
           at: .
+      - detect_root_change
       - run:
          name: Run Bundlewatch checks
          command: yarn bundlewatch
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ commands:
             then
               yarn << parameters.command_name >>
             else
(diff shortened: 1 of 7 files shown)


jamieomaguire (dismissed):

oliversweb (dismissed):

.circleci/config.yml
```diff
@@ -66,7 +66,7 @@ commands:
             then
               yarn << parameters.command_name >>
             else
-              yarn << parameters.command_name >> --filter=[master] --no-deps
+              yarn << parameters.command_name >> --filter=[master]
```
siggerzz, line 69: Removing this argument means that when you have a non-global change, packages that're dependant on your changed package will also get built. This will mean dependant `dist`'s are available to bundlewatch and it'll prevent it from falling over.

oliversweb (approved):

jamieomaguire (approved):

siggerzz:
This build is failing due to the async / await flakiness. This is being addressed in another ticket, so going to merge for now as this is blocking other PR's.

PR #2001, opened 2022-06-27
#trivial - Fix up build.

kevinrodrigues:
Fix up failing SB build.

### DIFF
--- a/packages/storybook/package.json
+++ b/packages/storybook/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/storybook",
   "private":true,
-  "version": "0.52.19",
+  "version": "0.52.21",
   "scripts": {
     "storybook:deploy": "storybook-to-ghpages --script storybook:build",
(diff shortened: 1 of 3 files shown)


packages/storybook/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/storybook",
   "private":true,
-  "version": "0.52.19",
```
kevinrodrigues, line 4: The versions are wrong so i've matched them now.

siggerzz (approved):

packages/storybook/vue.config.js
```diff
@@ -44,6 +44,7 @@ module.exports = {
                     const updateComponentsAndTypes = [
                         'atoms',
                         'molecules',
+                        'f-status-banner',
```
xander-marjoram, line 47: So just checking my understanding, does this mean that f-status-banner is the only `organism` that has been updated?

kevinrodrigues: Yup correct.

PR #2009, opened 2022-06-29
f-checkout@v3.33.0 - fix address mapping

FayeCarter:
### Fixed
- Address mapping from local storage.

### DIFF
--- a/packages/components/pages/f-checkout/src/store/checkout.module.js
+++ b/packages/components/pages/f-checkout/src/store/checkout.module.js
@@ -516,7 +516,14 @@ export default {
             }
 
             if (address) {
-                [state.address.line1, state.address.line2, state.address.line3, state.address.line4] = address.lines;
+                if (address.lines) {
+                    [state.address.line1, state.address.line2, state.address.line3, state.address.line4] = address.lines;
+                } else {
+                    state.address.line1 = address.line1;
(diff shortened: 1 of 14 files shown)


charliearlie (dismissed):

packages/components/pages/f-checkout/src/store/checkout.module.js
```diff
@@ -516,7 +516,14 @@ export default {
             }
 
             if (address) {
-                [state.address.line1, state.address.line2, state.address.line3, state.address.line4] = address.lines;
+                if (address.lines) {
```
jamieomaguire: does this require any additional testing?

jamieomaguire (dismissed):

packages/components/pages/f-checkout/src/store/checkout.module.js
BanksySan: You can use the null coalesce operator here.

BanksySan (dismissed):

packages/components/pages/f-checkout/src/store/checkout.module.js
FayeCarter: No longer using this bit of code 🙂

charliearlie (approved):

jamieomaguire (approved):

PR #2010, opened 2022-06-29
f-registration@v2.4.2 - increase password minlength to 10

jamieomaguire:
v2.4.2
*June 29, 2022*

### Added
- Increase password minimum length from 4 to 10 characters.

### DIFF
--- a/packages/components/pages/f-registration/src/components/_tests/Registration.test.js
+++ b/packages/components/pages/f-registration/src/components/_tests/Registration.test.js
@@ -469,7 +469,7 @@ describe('Registration', () => {
 
             it('should show error message and emit failure event when the password field is populated with too short an input', async () => {
                 // Arrange
-                wrapper.find('[data-test-id="formfield-password-input"]').setValue('dog');
+                wrapper.find('[data-test-id="formfield-password-input"]').setValue('foobarbaz');
 
                 // Act
                 await wrapper.vm.onFormSubmit();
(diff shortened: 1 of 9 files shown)


packages/components/pages/f-registration/src/components/_tests/Registration.test.js
```diff
@@ -469,7 +469,7 @@ describe('Registration', () => {
 
             it('should show error message and emit failure event when the password field is populated with too short an input', async () => {
                 // Arrange
-                wrapper.find('[data-test-id="formfield-password-input"]').setValue('dog');
+                wrapper.find('[data-test-id="formfield-password-input"]').setValue('foobarbaz');
```
jamieomaguire, line 472: A part of me wonders if we should extract the min-length to a constants file and reference it in all test files and the component. That way we'd only need to update the min length in 1 place and tests would keep up to date automatically. 

Keen to hear other people's thoughts on this. Playing devil's advocate it would add a slight layer of obscurity.

oliversweb: I think a bigger refactor would be better in that we move all the validation logic to a separate class rather than leaving it in the template ... we did this in the 'f-account-info' component.

oliversweb (approved):

packages/components/pages/f-registration/src/components/_tests/Registration.test.js
jamieomaguire: Thanks Billy - I'll take a look at `f-account-info` and see how that was done. Will leave this PR as it currently is and see if we could spin up a small ticket to refactor 👍

FayeCarter (approved):

PR #2012, opened 2022-06-30
f-wdio-utils@v1.0.0 - Add Vite and refactor base page object.

siggerzz:
**f-wdio-utils**
v1.0.0
*July 1, 2022*

### Changed
- Structure of page objects

**f-button**
v4.1.0
*July 1, 2022*

### Changed
- Update f-wdio-utils to v1.0.0
- Update component objects / specs to use ES6 syntax.


packages/components/atoms/f-button/test-utils/component-objects/f-button--action.component.js
```diff
@@ -1,6 +1,6 @@
-const Page = require('@justeat/f-wdio-utils/src/base.page');
+import BasePage from '@justeat/f-wdio-utils';
 
-module.exports = class ActionButton extends Page {
+class ActionButton extends BasePage.Page {
```
siggerzz: @tommcclean Is there a way this could be refactored to do the following that I'm not aware about?


```
import BasePage from '@justeat/f-wdio-utils';

class ActionButton extends BasePage {
```

JoshuaNg2332 (approved):

packages/components/atoms/f-button/test/component/f-button.component.spec.js
```diff
@@ -1,74 +1,57 @@
...
-    });
-
     it('should display the f-button action component', async () => {
         // Act
-        await button.load();
+        await ActionButton.load();
 
         // Assert
-        await expect(await button.isComponentDisplayed()).toBe(true);
+        await expect(await ActionButton.isComponentDisplayed()).toBe(true);
```
jamieomaguire, line 11: Tiny NIT: I'd personally probably store `await ActionButton.isComponentDisplayed()` as a variable just for readability. Entirely subjective of course and not a change request :)

packages/components/atoms/f-button/test/visual/f-button.visual.desktop.spec.js
```diff
@@ -67,14 +54,9 @@ describe('f-button Desktop visual tests', () => {
     });
 
     describe('link button', () => {
-        beforeEach(() => {
```
jamieomaguire, line 70: One thing just to be aware of is we're now using the exact same instance of the Button classes for each test. This might be fine, but it does increase the risk of corrupted/mutated state impacting each subsequent test if any actions are performed on the button.

Again it might not be an issue but could created weirdness down the line.

packages/services/f-wdio-utils/package.json
```diff
@@ -27,6 +30,7 @@
     "fozzie"
   ],
   "scripts": {
+    "build": "vite build",
```
jamieomaguire, line 33: 🔥

jamieomaguire (approved):

packages/services/f-wdio-utils/src/page.object.js
```diff
@@ -1,7 +1,7 @@
-const { source } = require('axe-core');
-const { buildUrl } = require('./storybook-extensions');
+import { source } from 'axe-core';
+import { buildUrl } from './storybook-extensions';
 
-class Page {
+export default class Page {
```
oliversweb, line 4: As per my slack comment:
```suggestion
export default class Page {
    constructor (componentType, componentName, componentSelector) {
        this.title = 'Component URLS';
        this.componentType = componentType;
        this.componentName = componentName;
        this.componentSelector = componentSelector;
        this.path = '';
    }

    get component () {
        return $(`[${this.componentSelector]`);
    }
```

packages/components/atoms/f-button/test/component/f-button.component.spec.js
siggerzz: Will be coming back to refactoring this further in another PR, just wanted to use this a test bed for the wdio changes 👍🏼

packages/components/atoms/f-button/test/visual/f-button.visual.desktop.spec.js
siggerzz: I _think_ we should be okay, as this is what WDIO recommend, but will bear that in mind for our refactoring work!

packages/services/f-wdio-utils/src/page.object.js
siggerzz: Will refactor f-wdio-utils in a later PR as there's some other changes we need to make, but can't until tests are refactored to async/await.

oliversweb (approved):

PR #2014, opened 2022-07-04
f-wdio-utils@v1.1.0 - Update Axe implementation to support async

siggerzz:
**f-wdio-utils**
v1.1.0
*July 4, 2022*

### Changed
- Axe Core implementation to use native WDIO plugin.

**f-button**
v4.2.0
*July 4, 2022*

### Changed
- Update f-wdio-utils to v1.1.0 to use new Axe Core implementation.
- Accessibility tests to be async.

### DIFF
--- a/packages/services/f-wdio-utils/src/axe-helper.js
+++ b/packages/services/f-wdio-utils/src/axe-helper.js
@@ -1,0 +1,62 @@
+const AxeBuilder = require('@axe-core/webdriverio').default;
+const fs = require('fs');
+const AxeReports = require('axe-reports');
+const { exec } = require('child_process');
+const { getTestConfiguration } = require ('../../../../test/configuration/configuration-helper');   
+const configuration = getTestConfiguration();
+
+export default new class AxeHelper {
+    /**
+     * Creates a CSV build artifact of the pages accessibility violations.
+     * @param {Object} results - The accessibility violations.
+     * @param {String} componentName - Name of the component that was tested.
+     */
+    processResults (results, componentName) {
+        console.log('Creating .CSV artifact for Axe violations');
+
+        const filePath = `${configuration.testType.violationCSVDirectory}/${componentName}-a11y-violations`;
+
+        // axe-reports can't create the CSV in CI due to permissions so we have to create the file ourselves.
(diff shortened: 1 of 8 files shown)


packages/services/f-wdio-utils/src/page.object.js
```diff
@@ -1,4 +1,4 @@
-import { source } from 'axe-core';
+import AxeHelper from './axe-helper';
```
siggerzz, line 1: This file is pretty much a copy paste from - https://github.com/justeat/fozzie-components/blob/master/test/utils/axe-helper.js

Figured it made sense to move into f-wdio-utils.

Once all components have been refactored the linked file will be removed 👍🏼

packages/services/f-wdio-utils/src/axe-helper.js
```diff
@@ -0,0 +1,62 @@
...
+        console.error(`Expected no accessibility violations. Found: ${results.violations.length}`);
+    };
+
+    /**
+     * Runs accessibility test results against a set of rules defined.
+     *
+     * @returns {ThenArg<ReturnType<BrowserCommandsType["executeAsync"]>>}
+     */
+    async getAccessibilityTestResults () {
+        const builder = new AxeBuilder({ client: browser })
```
siggerzz, line 51: This file is also mostly a copy-paste, with the only new functionality being the new `@axe-core/webdriverio` implementation.

All components using < 1.1.0 of f-wdio-utils will still be using the old implementation.

JoshuaNg2332 (approved):

oliversweb (approved):

packages/services/f-wdio-utils/src/axe-helper.js
```diff
@@ -0,0 +1,64 @@
...
+                        .withTags(['wcag21a', 'wcag21aa', 'wcag143', 'cat.color', 'cat.aria'])
+                        .disableRules(['duplicate-id', 'html-has-lang', 'aria-allowed-role', 'color-contrast-enhanced']);
```
zlatinivanov-je, line 53: Just wonder what is the reason for `duplicate-id` to be disabled?

PR #2015, opened 2022-07-04
Atoms - Refactor tests to use ES6 / async await #trivial

siggerzz:
### Changed
- Update f-wdio-utils to v1.1.0 to use new Axe Core implementation.
- Accessibility tests to be async.
- Specs to ES6 syntax

### DIFF
--- a/packages/components/atoms/f-card/babel.config.js
+++ b/packages/components/atoms/f-card/babel.config.js
@@ -10,12 +10,10 @@ module.exports = api => {
     if (!isTest) {
         api.cache(true); // Caches the computed babel config function – https://babeljs.io/docs/en/config-files#apicache
         presets.push(['@vue/app', { useBuiltIns: builtIns }]);
+    } else {
+        // use current node version for transpiling test files
+        presets.push(['@babel/env', { targets: { node: 'current' } }]);
     }
-
-    // Alias for @babel/preset-env
-    // Hooks into browserslist to provide smart Babel transforms
-    // https://babeljs.io/docs/en/babel-preset-env
-    presets.push('@babel/env');
 
     return {
         presets,
--- a/packages/components/atoms/f-card/test-utils/component-objects/f-card.component.js
+++ b/packages/components/atoms/f-card/test-utils/component-objects/f-card.component.js
@@ -4,6 +4,6 @@ class Card extends Page {
     constructor () {
         super('atom', 'card-component');
     }
-};
+}
 
 export default new Card();
--- a/packages/components/atoms/f-card/test-utils/component-objects/f-card.component.js
+++ b/packages/components/atoms/f-card/test-utils/component-objects/f-card.component.js
@@ -1,7 +1,9 @@
-const Page = require('@justeat/f-wdio-utils/src/base.page');
+import Page from '@justeat/f-wdio-utils';
 
-module.exports = class Card extends Page {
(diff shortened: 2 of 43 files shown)


packages/components/atoms/f-card/babel.config.js
```diff
@@ -10,13 +10,11 @@ module.exports = api => {
     if (!isTest) {
         api.cache(true); // Caches the computed babel config function – https://babeljs.io/docs/en/config-files#apicache
         presets.push(['@vue/app', { useBuiltIns: builtIns }]);
+    } else {
+        // use current node version for transpiling test files
+        presets.push(['@babel/env', { targets: { node: 'current' } }]);
```
siggerzz, line 15: This code is present in all other components, and was preventing ES6 syntax from being used in the tests.

packages/components/atoms/f-card/test-utils/component-objects/f-card.component.js
```diff
@@ -1,7 +1,9 @@
...
 
-module.exports = class Card extends Page {
+class Card extends Page {
     constructor () {
         super('atom', 'card-component');
     }
-};
+}
+
+export default new Card();
```
xander-marjoram, line 9: This is a bit different, what's the benefit of this over just `export default Card;`?

xander-marjoram: Well, naturally it means you don't have to create a new instance of the Card whenever you import it... I just haven't seen it done like this before 😄

xander-marjoram: As long as we don't store any state between tests, it would be okay, but IMO I think it would be fine to `new` up a `Card` for each test just to prevent any leaking between tests in future

siggerzz: Yeah, no state is shared across tests, so we'll be fine. Was just following the Page Object docs on WDIO's site and they follow this same structure - https://webdriver.io/docs/pageobjects/#get--ing-your-selectors

packages/components/atoms/f-popover/test/accessibility/axe-accessibility.spec.js
```diff
@@ -1,16 +1,12 @@
...
-        popover.load();
-    });
-    it('a11y - should test f-popover component WCAG compliance', () => {
+    it('a11y - should test f-popover component WCAG compliance', async () => {
+        // Arrange
+        await Popover.load();
+
         // Act
-        const axeResults = getAxeResults('f-popover');
+        const axeResults = await Popover.getAxeResults('f-popover');
```
xander-marjoram, line 9: Feels like some slight duplication here, like this should be possible:
```suggestion
        const axeResults = await Popover.getAxeResults();
```

Maybe something to look into in future

xander-marjoram (approved):

packages/components/atoms/f-popover/test/accessibility/axe-accessibility.spec.js
siggerzz: Yeah agreed, this is something that can be addressed quite easily. We've got more refactoring work coming, so can be done then 👍🏼

JoshuaNg2332 (approved):

PR #2023, opened 2022-07-07
f-checkout@v4.0.0 - newer sass syntax

jamieomaguire:
v4.0.0
-----------------------------
*July 07, 2022*

### Changed


packages/components/pages/f-checkout/src/components/Accordion.vue
```diff
@@ -60,32 +60,34 @@ export default {
...
+    padding: f.spacing(d) 0;
+    margin-top: f.spacing(d);
+    outline: none;
+    text-decoration: none;
+    text-align: left;
+    width: 100%;
+    .c-checkout-accordion-header-arrow {
+        height: $icon-size;
+        width: $icon-size;
     }
```
jamieomaguire, line 85: All I did was reduce the indentation here. 

Master version of this file: https://github.com/justeat/fozzie-components/blob/master/packages/components/pages/f-checkout/src/components/Accordion.vue

packages/components/pages/f-checkout/src/components/_tests/__snapshots__/Selector.test.js.snap
```diff
@@ -1,14 +1,14 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Selector computed :: orderMethod should show the collection label when the \`serviceType\` is \`collection\` 1`] = `
-<select id="formField-order-time-19" data-test-id="formfield-order-time-dropdown-select" name="order-time" class="FormDropdown_c-formDropdown-select_ROvjN FormDropdown_c-formField-field_2SIjk FormDropdown_c-formField-field--medium_3XEY1">
+<select id="formField-order-time-19" aria-required="true" data-test-id="formfield-order-time-dropdown-select" name="order-time" class="FormDropdown_c-formDropdown-select_ROvjN FormDropdown_c-formField-field_2SIjk FormDropdown_c-formField-field--medium_3XEY1">
```
jamieomaguire: updated due to newer version of `f-form-field` being used

kevinrodrigues (dismissed):

packages/components/pages/f-checkout/src/components/ErrorDialog.vue
```diff
@@ -1,6 +1,7 @@
 <template>
     <mega-modal
         ref="errorModal"
+        :is-text-aligned-center="true"
```
jamieomaguire: this preserves the current text-align center styles on the modal. This was made an opt-in behaviour in newer versions of `f-mega-modal`. If we want to switch to the default left-aligned style, we just need to remove this prop.

I've added it in as a part of this change to keep the percy tests consistent and reduce the scope of this PR.

kevinrodrigues (dismissed):

packages/components/pages/f-checkout/src/components/ErrorDialog.vue
DamianMullins: Minor thing (and feel free to ignore), but you can shorthand a `true` value by just specifying `is-text-aligned-center` as the attribute here.

DamianMullins (dismissed):

packages/components/pages/f-checkout/src/components/ErrorDialog.vue
jamieomaguire: you know what you're absolutely right and I now remember you telling me this exact thing last year on another PR haha - I'll update now. Thanks! :D

DamianMullins: I do love my attribute shorthands!

DamianMullins (dismissed):

packages/components/pages/f-checkout/package.json
```diff
@@ -66,16 +66,17 @@
...
-    "@justeat/f-button": "3.3.0",
-    "@justeat/f-card": "3.6.0",
-    "@justeat/f-card-with-content": "1.1.0",
-    "@justeat/f-error-message": "1.1.0",
+    "@justeat/f-alert": "6.0.0",
+    "@justeat/f-button": "4.2.0",
+    "@justeat/f-card": "4.1.0",
+    "@justeat/f-card-with-content": "3.0.0",
+    "@justeat/f-error-message": "2.1.0",
     "@justeat/f-form-field": "4.9.0",
```
jamieomaguire, line 74: not updating this because the asterisks changes to `f-form-field` probably warrant a good combing through which should be its own ticket

kevinrodrigues (approved):

FayeCarter (approved):

PR #2027, opened 2022-07-08
f-loyalty@v2.0.0 - new sass syntax

jamieomaguire:
v2.0.0
-----------------------------
*July 08, 2022*

### Changed


FayeCarter (dismissed):

packages/components/pages/f-loyalty/src/components/Terms.vue
```diff
@@ -24,13 +26,13 @@ export default {
...
 
 .c-loyalty-terms-link {
-    margin: spacing(e) 0;
+    margin: f.spacing(e) 0;
     text-align: center;
-    font-weight: $font-body-strong-s-weight;
+    font-weight: f.$font-body-strong-s-weight;
     text-decoration: none;
 
     &:hover {
```
kevinrodrigues, line 34: Worth adding a focus to this while we are here? `:focus`

kevinrodrigues (approved):

packages/components/pages/f-loyalty/src/components/Terms.vue
jamieomaguire: I'm not entirely sure - I had a look manually and it does utilise outlines in a clear and obvious way. Might be worth leaving out of this change but revisiting as part of the usability testing we've started doing on all these components?

FayeCarter (approved):

PR #2028, opened 2022-07-08
f-account-info@1.2.0 (and all other pages) - Move components to peerDependencies

xander-marjoram:
## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


packages/components/pages/f-checkout/package.json
```diff
@@ -62,18 +62,26 @@
...
-    "@justeat/f-error-message": "2.1.0",
-    "@justeat/f-form-field": "4.9.0",
-    "@justeat/f-link": "3.2.0",
-    "@justeat/f-mega-modal": "7.0.0",
+    "@justeat/f-alert": "6.x",
+    "@justeat/f-button": "4.x",
+    "@justeat/f-card": "4.x",
+    "@justeat/f-card-with-content": "3.x",
+    "@justeat/f-error-message": "2.x",
+    "@justeat/f-form-field": "4.x",
```
jamieomaguire: I think there needs to be some discussion with the checkout team here. By default I think recent versions of `f-form-field` mean that every field will be marked as required with an asterisk. I believe some of the fields are meant to be left as optional.

xander-marjoram: Oh okay, thanks.
Sounds like f-form-field should be updated so that the fields aren't required by default...

jamieomaguire: I think that could be a good idea. I vaguely recall that being proposed in refinement a while back but I think it may have been decided that we go with the reverse - might be worth us starting up that conversation again with the team.

xander-marjoram: This has been done as part of #2030

packages/components/pages/f-offers/package.json
```diff
@@ -54,19 +54,23 @@
...
+    "@justeat/f-trak": "0.x",
     "@braze/web-sdk": "3.3.0",
+    "js-cookie": "2.2.1",
     "vuex": ">=3.5.1"
   },
   "devDependencies": {
     "@braze/web-sdk": "3.3.0",
-    "@justeat/f-content-cards": "8.0.0-beta.4",
     "@justeat/f-button": "4.x",
+    "@justeat/f-content-cards": "8.0.0-beta.4",
```
jamieomaguire, line 69: I assume this beta can't be added as a peer dep too?

jamieomaguire (approved):

packages/components/pages/f-offers/package.json
xander-marjoram: I _think_ I tried `8.0.0-beta.x` but it didn't like it

packages/components/pages/f-contact-preferences/package.json
```diff
@@ -56,15 +56,20 @@
...
+    "@justeat/f-form-field": "6.x",
     "vuex": ">=3.5.1"
   },
   "devDependencies": {
+    "@justeat/f-alert": "6.x",
+    "@justeat/f-button": "4.x",
+    "@justeat/f-card": "4.x",
+    "@justeat/f-card-with-content": "3.x",
+    "@justeat/f-form-field": "6.x",
     "@justeat/f-vue-icons": "3.3.0",
```
siggerzz, line 72: Any reason for leaving this?

siggerzz (approved):

packages/components/pages/f-contact-preferences/package.json
xander-marjoram: My thinking was that I'd try and keep the complexity down to start with by sticking only to components for now

PR #2029, opened 2022-07-12
f-offers@v2.0.0-beta.1

jamieomaguire:
v2.0.0-beta.1
-----------------------------
*July 12, 2022*

### Changed


FayeCarter (dismissed):

packages/components/pages/f-offers/CHANGELOG.md
```diff
@@ -3,6 +3,15 @@
...
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+
+v1.12.0-beta.3
+-----------------------------
+*July 12, 2022*
+
+### Changed
+- **breaking changes** Update to `@use` and `@forward` SASS syntax
+
```
xander-marjoram: When the latest changelog entry is a `Latest` and you add another one, they should be merged together.

In other words, delete the `Latest` entry and add
```
- Refactor WebDriverIO tests to use async in order to support Node 16 using `codemod` utility.
```
to your `Changed` section

packages/components/pages/f-offers/package.json
```diff
@@ -62,12 +62,13 @@
   "devDependencies": {
     "@braze/web-sdk": "3.3.0",
     "@justeat/f-content-cards": "8.0.0-beta.4",
-    "@justeat/f-button": "3.0.2",
-    "@justeat/f-media-element": "2.0.0",
+    "@justeat/f-button": "4.2.0",
```
siggerzz: Update to `4.x`

packages/components/pages/f-offers/package.json
```diff
@@ -62,12 +62,13 @@
...
+    "@justeat/f-media-element": "3.1.0",
```
siggerzz: ditto, but `3.x`

packages/components/pages/f-offers/package.json
```diff
@@ -62,12 +62,13 @@
...
     "@justeat/f-searchbox": "6.7.0",
     "@justeat/f-trak": "0.7.1",
-    "@justeat/f-mega-modal": "3.0.0",
+    "@justeat/f-mega-modal": "7.1.0",
```
siggerzz: ditto but `7.x`

packages/storybook/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v0.52.31
+------------------------------
+*July 07, 2022*
```
xander-marjoram: Nit: 📆, and please remove leading `0`s from dates 😄

packages/components/pages/f-offers/src/assets/scss/common.scss
```diff
@@ -1 +1 @@
-@import '@justeat/fozzie/src/scss/fozzie';
+// Add styles here so it can be injected first via vue.config.js.
```
xander-marjoram, line 1: Would a `@forward` work here so fozzie doesn't have to be imported in each component file?

siggerzz (changes requested):
Minor comments around devDep versions

packages/storybook/CHANGELOG.md
jamieomaguire: Happy to - is that the general pattern everyone should be following or just a personal preference?

packages/components/pages/f-offers/src/assets/scss/common.scss
jamieomaguire: It would, but we don't want to because then it becomes less apparent where the values come from. When starting the work we felt that `f.spacing()` was more useful than `spacing()` etc

siggerzz (commented):
New package.json changes look good to me. Will leave the CSS changes to others :)

siggerzz (dismissed):
LGTM

packages/storybook/CHANGELOG.md
xander-marjoram: It was the original pattern, so I'm just trying to keep it consistent 👍🏻

kevinrodrigues (approved):

packages/components/pages/f-offers/package.json
kevinrodrigues: I thought we're only updating `peerDep` related packages that exist in `devDeps` too? I could be wrong as I can't fully remember now.

FayeCarter (approved):

PR #2030, opened 2022-07-12
f-form-field@6.0.0 - Replace `isRequired` prop with `required` attribute

xander-marjoram:
### Removed
- **BREAKING**: `isRequired` prop in favour of the standard HTML attribute.
- `aria-required` attributes as they were not adding anything.
  - `required` behaves as expected for all major screen reader / browser combinations, except Chrome and Firefox with Android Talkback. `aria-required` also does not work for these combinations.

### Changed
- Now handles `required` attribute and passes this through to each element, e.g., `input`, `select`, `textarea`.

### Added
- `isVisuallyRequired` prop.
  - When `true`, required fields will display a visual indicator that the field is required, e.g., an asterisk.
  - When `false`, fields can still be required but no visual indicator will be shown.
  - If the field is not required then the indicator will not be shown, regardless of this prop's value.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/atoms/f-form-field/src/components/FormDropdown.vue
+++ b/packages/components/atoms/f-form-field/src/components/FormDropdown.vue
@@ -11,7 +11,6 @@
                     [$style['c-formField-padding--iconLeading']]: hasIcon
                 }]"
             :disabled="attributes.disabled"
-            :aria-required="isRequired"
             :data-test-id="testId.select"
             v-bind="attributes"
             :value="value"
@@ -46,7 +45,7 @@ export default {
     props: {
         attributes: {
             type: Object,
-            default: () => {}
+            default: () => ({})
         },
         value: {
             type: String,
@@ -67,10 +66,6 @@ export default {
         hasIcon: {
             type: Boolean,
             default: false
-        },
-        isRequired: {
-            type: Boolean,
-            default: true
         }
     },
 
--- a/packages/components/atoms/f-form-field/src/components/FormField.vue
+++ b/packages/components/atoms/f-form-field/src/components/FormField.vue
@@ -15,7 +15,7 @@
                 v-if="shouldShowLabel"
                 :label-for="uniqueId"
                 :is-disabled="isDisabled"
-                :is-visually-required="$attrs.required && isVisuallyRequired"
+                :is-visually-required="showVisuallyRequiredIndicator"
                 v-bind="$props"
                 :data-test-id="testId.label">
                 {{ labelText }}
@@ -41,7 +41,6 @@
                 v-else-if="isSelectionControl"
                 :id="uniqueId"
                 :attributes="$attrs"
-                :is-visually-required="isVisuallyRequired"
                 v-bind="$props"
                 v-on="listeners" />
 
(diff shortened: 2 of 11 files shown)


xander-marjoram:
All browser-based behaviour is working as intended.
TODO: Test screen readers.
- [x] NVDA
- [ ] JAWS
- [ ] Voiceover
- [ ] Talkback

packages/components/atoms/f-form-field/src/components/FormDropdown.vue
```diff
@@ -46,7 +45,7 @@ export default {
     props: {
         attributes: {
             type: Object,
-            default: () => {}
+            default: () => ({})
```
jamieomaguire, line 48: good spot here!

jamieomaguire (dismissed):

packages/components/atoms/f-form-field/src/components/FormFieldAffixed.vue
```diff
@@ -53,7 +52,7 @@ export default {
     props: {
         attributes: {
             type: Object,
-            default: () => {}
+            default: () => ({})
```
ashleynolan, line 55: More out of interest than anything else – what does this change actually do? (mainly, was it breaking before because of the syntax error)

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -40,6 +41,7 @@
                 v-else-if="isSelectionControl"
                 :id="uniqueId"
                 :attributes="$attrs"
+                :is-visually-required="isVisuallyRequired"
```
ashleynolan: As the props are already being bound to this component (on line 45), shouldn't `isVisuallyRequired` already be available in the `form-selection-control` component?

packages/components/atoms/f-form-field/src/components/FormField.vue
```diff
@@ -15,6 +15,7 @@
                 v-if="shouldShowLabel"
                 :label-for="uniqueId"
                 :is-disabled="isDisabled"
+                :is-visually-required="$attrs.required && isVisuallyRequired"
```
ashleynolan: I'd be tempted to pull this out as a computed and then you could just pass that computedProp into `form-label` here and the `form-selection-control`.

Kinda subjective though, so if you thought there was more reason to leave it as it is, that's fine too.

packages/components/atoms/f-form-field/src/components/FormFieldAffixed.vue
xander-marjoram: It means the default value returns an empty object rather than undefined:
20220713T153949

packages/components/atoms/f-form-field/src/components/FormField.vue
xander-marjoram: Yep, good spot.

jamieomaguire (approved):

ashleynolan (approved):

PR #2031, opened 2022-07-13
f-header@10.2.0 - Add a theme to support header on Search

1itvinka:
### Added
- Theme to support header on Search

---

**Desktop view:**
Screenshot 2022-07-13 at 10 05 07

**Mobile view:**
Screenshot 2022-07-13 at 09 59 05

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

- [x] Safari
- [x] Firefox


packages/components/organisms/f-header/src/components/Header.vue
```diff
@@ -173,6 +174,10 @@ export default {
             return this.headerBackgroundTheme === 'transparent' && !this.mobileNavIsOpen;
         },
 
+        showSearchHeader () {
```
jamieomaguire: I'm not sure if we should add such domain-specific language here. I understand its solving a search problem, but is it possible to re-word it to be a little more generic/reusable?

1itvinka: happy to rename it, but I'm not sure what could be a good name for this theme. It was confirmed by the design team that this variant will be used only on search.

ashleynolan: Agree with @jamieomaguire – I'd name it closer to how the theme looks/behaves than the page it's used on. Just as it's only used on Search now, doesn't mean that it'll never be on another page in the future.

Is the main difference in the visual theme that it's orange on narrow screens and white on wide screens?

jamieomaguire:
Do we need to add a visual test to check the new variant?

jamieomaguire (changes requested):
Could we add desktop and mobile visual tests to support the new variant please?

packages/components/organisms/f-header/src/components/Header.vue
1itvinka: It's white on narrow now, and transparent with no gradient or `box-shadow` on wide screens.

ashleynolan: I'd probably call it something like `disappearingWhite`. As long as you document what it does on narrow and wide in the prop for the theme, then that should be good then.

jamieomaguire (approved):

DamianMullins (approved):

PR #2033, opened 2022-07-13
f-registration@v3.0.0 - Use new sass syntax.

kevinrodrigues:
### Changed
- **breaking changes** Update to `@use` and `@forward` SASS syntax.

### DIFF
--- a/packages/components/pages/f-registration/package.json
+++ b/packages/components/pages/f-registration/package.json
@@ -59,9 +59,9 @@
   },
   "devDependencies": {
     "@justeat/fozzie": "9.0.0-beta.6",
-    "@justeat/f-button": "4.2.0",
-    "@justeat/f-card": "4.1.0",
-    "@justeat/f-error-message": "2.1.0",
+    "@justeat/f-button": "4.x",
+    "@justeat/f-card": "4.x",
+    "@justeat/f-error-message": "2.x",
     "@justeat/f-form-field": "5.0.0",
     "@justeat/f-http": "0.8.0",
     "@justeat/f-link": "3.1.1",
--- a/packages/components/pages/f-registration/package.json
+++ b/packages/components/pages/f-registration/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-registration",
   "description": "Fozzie Registration Form Component",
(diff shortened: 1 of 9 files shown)


packages/components/pages/f-registration/package.json
```diff
@@ -58,9 +58,10 @@
     "@justeat/browserslist-config-fozzie": ">=1.1.1"
   },
   "devDependencies": {
-    "@justeat/f-button": "3.0.2",
```
jamieomaguire, line 61: I spoke to Ben earlier and my understanding is for these we should use `major.x` for now (excluding the fozzie beta)

jamieomaguire (dismissed):

packages/components/pages/f-registration/package.json
kevinrodrigues: Fixed.

jamieomaguire (approved):

packages/components/pages/f-registration/package.json
```diff
@@ -58,9 +58,10 @@
...
-    "@justeat/f-card": "3.0.0",
-    "@justeat/f-error-message": "1.0.0",
+    "@justeat/fozzie": "9.0.0-beta.6",
+    "@justeat/f-button": "4.x",
```
zlatinivanov-je, line 62: I may be missing something here, but what is the logic behind defining the version of the dependencies in such way?

zlatinivanov-je (approved):
Just a question to clarify the decision on versioning dependencies

packages/components/pages/f-registration/package.json
```diff
@@ -58,9 +58,10 @@
...
+    "@justeat/f-card": "4.x",
+    "@justeat/f-error-message": "2.x",
```
zlatinivanov-je, line 64: Don't see the use of this? `@justeat/f-error-message`

packages/components/pages/f-registration/package.json
kevinrodrigues: There's a PR which explains but in short the devDeps need to match the peerDep so yarn pulls down the correct versions. https://github.com/justeat/fozzie-components/pull/1956

PR #2038, opened 2022-07-15
f-alert@v6.1.0 (and other molecules) - Add Node 16 support

siggerzz:
This PR:
Adds Node 16 support to Molecule components
Refactors Storybook / Bundlewatch configs to use turborepo instead of lerna
Uses new Sass syntax in Storybook
Removes the `lerna` dependency


packages/storybook/config/storybook/story.config.js
```diff
@@ -4,20 +4,22 @@ let outputChangedComponentPackages;
...
 
-    const changedComponentPackages = changedPackagesArray
-        .filter(pkg => pkg.location.match(new RegExp('packages/components')));
+    const changedComponentPackages = changedPackagesArray.tasks
+        .filter(pkg => pkg.directory.match(new RegExp('packages/components')));
 
     const storyPaths = [];
 
-    changedComponentPackages.forEach(pkg => storyPaths.push(`${pkg.location}/stories/*.stories.@(js|mdx)`));
+    changedComponentPackages.forEach(pkg => storyPaths.push(`../../../../${pkg.directory}/stories/*.stories.@(js|mdx)`));
```
zlatinivanov-je, line 22: Is this part guaranteed to work everywhere `../../../../${pkg.directory}/stories/*.stories.@(js|mdx)`? 
Would something like `__dirname` be helpful?

zlatinivanov-je (approved):

packages/storybook/config/storybook/story.config.js
siggerzz: Yep, this is command is relative to this file, so the path won't change 👍🏼

JoshuaNg2332 (approved):

PR #2046, opened 2022-07-21
f-header@10.3.0 - Remove "disappearingWhite" theme

1itvinka:
In this PR we revert the changes made in https://github.com/justeat/fozzie-components/pull/2031. On Search we actually can use the default ('white') theme with minor style changes which can be added in the SW project. It will be easier to support than a separate theme.

## UI Review Checks

Screenshot 2022-07-21 at 11 01 44

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


jamieomaguire:
@1itvinka - if the theme you originally added works, and it's a valid design style that may be used on another application in the future then it might be worth keeping this theme. Otherwise another app will have to copy the changes in SearchWeb or create this PR again. 

If it's not going to be used anywhere else, then it would be good to understand why SearchWeb deviates from the existing themes? (a question for design I'd imagine)

It's also worth considering the visual regression testing that would be lost by not keeping the theme styling within Fozzie. SearchWeb probably won't have Percy capabilities for while, so might be another reason to keep it in here.

1itvinka:
Hey @jamieomaguire, thank you so much for having a look at this PR!

Happy to keep the theme, but it seems to me that it makes more sense to remove it, as I don't think we should keep extra code in the project just in case someone might need it in the future. We can always add it back.  

Great point about Percy tests! But because there're some other custom bits on the header on Search, we'll need to rely on manual visual testing anyway.

ashleynolan:
@1itvinka I think the reason it may be good to keep it, is it acts as a kind of documentation as to the variations of header styles across the microsites. That way if we do ever want to normalise the header styles, they're all in one place and we can see all the variations in Storybook.

yarn.lock
```diff
@@ -7290,6 +7290,13 @@ axios@0.21.4, axios@^0.21.1, axios@^0.21.2:
   dependencies:
     follow-redirects "^1.14.0"
 
+axios@0.24.0:
```
dawidchar, line 7293: This does not look right 🤔 maybe we can ignore the yarn.lock file from this PR as it seems unrelated 😄

dawidchar: I'm assuming one of the dependencies now requires Axios, so might be fine to keep 🤔

1itvinka:
> @1itvinka I think the reason it may be good to keep it, is it acts as a kind of documentation as to the variations of header styles across the microsites. That way if we do ever want to normalise the header styles, they're all in one place and we can see all the variations in Storybook.

Great point! Let's keep it then.

PR #2048, opened 2022-07-22
f-mfa@v0.3.0 - Added/Styled MFA page template

oliversweb:
### Added
- MFA Verification Page Template

___

**Approved by UX 👍** 

| Figma | Actual |
| --- | --- |
| image | image |


packages/components/pages/f-mfa/package.json
```diff
@@ -46,12 +46,19 @@
...
-    "@justeat/f-services": "1.16.0"
+    "@justeat/f-services": "1.16.0",
+    "@justeat/f-vue-icons": "3.3.0",
+    "@justeat/f-globalisation": "1.0.0"
   },
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.2.0"
   },
   "devDependencies": {
+    "@justeat/f-card": "4.1.0",
```
siggerzz: These should be:

```
    "@justeat/f-card": "4.x",
     "@justeat/f-button": "4.x",
     "@justeat/f-link": "3.x",
     "@justeat/f-card-with-content": "3.x",
```

packages/components/pages/f-mfa/package.json
```diff
@@ -46,12 +46,19 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-services": "1.16.0"
+    "@justeat/f-services": "1.16.0",
+    "@justeat/f-vue-icons": "3.3.0",
```
siggerzz: I _think_ we should be using `"@justeattakeaway/pie-icons-vue": "1.0.0"`?

Is this right, @ashleynolan?

siggerzz (changes requested):
Minor comments.

packages/components/pages/f-mfa/package.json
oliversweb: The icon I need (`BagSurfBgIcon`) not present in `@justeattakeaway/pie-icons-vue": "1.0.0"`

packages/components/pages/f-mfa/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-mfa",
   "description": "Fozzie Mfa - Multi-factor Authenticator Input Form",
-  "version": "0.2.0",
+  "version": "0.1.1",
```
tommcclean: This version is lower than the previous version

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -2,37 +2,147 @@
...
+            has-inner-spacing-large
+            card-size-custom="medium"
+            has-outline>
+            <bag-surf-bg-icon
+                :class="$style['c-mfa-icon']" />
+            <h2>
+                {{ $t('verificationPage.subTitle') }}
+            </h2>
+            <p
+                v-html="$t('verificationPage.instructionsPrimaryText', { email })" />
```
tommcclean, line 17: Thinking about this

tommcclean: I would personally recommend using a regular expression to validate the email before using it in a v-html. To prevent user input from being renderable in the page (script execution etc). It does come back from the API, but can we trust that this cannot be cheated up the stack?

I would either pull the email out of the v-html or validate it cautiously.

xander-marjoram: I think this might be better solved with _component interpolation_:

https://kazupon.github.io/vue-i18n/guide/interpolation.html#component-interpolation

oliversweb: @tommcclean there is a ticket for validation.   This value come will be delivered via the querystring and will be validated for script injection, etc.

packages/components/pages/f-mfa/package.json
ashleynolan: Yeh, it's fine to use f-vue-icons for those legacy images for now, until we work out how we want to serve those images and fully deprecate the old icons package (as these splash icons aren't in the new iconset)

packages/components/pages/f-mfa/package.json
ashleynolan: Should be `0.3.0`

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -43,11 +153,41 @@ export default {
...
-    min-height: 80vh;
-    width: 80vw;
     margin: auto;
-    border: 1px solid red;
-    font-family: f.$font-family-base;
-    @include f.font-size(heading-m);
+    text-align: center;
+
+    @include f.media('>=narrow') {
+        width: 80vw;
```
ashleynolan: Should there be a `max-width` set for this for users on very wide screens? Otherwise, I imagine it could get a bit wide at very large widths

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -43,11 +153,41 @@ export default {
...
+    }
+
+    h2 {
```
ashleynolan: I would add a class to the `h2` (`c-mfa-card`) and add a selector based on that. JUst means if the tag ever changes in the future for some reason, the CSS won't break

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -43,11 +153,41 @@ export default {
...
+}
+
+.c-mfa-card {
+    @include f.media('>=narrow') {
+        box-shadow: f.$elevation-01;
+    }
+}
+
+.c-mfa-formField {
+    text-align: left;
```
ashleynolan, line 198: Do you need to define text-align here? (I would have thought it was default left anyway)

ashleynolan: Thin I see why you need this, and it comes from adding `text-align: center` to the `c-mfa`?

One potential fix for this would be for us to add `text-align: left;` to the `f-form-field` component input so that this doesn't get overriden by styles set higher up like this one (don't need to do as part of this PR, but would be good to do this).

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -2,37 +2,147 @@
...
+                    required="true"
+                    :is-visually-required=false
+                    maxlength="10"
+                    :label-text="$t('verificationPage.formField.labelText')"
+                    :placeholder="$t('verificationPage.formField.placeholderText')" />
+
+                <f-button
+                    :class="$style['c-mfa-submit']"
+                    data-test-id="mfa-submit-button"
+                    :disabled="verificationCode.trim().length < 1"
```
kevinrodrigues: Could you move this to a computed? `isDisabled`

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -2,37 +2,147 @@
...
+                    button-type="primary"
```
kevinrodrigues: I might be wrong but i think f-button is default primary?

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -2,37 +2,147 @@
...
+                    action-type="submit"
+                    :is-loading="isSubmitting">
+                    {{ $t('verificationPage.submitButtonText') }}
+                </f-button>
+            </form>
+
+            <f-button
+                :class="$style['c-mfa-help-info']"
```
kevinrodrigues: Should the name of this and the line below be more in sync? i.e c-mfa-need-help-link

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -2,37 +2,147 @@
...
+                data-test-id="mfa-need-help-link"
+                button-type=link
```
kevinrodrigues: Does this need quotes? 🤔

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -2,37 +2,147 @@
...
+            isSubmitting: false,
+            showErrorPage: false
         };
+    },
+
+    mounted () {
+        this.initialise();
+    },
+
+    methods: {
```
kevinrodrigues, line 120: Will you be adding tests later?

kevinrodrigues: Actually ignore that i can see it's all test code anyway.

packages/components/pages/f-mfa/src/tenants/en-GB.js
```diff
@@ -1,4 +1,32 @@
-export default {
+const messages = {
     locale: 'en-GB',
-    text: 'I am a VMfa Component (GB)'
+    text: 'I am a VMfa Component (GB)',
+    verificationPage: {
+        subTitle: "We've emailed your login code",
+        instructionsPrimaryText: "We've sent a verification code to <b>{email}</b> to help keep your account secure. Enter it below to log in.",
```
kevinrodrigues: I think this should be `<strong>{email}</strong>`.

packages/components/pages/f-mfa/src/components/Mfa.vue
kevinrodrigues: I think you can also just do `!verificationCode.trim().length` without comparing to a integer. Probably makes it less readable but a comment in the computed will work or happy to leave as is.

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -2,37 +2,147 @@
...
+            <form
+                method="post"
+                novalidate
+                @submit.prevent="onFormSubmit">
+                <form-field
+                    v-model.lazy="verificationCode"
+                    :class="$style['c-mfa-formField']"
+                    data-test-id="mfa-verification-code-textbox"
+                    required="true"
+                    :is-visually-required=false
```
kevinrodrigues: Quotes?

packages/components/pages/f-mfa/src/components/Mfa.vue
tommcclean: I would think twice about using the query string for PII. I like the idea of component interpolation for that also.

When I discussed with them earlier in the month they suggested a GET endpoint would provide the email, better to do that than stick it in the logs

kevinrodrigues (dismissed):

packages/components/pages/f-mfa/src/tenants/en-GB.js
xander-marjoram: Will update this soon, at the same time as replacing the `v-html`

packages/components/pages/f-mfa/src/components/Mfa.vue
xander-marjoram: I've got another PR (#2063) to fix an issue with the visually required indicator, so I could try and include it in that.

packages/components/pages/f-mfa/src/components/Mfa.vue
xander-marjoram: This didn't seem to actually be doing much, so I've removed it and replaced with some different logic. Had to wrap the card contents in a div to get the width and padding right.

kevinrodrigues (dismissed):

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -2,37 +2,159 @@
...
+        /**
+        * TODO
+        */
+        async onFormSubmit () {
+            this.isSubmitting = true;
+            try {
+                // TODO
+                // eslint-disable-next-line no-promise-executor-return
+                await new Promise(resolve => setTimeout(resolve, 1000)) // DEBUG - Simulate network delay
+                .then(console.log('DEBUG - onFormSubmit', { verificationCode: this.verificationCode, code: this.code, email: this.email })); // eslint-disable-line no-console
```
Ariel-Rodriguez, line 144: Not sure if intended, probably a typo here.
Bear in mind this `console.log` is performing before the promise is fulfilled.

Given `await` is in place on previous line, you may opt to just not use `.then` at all on this line.

oliversweb: @Ariel-Rodriguez Point noted, it's just a bit of debug code ... I will be ripped out in next ticket 👍

tommcclean: You can just simulate this in your browser btw on the network tab

packages/components/pages/f-mfa/src/components/Mfa.vue
oliversweb: They are not going with an endpoint and are passing via querystring

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -2,37 +2,159 @@
...
+            <div :class="$style['c-mfa-card-content']">
+                <bag-surf-bg-icon
+                    :class="$style['c-mfa-icon']" />
+
+                <h2 :class="$style['c-mfa-card-heading']">
+                    {{ $t('verificationPage.subTitle') }}
+                </h2>
+
+                <i18n
+                    path="verificationPage.instructionsPrimaryText"
```
xander-marjoram, line 19: Further reading here: https://kazupon.github.io/vue-i18n/guide/interpolation.html#basic-usage

xander-marjoram (dismissed):

packages/components/pages/f-mfa/package.json
```diff
@@ -46,13 +46,21 @@
...
+    "@justeat/f-card": "4.x",
+    "@justeat/f-form-field": "6.x",
+    "@justeat/f-button": "4.x"
   },
   "devDependencies": {
-    "@justeat/f-wdio-utils": "1.x",
+    "@justeat/f-button": "4.x",
+    "@justeat/f-card": "4.x",
+    "@justeat/f-form-field": "6.x",
+    "@justeat/f-wdio-utils": "0.12.0",
```
siggerzz: Change to `1.x`

siggerzz (changes requested):
Minor comment regarding f-wdio-utils

xander-marjoram (approved):

siggerzz (approved):

PR #2052, opened 2022-07-25
f-wdio-utils@v1.2.0 (and other services) - Add node 16 support to services

siggerzz:


JoshuaNg2332 (dismissed):

packages/services/f-braze-adapter/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-braze-adapter",
   "description": "Fozzie Metadata Service",
-  "version": "5.0.1",
```
BanksySan, line 4: Should be `v0.6.0`.

BanksySan (changes requested):

packages/services/f-braze-adapter/package.json
siggerzz: Whoops. The typo was in the changelog. Will update it to be 5.1.0

JoshuaNg2332 (approved):

PR #2056, opened 2022-07-26
f-header@v10.4.0 (and other components) - Update to node 16 compatible dependencies

siggerzz:
Update the following components to have Node 16 compatible dependencies:

f-header
f-searchbox

### DIFF
--- a/packages/components/molecules/f-searchbox/package.json
+++ b/packages/components/molecules/f-searchbox/package.json
@@ -71,7 +71,7 @@
     "@justeat/f-button": "3.x",
     "@justeat/f-error-message": "1.x",
     "@justeat/f-globalisation": "0.x",
-    "@justeat/f-mega-modal": "3.x",
+    "@justeat/f-mega-modal": "7.x",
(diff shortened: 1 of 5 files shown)


packages/components/molecules/f-searchbox/package.json
```diff
@@ -63,16 +63,16 @@
     "@justeat/browserslist-config-fozzie": ">=1.1.1",
     "@justeat/f-button": "3.x",
     "@justeat/f-error-message": "1.x",
-    "@justeat/f-mega-modal": "3.x",
+    "@justeat/f-mega-modal": "7.x",
```
siggerzz, line 66: As mentioned, this version of Searchbox isn't being used in any consuming applications, but I've tested locally in Storybook with these new major versions and it seems to function fine!

Unfortunately, we have to upgrade all components to the latest deps, otherwise `yarn` will fall over when we eventually turn on Node 16.

kevinrodrigues (approved):

DamianMullins (approved):

PR #2062, opened 2022-07-27
fozzie-components@v7.21.0 - Update components to use Node 16 compatible version of f-services #trival #globalconfig

siggerzz:
This PR
- Updates any dependants of `@justeat/f-services` to use Node 16 compatible version.
- Refactors `@justeat/f-user-message` to use new major version of `@justeat/f-services`, so that it's inline with other components.

### DIFF
--- a/packages/components/pages/f-checkout/package.json
+++ b/packages/components/pages/f-checkout/package.json
@@ -3,7 +3,7 @@
   "description": "Fozzie Checkout - Fozzie Checkout Component",
   "version": "4.4.0",
   "main": "dist/f-checkout.umd.min.js",
-  "maxBundleSize": "130kB",
+  "maxBundleSize": "140kB",
   "files": [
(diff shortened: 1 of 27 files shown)


kevinrodrigues (approved):

packages/components/pages/f-checkout/package.json
```diff
@@ -3,7 +3,7 @@
   "description": "Fozzie Checkout - Fozzie Checkout Component",
   "version": "4.4.0",
   "main": "dist/f-checkout.umd.min.js",
-  "maxBundleSize": "130kB",
+  "maxBundleSize": "140kB",
```
siggerzz, line 6: Bundlesize increased as the dependencies use the new version of f-services which is slightly bigger.

jamieomaguire (approved):

PR #2064, opened 2022-07-28
f-registration@v3.4.0 - Update dependencies to be Node 16 compatible.

siggerzz:
v3.4.0
*July 28, 2022*

### Added
- Node 16 compatible version of `@justeat/f-services`.
- Node 16 compatible version of `@justeat/f-error-message`.
- Node 16 compatible version of `@justeat/f-vue-icons`.
- Node 16 compatible peerDependency version of `@justeat/f-button`.
- - Node 16 compatible peerDependency version of `@justeat/f-card`.

Note: Update f-card peerDep


packages/components/pages/f-registration/package.json
```diff
@@ -51,14 +51,14 @@
...
   ],
   "dependencies": {
-    "@justeat/f-services": "1.4.0",
+    "@justeat/f-services": "1.x",
     "vuelidate": "0.7.6"
   },
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.1.1",
-    "@justeat/f-button": "3.x",
+    "@justeat/f-button": "4.x",
```
siggerzz, line 59: This should've been updated to `4.x` a while back. As you can see on line 68, we already test f-registration with `f-button@4.0.0`, so we know it's compatible with this version 👍🏼

packages/components/pages/f-registration/package.json
```diff
@@ -51,14 +51,14 @@
...
     "@justeat/f-card": "3.x",
-    "@justeat/f-error-message": "1.x",
+    "@justeat/f-error-message": "2.x",
```
siggerzz, line 61: Although this updates to a new major version, the only thing that's been added to `f-error-message` is new Sass syntax. This dependency will continue to function as it did before.

packages/components/pages/f-registration/package.json
```diff
@@ -71,7 +71,7 @@
     "@justeat/f-form-field": "6.x",
     "@justeat/f-http": "0.x",
     "@justeat/f-link": "3.x",
-    "@justeat/f-vue-icons": "2.7.0",
+    "@justeat/f-vue-icons": "3.x",
```
siggerzz, line 74: ditto.

adammorr (dismissed):

kevinrodrigues (approved):

kathywklee (approved):

PR #2069, opened 2022-08-01
f-mfa@v0.4.0 - Added Api provider class

oliversweb:
### Added
- Api provider for call to AccountWeb (`PostChallenge()`)

### DIFF
--- a/packages/components/pages/f-mfa/src/constants.js
+++ b/packages/components/pages/f-mfa/src/constants.js
@@ -1,4 +1,4 @@
-export const SUBMIT_LOGIN_CHALLENGE_URL = 'api/account/mfa/login';
+export const SUBMIT_LOGIN_CHALLENGE_URL = 'applications/coreweb/validate-mfa-token';
 
 export const CONVERSATION_ID_HEADER_NAME = 'x-je-conversation';
 
--- a/packages/components/pages/f-mfa/src/constants.js
+++ b/packages/components/pages/f-mfa/src/constants.js
@@ -1,0 +1,5 @@
+export const SUBMIT_LOGIN_CHALLENGE_URL = 'api/account/mfa/login';
+
+export const CONVERSATION_ID_HEADER_NAME = 'x-je-conversation';
+
+export const CONVERSATION_ID_COOKIE_NAME = 'x-je-conversation';
--- a/packages/components/pages/f-mfa/src/services/_tests/AccountWeb.api.test.js
+++ b/packages/components/pages/f-mfa/src/services/_tests/AccountWeb.api.test.js
@@ -1,0 +1,82 @@
+import AccountWebApi from '../AccountWeb.api';
+
+import {
+    SUBMIT_LOGIN_CHALLENGE_URL,
+    CONVERSATION_ID_HEADER_NAME,
+    CONVERSATION_ID_COOKIE_NAME
+} from '../../constants';
+import {
+    baseUrl,
+    conversationId,
+    postChallengeBody
+} from '../../../test-utils/setup';
+
+let apiProvider;
+let httpClientMock;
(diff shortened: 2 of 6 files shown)


packages/components/pages/f-mfa/src/constants.js
```diff
@@ -0,0 +1,5 @@
+export const SUBMIT_LOGIN_CHALLENGE_URL = 'api/account/mfa/login';
```
siggerzz: I feel like this should perhaps be a prop that's provided to our consumers? At least that's what f-registration / f-checkout do - https://github.com/justeat/fozzie-components/blob/master/packages/components/pages/f-registration/src/components/Registration.vue#L258-L260

siggerzz (changes requested):
Minor comment.

siggerzz (dismissed):

siggerzz (approved):

packages/components/pages/f-mfa/src/services/_tests/AccountWeb.api.test.js
```diff
@@ -0,0 +1,82 @@
...
+            // Assert
+            expect(httpPostSpy).toHaveBeenCalledWith(expectedUri, expectedBody, expectedHeaders);
+        });
+
+        it('should set the cookie with a new conversation id if not provided with one', async () => {
+            // Act
+            await apiProvider.postValidateMfaToken(postBody);
+
+            // Assert
+            expect(cookiesSetSpy).toHaveBeenCalledWith(CONVERSATION_ID_COOKIE_NAME, expect.anything());
```
kevinrodrigues, line 71: How comes this is `expect.anything()`? Shouldn't it be at least expecting a string or some sort of formatted expectation?

packages/components/pages/f-mfa/src/services/_tests/AccountWeb.api.test.js
```diff
@@ -0,0 +1,82 @@
...
+        });
+
+        it('should not set the cookie if already provided with a conversation id', async () => {
```
kevinrodrigues, line 74: Minor but i think this description should read. `should not call cookieSet method if already provided with a conversation id`

kevinrodrigues (approved):

packages/components/pages/f-mfa/src/services/_tests/AccountWeb.api.test.js
oliversweb: @kevinrodrigues  Because this test is to test the conversation id is being created (i.e. calling the cookie.set()) and nothing else plus it is very difficult to pre-determine the random guid that will be created for the conversation id ... plus not that revealing even if we could.

packages/components/pages/f-mfa/src/services/_tests/AccountWeb.api.test.js
oliversweb: @kevinrodrigues I will change on one of the next pr's we do on this component, cool?

packages/components/pages/f-mfa/src/services/_tests/AccountWeb.api.test.js
kevinrodrigues: To me (and could just be me) looks like the test is checking to make sure cookieSetSpy is:
1. Being called correctly.
2. Is being called with the right inputs.

The correct inputs was my worry as it should reflect the expected type. The coversation ID i suspect would be a unique id generating a string so this test would only need to change to expect.any(String)`. Happy to leave as is tho.

oliversweb: Sure, will change to `expect.any(String)` on next pr 👍

PR #2077, opened 2022-08-04
fozzie-components@v7.23.0 - Add turborepo caching + optimise CI

siggerzz:
v7.23.0
*August 4, 2022*

### Added
- Turborepo local caching for `build` / `lint` / `test` / tasks
- New concurrency rules in CircleCI config
- Disabled npm publishing / Storybook deploy

### Changed
- Allure report generation to execute on fail to speed up CI.


.circleci/config.yml
```diff
@@ -359,38 +366,40 @@ jobs:
       - run:
           name: Allure - Generate Test Results
           command: yarn allure:generate
-          when: always
+          when: on_fail
```
xander-marjoram, line 369: 🚀

xander-marjoram:
What's the plan with getting storybook deployments to work again? We probably shouldn't have them disabled for long

siggerzz:
> What's the plan with getting storybook deployments to work again? We probably shouldn't have them disabled for long

For sure. Given we'll be starting work on the Github actions epic, we agreed not to spend time fixing it in Circle, as I've already spent a lot of time without success, and it seems to be a CI issue.

I'll double-check it works locally, but we may just have to manually deploy until we migrate.

turbo.json
```diff
@@ -3,50 +3,50 @@
   "baseBranch": "origin/master",
   "pipeline": {
     "build": {
-      "cache": false,
+      "cache": true,
```
kevinrodrigues, line 6: I think these default to `true`? Could these all be removed then?

turbo.json
```diff
@@ -154,15 +154,15 @@
       "dependsOn": []
     },
     "lint": {
-      "cache": false,
+      "cache": true,
       "dependsOn": []
     },
     "lint:fix": {
       "cache": false,
```
kevinrodrigues, line 161: What's the reason for leaving this as false for my understanding?

siggerzz: I just wanted to try and keep the changes minimal to task I know 100% benefit from caching. I'll be looking to add more in the future, but wanted to just enable caching gradually 👍🏼

turbo.json
siggerzz: It's more for visibility to be honest, just so it's really obvious where we're utilising it 👍🏼

kevinrodrigues (approved):

xander-marjoram (approved):

PR #2079, opened 2022-08-05
f-header@10.9.0 - Add Corporate Ordering

dawidchar:
### Fixed
- JE Logo vertical positioning.
- Country text & logo alignment in country selector

### Changed
- Friendly Name Truncation Tweak Points

### Added
- `showCorporateLink` - Shows "Corporate Ordering" Link.
- `isCondensed` - A flag to hide icons, reduce spacing between nav-links on mid screen size, reduce container padding on tweak points and apply stricter friendly name truncation.

_(Changes Approved By UX & Design)_

#### Design
image

### Narrow
image

#### **Without** Condensed Mid (Broken)
image
#### **With** Condensed Mid
image
image
image

### Wide
image
image

### Huge
image

#### Before
image

#### After
image

### Country Selector Alignment
Screenshot_Country Selector

Screenshot_Country Selector Mobile


packages/components/organisms/f-header/test/component/f-header.component.spec.js
```diff
@@ -3,7 +3,7 @@ import Header from '../../test-utils/component-objects/f-header.component';
 describe('Desktop - f-header component tests', () => {
     beforeEach(async () => {
         // Act
-        await Header.load({ showOffersLink: true, showDeliveryEnquiry: true });
+        await Header.load({ showOffersLink: true, showDeliveryEnquiry: true, showCorporateLink: true });
```
siggerzz, line 6: Thanks for this! 🚀

siggerzz:
Would it be possible to add Visual Regression tests for this?

You should be able to add a new object here for Desktop tests - https://github.com/justeat/fozzie-components/blob/master/packages/components/organisms/f-header/test/visual/f-header.visual.desktop.spec.js#L59-L63


And here for Mobile - https://github.com/justeat/fozzie-components/blob/master/packages/components/organisms/f-header/test/visual/f-header.visual.mobile.spec.js#L84-L88


Edit: Just noticed these tests check the prop is **not** displayed 🤦🏼.

If you happy for me too, I can add a isDisplayed visual test for this prop as part of this PR?

ashleynolan:
How does this look on the edge of breakpoints (say at 770px for instance?). It's already pretty tight with all links displayed currently. so if that could be a possibility, then worth checking for this scenario.

Attached image as an example:

Screenshot 2022-08-05 at 15 20 50

dawidchar:
Yea, with all the elements on, it does appear like there might be a problem on smaller desktop sizes 🤔😅 

image

dawidchar:
I guess this issue does appear to already be a bug on the current header 😅 

image

ashleynolan:
@dawidchar Worth chatting to design/product about that then, as if they want all the links displayed at once, will need a solution for that 😆 

Guessing the solution would be, don't show all the links at once, or design will need to come up with something for these mid sizes (I think extending the mobile nav to higher widths would look a bit strange potentially if it was that wide)

ashleynolan:
@dawidchar Urgh – I'll raise an issue for that bug with the country selector. I think that's because of the length of name being a bit longer than what we test with currently, so we'll need to update that and come up with something to sort that out.

dawidchar:
@ashleynolan Yea, I'll get in touch with my product manager about this issue and see what we can do about it 😅 

It might be worth for us to truncate the name at some width because there are probably users with much longer names who might be more prone to this issue.

ashleynolan:
@dawidchar Yeh, I've raised a ticket for it (will DM you the link). I think truncating the friendlyName will be the solution for that issue as you say, but may need additional space if we're also adding another link in as well.

siggerzz (dismissed):

packages/components/organisms/f-header/src/assets/scss/common.scss
```diff
@@ -42,3 +42,10 @@ $header--transparent-opacity        : 0.7;
         padding-top: 0;
     }
 }
+
+// Hides element in-between the mid and wide breakpoint
+.hide-on-mid {
+    @include f.media('>mid', '<wide') {
+        display: none !important;
```
sonnyprince, line 49: All LGTM - my only question, Is this `!important` needed?

dawidchar: Unfortunately yes, otherwise the `none` is overwritten by `flex`

Luckily, the context of that class, as a global util, makes sense because if you apply that class, you'd expect it to be forced 😄 

https://github.com/justeat/fozzie-components/blob/a00ac8ade7db0956850ae2e3021543797bfcf64a/packages/components/organisms/f-header/src/components/NavLink.vue#L198-L200

packages/components/organisms/f-header/src/components/Header.vue
```diff
@@ -91,6 +97,11 @@ export default {
             default: false
         },
 
+        isCondensedOnMid: {
```
kevinrodrigues: Does the name of this need to be associated to the breakpoint? Might be worth just calling it `isCondensed` that way if we decide to use it on another breakpoint it is a little more flexible?

packages/components/organisms/f-header/src/components/Header.vue
```diff
@@ -116,6 +127,11 @@ export default {
             default: false
         },
 
+        showCorporateLink: {
```
kevinrodrigues, line 130: Could you add these to the readme please? (if it isn't there already)

packages/components/organisms/f-header/src/components/Header.vue
dawidchar: Yea, makes sense - I'll update that now 😄 I have actually changed a tiny bit of the wide breakpoint, so that will suit better

packages/components/organisms/f-header/src/assets/scss/common.scss
jamieomaguire: does this kind of class warrant being added as a global utility class here?
https://github.com/justeat/fozzie-components/blob/master/packages/tools/fozzie/src/scss/trumps/_rwd.scss

sonnyprince: That's fair! TY for explanation :)

sonnyprince (dismissed):

kevinrodrigues:
LGTM just a few comments.

packages/components/organisms/f-header/src/components/Header.vue
kevinrodrigues: I think adding the point about it is currently used on mid in the readme will be sufficient here :)

packages/components/organisms/f-header/src/assets/scss/common.scss
dawidchar: > does this kind of class warrant being added as a global utility class here? https://github.com/justeat/fozzie-components/blob/master/packages/tools/fozzie/src/scss/trumps/_rwd.scss

Yea, that could be useful! 😄 @jamieomaguire

packages/components/organisms/f-header/src/components/Header.vue
dawidchar: I added "_Where appropriate_" in the README & `header.stories.js` and kept "_on mid_" in the `changelog.md` 😄

kevinrodrigues (approved):

packages/components/organisms/f-header/src/assets/scss/common.scss
kevinrodrigues: Could you get away with just using a show/hide statement inside vue to display what you need instead? i.e `v-show="isCondensed"`

dawidchar: Unfortuatley no, as that would hide the icons on all breakpoints, we only want to hide the icons on the mid breakpoint. I also want to stay away from getting the screen size through JS as that would cause the icons to flash on load while the JS is still loading 😄

sonnyprince (approved):

packages/components/organisms/f-header/src/components/NavLink.vue
```diff
@@ -7,10 +7,15 @@
             [$style['c-navLink']]: !isPopoverLink,
             [$style['c-navLink--popoverLink']]: isPopoverLink,
             [$style['c-nav-list-link--alt']]: isAltColour,
+            [$style['c-navLink--condensedMid']]: isCondensed,
```
ashleynolan, line 10: Sorry for the post-merge comment – with the update to the prop, probably makes sense to update the classname to just be `c-navLink--condensed` as well?

packages/components/organisms/f-header/src/components/Navigation.vue
```diff
@@ -695,6 +747,12 @@ export default {
     }
 }
 
+.c-nav-list--condensed .c-nav-list-item--horizontallyAlignedAboveMid {
```
ashleynolan, line 750: You could group this in the selector above:

```scss
.c-nav-list-item--horizontallyAlignedAboveMid {
    @include f.media('>mid') {
        float: left;
    }
    @include f.media('>wide') {
        margin-left: f.spacing(c);
    }

    .c-nav-list--condensed & {
        @include f.media('>mid', '<huge') {
            margin-left: 0;
        }
    }
}

packages/components/organisms/f-header/src/components/NavLink.vue
dawidchar: No worries, yea I spotted that after merging too, I've got another upcoming piece of work to do with f-header, so I'll make the change there, I'll also move the `hide-on-mid` class into fozzie as per @jamieomaguire's comment 😄 Alternatively I'll raise a patch PR for these small changes if I don't get around to that work soon. 😄

packages/components/organisms/f-header/src/components/Navigation.vue
dawidchar: Perfect! I'll get that added too! 😄 Thank you!

packages/components/organisms/f-header/src/components/NavLink.vue
ashleynolan: Nice one 👍

PR #2081, opened 2022-08-08
f-mfa@v0.6.0 - Added validation and full page error

oliversweb:
### Added
- Validation of the input data (querystring).
- Error page
- Emit Event upon success with the return url


packages/components/pages/f-mfa/src/components/_tests/Mfa.test.js
```diff
@@ -80,36 +106,195 @@ describe('Mfa', () => {
...
-        ['show', true],
-        ['NOT show', false]
-    ])('should %s error message when hasSubmitError is %s', async (_, hasSubmitError) => {
-        // Arrange
+        ['enable', 'populated', 'abc123', undefined],
+        ['disable', 'unpopulated', '', 'true']
+    ])('should %s the submit button if the otp field is %s', async (_, __, optValue, expected) => {
+        // Act
         wrapper = await mountSut();
+        await wrapper.setData({ otp: optValue });
```
xander-marjoram: Nit: typo in variable name, and it can also be simplified slightly
```suggestion
    ])('should %s the submit button if the otp field is %s', async (_, __, otp, expected) => {
        // Arrange
        wrapper = await mountSut();
        
        // Act
        await wrapper.setData({ otp });
```

packages/components/pages/f-mfa/src/components/_tests/Mfa.test.js
```diff
@@ -80,36 +106,195 @@ describe('Mfa', () => {
...
+                    }
+                }
+            };
+
+            // Act
+            wrapper = await mountSut({ mocks: sutOverrideMocks });
+
+            // Assert
+            const errorCard = wrapper.find('[data-test-id="mfa-error-page"]');
+            expect(errorCard.exists()).toBe(expected);
```
xander-marjoram: Probably a bit more readable to just do `.toBe(true)` and lose the variable

packages/components/pages/f-mfa/src/components/_tests/Mfa.test.js
```diff
@@ -80,36 +106,195 @@ describe('Mfa', () => {
...
+                    query: queryMock
+                }
+            };
+
+            // Act
+            wrapper = await mountSut({ mocks: sutOverrideMocks });
+
+            // Assert
+            const errorCard = wrapper.find('[data-test-id="mfa-error-page"]');
+            expect(errorCard.exists()).toBe(expected);
```
xander-marjoram: Same here

packages/components/pages/f-mfa/src/components/_tests/Mfa.test.js
```diff
@@ -80,36 +106,195 @@ describe('Mfa', () => {
...
+            // Assert
+            const validationError = wrapper.find('[data-test-id="mfa-form-validation-error"]');
+            expect(validationError.exists()).toBe(showValidationError);
+        });
+
+        describe('And the form data is valid', () => {
+            it('should call the postValidateMfaToken() method with the correct params', async () => {
+                // Arrange
+                mockPostValidateMfaToken = jest.fn(() => {}); // Add default mock to the spy
+                const expectedParams = { mfa_token: 'ABC123', otp: 'test-otp' }; // eslint-disable-line camelcase
```
xander-marjoram: Is the `mfa_token` needed for this test? Should it use the routeMock value by default?

packages/components/pages/f-mfa/src/components/_tests/Mfa.test.js
```diff
@@ -80,36 +106,195 @@ describe('Mfa', () => {
...
+                    wrapper = await mountSut();
+                    await wrapper.setData({ otp: 'test-otp' });
+
+                    // Act
+                    await wrapper.vm.onFormSubmit();
+
+                    // Assert
+                    expect(errorLogSpy).toHaveBeenCalledWith('Bad request when submitting MFA', err400, ['account-pages', 'mfa']);
+                    expect(wrapper.vm.isSubmitting).toBe(false);
+                });
```
xander-marjoram: I reckon you could combine some of these tests like this, for example:
(I haven't tested this so I wouldn't recommend committing directly)
```suggestion
                it.each(
                    ['bad request', 400, 'Bad request when submitting MFA'],
                    ['throttled request', 429, 'Throttled when submitting MFA'],
                )('should log %s if the error status is %s', async (errorMessage, status, exceptionMessage) => {
                    // Arrange
                    const error = new Error(errorMessage);
                    error.response = { status };
                    mockPostValidateMfaToken = jest.fn(() => { throw error; });
                    wrapper = await mountSut();
                    await wrapper.setData({ otp: 'test-otp' });

                    // Act
                    await wrapper.vm.onFormSubmit();

                    // Assert
                    expect(errorLogSpy).toHaveBeenCalledWith(exceptionMessage, error, ['account-pages', 'mfa']);
                    expect(wrapper.vm.isSubmitting).toBe(false);
                });
```

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -207,18 +243,60 @@ export default {
...
+        *
+        * @param {string} key - The key to search for in the query string.
+        * @param {string} field - The data field to store the value against.
+        * @param {string} regex - The regex to validate the value against.
+        */
+        processQueryString (key, field, regex) {
+            if (this.showErrorPage === true) {
+                return;
+            }
+            let err = new Error(`Missing key in querystring or failed regex [${key}]`);
```
kevinrodrigues: Should the creation of this be moved to within the error block? Also how does this work as it looks like it is reassigned to `error` later?

packages/components/pages/f-mfa/src/components/_tests/Mfa.test.js
```diff
@@ -80,36 +106,180 @@ describe('Mfa', () => {
...
+                    const expectedUrl = '/place/i/came/from';
+                    wrapper = await mountSut();
+                    await wrapper.setData({ otp: 'test-otp' });
+
+                    // Act
+                    await wrapper.vm.onFormSubmit();
+
+                    expect(wrapper.emitted()[REDIRECT_URL_EVENT_NAME]).toBeTruthy();
+                    expect(wrapper.emitted()[REDIRECT_URL_EVENT_NAME].length).toEqual(1);
+                    expect(wrapper.emitted()[REDIRECT_URL_EVENT_NAME][0]).toEqual([expectedUrl]);
```
kevinrodrigues, line 277: I think you only need this one. The others should be covered in this one expect (apart from the `.toEqual(1)` not sure how important that is?)

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -207,18 +243,60 @@ export default {
...
+        * supplied property indicated by the key.
+        * If the key is not present in the query string or the value is not valid, the
+        * showErrorPage flag is raised and the issue is logged.
+        *
+        * @param {string} key - The key to search for in the query string.
+        * @param {string} field - The data field to store the value against.
+        * @param {string} regex - The regex to validate the value against.
+        */
+        processQueryString (key, field, regex) {
+            if (this.showErrorPage === true) {
```
kevinrodrigues, line 258: here you can just do the below unless there some other value that comes back which you're trying to avoid?

`if (this.showErrorPage) return;`

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -207,18 +243,60 @@ export default {
...
+        emitRedirectEvent (url) {
+            const pathRaw = url.trim();
+            const path = `${pathRaw.charAt(0) === '/' ? pathRaw : `/${pathRaw}`}`;
+            this.$emit(REDIRECT_URL_EVENT_NAME, path);
+        },
+
+        /**
+        * Validates the otp value is within acceptable bounds (currently 1-10 characters).
+        */
         validateOtp () {
```
kevinrodrigues: Could this be called `isOtpValid()`.

packages/components/pages/f-mfa/src/components/Mfa.vue
oliversweb: Not a fan of inline `if`'s as they can make it harder to read.

packages/components/pages/f-mfa/src/components/Mfa.vue
oliversweb: Is does need to be declared outside but now I look at it I can give a better error message if I assign inside 👍

xander-marjoram (approved):

kevinrodrigues (approved):

PR #2082, opened 2022-08-08
fozzie-components@v7.24.0 - Create GitHub Action for building / deploying Storybook

siggerzz:
This PR adds the following GitHub Actions jobs to run against master:
- Install
- Build
- Storybook Deploy jobs

Tested Storybook deploy and working as expected.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ name: Fozzie Components CI
 
 on:
   pull_request:
-    branches: [ "master", "storybook-actions-test" ]
+    branches: [ "master" ]
 
 jobs:
   install:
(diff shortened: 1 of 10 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,124 @@
...
+on:
+  pull_request:
+    branches: [ "master" ]
+
+jobs:
+  install:
+    runs-on: ubuntu-latest
+    steps:
+    # Checkout the Repo
+    - name: Checkout
```
siggerzz, line 16: You'll notice some steps (`checkout` / `setup-node` etc) are duplicated across jobs.

This is because each job runs inside of its own container. Unlike CircleCI, GitHub Actions as far as I can tell, doesn't have the concept of persisting the state of the container for use in later jobs.

We do utilise caches though, so these steps shouldn't add any build time.

We also need to find a way to reuse these steps across jobs to reduce duplication.

kevinrodrigues (approved):
LGTM 🎉

ashleynolan (approved):

PR #2086, opened 2022-08-09
f-mfa@v0.8.1 - Storybook story with controls for props and api simulations

oliversweb:
### Changed
- Prepared the storybook story to handle changing of the properties and simulating the different api responses

### DIFF
--- a/packages/components/pages/f-mfa/stories/story.helper.js
+++ b/packages/components/pages/f-mfa/stories/story.helper.js
@@ -21,28 +21,27 @@ const apiStates = {
 
 // Mocks for the API calls
 const otpPOST200 = {
-    url: `/${SUBMIT_LOGIN_CHALLENGE_URL}`,
+    url: `http://localhost:8080/${SUBMIT_LOGIN_CHALLENGE_URL}`,
     method: httpVerbs.post,
     responseStatus: httpStatusCodes.ok,
-    // eslint-disable-next-line camelcase
-    requestData: { mfa_token: 'XYZ1234', otp: 'test-otp' },
-    responseData: null
+    requestData: { mfa_token: 'ABC123', otp: 'otp123' }, // eslint-disable-line camelcase
+    responseData: {}
 };
 
 const otpPOST400 = {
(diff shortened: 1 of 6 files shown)


packages/components/pages/f-mfa/stories/story.helper.js
```diff
@@ -0,0 +1,98 @@
...
+
+// Available states for the API mock
+const apiStates = {
+    none: 'no-issues',
+    apiPostFailed400: 'post-otp-fails-due-to-bad-request',
+    apiPostFailed429: 'post-otp-fails-due-to-throttled-request'
+};
+
+// Mocks for the API calls
+const otpPOST200 = {
```
kevinrodrigues, line 23: Do you need these definitions? You could just pass them as objects directly into the `apiDefinitions`? Will be all in once place and no need for a const as the `apiStates` indicate the status codes.

packages/components/pages/f-mfa/stories/story.helper.js
```diff
@@ -0,0 +1,98 @@
...
+} from '../src/constants';
+
+const httpStatusCodes = {
+    ok: 200,
+    badrequest: 400,
+    throttledrequest: 429
+};
+
+const httpVerbs = {
+    post: 'POST'
```
kevinrodrigues, line 12: could just pass this in directly without the const.

kevinrodrigues (approved):

packages/components/pages/f-mfa/stories/story.helper.js
oliversweb: Because in the future (and in other examples) you may want to mix and match GET/POST scenarios and assign them to different definitions ... so if you created them within the definitions then you would be duplicating code.

packages/components/pages/f-mfa/stories/story.helper.js
oliversweb: By creating then as an enumerate of constant values you can use them multiple times within the definition section and never be concerned about typo bugs.

xander-marjoram (approved):

PR #2088, opened 2022-08-10
f-mfa@0.9.0 - Expect complete validateUrl rather than building up from smartGatewayBaseUrl

xander-marjoram:
This follows the same pattern as f-registration which also works in the same way. The API requests will be made to the same domain as the website, so no base URL is required.

Also updated storybook to emit visible events (in the actions tab) when on a (simulated) successful request.

---

### Changed
- Take in the entire MFA `validateUrl` instead of building it up from `smartGatewayBaseUrl`.

### Fixed
- Pass `returnUrl` through to Help component.

### DIFF
--- a/packages/components/pages/f-mfa/stories/story.helper.js
+++ b/packages/components/pages/f-mfa/stories/story.helper.js
@@ -1,7 +1,3 @@
-import {
-    SUBMIT_LOGIN_CHALLENGE_URL
-} from '../src/constants';
-
 const httpStatusCodes = {
     ok: 200,
     badrequest: 400,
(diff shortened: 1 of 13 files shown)


kevinrodrigues (dismissed):

packages/components/pages/f-mfa/stories/story.helper.js
```diff
@@ -21,23 +17,23 @@ const apiStates = {
 
 // Mocks for the API calls
 const otpPOST200 = {
-    url: `http://localhost:8080/${SUBMIT_LOGIN_CHALLENGE_URL}`,
+    url: 'http://localhost:8080/mfa/validate',
     method: httpVerbs.post,
     responseStatus: httpStatusCodes.ok,
     requestData: { mfa_token: 'ABC123', otp: 'otp123' }, // eslint-disable-line camelcase
```
kevinrodrigues, line 23: Super minor (not for this pr), It might be worth updating this disabled line to a more specific eslint disable, i think it's camelCase? Same for the others

xander-marjoram: If I'm understanding you correctly, we've already got the `camelcase` specifier 😄

kevinrodrigues: My bad.

siggerzz (approved):

kevinrodrigues (approved):

PR #2097, opened 2022-08-16
storybook@v0.56.0 – Added custom viewports #trivial

ashleynolan:
### Added
- Added Custom Viewports for Storybook testing (based around breakpoints and example devices)


kevinrodrigues (approved):

ashleynolan:
Merging as it's only failing on a bundlewatch build error, and this package doesn't change any package bundle sizes

PR #2100, opened 2022-08-17
fozzie-components@7.26.0 - Add test components to bundles #trivial #globalconfig

siggerzz:


xander-marjoram:
We were seeing "Cannot use import statement outside a module" when using the test components, so we've had to add the test components to the bundle so they get transpiled.

It's not nice to have to increase the bundle sizes so much so we'll raise another ticket to look for a longer-term solution, but this is necessary for now to enable some important releases.

ashleynolan:
@xander-marjoram Does this increase the bundle size for the components in production too?

If so, would be good to understand what the problem is, as the bundle sizes are increasing a lot with this change.

siggerzz:
Closing this as it's no longer needed.

PR #2101, opened 2022-08-17
f-mfa@v0.10.0 - Added GTM Events (plus some minor fixes/tweaks)

oliversweb:
### Changed
- Changed mfa load success log message to `Info`.
- Help screen primary button text.
  
### Removed
- Mfa Form field validation/message.

### Added
- GTM events.
- Back button to error screen.

### DIFF
--- a/packages/components/pages/f-mfa/src/components/Help.vue
+++ b/packages/components/pages/f-mfa/src/components/Help.vue
@@ -48,7 +48,8 @@
                     button-type="link"
                     :href="loginLinkWithReturnUrl"
                     button-size="large"
-                    data-test-id="mfa-help-login-link">
+                    data-test-id="mfa-help-login-link"
+                    @click="recordAnalytics">
                     {{ $t('helpInfo.secondaryButtonText') }}
                 </f-button>
             </div>
@@ -68,6 +69,7 @@ import {
     BagSadBgIcon
 } from '@justeat/f-vue-icons';
 import tenantConfigs from '../tenants';
+import { buildEvent } from '../services/EventBuilder';
 
 export default {
     name: 'VMfa',
@@ -100,6 +102,12 @@ export default {
     computed: {
         loginLinkWithReturnUrl () {
             return `/account/login?returnUrl=${encodeURIComponent(this.returnUrl)}`;
+        }
+    },
+
+    methods: {
+        recordAnalytics () {
+            this.$gtm.pushEvent(buildEvent('s9'));
         }
     }
 };
--- a/packages/components/pages/f-mfa/src/components/Help.vue
+++ b/packages/components/pages/f-mfa/src/components/Help.vue
@@ -29,7 +29,10 @@
                     <i18n
                         path="helpInfo.instructionsPoint4Text"
                         tag="li">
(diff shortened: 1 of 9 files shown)


packages/components/pages/f-mfa/src/components/Help.vue
```diff
@@ -29,7 +29,10 @@
                     <i18n
                         path="helpInfo.instructionsPoint4Text"
                         tag="li">
-                        <a :href="$t('helpInfo.instructionsHelpLink')"><strong>{{ $t('helpInfo.instructionsHelpText') }}</strong></a>
+                        <a
```
tommcclean, line 32: ❓- Will the instructions help link be another page in coreweb? If so could consider a `nuxt-link`

packages/components/pages/f-mfa/src/services/EventBuilder.js
```diff
@@ -0,0 +1,89 @@
...
+                }
+            };
+        case 's4':
+            return {
+                EventName: 'form_error',
+                Parameters: {
+                    component_name: 'multiFactorAuthentication',
+                    component_type: 'form',
+                    form_action: 'error',
+                    error_message: `${data}`
```
kevinrodrigues: here you can just do `data`?

packages/components/pages/f-mfa/src/services/EventBuilder.js
```diff
@@ -0,0 +1,89 @@
+/* eslint-disable camelcase */
+// eslint-disable-next-line import/prefer-default-export
+export const buildEvent = (key, data = '') => {
+    switch (key) {
+        case 's1b':
```
tommcclean: What are these keys? Did we make them up - could the meanings / names be more obvious to somebody not involved?

kevinrodrigues (dismissed):

packages/components/pages/f-mfa/src/services/EventBuilder.js
tommcclean: ℹ️ - Could import keys from constants for consistency and less magic strings

kevinrodrigues: I was going to suggest adding to a constant file with the key describing what it is and the value as `s1b`

packages/components/pages/f-mfa/src/components/Help.vue
oliversweb: No, I think it goes to consumer help

packages/components/pages/f-mfa/src/services/EventBuilder.js
xander-marjoram: The only difference is that this does force it to be a string, even if someone were to pass an object in. TypeScript would save us here!

packages/components/pages/f-mfa/src/components/Help.vue
xander-marjoram: Nope, it'll be linking to a help centre article

kevinrodrigues (dismissed):

tommcclean (dismissed):

tommcclean (dismissed):

kevinrodrigues (approved):

tommcclean (approved):

PR #2103, opened 2022-08-18
f-registration@3.6.0 - Add mfa challenge issued event

luke-andrew-parry:
## Changed

Updated handling of 403 responses to emit an MfaChallengeIssued event.

### DIFF
--- a/packages/components/pages/f-registration/src/components/Registration.vue
+++ b/packages/components/pages/f-registration/src/components/Registration.vue
@@ -498,8 +498,17 @@ export default {
                         return;
                     }
 
-                    if (status === 403 || status === 401) {
+                    if (status === 401) {
                         this.$emit(EventNames.LoginBlocked);
+                        return;
+                    }
+
+                    if (status === 403) {
+                        this.$emit(EventNames.MfaChallengeIssued, {
+                            mfaToken: error.response.data.mfa_token,
+                            mfaTarget: error.response.data.mfa_target
+                        });
+
                         return;
                     }
                 }
--- a/packages/components/pages/f-registration/src/components/_tests/Registration.test.js
+++ b/packages/components/pages/f-registration/src/components/_tests/Registration.test.js
@@ -218,7 +218,7 @@ describe('Registration', () => {
                 expect(wrapper.emitted(EventNames.CreateAccountWarning).length).toBe(1);
             });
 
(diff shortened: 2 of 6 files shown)


packages/components/pages/f-registration/src/components/Registration.vue
```diff
@@ -498,10 +498,19 @@ export default {
                         return;
                     }
 
-                    if (status === 403 || status === 401) {
+                    if (status === 401) {
                         this.$emit(EventNames.LoginBlocked);
                         return;
                     }
+
+                    if (status === 403) {
```
tommcclean, line 506: LGTM, one ❓ - Is there any other reason why a 403 might be returned that won't contain MFA properties, should it potentially check for the existence of MFA data in the error response?

If not - why did we previously capture 403's

luke-andrew-parry: @tommcclean. No we've changed over the registration BFF to return a `401` for reCAPTCHA/Ravelin block. The only time a `403` will be returned now is when a MFA challenge is issued. The idea being that CoreWeb can hook into the event, get the token/target and redirect them to the new MFA page.

tommcclean: Awesome, thanks

tommcclean (approved):

kevinrodrigues (approved):

packages/components/pages/f-registration/src/components/_tests/Registration.test.js
```diff
@@ -247,7 +251,10 @@ describe('Registration', () => {
                 await flushPromises();
 
                 // Assert
-                expect(wrapper.emitted(EventNames.LoginBlocked).length).toBe(1);
+                const mfaChallengeIssuedEvent = wrapper.emitted(EventNames.MfaChallengeIssued);
+                expect(mfaChallengeIssuedEvent.length).toBe(1);
+                expect(mfaChallengeIssuedEvent[0][0].mfaToken).toBe('mfa-token');
+                expect(mfaChallengeIssuedEvent[0][0].mfaTarget).toBe('someone@somewhere.com');
```
xander-marjoram, line 257: We could also check that the `LoginBlocked` event isn't called, à la L258.

PR #2120, opened 2022-09-01
fozzie-components@v7.28.0 - Add linting and unit tests to GitHub Actions

siggerzz:
v7.28.0
*September 2, 2022*

### Added
- Unit testing / Linting to GitHub Actions.

Successful GitHub Actions job can be found here - https://github.com/justeat/fozzie-components/actions/runs/2973006752

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -125,11 +125,8 @@ jobs:
     - name: Build Components
       run: yarn build --cache-dir=".turbo"
     # Unit Test components.
-    - name: Build Components
+    - name: Unit Test Components
       run: yarn ci:test:${{ matrix.component-type}} --concurrency=10
-
-
-
     # Restore Storybook 'storybook-static'
     - name: Cache Storybook Static Directory
       if: matrix.component-type == 'atoms' || matrix.component-type == 'molecules' || matrix.component-type == 'organisms' || matrix.component-type == 'pages'
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,6 @@ on:
   push:
     branches: 
     - master
-    - gh-actions-unit-linting
 
 jobs:
   install:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -122,8 +124,12 @@ jobs:
     # Build components. Only components with out-of-date Turborepo hash will rebuilt, speeding up build time.
     - name: Build Components
       run: yarn build --cache-dir=".turbo"
+    # Unit Test components.
+    - name: Unit Test Components
+      run: yarn ci:test:${{ matrix.component-type}} --concurrency=10
     # Restore Storybook 'storybook-static'
     - name: Cache Storybook Static Directory
+      if: matrix.component-type == 'atoms' || matrix.component-type == 'molecules' || matrix.component-type == 'organisms' || matrix.component-type == 'pages'
```
siggerzz, line 132: In CircleCI, we have separate unit-test / browser-test jobs as you can see here.


However I've made the Unit tests part of the same job and added this conditional, as there's no way to specify:


`needs: unit-test-${{ matrix.component-type}}` like we do in CircleCI

zlatinivanov-je (approved):

.github/workflows/build.yml
tommcclean, line 132: Should it also contain layouts?

.github/workflows/build.yml
```diff
@@ -135,6 +141,7 @@ jobs:
           storybook-cache-
     # Run Component / A11y / Visual Tests
     - name: Run Component / A11y / Visual Tests
+      if: matrix.component-type == 'atoms' || matrix.component-type == 'molecules' || matrix.component-type == 'organisms' || matrix.component-type == 'pages'
```
tommcclean, line 144: Same as above

.github/workflows/build.yml
siggerzz: Not that I'm aware of, as we currently don't test these. However if and when we add tests, then yeah, this will needed to be added.

.github/workflows/build.yml
siggerzz: ditto

tommcclean (approved):

PR #2124, opened 2022-09-05
f-mfa@v0.13.0 - Clear opt field on mount()

oliversweb:
### Changed
- Clear the opt field on mount().
 ___
Test
https://www.loom.com/share/4547c9965cc04fbdb35d3e4d177b8b12

### DIFF
--- a/packages/components/pages/f-mfa/src/components/Mfa.vue
+++ b/packages/components/pages/f-mfa/src/components/Mfa.vue
@@ -195,6 +195,7 @@ export default {
     },
 
     mounted () {
+        this.otp = ''; // Clear the field
         this.initialise();
     },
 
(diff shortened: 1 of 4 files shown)


DamianMullins (approved):

kevinrodrigues (approved):

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -195,6 +195,7 @@ export default {
     },
 
     mounted () {
+        this.otp = ''; // Clear the field
```
kevinrodrigues, line 198: Should test be added to this? i.e on mounted it should not clear the otp out field.

oliversweb: This field will be blank on first load due to the prop being new and empty, the reason for the clearing here is that when you rock back/fwd on the browser buttons we need to clear field ... and this cannot be done via a unit test afaik.

PR #2126, opened 2022-09-06
f-mfa@0.15.0 - Fixes from staff beta

xander-marjoram:
### Changed
- Only make a new API call if there isn't one already in flight.
- Move query string validation from CSR to SSR.
- Logs for 400/429 errors to warnings.
- Protect against null/undefined query string values.

---

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11

### DIFF
--- a/packages/components/pages/f-mfa/src/components/Mfa.vue
+++ b/packages/components/pages/f-mfa/src/components/Mfa.vue
@@ -194,7 +194,7 @@ export default {
         }
     },
 
-    mounted () {
+    created () {
         this.initialise();
     },
 
--- a/packages/components/pages/f-mfa/src/components/Mfa.vue
+++ b/packages/components/pages/f-mfa/src/components/Mfa.vue
@@ -188,7 +188,7 @@ export default {
 
         errorPrimaryButton () {
             return {
-                href: this.cleanUrl(this.returnUrl),
+                href: this.cleanUrl(this.returnUrl || '/'),
                 text: this.$t('errorMessages.loading.primaryButtonText')
             };
         }
@@ -210,7 +210,7 @@ export default {
         initialise () {
             this.showErrorPage = false;
             this.validateProperty(this.returnUrl, 'returnUrl', RETURN_URL_REGEX);
-            this.validateProperty(this.email.toLowerCase(), 'email', EMAIL_RFC5322_REGEX);
+            this.validateProperty(this.email?.toLowerCase(), 'email', EMAIL_RFC5322_REGEX);
             this.validateProperty(this.code, 'code', MFA_CODE_REGEX);
 
             if (this.showErrorPage) {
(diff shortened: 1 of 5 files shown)


ashleynolan (dismissed):

packages/components/pages/f-mfa/src/components/Mfa.vue
```diff
@@ -189,14 +189,18 @@ export default {
...
     },
 
+    created () {
+        this.initialise(); // Query string validation should happen during SSR to prevent reflecting malicious input
+    },
+
     mounted () {
-        this.initialise();
+        this.otp = ''; // Clear the field
     },
```
oliversweb, line 204: I've removed this need to clear the field now ... unsure why it's still appearing on your pr tho

ashleynolan (approved):

packages/components/pages/f-mfa/src/components/_tests/Mfa.test.js
```diff
@@ -267,5 +267,28 @@ describe('Mfa', () => {
...
+            });
+            pushEventSpy.mockClear();
+
+            // Act
+            await wrapper.vm.onFormSubmit();
+
+            // Assert
+            expect(mockPostValidateMfaToken).not.toHaveBeenCalled();
+            expect(pushEventSpy).not.toHaveBeenCalled();
+            expect(wrapper.vm.isSubmitting).toBe(true);
```
kevinrodrigues, line 291: I don't think you need this line.

oliversweb (approved):

packages/components/pages/f-mfa/src/components/_tests/Mfa.test.js
oliversweb: will fix on next pr, nw

packages/components/pages/f-mfa/src/components/Mfa.vue
oliversweb: will fix on next pr, nw

packages/components/pages/f-mfa/src/components/_tests/Mfa.test.js
xander-marjoram: @kevinrodrigues @oliversweb This line is helpful, it checks that we haven't re-enabled the button. It's relevant because of the `finally` block, which should be skipped over if the button is clicked for a second time, and should only be called when the first request comes back.

packages/components/pages/f-mfa/src/components/Mfa.vue
xander-marjoram: Probably snuck through a merge conflict

packages/components/pages/f-mfa/src/components/_tests/Mfa.test.js
kevinrodrigues: Ok i see that makes sense now, I hadn't realised it was related to the early return. 👍  I would suggest splitting tests out with a description, that way it's a little easier to understand the unit of test.

```javascript
describe('And "isSubmitting" is truthy', () => ..
   it('should exit early to prevent multiple requests')
````

PR #2137, opened 2022-09-16
f-http@v1.0.0 - Pass Conversation ID Headers

tommcclean:
## Q&A

<details>
    <summary>Why accept a function rather than just check cookies?</summary>
Because this needs to work isomorphically, it isn't possible to get cookies serverside without access to req + res, it isn't possible to access res/req here so providing a function from higher in the lifecycle is needed
</details>

<details>
    <summary>Why accept a function wrapper rather than an instance of cookies-universal</summary>
Because not all clients will make the same choice around libraries they use. So a simple wrapper makes this a bit simpler
</details>

image

## Changes

v0.11.0
*September 16, 2022*

### Added
- Now accepts a function to load cookies
- Now attempts to append conversation id headers to each call (driven by cookie value)

### Fixed
- Isomorphism error when using http serverside

### DIFF
--- a/packages/services/f-http/src/createClient.js
+++ b/packages/services/f-http/src/createClient.js
@@ -1,4 +1,3 @@
-import axios from 'axios';
 import defaultOptions from './defaultOptions';
 import setAuthorisationToken from './authorisationHandler';
 import httpVerbs from './httpVerbs';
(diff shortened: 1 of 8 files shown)


packages/services/f-http/src/createClient.js
```diff
@@ -55,7 +75,13 @@ export default class HttpClient {
      * @return {object} - Returns data from response
      */
     async post (resource, body, headers = {}) {
-        return this.sendRequestWithBody(httpVerbs.POST, resource, body, headers);
+        console.log(headers);
```
tommcclean: Bugger

oliversweb (approved):

kevinrodrigues (approved):

PR #2138, opened 2022-09-16
f-analytics@v0.19.0 - pushEvent() - Stripped objects of reflection details

oliversweb:
### Changed
- Converted the objects being passed to `pushEvent()` to simple object which removes any unwanted reflection details
---
Before
image

After
image

### DIFF
--- a/packages/services/f-analytics/src/services/analytics.service.js
+++ b/packages/services/f-analytics/src/services/analytics.service.js
@@ -157,7 +157,10 @@ export default class AnalyticService {
         if (isDataLayerPresent()) {
             const events = [...this.store.state[`${this.options.namespace}`].events];
 
-            events.forEach(e => window.dataLayer.push({ ...e }));
+            // Note; we need to remove the unwanted reflection details from being pushed
+            // along with the simple object.
+            // So by converting to plain simple object using stringify/parse removes these details.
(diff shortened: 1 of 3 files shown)


packages/services/f-analytics/src/services/analytics.service.js
```diff
@@ -157,7 +157,10 @@ export default class AnalyticService {
         if (isDataLayerPresent()) {
             const events = [...this.store.state[`${this.options.namespace}`].events];
 
-            events.forEach(e => window.dataLayer.push({ ...e }));
+            // Note; we need to remove the unwanted reflection details from being pushed
+            // along with the simple object.
+            // So by converting to plain simple object using stringify/parse removes these details.
+            events.forEach(e => window.dataLayer.push(JSON.parse(JSON.stringify(e))));
```
jamieomaguire, line 163: is there any schema / test that needs updating to reflect this?

jamieomaguire (approved):

packages/services/f-analytics/src/services/analytics.service.js
oliversweb: No, as in there is a test already to see if the correct obj is pushed but it's only when it travels thru gtm and hits the datalayer, then viewed via the dev console does it show the unwanted reflection details, so a tad too difficult to test.

xander-marjoram (approved):

PR #2148, opened 2022-09-28
f-checkout@4.6.1 - Add minor improvements

kathywklee:
v4.6.1
*September 28, 2022*

### Added
- New props `getNoteConfigUrl` and `checkoutFeatures` to README

### DIFF
--- a/packages/components/pages/f-checkout/CHANGELOG.md
+++ b/packages/components/pages/f-checkout/CHANGELOG.md
@@ -11,9 +11,6 @@ v4.6.1
 ### Added
 - New props `getNoteConfigUrl` and `checkoutFeatures` to README
 
-### Changed
-- Get checkout failure now error instead of warning
-
 
 v4.6.0
 ------------------------------
--- a/packages/components/pages/f-checkout/CHANGELOG.md
+++ b/packages/components/pages/f-checkout/CHANGELOG.md
@@ -9,7 +9,7 @@ v4.6.1
 *September 28, 2022*
 
 ### Added
-- New props to README
+- New props `getNoteConfigUrl` and `checkoutFeatures` to README
 
 ### Changed
 - Get checkout failure now error instead of warning
--- a/packages/components/pages/f-checkout/CHANGELOG.md
+++ b/packages/components/pages/f-checkout/CHANGELOG.md
@@ -4,6 +4,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v4.6.1
+------------------------------
+*September 28, 2022*
+
+### Added
+- New props to README
+
+### Updated
+- Get checkout failure now error instead of warning
+
+
 v4.6.0
 ------------------------------
 *August 25, 2022*
--- a/packages/components/pages/f-checkout/CHANGELOG.md
+++ b/packages/components/pages/f-checkout/CHANGELOG.md
@@ -11,7 +11,7 @@ v4.6.1
 ### Added
 - New props to README
 
-### Updated
+### Changed
 - Get checkout failure now error instead of warning
 
 
--- a/packages/components/pages/f-checkout/README.md
+++ b/packages/components/pages/f-checkout/README.md
@@ -85,6 +85,8 @@ The props that can be defined are as follows:
 | `loginUrl` | `String` | `-` | URL to navigate to if the user wishes to change account. |
 | `paymentPageUrlPrefix` | `String` | `-` | URL prefix to navigate to after the order has been successfully placed, so the user can pay. The `orderId` will be appended to this URL to form the full URL. |
 | `applicationName` | `String` | `-` | The name of the application using this component. |
+| `getNoteConfigUrl` | `String` | - | URL for the API called to get the note configuration for split notes |
+| `checkoutFeatures` | `Object` | - | Object containing relevant feature flags set in CoreWeb configuration files |
 
 ### Events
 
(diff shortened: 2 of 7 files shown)


packages/components/pages/f-checkout/README.md
```diff
@@ -85,6 +85,8 @@ The props that can be defined are as follows:
 | `loginUrl` | `String` | `-` | URL to navigate to if the user wishes to change account. |
 | `paymentPageUrlPrefix` | `String` | `-` | URL prefix to navigate to after the order has been successfully placed, so the user can pay. The `orderId` will be appended to this URL to form the full URL. |
 | `applicationName` | `String` | `-` | The name of the application using this component. |
+| `getNoteConfigUrl` | `String` | - | URL for the API called to get the note configuration for split notes |
+| `checkoutFeatures` | `Object` | - | Object containing relevant feature flags set in CoreWeb configuration files |
```
siggerzz: Is it likely we'll need to add any browser / visual tests for these props? For example if you supply the `checkoutFeatures` object, can that change the UI state at all?

kathywklee: This prop was added when implementing the split note functionality. From my understanding it gives us flexibility if we introduces more feature flags in the future, but for now we only have boolean `isSplitNotesEnabled` there.

As feature flag is a temporary thingy, I think it is fine not to add the tests based on this. What is your general practice?

packages/components/pages/f-checkout/src/log-events.js
```diff
@@ -63,7 +63,7 @@ export default {
     },
     CheckoutGetFailure: {
         logMessage: 'Get Checkout Failure',
-        logMethod: 'warn'
+        logMethod: 'error'
```
oliversweb: I presume this will now start paging the on-call engineers, who should they wake when this happens?

kathywklee: This event belongs to an API of getting the details to fulfill an order, which is maintained by the checkout platform team.

packages/components/pages/f-checkout/CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
 
 
+v4.6.1
+------------------------------
+*September 28, 2022*
+
+### Added
+- New props to README
+
+### Updated
```
kevinrodrigues: `Changed`

kevinrodrigues (dismissed):

jamieomaguire (dismissed):

packages/components/pages/f-checkout/CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v4.6.1
+------------------------------
+*September 28, 2022*
+
+### Added
+- New props to README
```
xander-marjoram: Please could you mention the names of the new props here? It makes it a bit easier when you're skimming the changelog

packages/components/pages/f-checkout/README.md
xander-marjoram: I probably wouldn't mention the names of internal components on a public repo as it's not necessary, and this component doesn't need to know where it's being consumed

packages/components/pages/f-checkout/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-checkout",
   "description": "Fozzie Checkout - Fozzie Checkout Component",
-  "version": "4.6.0",
+  "version": "4.6.1",
```
xander-marjoram, line 4: Adding new props is definitely worth a minor bump, so `4.7.0` please

packages/components/pages/f-checkout/src/log-events.js
xander-marjoram: Where are the alerts configured? I'm still a little concerned that we'll end up getting paged for this, even if it's owned by the checkout platform team

packages/services/f-http/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v1.0.2
+------------------------------
+*October 4, 2022*
+
+### Added
+- a build script for storybook
```
xander-marjoram: Always good to be specific 😄 
```suggestion
- `storybook:build` script to package.json
```

packages/components/pages/f-checkout/README.md
kathywklee: I see. In this case I will just add `getNoteConfigUrl` here instead.

packages/components/pages/f-checkout/package.json
kathywklee: The props have been added in the past, but at that time the documentation was missing. I think it's fine to make it a patch bump. What do you think?

packages/components/pages/f-checkout/README.md
xander-marjoram: If the prop exists in the component, it should definitely be listed in the readme, we just don't have to say exactly _where_ the features will be provided from, only **what** the component is expecting.

packages/components/pages/f-checkout/package.json
xander-marjoram: Ah right, if you're just adding missing documentation then a patch is fine, carry on 😄

xander-marjoram (approved):

packages/components/pages/f-checkout/src/log-events.js
kathywklee: After discussion we agreed to keep the alert type as warn. We may revisit this again if there's a need to change it.

kevinrodrigues (approved):

PR #2165, opened 2022-10-13
f-rating@0.2.0 - Add stars implementation to initial generator code.

kevinrodrigues:
### Added
- `pie-icons-vue` repo as `dep` to pull in the star icons.
- Stories component added.
- Prop `starRating` added to allow the consumer to pass in a value.
- Readme prop table.
- Snapshot test.

### Screenshot
Screenshot 2022-10-13 at 13 54 42

### DIFF
--- a/packages/components/molecules/f-rating/README.md
+++ b/packages/components/molecules/f-rating/README.md
@@ -69,10 +69,10 @@ There may be props that allow you to customise its functionality.
 
 The props that can be defined are as follows (if any):
 
-| Prop          |   Type   | Required | Default | Description                                           |
-| :---          |:--------:|:--------:|:-------:|:------------------------------------------------------|
-| `starRating`  | `Number` |   Yes    |    -    | Sets the number of stars out of 5                     |
-| `maxStarRating`  | `Number` |    No    |    5    | Sets the number of max stars the consumer should use. |
+| Prop          |   Type   | Required | Default | Description                                                                                                            |
+| :---          |:--------:|:--------:|:-------:|:-----------------------------------------------------------------------------------------------------------------------|
+| `starRating`  | `Number` |   Yes    |    -    | Sets the displayed rating (filled stars). i.e. for `2 out of x`, 2 would be the `starRating`                           |
+| `maxStarRating`  | `Number` |    No    |    5    | Sets the maximum number of stars that the rating is set against. i.e. for `x out of 5`, 5 would be the `maxStarRating` |
 
 ### Events
 
--- a/packages/components/molecules/f-rating/README.md
+++ b/packages/components/molecules/f-rating/README.md
@@ -69,8 +69,9 @@ There may be props that allow you to customise its functionality.
 
 The props that can be defined are as follows (if any):
 
-| Prop  | Type  | Default | Description |
-| ----- | ----- | ------- | ----------- |
+| Prop          |   Type   | Required | Default | Description                       |
+| :---          |:--------:|:--------:|:-------:|:----------------------------------|
+| `starRating`  | `Number` |   Yes    |    -    | Sets the number of stars out of 5 |
 
 ### Events
 
--- a/packages/components/molecules/f-rating/README.md
+++ b/packages/components/molecules/f-rating/README.md
@@ -69,9 +69,10 @@ There may be props that allow you to customise its functionality.
 
 The props that can be defined are as follows (if any):
 
-| Prop          |   Type   | Required | Default | Description                       |
-| :---          |:--------:|:--------:|:-------:|:----------------------------------|
-| `starRating`  | `Number` |   Yes    |    -    | Sets the number of stars out of 5 |
+| Prop          |   Type   | Required | Default | Description                                           |
+| :---          |:--------:|:--------:|:-------:|:------------------------------------------------------|
+| `starRating`  | `Number` |   Yes    |    -    | Sets the number of stars out of 5                     |
+| `maxStarRating`  | `Number` |    No    |    5    | Sets the number of max stars the consumer should use. |
 
 ### Events
 
--- a/packages/components/molecules/f-rating/package.json
+++ b/packages/components/molecules/f-rating/package.json
@@ -45,8 +45,9 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
+    "@justeat/f-globalisation": "1.x",
     "@justeat/f-services": "1.x",
-    "@justeattakeaway/pie-icons-vue": "1.x"
+    "@justeattakeaway/pie-icons-vue": "2.0.0-beta.0"
   },
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.2.0"
--- a/packages/components/molecules/f-rating/package.json
+++ b/packages/components/molecules/f-rating/package.json
@@ -45,12 +45,12 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-globalisation": "1.x",
     "@justeat/f-services": "1.x",
     "@justeattakeaway/pie-icons-vue": "2.0.0-beta.0"
   },
   "peerDependencies": {
-    "@justeat/browserslist-config-fozzie": ">=1.2.0"
+    "@justeat/browserslist-config-fozzie": ">=1.2.0",
+    "@justeat/f-globalisation": "1.x"
   },
   "devDependencies": {
     "@justeat/f-wdio-utils": "1.x",
--- a/packages/components/molecules/f-rating/package.json
+++ b/packages/components/molecules/f-rating/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-rating",
   "description": "Fozzie Rating - Global Rating component",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "main": "dist/f-rating.umd.min.js",
   "maxBundleSize": "20kB",
   "files": [
@@ -45,7 +45,8 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-services": "1.x"
+    "@justeat/f-services": "1.x",
+    "@justeattakeaway/pie-icons-vue": "1.x"
   },
   "peerDependencies": {
     "@justeat/browserslist-config-fozzie": ">=1.2.0"
--- a/packages/components/molecules/f-rating/src/components/Rating.vue
+++ b/packages/components/molecules/f-rating/src/components/Rating.vue
@@ -2,13 +2,14 @@
     <div
         :class="$style['c-rating']"
         data-test-id="rating-component">
-        <ul :class="$style['c-rating-star-wrapper']">
+        <ul
+            :class="$style['c-rating-starWrapper']">
             <li
                 v-for="star in maxStarRating"
                 :key="star"
                 :class="$style['c-rating-star']">
                 <star-filled-icon
-                    v-if="hasRating(star)"
+                    v-if="isRatingStarFilled(star)"
                     :class="$style['c-rating-star-filled']" />
                 <star-icon
                     v-else
@@ -57,8 +58,7 @@ export default {
 
     data () {
         return {
-            tenantConfigs,
-            rating: this.starRating
+            tenantConfigs
         };
     },
 
@@ -72,12 +72,12 @@ export default {
         getRatingDescription () {
             return this.starRating < 2
                 ? this.$tc('ratings.starsDescription', 1, {
-                    rating: this.rating,
+                    rating: this.starRating,
                     total: this.maxStarRating
                 })
 
                 : this.$tc('ratings.starsDescription', 2, {
-                    rating: this.rating,
+                    rating: this.starRating,
                     total: this.maxStarRating
                 });
         }
@@ -87,11 +87,11 @@ export default {
         /**
          * Check `star` against value passed by consumer to allow empty stars to render.
          *
-         * @param star
+         * @param star {Number}
          * @returns {boolean}
          */
-        hasRating (star) {
-            return star <= this.rating;
+        isRatingStarFilled (star) {
+            return star <= this.starRating;
         }
     }
 };
@@ -100,12 +100,11 @@ export default {
 <style lang="scss" module>
 @use '@justeat/fozzie/src/scss/fozzie' as f;
 
-.c-rating {
-    .c-rating-star-wrapper {
-        margin: 0;
-        padding: 0;
-        list-style-type: none;
-    }
+.c-rating-starWrapper {
+    margin: 0;
+    padding: 0;
+    list-style-type: none;
+}
 
     .c-rating-star {
         display: inline-block;
@@ -123,5 +122,4 @@ export default {
             fill: f.$color-mozzarella-50;
         }
     }
-}
 </style>
--- a/packages/components/molecules/f-rating/src/components/Rating.vue
+++ b/packages/components/molecules/f-rating/src/components/Rating.vue
@@ -8,7 +8,7 @@
                 :key="star"
                 :class="$style['c-rating-star']">
                 <star-filled-icon
-                    v-if="getRating(star)"
+                    v-if="hasRating(star)"
                     :class="$style['c-rating-star-filled']" />
                 <star-icon
                     v-else
@@ -77,12 +77,12 @@ export default {
 
     methods: {
         /**
-         * Check if `star` against value passed by consumer to allow empty stars to render.
+         * Check `star` against value passed by consumer to allow empty stars to render.
          *
          * @param star
          * @returns {boolean}
          */
-        getRating (star) {
+        hasRating (star) {
             return star <= this.rating;
         }
     }
--- a/packages/components/molecules/f-rating/src/components/Rating.vue
+++ b/packages/components/molecules/f-rating/src/components/Rating.vue
@@ -14,12 +14,12 @@
                     v-else
                     :class="$style['c-rating-star-empty']" />
             </li>
-            <span
-                data-test-id="c-rating-description"
-                class="is-visuallyHidden">
-                {{ starRating }} {{ setRatingDescription }} {{ maxStarRating }}
-            </span>
         </ul>
+        <span
+            data-test-id="c-rating-description"
+            class="is-visuallyHidden">
+            {{ getRatingDescription }}
+        </span>
     </div>
 </template>
 
@@ -28,7 +28,7 @@ import {
     StarFilledIcon,
     StarIcon
 } from '@justeattakeaway/pie-icons-vue';
-import { globalisationServices } from '@justeat/f-services';
+import { VueGlobalisationMixin } from '@justeat/f-globalisation';
 import tenantConfigs from '../tenants';
 
 export default {
@@ -37,6 +37,9 @@ export default {
         StarFilledIcon,
         StarIcon
     },
+
+    mixins: [VueGlobalisationMixin],
+
     props: {
         locale: {
             type: String,
@@ -53,25 +56,30 @@ export default {
     },
 
     data () {
-        const locale = globalisationServices.getLocale(tenantConfigs, this.locale, this.$i18n);
-        const localeConfig = tenantConfigs[locale];
-
         return {
-            copy: { ...localeConfig },
+            tenantConfigs,
             rating: this.starRating
         };
     },
 
     computed: {
         /**
-         * Return singular description if one star.
+         * Return description using `vue-i18n Pluralization` if one star or an alternative if more.
+         *
          *
          * @returns {string|*}
          */
-        setRatingDescription () {
+        getRatingDescription () {
             return this.starRating < 2
-                ? this.copy.ratings.starsDescriptionSingular
-                : this.copy.ratings.starsDescription;
+                ? this.$tc('ratings.starsDescription', 1, {
+                    rating: this.rating,
+                    total: this.maxStarRating
+                })
+
+                : this.$tc('ratings.starsDescription', 2, {
+                    rating: this.rating,
+                    total: this.maxStarRating
+                });
         }
     },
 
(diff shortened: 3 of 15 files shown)


packages/components/molecules/f-rating/src/tenants/en-GB.js
```diff
@@ -1,4 +1,8 @@
 export default {
     locale: 'en-GB',
-    text: 'I am a VRating Component (GB)'
+    text: 'I am a VRating Component (GB)',
+    ratings: {
+        starsDescription: 'stars out of',
```
jamieomaguire: we can use vue i18n pluralization here:
https://kazupon.github.io/vue-i18n/guide/pluralization.html

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,30 +2,89 @@
...
         };
+    },
+
+    computed: {
+        /**
+         * Return singular description if one star.
+         *
+         * @returns {string|*}
+         */
+        setRatingDescription () {
```
jamieomaguire: I think as per my other comment we can use i18n pluralization here. However just to be nitpicky I'm not sure if `set` is the correct verb to use in this method. Feels more like either a `get` or just `ratingDescription` to me - purely my opinion though!

packages/components/molecules/f-rating/README.md
```diff
@@ -69,8 +69,9 @@ There may be props that allow you to customise its functionality.
 
 The props that can be defined are as follows (if any):
 
-| Prop  | Type  | Default | Description |
-| ----- | ----- | ------- | ----------- |
+| Prop          |   Type   | Required | Default | Description                       |
+| :---          |:--------:|:--------:|:-------:|:----------------------------------|
+| `starRating`  | `Number` |   Yes    |    -    | Sets the number of stars out of 5 |
```
ashleynolan: More out of interest than anything else. Are we migrating all platforms to use 5 stars using this component, or will there be a use-case for 6 stars in the interim?

packages/components/molecules/f-rating/package.json
```diff
@@ -45,7 +45,8 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-services": "1.x"
+    "@justeat/f-services": "1.x",
+    "@justeattakeaway/pie-icons-vue": "1.x"
```
ashleynolan: Did you try using the beta package for the icons?

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,30 +2,89 @@
...
+                :key="star"
+                :class="$style['c-rating-star']">
+                <star-filled-icon
+                    v-if="getRating(star)"
+                    :class="$style['c-rating-star-filled']" />
+                <star-icon
+                    v-else
+                    :class="$style['c-rating-star-empty']" />
+            </li>
+            <span
```
ashleynolan: I don't think this will be valid HTML – I would think the `span` will need to sit outside the `ul` tag

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,30 +2,89 @@
...
+    },
+
+    methods: {
+        /**
+         * Check if `star` against value passed by consumer to allow empty stars to render.
+         *
+         * @param star
+         * @returns {boolean}
+         */
+        getRating (star) {
```
jamieomaguire: if I understand this correctly, I think a clearer name may be `hasRating`. I did have trouble understanding the JS doc comment though

packages/components/molecules/f-rating/README.md
ashleynolan: Looks like you also have a prop for `maxStarRating` too?

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -34,13 +93,27 @@ export default {
...
+
+        .c-rating-star-filled {
+            & path {
+                fill: f.$color-support-brand-01;
+            }
+        }
+
+        .c-rating-star-empty {
+            & path {
+                fill: f.$color-mozzarella-50;
```
jamieomaguire: is there a more suitable alias we could use here?

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -34,13 +93,27 @@ export default {
...
+            }
+        }
+    }
```
ashleynolan: Because of the naming scheme, you shouldn't need to nest these inside `c-rating` (just have it outside so the selector specificity is as low as possible)

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -34,13 +93,27 @@ export default {
...
-    margin: auto;
-    border: 1px solid red;
-    font-family: f.$font-family-base;
-    @include f.font-size(heading-m);
+    .c-rating-star-wrapper {
+        margin: 0;
+        padding: 0;
+        list-style-type: none;
+
+        .c-rating-star {
```
ashleynolan: Same for these – no need to nest the classes

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -34,13 +93,27 @@ export default {
...
+            display: inline-block;
+            width: 15px; // Todo - decide on how to size these. Will create a ticket around this.
+        }
+
+        .c-rating-star-filled {
```
ashleynolan: And the same for these classes

packages/components/molecules/f-rating/src/components/Rating.vue
ashleynolan: One to go back to design on potentially. We should have an alias for this, but it doesn't look like we do.

So we either need to adjust the colour so that it uses an alias, or an alias needs adding to the tokens.

ashleynolan (changes requested):

packages/components/molecules/f-rating/src/components/Rating.vue

packages/components/molecules/f-rating/src/tenants/en-GB.js
```diff
@@ -1,4 +1,7 @@
 export default {
     locale: 'en-GB',
-    text: 'I am a VRating Component (GB)'
+    ratings: {
+        starsDescription: 'stars out of',
+        starsDescriptionSingular: 'star out of'
```
jamieomaguire: we can also do something cool here to pass the value into the message I think. Sending you an example offline

packages/components/molecules/f-rating/README.md
kevinrodrigues: That's my understand. Anyone on 6 probably will keep their current implementation until they move across although i believe they may have no choice if the api is being worked on. Ill double check this tho

packages/components/molecules/f-rating/package.json
kevinrodrigues: Should we be using the beta? Happy to swap over.

packages/components/molecules/f-rating/src/tenants/en-GB.js
kevinrodrigues: Yeah so you can use a template. The issue from what I can see (including the related comments about plural/singular copy) is that we're not using the vue globalisation package which gives us the template helper. Also using the plural helper from i18n means we'd probably have to modify that package so it exposes it for us. Will have a look further into it tho.

packages/components/molecules/f-rating/package.json
ashleynolan: Yes please – it's the same package, it's just in beta while we wait for the name changes from design.

If it works ok in this though, I'll change the readme of the old package to point to the new one

packages/components/molecules/f-rating/src/tenants/en-GB.js
jamieomaguire: If the package we use is preventing us from properly using the i18n features then yeah we should probably look into extending the package - happy to help with this if you like!

packages/components/molecules/f-rating/package.json
kevinrodrigues: Added.

packages/components/molecules/f-rating/README.md
kevinrodrigues: Fixed.

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: Shall i create a ticket for this or does this come in with the token updates?

packages/components/molecules/f-rating/src/tenants/en-GB.js
kevinrodrigues: Looks like i18n works out the box with the `tc` version so added that :)

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: Fixed

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: Added a comment to Ash's comment.

packages/components/molecules/f-rating/README.md
```diff
@@ -69,8 +69,10 @@ There may be props that allow you to customise its functionality.
 
 The props that can be defined are as follows (if any):
 
-| Prop  | Type  | Default | Description |
-| ----- | ----- | ------- | ----------- |
+| Prop          |   Type   | Required | Default | Description                                           |
+| :---          |:--------:|:--------:|:-------:|:------------------------------------------------------|
+| `starRating`  | `Number` |   Yes    |    -    | Sets the number of stars out of 5                     |
+| `maxStarRating`  | `Number` |    No    |    5    | Sets the number of max stars the consumer should use. |
```
ashleynolan:
```suggestion
| `starRating`  | `Number` |   Yes    |    -    | Sets the displayed rating (filled stars). i.e. for `2 out of x`, 2 would be the `starRating`                    |
| `maxStarRating`  | `Number` |    No    |    5    | Sets the maximum number of stars that the rating is set against. i.e. for `x out of 5`, 5 would be the `maxStarRating`  |
```

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,30 +2,97 @@
     <div
         :class="$style['c-rating']"
         data-test-id="rating-component">
-        {{ copy.text }}
+        <ul :class="$style['c-rating-star-wrapper']">
```
ashleynolan:
```suggestion
        <ul :class="$style['c-rating-starWrapper']">
```

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -34,13 +101,27 @@ export default {
...
 .c-rating {
-    display: flex;
-    justify-content: center;
-    min-height: 80vh;
-    width: 80vw;
-    margin: auto;
-    border: 1px solid red;
-    font-family: f.$font-family-base;
-    @include f.font-size(heading-m);
+    .c-rating-star-wrapper {
```
ashleynolan:
```suggestion
    .c-rating-starWrapper {
```

ashleynolan: Looks like the nesting is still here – can just move all the classes to be individually specified (maybe drop me a message if it helps - not the easiest thing to explain sometimes in comments!)

ashleynolan (commented):
Couple of super small comments - otherwise LGTM

packages/components/molecules/f-rating/src/components/Rating.vue
ashleynolan: I would just discuss it separately with Rachel or Jordi – don't think it needs a ticket. If it gets updated as you say, it'll fall into token updates

packages/components/molecules/f-rating/package.json
```diff
@@ -45,7 +45,9 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-services": "1.x"
+    "@justeat/f-globalisation": "1.x",
```
jamieomaguire: I always forget, should these be peerDeps rather than regular dependencies?

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,30 +2,97 @@
...
+                :class="$style['c-rating-star']">
+                <star-filled-icon
+                    v-if="hasRating(star)"
+                    :class="$style['c-rating-star-filled']" />
+                <star-icon
+                    v-else
+                    :class="$style['c-rating-star-empty']" />
+            </li>
+        </ul>
+        <span
```
jamieomaguire, line 19: could this be an `aria` attribute (such as label) on the `ul` element itself?

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,30 +2,97 @@
...
     },
-    data () {
-        const locale = globalisationServices.getLocale(tenantConfigs, this.locale, this.$i18n);
-        const localeConfig = tenantConfigs[locale];
 
+    data () {
         return {
-            copy: { ...localeConfig }
+            tenantConfigs,
+            rating: this.starRating
```
jamieomaguire: Is there a reason not to directly use the prop rather than duplicate it?

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,30 +2,97 @@
...
         :class="$style['c-rating']"
         data-test-id="rating-component">
-        {{ copy.text }}
+        <ul :class="$style['c-rating-star-wrapper']">
+            <li
+                v-for="star in maxStarRating"
+                :key="star"
+                :class="$style['c-rating-star']">
+                <star-filled-icon
+                    v-if="hasRating(star)"
```
jamieomaguire: If the current key is 3 (for example) then the code reads like:

`v-if="hasRating(3)"` - is that right? 

Sorry, I just find it a little hard to read (I might need a coffee) :)

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,30 +2,97 @@
...
+                    :class="$style['c-rating-star-filled']" />
+                <star-icon
```
jamieomaguire, line 14: does this work for decimals? i.e. 3.5 stars being 3 filled and a half filled star etc

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,30 +2,97 @@
...
+                    total: this.maxStarRating
+                });
+        }
+    },
+
+    methods: {
+        /**
+         * Check `star` against value passed by consumer to allow empty stars to render.
+         *
+         * @param star
```
jamieomaguire: I think we should give a type for star, which I think is number?

packages/components/molecules/f-rating/src/components/Rating.vue
jamieomaguire: I wonder if something like `v-if="shouldFillStar(index)"` might be clearer

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
```diff
@@ -1,10 +1,117 @@
...
 
 describe('Rating', () => {
+    let propsData;
+    let wrapper;
+    const $tc = jest.fn();
+
+    beforeEach(() => {
+        propsData = {
+            starRating: 2,
+            maxStarRating: 5
```
jamieomaguire, line 17: if we're not going to define these on a test by test basis, I'd declare them outside the beforeEach so tests can access them.

That way we can add a test to check that when the starRating provided is equal to maxStarRating, it behaves correctly.

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
```diff
@@ -1,10 +1,117 @@
...
+                const result = wrapper.find('[data-test-id="c-rating-description"]');
+
+                // Assert
+                expect(result).toMatchSnapshot();
+            });
+
+            describe('when invoked', () => {
+                it('should return truthy when `starRating` is less than or equal to `rating`', () => {
+                    // Act
+                    const result = wrapper.vm.hasRating(2);
```
jamieomaguire: NIT: It's a little hard to know what's going on in the test because we're 'Arranging' a lot of stuff outside of each test (props etc) - it would be duplication but perhaps easier to understand each test if we created a new wrapper for every test.

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
```diff
@@ -1,10 +1,117 @@
...
+            it('should contain a description `c-rating-description`', () => {
+                // Act
+                const result = wrapper.find('[data-test-id="c-rating-description"]');
+
+                // Assert
+                expect(result).toMatchSnapshot();
+            });
+
+            describe('when invoked', () => {
+                it('should return truthy when `starRating` is less than or equal to `rating`', () => {
```
jamieomaguire: Should we add a test checking when `starRating` is less than rating? I only see a test for equal here

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
```diff
@@ -1,10 +1,117 @@
...
+                        i18n
+                    });
+
+                    // Act & Assert
+                    expect(wrapper.vm.getRatingDescription).toBe(description);
+                });
+
+                it('should return a plural description if the rating is greater than 1', () => {
+                    // Arrange
+                    const description = '2 stars out of 5';
```
jamieomaguire: entirely up to you - would a snapshot test mean less maintenance overhead here if the copy changes?

packages/components/molecules/f-rating/src/tenants/en-AU.js
```diff
@@ -1,4 +1,6 @@
 export default {
     locale: 'en-AU',
-    text: 'I am a VRating Component (AU)'
+    ratings: {
+        starsDescription: '{rating} star out of {total} | {rating} stars out of {total}'
```
jamieomaguire, line 4: 🔥

packages/components/molecules/f-rating/src/tenants/es-ES.js
```diff
@@ -1,4 +1,6 @@
 export default {
     locale: 'es-ES',
-    text: 'I am a VRating Component (ES)'
+    ratings: {
+        starsDescription: '{rating} star out of {total} | {rating} stars out of {total}'
```
jamieomaguire: should we change all languages other than English to something like "TODO - GET APPROPRIATE TRANSLATION" - otherwise it might get forgotten about

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: Ill change it to `hasRatingStarFilled`.  So it matched the class somewhat.

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: Not yet. I have a note about a ticket for this. Also css was used before from what i remember and this is purely using svg no css.

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
kevinrodrigues: I think it's ok at the moment :) What is difficult to understand? I'm happy to explain on a call but essentially we only have one set up at the top which will require the translations mock, a vue instance (because we need to tell vue about the translations) & props. I do agree it makes it easier to move it into each describe but only when it makes sense, I think it's quite straightforward here, if we ended up modifying the props data or wrapper in each test but having the setup right at the top then  it makes it harder to follow and therefore moving into each test unit would be better.

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: That's a good suggestion. I'll have a think and add that.

kevinrodrigues: I've had a play with this and it turns out (on Mac anyway) voiceover works much better with how it is right now. Using an aria label doesn't really work as good. Thoughts?

packages/components/molecules/f-rating/package.json
kevinrodrigues: I think it could actually be a peer as the consuming app most likely a page component will import the globalisation anyway. Ill move it 👍

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
kevinrodrigues: I think the current test covers this? If the code was modified and the `=` removed the test would fail in that instance.

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
kevinrodrigues: Good shout.

packages/components/molecules/f-rating/src/tenants/it-IT.js
```diff
@@ -1,4 +1,6 @@
 export default {
     locale: 'it-IT',
-    text: 'I am a VRating Component (IT)'
+    ratings: {
+        starsDescription: '{rating} star out of {total} | {rating} stars out of {total}'
+    }
```
jamieomaguire:
```suggestion
    ratings: {
        starsDescription: '{rating} ADD TRANSLATION {total} | {rating} ADD TRANSLATION {total}'
    }
```

jamieomaguire (dismissed):

jamieomaguire (approved):

packages/components/molecules/f-rating/src/components/Rating.vue
jamieomaguire: Sorry I only just saw this - yep, if the UX is preferable using the current approach then go for it! Out of interest what was the difference?

kevinrodrigues: I'm going to take a look at this separately as i will look to improve this in the next ticket. :)

ashleynolan (approved):

PR #2166, opened 2022-10-14
f-loyalty@v2.6.0 - StampCards more markets

matthewhardern:
Adds updates to tenant files, additional changes to use cloudinary. Also moving the bag amounts into each tenant. Additional tests added for the new tenants. 

## Browsers Tested

- [x] Chrome (latest)


packages/components/pages/f-loyalty/src/components/HowItWorks.vue
```diff
@@ -152,44 +143,42 @@ export default {
             // _Very_ naive way of implementing ordinals but will suit our purposes here
             const ordinal = { 0: 'st', 1: 'nd', 2: 'rd' };
 
-            return this.bagsData.map((bagValue, index) => ({
+            return [...this.$t('howItWorks.bagValues')].map((bagValue, index) => ({
                 number: `${index + 1}${ordinal[index] || 'th'}`,
```
juan-amor, line 161: What is this number used for? If it's just a position, shouldn't be just an int/short? If it's used anywhere, any potential issues for other languages other than English?

packages/components/pages/f-loyalty/src/tenants/en-AU.js
```diff
@@ -33,6 +33,14 @@ const messages = {
         loginButtonText: 'Log in'
     },
     howItWorks: {
+        imagePath: 'https://just-eat-prod-eu-res.cloudinary.com/image/upload/v1630068495/Experiments/Homeweb-Coreweb/',
+        bagValues: [
```
juan-amor: Not sure I understand the logic of the bagValues, can you please explain them?

packages/components/pages/f-loyalty/src/components/HowItWorks.vue
```diff
@@ -152,44 +143,42 @@ export default {
             // _Very_ naive way of implementing ordinals but will suit our purposes here
             const ordinal = { 0: 'st', 1: 'nd', 2: 'rd' };
 
-            return this.bagsData.map((bagValue, index) => ({
+            return [...this.$t('howItWorks.bagValues')].map((bagValue, index) => ({
```
juan-amor: seems weird to read the values out of some localisation values - What if the property in say Italy doesn't exist? Is there a way to move the bagvalues outside the <tenant?.js file?

packages/components/pages/f-loyalty/src/tenants/en-AU.js
matthewhardern: each tenant has their own amounts they want shown in the bags diagram in how it works page, these numbers are used to find the correct bag image to show.

packages/components/pages/f-loyalty/src/components/HowItWorks.vue
matthewhardern: stated in another comment

packages/components/pages/f-loyalty/src/components/HowItWorks.vue
matthewhardern: yea maybe it would be better to have an object that uses tenant as the key returning the array of numbers?

packages/components/pages/f-loyalty/src/tenants/en-AU.js
juan-amor: Yeah had already made the connection with the images, was more about what the values in the array represent.
```
 bagValues: [
            40,
            50,
            45,
            40,
            55
        ],
```

So what relationship between the $40 image and the order value for example?

packages/components/pages/f-loyalty/src/components/HowItWorks.vue
juan-amor: Yeah, I think that'd be better - I don't think localisation should interfere with the business logic. 

Also, you can have a `tenantBagFees.js` mapper that will have the config for you centralised, as the component doesn't need to know about it, just requires some values to parse

packages/components/pages/f-loyalty/src/tenants/en-AU.js
matthewhardern: $40 is $40 order value. Each market wants different order values as examples.

juan-amor: Ah excellent, thanks

packages/components/pages/f-loyalty/src/components/HowItWorks.vue
matthewhardern: agreed I will take this approach.

ashleynolan (dismissed):

packages/components/pages/f-loyalty/src/tenants/en-NZ.js
```diff
@@ -33,12 +33,13 @@ const messages = {
         loginButtonText: 'Log in'
     },
     howItWorks: {
+        imagePath: 'https://just-eat-prod-eu-res.cloudinary.com/image/upload/v1630068495/Experiments/Homeweb-Coreweb/',
```
minrwhite: Is this correct? Should we have a NZ-local domain, or does cloudinary handle this by having different DNS responses for requests in different geolocales?

minrwhite (dismissed):

packages/components/pages/f-loyalty/src/tenants/en-NZ.js
matthewhardern: replied in slack, but will ask content team as that is where I got it from

packages/components/pages/f-loyalty/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v2.6.0
+------------------------------
+*October 12, 2022*
```
ashleynolan: Worth updating?

ashleynolan (dismissed):

packages/components/pages/f-loyalty/CHANGELOG.md
matthewhardern: yep will do

ashleynolan (approved):

minrwhite (approved):

PR #2167, opened 2022-10-14
fozzie-components@v7.38.0 - Atoms - Remove duplicate dependencies #globalconfig

siggerzz:
This PR removes the following dev dependencies from `atom` components, as there are versions already defined in the root of the project.
```
    "@vue/cli-plugin-babel"
    "@vue/cli-plugin-unit-jest"
    "@vue/test-utils"
```

It also modifies the `test` command to call `jest` directly, rather than using the `vue-cli-service test:unit` command.

This is because for some reason, now that `@vue/test-utils` lives at the root, the following error gets returned:

```
@justeat/f-button:test: $ vue-cli-service test:unit
@justeat/f-button:test:  ERROR  command "test:unit" does not exist.
```

I'll try to find a fix for this, however calling `jest` directly seems to work the same way 👍🏼

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v7.37.0
 ------------------------------
-*October 14, 2022*
+*October 17, 2022*
 
 ### Removed
 - Duplicate dependencies of `atoms` packages.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v7.37.0
+------------------------------
+*October 14, 2022*
+
+### Removed
+- Duplicate dependencies of `atoms` packages.
+
+
 v7.36.0
 ------------------------------
 *October 13, 2022*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 v7.38.0
 ------------------------------
 *October 24, 2022*
+
 ### Removed
 - Duplicate dependencies of `atoms` packages.
-*October 19, 2022*
 
 
 v7.37.0
 ------------------------------
-*October 17, 2022*
+*October 19, 2022*
 
 ### Changed
 - Updated Browser Support page by removing Internet Explorer 11 and versions.
--- a/packages/components/atoms/f-form-field/package.json
+++ b/packages/components/atoms/f-form-field/package.json
@@ -56,6 +56,7 @@
   },
   "devDependencies": {
     "@justeat/f-wdio-utils": "1.x",
-    "postcss-assets": "5.0.0"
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,18 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+v7.37.0
+------------------------------
+*October 14, 2022*
+
+###
+Changed
```
xander-marjoram: Nit: newline

ashleynolan:
@siggerzz Just a note – it wouldn't be an issue to leave the test script as calling jest directly. I think the less dependency on vue-cli, the better tbh, as it means we aren't tied to them when we want to update to the latest version of jest then.

So if you want to remove `@vue/cli-plugin-unit-jest` entirely rather than move it to the root, that's likely completely fine too.

CHANGELOG.md
ashleynolan:
```suggestion
### Changed
```

ashleynolan:
Looks good – just as a caution to the `vue-cli` packages (not sure if it's related to the build error). The reason these were in the packages rather than root was because the format of them meant that they weren't compatible with being in the root `package.json`. It was so long ago that I can't remember exactly what it was about them that meant that they couldn't be moved. I remember working with @dandel10n on this though, so not sure if she may be able to remember better than me!

siggerzz:
> @siggerzz Just a note – it wouldn't be an issue to leave the test script as calling jest directly. I think the less dependency on vue-cli, the better tbh, as it means we aren't tied to them when we want to update to the latest version of jest then.
> 
> So if you want to remove `@vue/cli-plugin-unit-jest` entirely rather than move it to the root, that's likely completely fine too.

No worries at all. Will probably remove this dep once all the packages are calling Jest directly :)

packages/components/atoms/f-form-field/package.json
```diff
@@ -55,10 +55,8 @@
     "@justeat/browserslist-config-fozzie": ">=1.1.1"
   },
   "devDependencies": {
-    "@vue/cli-plugin-babel": "4.5.16",
-    "@vue/cli-plugin-unit-jest": "4.5.16",
-    "@vue/test-utils": "1.0.3",
     "@justeat/f-wdio-utils": "1.x",
-    "postcss-assets": "5.0.0"
+    "postcss-assets": "5.0.0",
+    "@vue/cli-plugin-babel": "4.5.16"
```
siggerzz, line 60: Had to leave this dependency as it was causing issues when moved to the root. We believe it's just an issue with optional chaining and our babel config, so we can address this as part of a later ticket if other components exhibit the same behaviour.

packages/components/molecules/f-rating/package.json
```diff
@@ -56,6 +55,7 @@
     "@justeat/f-wdio-utils": "1.x",
     "@vue/cli-plugin-babel": "4.5.15",
     "@vue/cli-plugin-unit-jest": "4.5.15",
-    "@vue/test-utils": "1.2.2"
+    "@vue/test-utils": "1.2.2",
+    "@justeattakeaway/pie-icons-vue": "1.x"
```
xander-marjoram: Was this meant to be a downgrade from v2 beta to v1?

xander-marjoram (approved):

jamieomaguire (approved):

PR #2170, opened 2022-10-19
f-rating@0.4.0 - Refactor component to allow fractional stars.

kevinrodrigues:
### Added
- `starRatingSize` to allow component sizing.
- Prop validation.

### Changed
- Refactored markup to allow fractional rating sizes.

#### 5 stars out of 5
Screenshot 2022-10-19 at 09 20 34

#### 2.3 stars out of 5
Screenshot 2022-10-19 at 10 13 12

#### 2.5 stars out of 5

Screenshot 2022-10-19 at 09 14 39

#### 5 stars out of 5
Screenshot 2022-10-19 at 09 19 57

#### 2.3 stars out of 5
Screenshot 2022-10-19 at 10 12 13

#### 3 stars out of 5
Screenshot 2022-10-19 at 09 17 48

#### 5 stars out of 5
Screenshot 2022-10-19 at 10 10 41

#### 2.3 stars out of 5
Screenshot 2022-10-19 at 10 11 04

#### 2.3 stars out of 5
Screenshot 2022-10-19 at 10 19 42

#### 2.5 stars out of 5
Screenshot 2022-10-19 at 10 20 42

#### 5 stars out of 5
Screenshot 2022-10-19 at 10 21 04

### Default (small)
Screenshot 2022-10-19 at 10 22 30

### Medium
Screenshot 2022-10-19 at 10 22 02

### Large

Screenshot 2022-10-19 at 10 23 07

### DIFF
--- a/packages/components/molecules/f-rating/README.md
+++ b/packages/components/molecules/f-rating/README.md
@@ -73,6 +73,7 @@ The props that can be defined are as follows (if any):
 | :---          |:--------:|:--------:|:-------:|:-----------------------------------------------------------------------------------------------------------------------|
 | `starRating`  | `Number` |   Yes    |    -    | Sets the displayed rating (filled stars). i.e. for `2 out of x`, 2 would be the `starRating`                           |
 | `maxStarRating`  | `Number` |    No    |    5    | Sets the maximum number of stars that the rating is set against. i.e. for `x out of 5`, 5 would be the `maxStarRating` |
+| `starRatingSize`  | `String` |    No    | 'small' | Sets the component size. By default the component will use `small` the other options are `medium` & `large`            |
 
 ### Events
 
--- a/packages/components/molecules/f-rating/src/components/Rating.vue
+++ b/packages/components/molecules/f-rating/src/components/Rating.vue
@@ -18,10 +18,11 @@
                     $style['c-rating-mask'],
                     $style['c-rating-container']
                 ]"
-                :style="getRatingStarPercentage">
+                :style="`--starRatingPercentage: ${getRatingStarPercentage}`">
                 <star-filled-icon
                     v-for="star in maxStarRating"
                     :key="star"
+                    :style="`--starRatingSize: c-rating-star--${starRatingSize}`"
                     :class="[
                         $style['c-rating-star-filled'],
                         $style[`c-rating-star--${starRatingSize}`]
@@ -72,7 +73,7 @@ export default {
         starRatingSize: {
             type: String,
             default: 'small',
-            validator: value => VALID_STAR_RATING_SIZES.includes(value)
+            validator: value => !!VALID_STAR_RATING_SIZES[value]
         }
     },
 
@@ -108,9 +109,7 @@ export default {
          * @returns {string}
          */
         getRatingStarPercentage () {
-            const percentage = (this.starRating / this.maxStarRating) * 100;
-
-            return `width: ${percentage}%`;
+            return `${(this.starRating / this.maxStarRating) * 100}%`;
         }
     }
 };
@@ -160,6 +159,7 @@ export default {
         top: 0;
         left: 0;
         overflow: hidden;
+        width: var(--starRatingPercentage);
 
         svg {
             flex-shrink: 0;
--- a/packages/components/molecules/f-rating/src/components/Rating.vue
+++ b/packages/components/molecules/f-rating/src/components/Rating.vue
@@ -2,25 +2,38 @@
     <div
         :class="$style['c-rating']"
         data-test-id="rating-component">
-        <ul
-            :class="$style['c-rating-starWrapper']">
-            <li
-                v-for="star in maxStarRating"
-                :key="star"
-                :class="$style['c-rating-star']">
-                <star-filled-icon
-                    v-if="isRatingStarFilled(star)"
-                    :class="$style['c-rating-star-filled']" />
+        <div :class="$style['c-rating-starWrapper']">
+            <div :class="$style['c-rating-container']">
                 <star-icon
-                    v-else
-                    :class="$style['c-rating-star-empty']" />
-            </li>
-        </ul>
-        <span
-            data-test-id="c-rating-description"
-            class="is-visuallyHidden">
-            {{ getRatingDescription }}
-        </span>
+                    v-for="star in maxStarRating"
+                    :key="star"
+                    :class="[
+                        $style['c-rating-star-empty'],
+                        $style[`c-rating-star--${starRatingSize}`]
+                    ]" />
+            </div>
+
+            <div
+                :class="[
+                    $style['c-rating-mask'],
+                    $style['c-rating-container']
+                ]"
+                :style="getRatingStarPercentage">
+                <star-filled-icon
+                    v-for="star in maxStarRating"
+                    :key="star"
+                    :class="[
+                        $style['c-rating-star-filled'],
+                        $style[`c-rating-star--${starRatingSize}`]
+                    ]" />
+            </div>
+
+            <span
+                data-test-id="c-rating-description"
+                class="is-visuallyHidden">
+                {{ getRatingDescription }}
+            </span>
+        </div>
     </div>
 </template>
 
@@ -31,6 +44,7 @@ import {
 } from '@justeattakeaway/pie-icons-vue';
 import { VueGlobalisationMixin } from '@justeat/f-globalisation';
 import tenantConfigs from '../tenants';
+import { VALID_STAR_RATING_SIZES } from '../constants';
 
 export default {
     name: 'VRating',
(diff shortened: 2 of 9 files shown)


packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,25 +2,38 @@
...
-        </ul>
-        <span
-            data-test-id="c-rating-description"
-            class="is-visuallyHidden">
-            {{ getRatingDescription }}
-        </span>
+                    v-for="star in maxStarRating"
+                    :key="star"
+                    :class="[
+                        $style['c-rating-star-empty'],
```
jamieomaguire, line 11: should this be `c-rating-star--empty` where `empty` is a modifier?

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -80,18 +100,17 @@ export default {
...
-         * @param star {Number}
-         * @returns {boolean}
+         * @returns {string}
          */
-        isRatingStarFilled (star) {
-            return star <= this.starRating;
+        getRatingStarPercentage () {
+            const percentage = (this.starRating / this.maxStarRating) * 100;
+
+            return `width: ${percentage}%`;
```
jamieomaguire: we could keep the same logic here but instead of inlining a width declaration, set the width value as a css variable in the style declaration above i.e. `--starRatingPercentage`. 

The variable will be scoped to the element you add the inline style declaration to and it _might_ be a little more maintenance friendly in the future as you can just set (as an example):

```
.c-rating-mask {
  width: var(--starRatingPercentage);
}
```

The benefit is the actual width rule stays in the css. We're just changing a variable value in the template which (in my opinion) is a bit neater :)

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,25 +2,38 @@
...
+                        $style[`c-rating-star--${starRatingSize}`]
+                    ]" />
+            </div>
+
+            <div
+                :class="[
+                    $style['c-rating-mask'],
+                    $style['c-rating-container']
+                ]"
+                :style="getRatingStarPercentage">
```
jamieomaguire: follow up from my comment below. This could be something like:

```
:style="`--starRatingPercentage: ${getRatingStarPercentage}`"
```

Just guessing the syntax there but I think it's close 🤣

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -103,12 +122,25 @@ export default {
...
+            width: 12px;
+        }
+
+        &--medium {
+            width: 16px;
+        }
+
+        &--large {
+            width: 28px;
+        }
```
jamieomaguire, line 142: Possibly same here in regards to leveraging css variables. We could just set the width to `var(--starRatingSize)` and simply set the variable value based on the size prop provided. It means we can remove all these modifiers.

packages/components/molecules/f-rating/src/constants.js
```diff
@@ -0,0 +1,2 @@
+// eslint-disable-next-line import/prefer-default-export
+export const VALID_STAR_RATING_SIZES = ['small', 'medium', 'large'];
```
jamieomaguire: Probably doesn't matter for an array this size but you could squeeze some performance out of this by switching to using an object as the key look up would be a constant time lookup (`o(1)`) rather than linear which would get worse if the collection grows. 

Pragmatically, for a 3 item array it likely doesn't matter but it would be something like:

```
// I think any truthy value works. We just wanted to check that the object has a key of the value provided to the validator i.e. 'small'
export const VALID_STAR_RATING_SIZES = {
  small: true,
  medium: true,
  large: true
}
```

Then in your validator something like: 

```
validator: value => !!VALID_STAR_RATING_SIZES[value]
```

Again, probably not a big deal here but if the component could be added to a page hundreds of times then it might make a bit of a difference?

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: It's not really a modifier as the icon itself is empty. It's mostly a reference and hook for the empty star icon so it can be styled.

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: Didn't know you could do that. Ill give that a try 👍

packages/components/molecules/f-rating/src/constants.js
jamieomaguire: We discussed offline but just for reference incase others read this - arrays _might_ be better optimised for smaller collections like this. Definitely don't want to mislead anyone into thinking key look-ups are always preferable :)

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: As discussed I think for now we leave this one as it is. I think once we have a set of defined sizes from design we could pull that from fozzie or set them in a constant file instead.

jamieomaguire (approved):

kevinrodrigues:
Will merge this but if there's any other comments please add them here and ill make sure ill add them to the next pr im currently working on.

packages/components/molecules/f-rating/README.md
```diff
@@ -73,6 +73,7 @@ The props that can be defined are as follows (if any):
 | :---          |:--------:|:--------:|:-------:|:-----------------------------------------------------------------------------------------------------------------------|
 | `starRating`  | `Number` |   Yes    |    -    | Sets the displayed rating (filled stars). i.e. for `2 out of x`, 2 would be the `starRating`                           |
 | `maxStarRating`  | `Number` |    No    |    5    | Sets the maximum number of stars that the rating is set against. i.e. for `x out of 5`, 5 would be the `maxStarRating` |
+| `starRatingSize`  | `String` |    No    | 'small' | Sets the component size. By default the component will use `small` the other options are `medium` & `large`            |
```
ashleynolan, line 76: More out of interest – is this sizing taken from design guidelines, or is it based on the star ratings that already exist?

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,25 +2,39 @@
...
-        <ul
-            :class="$style['c-rating-starWrapper']">
-            <li
-                v-for="star in maxStarRating"
-                :key="star"
-                :class="$style['c-rating-star']">
-                <star-filled-icon
-                    v-if="isRatingStarFilled(star)"
-                    :class="$style['c-rating-star-filled']" />
+        <div :class="$style['c-rating-starWrapper']">
```
ashleynolan, line 5: I think the naming can be tidied up here a bit. May be worth pairing to go through it, but I imagine this could just be `c-rating-stars` (as it seems like that's what it's wrapping?)

Give me a shout though and we can catch up and go through it.

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -103,12 +121,25 @@ export default {
...
+    display: inline-block;
 }
+    .c-rating-container {
+        display: flex;
+    }
 
     .c-rating-star {
-        display: inline-block;
-        width: 15px; // Todo - decide on how to size these. Will create a ticket around this.
+        &--small {
```
ashleynolan, line 132: We usually tend to avoid using & for name concatenation, mainly as it makes the code less searchable (so if I did a find inn the codebase for `c-rating-star--small` this wouldn't show up for instance).

I'd probably just write these classes longform for that reason.

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -2,25 +2,39 @@
...
-        </ul>
-        <span
-            data-test-id="c-rating-description"
-            class="is-visuallyHidden">
-            {{ getRatingDescription }}
-        </span>
+                    v-for="star in maxStarRating"
+                    :key="star"
+                    :class="[
+                        $style['c-rating-star-empty'],
```
ashleynolan, line 11: Linked to the name tidy-up mentioned above – atm the moment there doesn't seem to be a base `c-rating-star` class that these other classnames link to (with the modifier etc).

Also, there aren't any styles linked to `c-rating-star-empty` atm by the looks of it either?

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: I was actually thinking that whilst adding these. Will do 👍

packages/components/molecules/f-rating/README.md
kevinrodrigues: Good question as I wasn't actually sure what we're doing with the sizing at the moment. I looked at the snacks component and they have a slightly different size setting. So for now I've decided to stick with legacy JE sizes as that makes the most sense. I think we'd need to go to design and ask them for specific sizes at some point though. What do you think?

ashleynolan: Yeh, this should definitely be documented in figma somewhere (alongside the component designs, if they exist), so should be something we raise to the design team.

kevinrodrigues: Ticket refined for this 👍

kevinrodrigues:
@ashleynolan fixes in this pr: https://github.com/justeat/fozzie-components/pull/2174

PR #2173, opened 2022-10-24
f-checkout@4.7.0 - Add retrieve address from cookie logic

kathywklee:
v4.7.0
*October 24, 2022*

### Added
- Get address from cookies logic for international tenants
- Suburb field for address format in New Zealand

### Removed
-  Get address from local storage logic in the handler of UPDATE_STATE mutation

### DIFF
--- a/packages/components/pages/f-checkout/src/components/Checkout.vue
+++ b/packages/components/pages/f-checkout/src/components/Checkout.vue
@@ -208,6 +208,11 @@ export default {
         checkoutFeatures: {
             type: Object,
             default: () => ({})
+        },
+
+        shouldLoadAddressFromLocalStorage: {
+            type: Boolean,
+            default: true
         }
     },
 
@@ -273,7 +278,7 @@ export default {
                 (!this.address || !this.address.line1);
         },
 
-        /* If phone number is missing both from chckout api and from
+        /* If phone number is missing both from checkout api and from
         * `state.AuthToken`, then retrieve the phone number from customer api
         * This can happen for newly created guest */
         shouldLoadCustomer () {
@@ -429,8 +434,8 @@ export default {
             this.isLoading = true;
 
             const promises = this.isLoggedIn
-                ? [this.loadBasket(), this.loadCheckout(), this.loadAvailableFulfilment()]
-                : [this.loadBasket(), this.loadAddressFromLocalStorage(), this.loadAvailableFulfilment()];
+                ? [this.loadBasket(), this.loadAddressFromStorage(), this.loadCheckout(), this.loadAvailableFulfilment()]
+                : [this.loadBasket(), this.loadAddressFromStorage(), this.loadAvailableFulfilment()];
 
             await Promise.all(promises);
 
@@ -449,11 +454,11 @@ export default {
         },
 
         /**
-         * Update address lines if localStorage is populated.
+         * Update address lines if it is saved in storage.
          *
          * */
-        loadAddressFromLocalStorage () {
-            const address = addressService.getAddressFromLocalStorage(this.tenant);
+        loadAddressFromStorage () {
+            const address = addressService.getAddressFromStorage(this.tenant, this.$cookies, this.shouldLoadAddressFromLocalStorage);
 
             if (address) {
                 this.updateAddress(address);
@@ -773,7 +778,9 @@ export default {
                             url: this.getGeoLocationUrl,
                             postData: locationData,
                             timeout: this.checkoutTimeout,
-                            tenant: this.tenant
+                            tenant: this.tenant,
+                            cookies: this.$cookies,
+                            shouldLoadAddressFromLocalStorage: this.shouldLoadAddressFromLocalStorage
                         });
                     }
                 } catch (error) {
--- a/packages/components/pages/f-checkout/src/services/addressService/index.js
+++ b/packages/components/pages/f-checkout/src/services/addressService/index.js
@@ -14,7 +14,7 @@ const tenants = {
     ie
 };
 
-const storedLocationKey = 'je-full-address-details';
+const localStorageStoredLocationKey = 'je-full-address-details';
 
 function toFormattedPostcode (postcode, tenant) {
     const shouldFormatPostCode = tenant === 'uk';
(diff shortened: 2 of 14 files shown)


packages/components/pages/f-checkout/src/store/checkout.module.js
```diff
@@ -412,39 +423,35 @@ export default {
...
-                    window.localStorage.setItem('je-full-address-details', JSON.stringify({
-                        PostalCode: addressDetails.postcode,
-                        Line1: addressDetails.line1,
-                        Line2: addressDetails.line2,
-                        Line3: addressDetails.line3,
-                        Line4: addressDetails.line4,
-                        administrativeArea: addressDetails.administrativeArea,
-                        City: addressDetails.locality
-                    }));
+            if (shouldLoadAddressFromLocalStorage) {
```
matthewhardern: this is a lot of nested if statements

kathywklee: We may improve the nested if case by combining the conditions from if clauses, and replace else to another condition. What do you think?

matthewhardern: well I feel like the work should be done in the address service so instead here you just have for example `addressService.getAddress` whether the address is returned from cookie or localstorage is the responsibility of the address service.

matthewhardern: this would avoid duplicating 
```
const storedAddress = addressService.getAddressFromLocalStorage(tenant, false);
addressCoords = getGeoLocationFromAddress(addressDetails, storedAddress);

File: packages/components/pages/f-checkout/src/store/checkout.module.js [hunk 2]
```diff
@@ -412,39 +423,35 @@ export default {
...
+                if (isAddressInLocalStorage) {
+                    const storedAddress = addressService.getAddressFromLocalStorage(tenant, false);
+
+                    addressCoords = getGeoLocationFromAddress(addressDetails, storedAddress);
```
matthewhardern: Just out of curiosity what is this bit doing in grand scheme of things?

packages/components/pages/f-checkout/src/components/Checkout.vue
```diff
@@ -449,11 +449,14 @@ export default {
...
 
         /**
-         * Update address lines if localStorage is populated.
+         * Update address lines if it is saved in storage (cookie or local storage), varies among tenants.
          *
          * */
-        loadAddressFromLocalStorage () {
-            const address = addressService.getAddressFromLocalStorage(this.tenant);
+        loadAddressFromStorage () {
+            const shouldLoadAddressFromLocalStorage = addressService.shouldLoadAddressFromLocalStorage(this.tenant);
```
juan-amor: IMHO the Vue component doesn't need to know how the address is retrieved - the service needs to decide where to get it from (localstorage or cookie). Otherwise, you have the getting address logic split and any change in the service may require changing the component too.

It's cleaner to do 
```
loadAddressFromStorage () {
     const address = addressService.getAddress(this.tenant) 
...
```
In reality, the component needs an address and it's the service's responsibility to provide one, hiding the implementation details.

packages/components/pages/f-checkout/src/store/checkout.module.js
kathywklee: Our logic to obtain geolocation is: 1. get it from storage 2. if it doesn't exist in 1, get it from API.
For now we only support restoring geolocation from local storage only. However for some tenants we store such an info in cookies instead. Therefore we need to improve it, so that the geolocation data can be retrieved and use it later on.

packages/components/pages/f-checkout/src/services/addressService/index.js
```diff
@@ -90,25 +90,31 @@ function getAddressClosestToPostcode (currentPostcode, addressData, tenant) {
     return address || addresses?.find(a => a && a.ZipCode.startsWith(postcode));
 }
 
+const shouldLoadAddressFromLocalStorage = tenant => tenant === 'uk';
```
juan-amor: Can this tenant comparison be made outside the service (i.e. via feature toggle or JET FM)? I don't think tenant-specific logic should live in the code if possible

packages/components/pages/f-checkout/src/store/checkout.module.js
kathywklee: I see. In this case I think this would also work.

packages/components/pages/f-checkout/src/services/addressService/index.js
```diff
@@ -90,25 +90,31 @@ function getAddressClosestToPostcode (currentPostcode, addressData, tenant) {
...
+
 /**
  * Checks whether the user has an address in local storage available to us
  * @returns {boolean} - whether the address is in local storage or now
  */
 const isAddressInLocalStorage = () => {
     if (window.localStorage) {
-        const address = window.localStorage.getItem(storedLocationKey) ? JSON.parse(window.localStorage.getItem(storedLocationKey)) : null;
+        const address = window.localStorage.getItem(localStorageStoredLocationKey) ? JSON.parse(window.localStorage.getItem(localStorageStoredLocationKey)) : null;
```
juan-amor: Calling twice getting the item from local storage, why not:
```
const item  = window.localStorage?.getItem(localStorageStoredLocationKey);
const address = !!item ? JSON.parse(item) : null;
...
```

packages/components/pages/f-checkout/src/services/addressService/index.js
```diff
@@ -168,13 +188,76 @@ function getClosestAddress (addressData, tenant, currentPostcode) {
...
+ * @param tenant - The current tenant
+ * @param cookies - Cookies object
+ * @param mapped - boolean deciding whether to return the address in a mapped state
+ * @returns address - The address from cookie
+ */
+function getAddressFromCookie (tenant, cookies, mapped = true) {
+    if (!cookies || typeof cookies.get !== 'function') return null;
+    let storedAddress = null;
+
+    const houseNumber = cookies.get('je-last_houseNo_used');
```
juan-amor, line 194: Can all the address info set in one cookie rather than one cookie per field? Seems excessive...

kathywklee: Unfortunately no, for now we have to use the existing mechanism in legacy repository.

packages/components/pages/f-checkout/src/services/addressService/index.js
kathywklee: I created a prop instead because this flag depends on the architecture of the existing system. It doesn't sound like a feature.

juan-amor (dismissed):
LGTM

matthewhardern (dismissed):
LGTM

matthewhardern (approved):

juan-amor (approved):

PR #2174, opened 2022-10-24
f-rating@0.5.0 - Add description to rating component & tests.

kevinrodrigues:
### Added
- Visual tests.
- Prop validation.
- Props `ratingDisplayType` & `totalReviews`.
- Storybook contains ratings in different sizes & formats.

### Changed
- Class names to be more suitable.

## Screenshots
(The alignment for small is a bit off but for now ignore it as we need to ask design about displaying text next to small versions)
Screenshot 2022-10-24 at 17 34 14
Screenshot 2022-10-24 at 17 34 23
Screenshot 2022-10-24 at 17 34 35
Screenshot 2022-10-24 at 17 34 47

### DIFF
--- a/packages/components/molecules/f-rating/README.md
+++ b/packages/components/molecules/f-rating/README.md
@@ -75,7 +75,7 @@ The props that can be defined are as follows (if any):
 | `maxStarRating`  | `Number` |    No    |    5    | Sets the maximum number of stars that the rating is set against. i.e. for `x out of 5`, 5 would be the `maxStarRating` | -
 | `starRatingSize`  | `String` |    No    | 'small' | Sets the component size. By default the component will use `small`.                                                    | `small` `medium` `large`
 | `ratingDisplayType`  | `String` |    No    |  null   | Sets the descriptive text which will be displayed next to the component. By default there will be no text.             | `noRating` `short` `medium` `long`
-| `totalReviews`  | `Number` |    No    |  null   | Sets the total number of reviews to display. Example: `View 499 reviews` where `499` is the total.                     | -
+| `reviewCount`  | `Number` |    No    |  null   | Sets the total number of reviews to display. Example: `View 499 reviews` where `499` is the total.                     | -
 
 ### Events
 
--- a/packages/components/molecules/f-rating/README.md
+++ b/packages/components/molecules/f-rating/README.md
@@ -69,11 +69,13 @@ There may be props that allow you to customise its functionality.
 
 The props that can be defined are as follows (if any):
 
-| Prop          |   Type   | Required | Default | Description                                                                                                            |
-| :---          |:--------:|:--------:|:-------:|:-----------------------------------------------------------------------------------------------------------------------|
-| `starRating`  | `Number` |   Yes    |    -    | Sets the displayed rating (filled stars). i.e. for `2 out of x`, 2 would be the `starRating`                           |
-| `maxStarRating`  | `Number` |    No    |    5    | Sets the maximum number of stars that the rating is set against. i.e. for `x out of 5`, 5 would be the `maxStarRating` |
-| `starRatingSize`  | `String` |    No    | 'small' | Sets the component size. By default the component will use `small` the other options are `medium` & `large`            |
+| Prop          |   Type   | Required | Default | Description                                                                                                            | Options |
+| :---          |:--------:|:--------:|:-------:|:-----------------------------------------------------------------------------------------------------------------------|---------|
+| `starRating`  | `Number` |   Yes    |    -    | Sets the displayed rating (filled stars). i.e. for `2 out of x`, 2 would be the `starRating`                           | -
+| `maxStarRating`  | `Number` |    No    |    5    | Sets the maximum number of stars that the rating is set against. i.e. for `x out of 5`, 5 would be the `maxStarRating` | -
+| `starRatingSize`  | `String` |    No    | 'small' | Sets the component size. By default the component will use `small`.                                                    | `small` `medium` `large`
+| `ratingDisplayType`  | `String` |    No    |  null   | Sets the descriptive text which will be displayed next to the component. By default there will be no text.             | `noRating` `short` `medium` `long`
+| `totalReviews`  | `Number` |    No    |  null   | Sets the total number of reviews to display. Example: `View 499 reviews` where `499` is the total.                     | -
 
 ### Events
 
--- a/packages/components/molecules/f-rating/src/components/Rating.vue
+++ b/packages/components/molecules/f-rating/src/components/Rating.vue
@@ -23,7 +23,6 @@
                 <star-filled-icon
                     v-for="star in maxStarRating"
                     :key="star"
-                    :style="`--starRatingSize: c-rating-star--${ratingDisplayType}`"
                     :class="[
                         $style['c-rating-star'],
                         $style['c-rating-star--filled'],
--- a/packages/components/molecules/f-rating/src/components/Rating.vue
+++ b/packages/components/molecules/f-rating/src/components/Rating.vue
@@ -90,7 +90,7 @@ export default {
         },
         ratingDisplayType: {
             type: String,
-            default: 'noRating',
+            default: null,
             validator: value => VALID_STAR_RATING_DISPLAY_TYPE.includes(value)
         },
         totalReviews: {
--- a/packages/components/molecules/f-rating/src/components/Rating.vue
+++ b/packages/components/molecules/f-rating/src/components/Rating.vue
@@ -137,7 +137,7 @@ export default {
 
     methods: {
         /**
-         * Display review rating message format.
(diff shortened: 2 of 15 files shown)


packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
```diff
@@ -157,5 +158,97 @@ describe('Rating', () => {
...
+                            propsData,
+                            localVue,
+                            i18n
+                        });
+
+                        // Act
+                        const result = wrapper.vm.getRatingDisplayFormat();
+
+                        // Assert
+                        expect(result).toBe('No ratings yet');
```
kevinrodrigues: I will change this tomorrow because the default rating component should display just the stars with no text.

kevinrodrigues: Fixed.

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -111,44 +133,67 @@ export default {
...
+         */
+        getRatingDisplayFormat () {
+            return this.ratingDisplayType
+                ?
+                this.$t(`ratings.ratingDisplayType.${this.ratingDisplayType}`, {
+                    rating: this.starRating,
+                    maxStarRating: this.maxStarRating,
+                    totalReviews: this.totalReviews
+                })
+                : false;
```
juan-amor: Quick question but why return false rather than null or an empty string?

packages/components/molecules/f-rating/src/tests/constants.test.js
```diff
@@ -1,7 +1,10 @@
-import { VALID_STAR_RATING_SIZES } from '../constants';
+import {
+    VALID_STAR_RATING_SIZES,
+    VALID_STAR_RATING_DISPLAY_TYPE
+} from '../constants';
 
 describe('`constants`', () => {
-    describe('VALID_STAR_RATING_SIZES', () => {
+    describe('`VALID_STAR_RATING_SIZES`', () => {
```
matthewhardern, line 7: any reason for these back ticks

packages/components/molecules/f-rating/src/tests/constants.test.js
```diff
@@ -14,4 +17,17 @@ describe('`constants`', () => {
             expect(VALID_STAR_RATING_SIZES).toEqual(values);
         });
     });
+
+    describe('`VALID_STAR_RATING_DISPLAY_TYPE`', () => {
```
matthewhardern, line 21: same as above

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -111,44 +133,67 @@ export default {
...
-    .c-rating-container {
+    .c-rating-stars-icons {
         display: flex;
     }
 
     .c-rating-star {
-        &--small {
-            width: 12px;
-        }
+        width: 12px;
```
matthewhardern, line 176: feel these should be variables at the top so they could be easily changed, more ideally / out of scope for this is that icon component has a size parameter.

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
```diff
@@ -157,5 +158,97 @@ describe('Rating', () => {
...
+        describe('`ratingDisplayType`', () => {
+            it.each(VALID_STAR_RATING_DISPLAY_TYPE)('should allow correct value `%s`', displayType => {
+                // Act
+                const { validator } = VRating.props.ratingDisplayType;
+
+                // Assert
+                expect(validator(displayType)).toBe(true);
+            });
+
+            describe('when `ratingDisplayType` is truthy', () => {
```
juan-amor, line 171: Does it require a test when the `ratingDisplayType` is `falsy` ?

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -111,44 +133,67 @@ export default {
...
+        /**
+         * Display review rating message format.
+         *
+         * @todo - If the component is using `short` as a `starRatingSize` we shouldn't display text
+         * alongside it for now. (TBC with design - ticket in backlog).
+         *
+         * @returns {string|boolean}
+         */
+        getRatingDisplayFormat () {
+            return this.ratingDisplayType
```
matthewhardern: seems unnecessary as it is only being called when v-if check for `ratingDisplayType` is true?

packages/components/molecules/f-rating/src/tests/constants.test.js
kevinrodrigues: Just a convention which i follow, in a test runner the back ticks highlight so it's easier to read the tests when they run. Can be removed tho

packages/components/molecules/f-rating/src/components/Rating.vue
jamieomaguire: I'd caution against returning a boolean here if the other return type is a string. Can lead to a bit of confusion down the line! I'd just return the copy and rely on other logic to decide if it should be used or not

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: There's a ticket in our backlog about sizing so will be tackled there :)

jamieomaguire: I agree with this :)

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: I see what you mean. Will fix. 👍

packages/components/molecules/f-rating/README.md
```diff
@@ -69,11 +69,13 @@ There may be props that allow you to customise its functionality.
...
+| :---          |:--------:|:--------:|:-------:|:-----------------------------------------------------------------------------------------------------------------------|---------|
+| `starRating`  | `Number` |   Yes    |    -    | Sets the displayed rating (filled stars). i.e. for `2 out of x`, 2 would be the `starRating`                           | -
+| `maxStarRating`  | `Number` |    No    |    5    | Sets the maximum number of stars that the rating is set against. i.e. for `x out of 5`, 5 would be the `maxStarRating` | -
+| `starRatingSize`  | `String` |    No    | 'small' | Sets the component size. By default the component will use `small`.                                                    | `small` `medium` `large`
+| `ratingDisplayType`  | `String` |    No    |  null   | Sets the descriptive text which will be displayed next to the component. By default there will be no text.             | `noRating` `short` `medium` `long`
+| `totalReviews`  | `Number` |    No    |  null   | Sets the total number of reviews to display. Example: `View 499 reviews` where `499` is the total.                     | -
```
xander-marjoram: Would `reviewCount` be a bit more succinct?

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -111,44 +133,67 @@ export default {
         getRatingStarPercentage () {
             return `${(this.starRating / this.maxStarRating) * 100}%`;
```
xander-marjoram: I know this is from a different PR, but would it make more sense to have this return a number, and then add the percentage sign in the style tag?

packages/components/molecules/f-rating/README.md
kevinrodrigues: Not great with naming, ill change it 👍

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: I think i had something like this before and then was asked to change it, ill have a look at the previous comments and see if i can move it back 👍

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: Will fix.

packages/components/molecules/f-rating/src/components/Rating.vue

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: Removed.

packages/components/molecules/f-rating/README.md
kevinrodrigues: Fixed.

packages/components/molecules/f-rating/src/components/Rating.vue

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
kevinrodrigues: Fixed.

juan-amor (approved):

ashleynolan (approved):

PR #2175, opened 2022-10-24
fozzie-components@v7.40.0 - Pages / Templates / Services / Tools - Remove duplicate dependencies #globalconfig

siggerzz:


zlatinivanov-je:
I looked into this and it looks ok, but struggling to understand what is it trying to achieve, a description would help. 
Also how is it ensuring no regressions are introduced conflicting versions of dependencies are avoided in host apps.

siggerzz:
> I looked into this and it looks ok, but struggling to understand what is it trying to achieve, a description would help.
> Also how is it ensuring no regressions are introduced conflicting versions of dependencies are avoided in host apps.

Will add a description once I've managed to iron out these outstanding issues.

The general idea is to reduce the duplication of dependencies where we have multiple versions across components. Where possible, we should have a single version of a dependency defined in the root that is used by all components. Doing so will make it easier for tools like renovate to update deps without issue 👍🏼

siggerzz:
Closing in favour of https://github.com/justeat/fozzie-components/pull/2208

PR #2178, opened 2022-10-26
f-rating@0.6.0 - add single star variant

kevinrodrigues:
### Added
- rating.scss to shared styles between components.
- Unit tests to cover changes.
- Visual tests.

### Changed
- Split out from main rating file into Five & Single star components.

## Screenshots
Screenshot 2022-10-26 at 16 11 58
Screenshot 2022-10-26 at 16 13 46
Screenshot 2022-10-26 at 16 13 54
Screenshot 2022-10-26 at 16 14 02


jamieomaguire:
Screenshot 2022-10-27 at 09 55 41
is this variant in there @kevinrodrigues ? 

Tiny NIT: I think we need a bit more whitespace between the star and the text

Also the alignment for the really small size seems a little off? Is that intentional?

jamieomaguire:
Screenshot 2022-10-27 at 09 57 07
I dont think this variant makes sense.

Do we have this? 
Screenshot 2022-10-27 at 09 57 23

jamieomaguire:
Screenshot 2022-10-27 at 09 58 28
Is this a valid variant? I can't imagine this making much sense to the user - it looks a bit like it's 499 1 star reviews

packages/components/molecules/f-rating/src/assets/scss/rating.scss
```diff
@@ -0,0 +1,41 @@
...
+.c-rating-star {
+    width: 12px;
+}
+
+.c-rating-star--medium {
+    width: 16px;
+}
+
+.c-rating-star--large {
+    width: 28px;
```
jamieomaguire, line 28: can we use `spacing` here?

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -3,32 +3,11 @@
...
-                <star-filled-icon
-                    v-for="star in maxStarRating"
-                    :key="star"
-                    :class="[
-                        $style['c-rating-star'],
-                        $style['c-rating-star--filled'],
-                        $style[`c-rating-star--${starRatingSize}`]
-                    ]" />
-            </div>
+            <component
```
jamieomaguire, line 6: 🔥

kevinrodrigues:
> Screenshot 2022-10-27 at 09 55 41
> 
> is this variant in there @kevinrodrigues ?
> Tiny NIT: I think we need a bit more whitespace between the star and the text
> 
> Also the alignment for the really small size seems a little off? Is that intentional?

Where is this version available? For the alignment the sizes are going to be dealt with in another ticket and there probably wont be any text next to the small star so i have left it as is for now until then.

packages/components/molecules/f-rating/src/assets/scss/rating.scss
kevinrodrigues: Sizing will be in another ticket.

kevinrodrigues:
> Screenshot 2022-10-27 at 09 57 07
> 
> I dont think this variant makes sense.
> Do we have this? Screenshot 2022-10-27 at 09 57 23

This one is just a SB demo. I can change it but basically there's a drop down allowing you to view all variants against all stars.

packages/components/molecules/f-rating/src/components/RatingSingleStarVariant.vue
```diff
@@ -0,0 +1,58 @@
...
+        StarIcon
+    },
+
+    props: {
+        starRating: {
+            type: Number,
+            default: 0
+        },
+        starRatingSize: {
+            type: String,
```
jamieomaguire, line 42: is `String` the appropriate type here? It will usually come down as a number I believe

packages/components/molecules/f-rating/src/components/RatingFiveStarVariant.vue
```diff
@@ -0,0 +1,89 @@
...
+    },
+
+    props: {
+        starRating: {
+            type: Number,
+            default: 0
+        },
+        maxStarRating: {
+            type: Number,
+            default: 5
```
xander-marjoram: If this can be configured, should this be a "multi-star variant" rather than specifically a five-star variant?

packages/components/molecules/f-rating/src/components/RatingSingleStarVariant.vue
```diff
@@ -0,0 +1,58 @@
...
+    }
+};
+</script>
+
+<style lang="scss" module>
+@use '@justeat/fozzie/src/scss/fozzie' as f;
+@use '../../src/assets/scss/rating';
+
+.c-rating-star--empty {
+    & path {
```
xander-marjoram: Nit: does the `&` add anything here?

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
```diff
@@ -43,11 +43,26 @@ describe('Rating', () => {
...
+                        starRating: 1
+                    };
+                    wrapper = shallowMount(VRating, {
+                        propsData,
+                        localVue,
+                        i18n
+                    });
+
+                    // Act & Assert
+                    expect(wrapper.vm.getRatingDescription).toMatchSnapshot();
```
xander-marjoram: I know this test wasn't actually changed as part of this PR, but IMO a description is fairly simple so the expected text should live within the test so it's easy to see at a glance. For large and complex objects, snapshots are great.

Also, how does this work for decimals? I would expect `1.8 stars` rather than `1.8 star`. Might be worth raising a ticket to look into this.

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
```diff
@@ -75,38 +90,44 @@ describe('Rating', () => {
...
+                        wrapper = shallowMount(VRating, {
+                            propsData,
+                            localVue,
+                            i18n
+                        });
 
-                    // Act & Assert
-                    expect(wrapper.vm.getRatingDescription).toMatchSnapshot();
+                        // Act & Assert
+                        expect(wrapper.vm.setRatingVariant).toBe('rating-single-star');
```
xander-marjoram: You can also check for the actual component by importing it and doing something like:
```javascript
expect(wrapper.findComponent(RatingSingleStar).exists()).toBe(true);
```

packages/components/molecules/f-rating/src/components/_tests/RatingFiveStarVariant.test.js
```diff
@@ -0,0 +1,66 @@
...
+                    localVue,
+                    i18n
+                });
+
+                // Act & Assert
+                expect(wrapper.vm.getRatingStarPercentage).toBeDefined();
+            });
+
+            describe('when invoked', () => {
+                it('should return a correct calculated number from a combination of `starRating` and `maxStarRating`', () => {
```
xander-marjoram, line 49: Just a passing thought, it would be great if we could test the sass variable too (obviously not expecting you to cover that in this PR!)

packages/components/molecules/f-rating/test/visual/f-rating.visual.spec.js
```diff
@@ -123,5 +123,27 @@ devices.forEach(device => {
                 });
             });
         });
+
+        describe('`isSingleStarVariant`', () => {
+            describe('when populated with `false`', () => {
+                it('should display the five star variant', async () => {
```
xander-marjoram, line 129: Are we planning on adding visual tests for more ratings in future?

packages/components/molecules/f-rating/src/components/RatingSingleStarVariant.vue
xander-marjoram: I think it means `'small'`, `'medium'` or `'large'`? The parent component has a validator on the prop so only those values should get through.
Although... taking a look at the constants file it looks like an object, I was expecting to see an array containing those three values

kevinrodrigues: Yes it was originally an array. @jamieomaguire suggested an object, but for the second constant i went back to using arrays and would like to revert this object one too now. I feel it would be small enough to not worry about loops etc (something that came up in another pr) thoughts?

kevinrodrigues: So to answer the first question, it would be in the form of a string as Xander mentioned.

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
kevinrodrigues: Originally had the text instead of snapshot. Will change back. With the decimals that's a good spot. Ill add a ticket for this.

packages/components/molecules/f-rating/src/components/RatingSingleStarVariant.vue
xander-marjoram: Personally I think an array is a bit cleaner and results in some simpler code, `VALID_STAR_RATING_SIZES.includes(value)` rather than `!!VALID_STAR_RATING_SIZES[value]`.

I do understand there might be performance benefits, especially for a page like SERP where this component could appear hundreds of times. This could be a good candidate for benchmarking?

packages/components/molecules/f-rating/src/components/RatingFiveStarVariant.vue
kevinrodrigues: Yeah. I may also change the component name too as having a flexible amount may be a good idea.

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
kevinrodrigues: Nice

packages/components/molecules/f-rating/src/components/_tests/RatingFiveStarVariant.test.js
kevinrodrigues: Will add to the next PR if that's ok?

packages/components/molecules/f-rating/test/visual/f-rating.visual.spec.js
kevinrodrigues: Will add.

packages/components/molecules/f-rating/src/components/RatingFiveStarVariant.vue
kevinrodrigues: Have updated just the name of the files now. I think i'll leave default as 5 but it can actually work with any number. Will update the readme too.
Screenshot 2022-10-27 at 18 39 05

packages/components/molecules/f-rating/src/components/RatingSingleStarVariant.vue
kevinrodrigues: Reverted.

packages/components/molecules/f-rating/test/visual/f-rating.visual.spec.js
kevinrodrigues: Will bundle these up in another PR if that's ok.

packages/components/molecules/f-rating/src/assets/scss/rating.scss
```diff
@@ -0,0 +1,37 @@
...
+    padding: 0;
+    position: relative;
+    display: inline-block;
+}
+.c-rating-stars-icons,
+.c-rating-stars-icon {
+    display: flex;
+}
+
+.c-rating-star {
```
juan-amor, line 19: should this be `c-rating-star-small` instead to make it consistent with the other classes?

packages/components/molecules/f-rating/src/components/RatingMultiStarVariant.vue
```diff
@@ -0,0 +1,89 @@
...
+            type: Number,
+            default: 0
+        },
+        maxStarRating: {
+            type: Number,
+            default: 5
+        },
+        starRatingSize: {
+            type: String,
+            default: ''
```
juan-amor, line 57: equivalently this should be set to one of the values in the array?

packages/components/molecules/f-rating/src/components/RatingMultiStarVariant.vue
```diff
@@ -0,0 +1,89 @@
...
+    },
+
+    computed: {
+        /**
+         * Calculate a percentage from the `starRating` value passed in by the consuming application.
+         *
+         * @returns {Number}
+         */
+        getRatingStarPercentage () {
+            return (this.starRating / this.maxStarRating) * 100;
```
juan-amor, line 68: caution if the `maxStarRating` prop is 0; probably a good idea to check this

xander-marjoram: We could have a validator on `maxStarRating` to only allow positive integers

packages/components/molecules/f-rating/src/assets/scss/rating.scss
kevinrodrigues: Was part of a discussion with @ashleynolan. Had initially had that but we decided to use the default class as the smaller variant.

ashleynolan (dismissed):

packages/components/molecules/f-rating/src/assets/scss/rating.scss
xander-marjoram: Sounds good, might as well make 'small' the default value for `starRatingSize` then 😄

packages/components/molecules/f-rating/src/components/RatingMultiStarVariant.vue
kevinrodrigues: Good shout. Will add.

packages/components/molecules/f-rating/src/components/RatingMultiStarVariant.vue
kevinrodrigues: So i dont think i need to here as the main component rating.vue passes down the prop default `small`. Here it just needs to be there so i can reference it unless there is a better way? Alternatively i could just set it for clarity?

xander-marjoram: I'm happy with that being the case

packages/components/molecules/f-rating/src/components/RatingMultiStarVariant.vue
kevinrodrigues: Fixed. Added test too.

juan-amor (dismissed):

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -81,12 +58,13 @@ export default {
         },
         maxStarRating: {
             type: Number,
-            default: 5
+            default: 5,
+            validator: value => value > 0
```
xander-marjoram: I think at the moment it would also allow something like 2.5? Happy for you to address in a future PR though

kevinrodrigues: Nice spot! Fixed.

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
```diff
@@ -257,14 +257,22 @@ describe('Rating', () => {
             });
 
             describe('validator', () => {
-                it('should only allow positive integers', () => {
+                it('should only allow positive whole integers', () => {
```
xander-marjoram: "Integer" already implies a "whole" number

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -59,7 +59,7 @@ export default {
         maxStarRating: {
             type: Number,
             default: 5,
-            validator: value => value > 0
+            validator: value => value > 0 && value % 1 === 0 // Only allow positive `whole` numbers larger than zero and exclude numbers like 2.5.
```
xander-marjoram: This is a bit more concise
```suggestion
validator: value => value > 0 && value % 1 === 0 // Positive integers
```

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
kevinrodrigues: Fixed.

xander-marjoram (approved):

juan-amor (approved):

packages/components/molecules/f-rating/test/visual/f-rating.visual.spec.js
kevinrodrigues: Covered here: https://github.com/justeat/fozzie-components/pull/2183

PR #2179, opened 2022-10-26
f-registration@3.8.0 Use Vue I18n

adammorr:
---

This PR updates f-registration to use Vue I18n for text localization as opposed to manually pulling them in for the current locale and exposing through a `copy` property. We now also show the correct form error message when we have 1 or many errors.

The reason for this change was two fold; firstly to bring f-registration into line with other components and secondly to leverage translated interpolated strings using the I18n framework.

---

## UI Review Checks

I haven't changed the UI with my change but thought it useful to show the translations

_en-GB_
image

_en-GB - single error_
image

_it-IT_
image

_it-IT - single error_
image

_en-NZ_
image

_es-ES_
image

_es-ES - single error_
image

### DIFF
--- a/packages/components/pages/f-registration/package.json
+++ b/packages/components/pages/f-registration/package.json
@@ -3,7 +3,7 @@
   "description": "Fozzie Registration Form Component",
   "version": "3.7.3",
   "main": "dist/f-registration.umd.min.js",
-  "maxBundleSize": "70kB",
+  "maxBundleSize": "75kB",
   "files": [
     "dist",
     "test-utils",
--- a/packages/components/pages/f-registration/package.json
+++ b/packages/components/pages/f-registration/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-registration",
   "description": "Fozzie Registration Form Component",
-  "version": "3.7.2",
+  "version": "3.7.3",
   "main": "dist/f-registration.umd.min.js",
   "maxBundleSize": "70kB",
   "files": [
--- a/packages/components/pages/f-registration/package.json
+++ b/packages/components/pages/f-registration/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@justeat/f-registration",
   "description": "Fozzie Registration Form Component",
-  "version": "3.7.3",
+  "version": "3.8.0",
   "main": "dist/f-registration.umd.min.js",
   "maxBundleSize": "75kB",
   "files": [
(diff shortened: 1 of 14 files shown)


ZsofiaS (dismissed):

siggerzz:
@adammorr Looks like bundlewatch is failing because the bundlesize has increased. Could you update the `maxBundleSize` property in `package.json` to 75kb?

packages/components/pages/f-registration/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "@justeat/f-registration",
   "description": "Fozzie Registration Form Component",
-  "version": "3.7.2",
+  "version": "3.7.3",
```
xander-marjoram: Nit: This should probably be a minor version, so `3.8.0`

packages/components/pages/f-registration/src/tenants/en-NZ.js
```diff
@@ -1,62 +1,59 @@
...
-            maxLengthError: 'Email address exceeds 50 characters',
-            invalidEmailError: 'Please enter your email address correctly',
-            alreadyExistsError: 'Email address is already registered'
-        },
-
-        password: {
-            requiredError: 'Please enter a password',
-            minLengthError: 'Password is less than ten characters '
+        validationMessages: {
+            formError: 'There are {errorCount} errors in the form',
```
xander-marjoram: Small thing that we can maybe address in future, if there is only 1 error the text will say "There are 1 errors in the form"

hwoodiwiss (dismissed):

packages/components/pages/f-registration/src/tenants/en-NZ.js
ZsofiaS: The same goes for the Spanish & Italian translations (to be confirmed with native speakers)

ZsofiaS (dismissed):

packages/components/pages/f-registration/src/components/_tests/helpers/setup.js
```diff
@@ -0,0 +1,20 @@
+import tenantConfigs from '../../../tenants';
+
+const defaultPropData = {
+    locale: 'en-GB',
+    createAccountUrl: 'http://localhost/account/register',
+    showLoginLink: true,
+    loginUrl: '/account/register'
```
xander-marjoram: If you change this to `/account/login` does it break any tests? I know it's just a placeholder value but it would probably be better if it's somewhat correct 😄

adammorr: I didn’t even notice that 🙈 I’ll change and see

packages/components/pages/f-registration/src/tenants/en-NZ.js
adammorr: We can do this now as there are appropriate translations in f-checkout

xander-marjoram (approved):

ZsofiaS (approved):

PR #2183, opened 2022-11-01
f-rating@0.7.0 - Add way to handle visually hidden text.

kevinrodrigues:
### Added
- Visual tests to cover changes in v0.6.0
- `hasRatingAvailable` computed value to handle when to display visually hidden text.


packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -1,7 +1,7 @@
 <template>
     <div
         :class="$style['c-rating']"
-        data-test-id="rating-component">
+        :data-test-id="`${setRatingVariant}-component`">
```
xander-marjoram: Isn't this computed prop more of a `get` than a `set`?

packages/components/molecules/f-rating/stories/Rating.stories.js
```diff
@@ -33,25 +30,14 @@ export const RatingComponent = (args, { argTypes }) => ({
         </div>`
 });
 
-RatingComponent.storyName = 'f-rating';
+RatingComponent.storyName = 'f-rating Multi Star';
```
xander-marjoram, line 33: Nit: If we're updating this do we want to update the filename?

packages/components/molecules/f-rating/src/components/Rating.vue

packages/components/molecules/f-rating/stories/Rating.stories.js
kevinrodrigues: Fixed.

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -112,10 +113,20 @@ export default {
...
+        },
+
+        /**
+         * Allow only valid rating display types, so we show
+         * visually hidden text in those cases only.
+         *
+         * @returns {boolean}
+         */
+        hasRatingAvailable () {
+            return this.ratingDisplayType !== 'noRating';
```
xander-marjoram, line 129: Had a chat with @kevinrodrigues and we're gonna look to change this to something like `return this.reviewCount > 0;`.

kevinrodrigues: Will create a ticket around this and pick it up next. 👍

packages/components/molecules/f-rating/test/visual/f-rating-singleStar.visual.spec.js
```diff
@@ -0,0 +1,142 @@
...
+            });
+        });
+
+        describe('`starRatingSize`', () => {
+            it('should be displayed at the correct size', async () => {
+                // Act
+                await RatingSingleStar.load({ isSingleStarVariant: true, starRatingSize: 'small' });
+
+                // Assert
+                await browser.percyScreenshot('f-rating Single Star - Visual Test for Prop - starRatingSize = small', device);
```
xander-marjoram, line 55: In future it would be good to combine these few tests with `it.each`

xander-marjoram (approved):

packages/components/molecules/f-rating/test/visual/f-rating-singleStar.visual.spec.js
kevinrodrigues: I did initially try that but as the loop at the top has a `each` attached to it, it caused issues. I'll take a look at resolving that next.

packages/components/molecules/f-rating/src/components/Rating.vue
dawidchar: Just a little improvement, that could be added to that ticket too. 😄 

https://github.com/justeat/fozzie-components/blob/ddae7b31ef217ba30af775058149a3c93746329c/packages/components/molecules/f-rating/src/tenants/en-GB.js#L4
https://github.com/justeat/fozzie-components/blob/ddae7b31ef217ba30af775058149a3c93746329c/packages/components/molecules/f-rating/src/components/Rating.vue#L98-L109

Updated to
``` javascript
getRatingDescription () {
  return this.$tc('ratings.starsDescription', this.starRating , {
      rating: this.starRating,
      maxStarRating: this.maxStarRating
  })
}
```

[i18n Docs](https://kazupon.github.io/vue-i18n/guide/pluralization.html#accessing-the-number-via-the-pre-defined-argument)

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
```diff
@@ -133,6 +133,73 @@ describe('Rating', () => {
...
+                            propsData,
+                            localVue,
+                            i18n
+                        });
+                    });
+
+                    it('should return truthy', () => {
+                        // Act & Assert
+                        expect(wrapper.vm.hasRatingAvailable).toBe(false);
+                    });
```
dawidchar, line 191: NIT: "truthy" but expects false 😄

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: Will be looking at this method in another ticket. Will look at improving there :)

juan-amor (approved):

packages/components/molecules/f-rating/test/visual/f-rating-singleStar.visual.spec.js
```diff
@@ -0,0 +1,142 @@
...
+            });
+        });
+
+        describe('`maxStarRating`', () => {
+            it('should be displayed', async () => {
+                // Act
+                await RatingSingleStar.load({ isSingleStarVariant: true, starRating: 5 });
+
+                // Assert
+                await browser.percyScreenshot('f-rating Single Star - Visual Test for Prop - maxStarRating = 5', device);
```
dawidchar, line 35: Testing `maxStarRating` but prop is `starRating`

packages/components/molecules/f-rating/test/visual/f-rating.visual.spec.js
```diff
@@ -32,7 +32,7 @@ devices.forEach(device => {
                 await Rating.load({ starRating: 5 });
 
                 // Assert
-                await browser.percyScreenshot('f-rating - Visual Test for Prop - maxStarRating = 5', device);
+                await browser.percyScreenshot('f-rating Multi Star - Visual Test for Prop - maxStarRating = 5', device);
```
dawidchar, line 35: Testing `maxStarRating` but prop is `starRating` 😄

kevinrodrigues: Good spot on these 👍

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: @dawidchar I was looking at this comment in my current ticket as im in the same area but i'm not sure how this would work because the numbers (1, 2) target the translations via the delimator `|` So it needs to be either 1 or 2 and not `starRating` as that number could run to any number. Unless i've read this wrong?

PR #2194, opened 2022-11-08
f-rating@v0.8.0 - Simplify prop.

kevinrodrigues:
### Fixed
- Some comments from version 0.7.0.
- Tests so they work with new changes.
- Simplify prop `ratingDisplayType` so now only accepts 3 options.


packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -69,7 +69,7 @@ export default {
         },
         ratingDisplayType: {
             type: String,
-            default: null,
+            required: true,
```
ashleynolan: If this is required, wouldn't leaving it empty (to display no text) be impossible? Also, the README says that it's not required?

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
```diff
@@ -174,18 +180,19 @@ describe('Rating', () => {
                     });
                 });
 
-                describe('and a rating type is set to `noRating`', () => {
+                describe('and a `reviewCount` is falsey', () => {
```
ashleynolan, line 183: Do you need a test here for if `ratingDisplayType` is null? (depends if it's required or not)

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: That's a good point. Will have a think over this. 👍

packages/components/molecules/f-rating/src/components/Rating.vue
```diff
@@ -126,20 +126,24 @@ export default {
...
         /**
          * Gets the correct rating display format from translations.
          *
-         * @todo - If the component is using `short` as a `starRatingSize` we shouldn't display text
-         * alongside it for now. (TBC with design - ticket in backlog).
+         * 1. If reviewCount is Zero, return "No ratings yet".
+         * 2. If reviewCount is available, return `ratingDisplayType` .
          *
-         * @returns {string}
+         * @returns {String}
```
zlatinivanov-je: Is the case change necessary? Seems to break consistency from above

zlatinivanov-je (commented):
LGTM 
Subject to comments

packages/components/molecules/f-rating/src/components/_tests/Rating.test.js
kevinrodrigues: I think this one is covered [here](https://github.com/justeat/fozzie-components/pull/2194/files#diff-bd17adcec33ca1bc737ba12a318b084ea668c55175957ffab8198ceeba031481R419).

packages/components/molecules/f-rating/src/components/Rating.vue
kevinrodrigues: My bad.

packages/components/molecules/f-rating/src/components/Rating.vue

dawidchar (approved):

ashleynolan (approved):

PR #2202, opened 2022-11-16
f-rating@v0.10.0 - Adds new prop size for component

kevinrodrigues:
### Added

- `xsmall` prop option.
- Visual tests to cover `xsmall` prop size.
- Unit tests to constant file to cover `xsmall` prop size.
- Note about usage for sizes.

## Screenshots

Screenshot 2022-11-16 at 11 19 03
Screenshot 2022-11-16 at 11 19 15
Screenshot 2022-11-16 at 11 18 53

Screenshot 2022-11-16 at 11 21 31


packages/components/molecules/f-rating/src/assets/scss/rating.scss
```diff
@@ -16,16 +17,21 @@
     display: flex;
 }
 
+// There are four sizes available, the default size starts at 20px. See readme prop `starRatingSize` for more info.
 .c-rating-star {
-    width: 12px;
-}
+  width: 20px;
```
juan-amor: I think the default value is inconsistent with the `VALID_STAR_RATING_SIZES` array below. IMHO I think this should change to `small`, as everywhere else in the code uses `small` rather than `default`. I know that this was discussed in the past, but with more sizes seems more relevant to match them to their class name

kevinrodrigues: Yes i was in two minds this morning about it too. Ill add that in as i think it will make things clearer.

kevinrodrigues: My branch messed up so have made a new one with the change here: https://github.com/justeat/fozzie-components/pull/2204

PR #2203, opened 2022-11-16
fozzie@10.9.0 - Fix failing tests

xander-marjoram:
### Fixed
- Failing tests.
- Linting errors.
  - Including renaming keyframe animation names to kebab-case.


xander-marjoram:
As far as I'm aware, the animation names shouldn't be used outside of this package, so renaming them shouldn't be an issue?

jamieomaguire (dismissed):

packages/tools/fozzie/src/test/scss/scss-setup.spec.js
```diff
@@ -11,13 +11,15 @@ const sassTrue = require('sass-true');
 const glob = require('glob');
 
 describe('Sass', () => {
-    // All Scss unit tests should follow this naming convention
-    const testFileGlob = '**/*.spec.scss';
-    const testFilePathGlob = path.resolve(process.cwd(), testFileGlob);
+    it('Find and run tests', () => {
```
jamieomaguire: nice change - not sure why I didn't do this before

jamieomaguire: NIT: can we rename to 'Find and run SCSS unit tests'?

kevinrodrigues: If you change the main describe to scss then you won't need to add it here again.

kevinrodrigues (dismissed):

packages/tools/fozzie/src/test/scss/scss-setup.spec.js
jamieomaguire: So `describe('SCSS')` then `it('Find and run unit tests')`? Sounds good to me if so!

kevinrodrigues: Yeah, so it will read:

`SCSS >  should find and run unit tests`

kevinrodrigues (approved):

jamieomaguire (approved):

PR #2206, opened 2022-11-17
#trivial - fix up build

kevinrodrigues:
### Added
- Revert CI change which caused tests not to run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -50,6 +50,7 @@ jobs:
       run: yarn --prefer-offline
 
   danger:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -50,7 +50,6 @@ jobs:
       run: yarn --prefer-offline
 
   danger:
-    if: ${{ github.ref != 'refs/heads/master' }}
```
siggerzz: We need to keep this :)

siggerzz (changes requested):
Minor comment!

siggerzz (approved):

PR #2213, opened 2022-11-22
f-content-cards@v10.2.0 - Updating tests and adding analytics tracking

matthewhardern:
Adding in Analytics events to detect card view times and card click events as per requirements, also adding in additional unit tests

### DIFF
--- a/packages/components/organisms/f-content-cards/CHANGELOG.md
+++ b/packages/components/organisms/f-content-cards/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 v10.2.0
 ------------------------------
-*November 22, 2022*
+*December 01, 2022*
 
 ### Added
 - Unit tests for extra states
--- a/packages/components/organisms/f-content-cards/CHANGELOG.md
+++ b/packages/components/organisms/f-content-cards/CHANGELOG.md
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v10.2.0
+------------------------------
+*November 22, 2022*
+
+### Added
+- Unit tests for extra states
+- Analytic tracking events
+
+
 v10.1.1
 ------------------------------
 *November 21, 2022*
--- a/packages/components/organisms/f-content-cards/src/components/ContentCards.js
+++ b/packages/components/organisms/f-content-cards/src/components/ContentCards.js
@@ -41,6 +41,20 @@ export default {
         cardObserver: {}
     }),
 
+    watch: {
+        errors (errors) {
+            if (errors.filter(e => e === STATE_ERROR).length === this.adapters.length) {
+                this.state = STATE_ERROR;
+            }
+            if (errors.filter(e => e === STATE_NO_CARDS).length === this.adapters.length) {
+                this.state = STATE_NO_CARDS;
+            }
+            if (errors.length === this.adapters.length) {
+                this.state = STATE_ERROR;
+            }
+        }
+    },
+
     computed: {
         xOffsets () {
             return uniq(this.$children.map(({ $el }) => $el.offsetLeft)).sort((a, b) => a - b);
@@ -149,6 +163,7 @@ export default {
 
     methods: {
         runAdapterCallbacks (successCallback, errorCallback) {
+            this.state = STATE_LOADING;
             this.adapters.forEach(adapter => {
                 adapter.initialise(this.filters, cards => {
                     if (cards === undefined) {
@@ -217,6 +232,7 @@ export default {
             );
         }, (source, error) => {
             if (error.type === 'NoCards') {
+                this.errors.push(STATE_NO_CARDS);
                 this.$emit(HAS_LOADED, { adapter: source, cards: [] });
                 this.$log.info(
                     `Content Cards Adapters (${source}) - No content cards. Key: (${this.loggingKey})`,
@@ -229,6 +245,7 @@ export default {
                     }
                 );
             } else {
+                this.errors.push(STATE_ERROR);
                 this.$emit(ON_ERROR, { adapter: source });
                 this.$log.info(
                     `Content Cards Adapters (${source}) - Error receiving content cards. Key: (${this.loggingKey})`,
--- a/packages/components/organisms/f-content-cards/src/components/ContentCards.js
+++ b/packages/components/organisms/f-content-cards/src/components/ContentCards.js
@@ -1,4 +1,6 @@
 import { globalisationServices } from '@justeat/f-services';
+import differenceInMilliseconds from 'date-fns/differenceInMilliseconds';
+import { uniq } from 'lodash';
 import tenantConfigs from '../tenants';
 import {
     HAS_LOADED,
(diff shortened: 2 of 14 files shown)


minrwhite (dismissed):

packages/components/organisms/f-content-cards/src/components/_tests/ContentCards.test.js
```diff
@@ -151,41 +154,293 @@ describe('ContentCards', () => {
             expect(cards.length).toEqual(0);
         });
 
+        it('should have 0 cards in the cards array', () => {
```
xander-marjoram: This looks like a duplicate of the above test?

packages/components/organisms/f-content-cards/src/components/_tests/ContentCards.test.js
```diff
@@ -151,41 +154,293 @@ describe('ContentCards', () => {
...
+            expect(slot).toBeDefined();
+        });
 
+        it('should return these cards in the scoped slot variable', () => {
+            // Act
+            const cards = instance.findAll('[data-test-id="__MOCK_TYPE__"]');
+            const cards2 = instance.findAll('[data-test-id="__MOCK_TYPE_2__"]');
 
+            // Assert
+            expect(cards.length + cards2.length).toEqual(2);
```
xander-marjoram: Might it be better to have two expectations?
```suggestion
            expect(cards).toHaveLength(1);
            expect(cards2).toHaveLength(1);
```

packages/components/organisms/f-content-cards/src/components/_tests/ContentCards.test.js
matthewhardern: yep good catch will remove

packages/components/organisms/f-content-cards/src/components/_tests/ContentCards.test.js
```diff
@@ -151,41 +154,293 @@ describe('ContentCards', () => {
...
+            expect(slot).toBeDefined();
+        });
+
+        it('should return these cards in the scoped slot variable', () => {
+            // Act
+            const cards = instance.findAll('[data-test-id="__MOCK_TYPE__"]');
+            const cards2 = instance.findAll('[data-test-id="__MOCK_TYPE_2__"]');
+
+            // Assert
+            expect(cards.length + cards2.length).toEqual(1);
```
xander-marjoram: Same here, it's not obvious which one should have zero cards

packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -35,10 +37,33 @@ export default {
...
         cards: [],
-        errors: []
+        errors: [],
+        cardObserver: {}
     }),
 
+    watch: {
+        errors (errors) {
+            if (errors.filter(e => e === STATE_ERROR).length === this.adapters.length) {
+                this.state = STATE_ERROR;
```
xander-marjoram, line 46: So it looks like these states are only set when all adapters are erroring or returning no cards, is that right?

packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -35,10 +37,33 @@ export default {
...
+            }
+            if (errors.length === this.adapters.length) {
+                this.state = STATE_ERROR;
+            }
+        }
+    },
+
     computed: {
+        xOffsets () {
+            return uniq(this.$children.map(({ $el }) => $el.offsetLeft)).sort((a, b) => a - b);
```
xander-marjoram: A few things here:
1. Could you use a `Set` to make the array unique which would remove the need for the import?
2. Incidentally, I can't see lodash in the package.json, how are we importing it?
3. I think `.sort((a, b) => a - b)` behaves the same as just `.sort()`.

packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -69,19 +94,56 @@ export default {
...
+                        ...(entry.isIntersecting ? { intersectionTime: new Date() } : { intersectionTime: null })
+                    }
+                };
+
+                if (entry.isIntersecting) {
+                    this.handleCardView({
+                        ...card,
+                        metadata: {
+                            coordinates: {
+                                x: this.xOffsets.findIndex(offset => offset === $el.offsetLeft) + 1,
```
xander-marjoram, line 131: This returns the index of the coordinate + 1, rather than the actual coordinate, right?

packages/components/organisms/f-content-cards/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v10.2.0
+------------------------------
+*November 22, 2022*
```
xander-marjoram: Nit 📆

packages/components/organisms/f-content-cards/src/components/ContentCards.js
matthewhardern: yes that is correct as we would still want it to display cards if one of the adapters has succeeded

packages/components/organisms/f-content-cards/src/components/ContentCards.js
matthewhardern: yes they wanted index + 1

packages/components/organisms/f-content-cards/src/components/ContentCards.js
matthewhardern: 1) Awesome idea, will make the change, 2) agreed, 3) good point

packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -69,19 +94,56 @@ export default {
         const locale = globalisationServices.getLocale(tenantConfigs, this.locale, this.$i18n);
         const localeConfig = tenantConfigs[locale];
 
+        const options = {
+            root: null,
+            rootMargin: '0px',
+            threshold: 0.7
+        };
+
+        const observerCallback = entries => {
```
ashleynolan, line 113: Would be good to have a few more comments about what's happening in this method, as there's a lot going on in this `provide()` now.

At a glance, it looks like it's reporting when a card gets intersected with onscreen and provides data on that, but would be good to give a bit of a non-code overview somewhere.

matthewhardern: your glance is correct, I will add a comment above to indicate what it's doing from a non code perspective

matthewhardern: I have added an extended comment

minrwhite (dismissed):

packages/components/organisms/f-content-cards/src/components/ContentCards.js
```diff
@@ -69,19 +93,67 @@ export default {
...
+            entries.forEach(entry => {
+                const { card } = entry.target.__vue__.$options.propsData;
+                const { $el } = entry.target.__vue__;
+
+                this.cardObserver = {
+                    ...this.cardObserver,
+                    [card.id]: {
+                        isIntersecting: entry.isIntersecting,
+                        ...(entry.isIntersecting ? { intersectionTime: new Date() } : { intersectionTime: null })
```
xander-marjoram: A little simpler
```suggestion
                        intersectionTime: entry.isIntersecting ? new Date() : null
```

packages/components/organisms/f-content-cards/src/components/ContentCards.js
xander-marjoram: Ah right, I finally understood this, so this turns pixel offsets into grid coordinates, i.e., 1st column, 2nd row

xander-marjoram (approved):

ashleynolan (approved):

PR #2226, opened 2022-12-06
f-checkout@4.9.1 - Fix postalCode type

kathywklee:
v4.9.1
*December 6, 2022*

### Fixed
- Incorrect `PostalCode` type when restoring address from cookies

### DIFF
--- a/packages/components/pages/f-checkout/src/services/addressService/index.js
+++ b/packages/components/pages/f-checkout/src/services/addressService/index.js
@@ -209,7 +209,7 @@ function getAddressFromCookie (tenant, cookies, mapped = true) {
             Line1: `${houseNumber} ${street}`,
             Line2: unitNumber,
             Line3: sublocality,
-            PostalCode: postalCode,
+            PostalCode: postalCode.toString(),
             administrativeArea: state,
             City: city,
             Field1: latitude,
--- a/packages/components/pages/f-checkout/src/services/tests/addressService.test.js
+++ b/packages/components/pages/f-checkout/src/services/tests/addressService.test.js
@@ -347,14 +347,14 @@ describe('addressService', () => {
         describe('when the address does exist in cookies', () => {
             let get;
             let cookies;
-            const houseNumber = '35';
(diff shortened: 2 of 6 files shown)


packages/components/pages/f-checkout/src/services/tests/addressService.test.js
```diff
@@ -381,7 +381,7 @@ describe('addressService', () => {
                     line2: unitNumber,
                     locality: city,
                     administrativeArea: state,
-                    postcode: postalCode
+                    postcode: postalCode.toString()
```
kevinrodrigues, line 384: I would have thought this would be a string already?

packages/components/pages/f-checkout/src/services/addressService/index.js
```diff
@@ -209,7 +209,7 @@ function getAddressFromCookie (tenant, cookies, mapped = true) {
             Line1: `${houseNumber} ${street}`,
             Line2: unitNumber,
             Line3: sublocality,
-            PostalCode: postalCode,
+            PostalCode: postalCode.toString(),
```
kevinrodrigues, line 212: Same as my other comment.

packages/components/pages/f-checkout/src/services/tests/addressService.test.js
kathywklee: No, the cookie parser will convert the cookie value to number type if it is number-like.

kevinrodrigues (approved):

packages/components/pages/f-checkout/src/services/tests/addressService.test.js
kevinrodrigues: I do think that's a little odd. Maybe we should look at this at the parser level as it would probably be the same issue in other areas too.

kathywklee: I read the documentation of the package, unfortunately there is no such a setting.

PR #2230, opened 2022-12-12
f-rating@0.14.0

kevinrodrigues:
### Added
- Props:
  - `shouldAlignRatingLeft` to allow positioning of accompanying text and SVG icons.
  - `ratingFontSize` to allow large and default font sizes to be used.
- Visual tests to cover above additions.

### Removed
- Visual tests that weren't useful.

### Changed
- Updated to the new `pie-icons-vue` beta release.

## Screenshots:
Screenshot 2022-12-12 at 10 15 46

Screenshot 2022-12-12 at 10 15 54


packages/components/molecules/f-rating/README.md
```diff
@@ -69,14 +69,16 @@ There may be props that allow you to customise its functionality.
...
+| `reviewCount`       | `Number`  |    No    |   null    | Sets the total number of reviews to display. Example: `View 499 reviews` where `499` is the total.                                                                                                                                                                                                           | -
+| `isUserRating`      | `Boolean` |    No    |   false   | Sets the rating as `(You)` to indicate the rating viewed was left by the user.                                                                                                                                                                                                                               | -
+| `shouldAlignRatingLeft`      | `Boolean` |    No    |   false   | When `true` is passed in, the rating display text will be set to the left hand side of the SVG stars.                                                                                                                                                                                                          | -
```
ashleynolan: I'd be inclined to call this `shouldAlignRatingTextLeft` just to make it really explicit that it's the text part of the rating being aligned

ashleynolan (commented):
Looks good – one super small comment, but otherwise LGTM

packages/components/molecules/f-rating/README.md
kevinrodrigues: Fixed 👍

packages/components/molecules/f-rating/src/tests/constants.test.js
```diff
@@ -26,4 +27,17 @@ describe('`constants`', () => {
             expect(result).toBe(true);
         });
     });
+
+    describe('`VALID_STAR_FONT_SIZES`', () => {
```
juan-amor, line 28: What's the purpose or value of this test? Isn't this testing the definition of the constant itself?

kevinrodrigues: Could remove but usually I prefer to make sure the constant file is protected from inadvertent changes, this test will make sure the constant file contains the correct  expected values at all times.

juan-amor: ok, tbh I've never tested a constant like this - if you were testing a validator then it would make sense to test the validator returns true if it matches the constant and false otherwise, but in this case, you could set in the test like the following and it would pass too:

```
const values = ['default', 'large', 'foo', 'bar'];
```

No strong feelings though, so will approve :D

juan-amor (dismissed):

packages/components/molecules/f-rating/src/tests/constants.test.js
kevinrodrigues: No that raises a good point. I'll take a look at this tomorrow. 👍

kevinrodrigues: @juan-amor Fixed now.

juan-amor (approved):

PR #2255, opened 2023-03-22
fozzie-components@7.49.0 - Update fozzie and components to use latest pie-design-tokens for new focus styles

xander-marjoram:
To-do:

- [ ] ~Try changing silent extender to include `&:focus` and `&:focus-visible`~
- [x] Make `%u-elementFocus--boxShadow` the default because the default style uses box-shadow
- [x] Package.json and changelog updates for changed components
- [x] Test in other major browsers

## UI Review Checks

## Browsers Tested

- [x] Chrome (latest)
- [ ] Edge
- [x] Safari
- [x] Firefox


packages/components/atoms/f-form-field/src/assets/scss/common.scss
```diff
@@ -16,7 +16,7 @@ $form-input-borderColour                : f.$color-border-default;
 $form-input-borderRadius                : f.$radius-rounded-c;
 $form-input-borderColour--invalid       : f.$color-support-error;
 
-$form-input-focus--boxShadow            : 0 0 0 2px f.$color-focus;
+$form-input-focus--boxShadow            : 0 0 0 2px f.$color-focus-inner, 0 0 0 4px f.$color-focus-outer;
```
xander-marjoram, line 19: Using `%u-elementFocus` here would mess up the positioning of the checkboxes, etc.

.yarnrc
```diff
@@ -1,5 +0,0 @@
-# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
-# yarn lockfile v1
-
-
-yarn-path ".yarn/releases/yarn-1.19.1.js"
```
ashleynolan, line 5: Isn't this used to make sure everyone is running the same version of `yarn`? I know we recommend using volta now, but if there's no harm in having this too as the default way yarn handles version locking, I'd probably have both settings (just set to the same version).

packages/components/atoms/f-button/src/components/Button.vue
```diff
@@ -299,8 +298,13 @@ $btn-icon-sizeXSmall-buttonSize        : 32px;
     outline: 0;
 
     // Show focus styles on keyboard focus.
+    &:focus,
     &:focus-visible {
-        box-shadow: 0 0 0 2px $btn-default-outline-color;
+        @extend %u-elementFocus;
+
+        &, &:after {
```
ashleynolan, line 305: I think you've got this as a TODO by the looks of it, but would be good to add a silent extend for this as looks like it's used in quite a few places.

.yarnrc
xander-marjoram: With Volta we're pinning to yarn 1.22.19 so I don't think this file would be used anyway

packages/components/atoms/f-button/src/components/Button.vue
xander-marjoram: Yep, I can add that in the next PR?

xander-marjoram: Ideally I'd have it as some sort of mixin where you can specify what you want the radius to be (similar to what we do with spacing or font size), what do you reckon?

.yarnrc
ashleynolan: Yeh, but you can pin the yarn version to `1.22.19` by downloading the v1.22.19 release file and including that instead. Then yarn-path should also point to that file (as a backup for anyone now using Volta)

v1.22.19 file [is available here](https://github.com/yarnpkg/yarn/releases/download/v1.22.19/yarn-1.22.19.js)

packages/components/atoms/f-button/src/components/Button.vue
ashleynolan: Yeh, I think that sounds good 👍

packages/tools/fozzie/src/scss/objects/_buttons.scss
```diff
@@ -108,9 +107,6 @@
         color: $btn-default-color;
         text-decoration: none;
 
-        &:focus {
```
ashleynolan, line 111: Should this be replaced by a new focus style for fozzie buttons? (these are likely used in things like Payments/OrderWeb). Not critical, but may need something added for these in a subsequent PR.

ashleynolan (commented):
Couple of small comments – looks good though

ashleynolan (approved):

siggerzz (approved):

packages/tools/fozzie/src/scss/objects/_buttons.scss
xander-marjoram: We'll check this post-merge, and patch if it does break

PR #2257, opened 2023-03-22
fozzie-components@v7.48.1 - Deploy fix storybook

siggerzz:
Fixed issues where storybook-static wasn't being cached.

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -130,11 +130,13 @@
     },
     "storybook:build":  {
       "cache": true,
-      "dependsOn": []
+      "dependsOn": [],
+      "outputs": ["storybook-static"]
     },
     "storybook:build-changed":  {
       "cache": true,
-      "dependsOn": []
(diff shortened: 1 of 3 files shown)


turbo.json
```diff
@@ -130,11 +130,13 @@
     },
     "storybook:build":  {
       "cache": true,
-      "dependsOn": []
+      "dependsOn": [],
+      "outputs": ["storybook-static"]
```
jamieomaguire, line 134: is there any way to catch issues like this before merging, in the future?

siggerzz: PR Preview deployments to S3 😉

kevinrodrigues (approved):

turbo.json
jamieomaguire: that'll do it!

jamieomaguire (approved):

PR #2263, opened 2023-03-27
f-skeleton-loader@v3.3.0 - Add a 5 star toggle to Ratings Skeleton

dawidchar:
### Added
 - An optional `options` prop
 - Added an optional option `options.use5Stars` to show 5 stars instead of 6 in ratings skeleton.

image
image


packages/components/molecules/f-skeleton-loader/src/components/skeletons/Rating.vue
```diff
@@ -1,13 +1,22 @@
...
         :class="$style['c-rating']">
-        ★★★★★★
+        {{ stars }}
     </div>
 </template>
 
 <script>
 export default {
-    name: 'SkeletonRating'
+    name: 'SkeletonRating',
```
jamieomaguire, line 10: if the default going forward is to go with 5 stars, then I think this logic should be flipped to opt into using 6 stars and fallback to 5

jamieomaguire: then perhaps also bump this to v4

dawidchar: It might still be a bit of time before this becomes the default across all tenants, so for now I think it's still worth having 5 stars as opt in 😄 But happy to create a v4 if you think it's valuable.

sonnyprince (approved):

Charlotte-91 (approved):

PR #2267, opened 2023-04-06
f-mega-modal@7.4.0 - Add back button and rtl config.

kevinrodrigues:
### Added
- Prop `closeButtonStyle` allows either a chevron icon `chevron-left-icon` or close icon `close-small-icon` to be used.
- `isModeRightToLeft` prop to allow a chevron icon to be used in a right to left tenant.

### Screenshots:

Screenshot 2023-04-06 at 15 37 49

### RTL:
Screenshot 2023-04-06 at 15 38 01

### DIFF
--- a/packages/components/molecules/f-mega-modal/CHANGELOG.md
+++ b/packages/components/molecules/f-mega-modal/CHANGELOG.md
@@ -4,12 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-v7.4.0
+v8.0.0 ** Breaking Change **
 ------------------------------
-*April 06, 2023*
+*April 12, 2023*
+
+### Changed
+- Prop `hasCloseButton` to `closeButtonStyle` allow either a chevron icon or close icon to be used.
+  - `closeButtonStyle` options: `none`, `close-small-icon`, `chevron-left-icon`.
 
 ### Added
-- `hasBackButton` prop to allow a chevron icon to be used instead of a close icon.
 - `isModeRightToLeft` prop to allow a chevron icon to be used in a right to left tenant.
 
 
--- a/packages/components/molecules/f-mega-modal/CHANGELOG.md
+++ b/packages/components/molecules/f-mega-modal/CHANGELOG.md
@@ -10,7 +10,7 @@ v7.4.0
 
 ### Added
 - `hasBackButton` prop to allow a chevron icon to be used instead of a close icon.
-- `hasModeRightToLeft` prop to allow a chevron icon to be used in a right to left tenant.
+- `isModeRightToLeft` prop to allow a chevron icon to be used in a right to left tenant.
 
 
 v7.3.0
--- a/packages/components/molecules/f-mega-modal/CHANGELOG.md
+++ b/packages/components/molecules/f-mega-modal/CHANGELOG.md
@@ -10,7 +10,7 @@ v8.0.0 ** Breaking Change **
 
 ### Changed
 - Prop `hasCloseButton` to `closeButtonStyle` allow either a chevron icon or close icon to be used.
-  - `closeButtonStyle` options: `none`, `close-small-icon`, `chevron-left-icon`.
+  - `closeButtonStyle` options: `''`, `close-small-icon`, `chevron-left-icon`.
 
 ### Added
 - `isModeRightToLeft` prop to allow a chevron icon to be used in a right to left tenant.
--- a/packages/components/molecules/f-mega-modal/CHANGELOG.md
+++ b/packages/components/molecules/f-mega-modal/CHANGELOG.md
@@ -9,7 +9,7 @@ v8.0.0 ** Breaking Change **
 *April 12, 2023*
 
 ### Changed
-- Prop `hasCloseButton` to `closeButtonStyle` allow either a chevron icon or close icon to be used.
+- Prop `hasCloseButton` to `closeButtonStyle` allows either a chevron icon or close icon to be used or the choice not to display anything.
   - `closeButtonStyle` options: `''`, `close-small-icon`, `chevron-left-icon`.
 
 ### Added
--- a/packages/components/molecules/f-mega-modal/CHANGELOG.md
+++ b/packages/components/molecules/f-mega-modal/CHANGELOG.md
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v7.4.0
+------------------------------
+*April 06, 2023*
+
+### Added
+- `hasBackButton` prop to allow a chevron icon to be used instead of a close icon.
+- `hasModeRightToLeft` prop to allow a chevron icon to be used in a right to left tenant.
+
+
 v7.3.0
 ------------------------------
 *March 27, 2023*
--- a/packages/components/molecules/f-mega-modal/CHANGELOG.md
+++ b/packages/components/molecules/f-mega-modal/CHANGELOG.md
@@ -4,15 +4,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
-v8.0.0 ** Breaking Change **
+v7.4.0
 ------------------------------
 *April 12, 2023*
 
-### Changed
-- Prop `hasCloseButton` to `closeButtonStyle` allows either a chevron icon or close icon to be used or the choice not to display anything.
-  - `closeButtonStyle` options: `''`, `close-small-icon`, `chevron-left-icon`.
-
 ### Added
+- Prop `closeButtonStyle` allows either a chevron icon `chevron-left-icon` or close icon `close-small-icon` to be used.
 - `isModeRightToLeft` prop to allow a chevron icon to be used in a right to left tenant.
 
 
--- a/packages/components/molecules/f-mega-modal/README.md
+++ b/packages/components/molecules/f-mega-modal/README.md
@@ -119,7 +119,7 @@ The props that can be defined are as follows:
 | `is-close-fixed`       | `Boolean` | `false`         | Sets the modal close button position to `fixed`.                                                                                                                  |
 | `is-positioned-bottom` | `Boolean` | `false`         | Sets the modal position to the bottom of the viewport for all screen widths.                                                                                      |
 | `has-overlay`          | `Boolean` | `true`          | Controls whether or not to display an overlay behind the modal.                                                                                                   |
-| `close-button-style`   | `String`  | `close-small-icon`          | Controls the style of the button. Accepts `close-small-icon` & `chevron-left-icon`. If you pass an empty string '', no styles for the close button will be shown. |
+| `close-button-style`   | `String`  | `close-small-icon`          | Controls the style of the button. Accepts `close-small-icon` & `chevron-left-icon`. |
 | `close-on-blur`        | `Boolean` | `true`          | Controls whether or not to close the modal when the user clicks outside of the modal.                                                                             |
 | `close-button-copy`    | `String`  | `"Close modal"` | Sets the hidden text value for the close button which is used by screen-readers.                                                                                  |
 | `title`                | `String`  | `''`            | When set, will add a title to the top of the component. If you need to define a custom heading, ignore this prop.                                                 |
--- a/packages/components/molecules/f-mega-modal/README.md
+++ b/packages/components/molecules/f-mega-modal/README.md
@@ -119,7 +119,7 @@ The props that can be defined are as follows:
 | `is-close-fixed`       | `Boolean` | `false`         | Sets the modal close button position to `fixed`.                                                                                       |
 | `is-positioned-bottom` | `Boolean` | `false`         | Sets the modal position to the bottom of the viewport for all screen widths.                                                           |
 | `has-overlay`          | `Boolean` | `true`          | Controls whether or not to display an overlay behind the modal.                                                                        |
-| `close-button-style`   | `String`  | `close-small-icon`          | Controls the style of the button. Accepts `cross` & `chevron`.                                                               |
+| `close-button-style`   | `String`  | `cross`          | Controls the style of the button. Accepts `cross` & `chevron`.                                                               |
 | `close-on-blur`        | `Boolean` | `true`          | Controls whether or not to close the modal when the user clicks outside of the modal.                                                  |
 | `close-button-copy`    | `String`  | `"Close modal"` | Sets the hidden text value for the close button which is used by screen-readers.                                                       |
 | `title`                | `String`  | `''`            | When set, will add a title to the top of the component. If you need to define a custom heading, ignore this prop.                      |
--- a/packages/components/molecules/f-mega-modal/README.md
+++ b/packages/components/molecules/f-mega-modal/README.md
@@ -108,22 +108,24 @@ The following rudimentary example can be used as a guide for implementing this c
 
 The props that can be defined are as follows:
 
-| Prop  | Type  | Default | Description |
-| ----- | ----- | ------- | ----------- |
-| `is-open` | `Boolean` | `false` | Sets the modal to open or closed state. |
-| `is-narrow` | `Boolean` | `false` | Use the narrow visual style. |
-| `is-wide` | `Boolean` | `false` | Use the wide visual style. |
-| `is-flush` | `Boolean` | `false` | Removes passing around the modal content. |
-| `is-full-height` | `Boolean` | `false` | Sets the modal content to full height of the screen.<br/><br/><blockquote>Note this only applies to small screen devices.</blockquote> |
-| `is-scrollable` | `Boolean` | `false` | Makes the modal content scrollable. |
-| `is-close-fixed` | `Boolean` | `false` | Sets the modal close button position to `fixed`. |
-| `is-positioned-bottom` | `Boolean` | `false` | Sets the modal position to the bottom of the viewport for all screen widths. |
-| `has-overlay` | `Boolean` | `true` | Controls whether or not to display an overlay behind the modal. |
-| `has-close-button` | `Boolean` | `true` | Controls whether or not to display the modal close button. |
-| `close-on-blur` | `Boolean` | `true` | Controls whether or not to close the modal when the user clicks outside of the modal. |
-| `close-button-copy` | `String` | `"Close modal"` | Sets the hidden text value for the close button which is used by screen-readers. |
-| `title` | `String` | `''` | When set, will add a title to the top of the component. If you need to define a custom heading, ignore this prop.  |
-| `titleHtmlTag` | `String` | `h3` | Sets the tag for the component's title.<br><br>Allowed values are `h1`, `h2`, `h3`, and `h4` |
+| Prop                   | Type  | Default         | Description                                                                                                                            |
+|------------------------| ----- |-----------------|----------------------------------------------------------------------------------------------------------------------------------------|
+| `is-open`              | `Boolean` | `false`         | Sets the modal to open or closed state.                                                                                                |
+| `is-narrow`            | `Boolean` | `false`         | Use the narrow visual style.                                                                                                           |
+| `is-wide`              | `Boolean` | `false`         | Use the wide visual style.                                                                                                             |
+| `is-flush`             | `Boolean` | `false`         | Removes passing around the modal content.                                                                                              |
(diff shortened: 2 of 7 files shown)


packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -15,6 +15,7 @@
                 [$style['c-megaModal-content--narrow']]: isNarrow,
                 [$style['c-megaModal-content--wide']]: isWide,
                 [$style['c-megaModal-content--flush']]: isFlush,
+                [$style['c-megaModal-content--right-to-left']]: hasModeRightToLeft,
```
ashleynolan:
```suggestion
                [$style['c-megaModal-content--rightToLeft']]: hasModeRightToLeft,
```

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -342,6 +365,10 @@ export default {
     transform: translate(50%, -50%);
     width: 95%;
 
+    &.c-megaModal-content--right-to-left {
```
ashleynolan: Once you've moved the class onto the base element, this could just become:

```scss
.c-megaModal--modeRTL & {
  …
}
```

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -33,7 +34,10 @@
                         :is="titleHtmlTag"
                         v-if="title"
                         :id="uid"
-                        :class="['c-megaModal-title', $style['c-megaModal-title']]"
+                        :class="['c-megaModal-title', $style['c-megaModal-title'], {
+                            [$style['c-megaModal-hasBackBtn']]: hasBackButton,
```
ashleynolan: I would put these modifiers up onto the `c-megaModal` `div` element. The modifier class should then be `c-megaModal--hasBackBtn` too.

Just means if any further styling modifications are needed, you can style it so that

```scss
.c-megaModal--hasBackBtn {
  & .c-megaModal-title { … }
  … // any other stuff here too
}

File: packages/components/molecules/f-mega-modal/src/components/MegaModal.vue [hunk 4]
```diff
@@ -33,7 +34,10 @@
...
+                            [$style['c-megaModal-modeRTL']]: hasModeRightToLeft
```
ashleynolan: Same for this one – I'd put this onto the `c-megaModal` element.

Also, `isModeRightToLeft` feels more grammatically correct I think for the prop name.

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -43,13 +47,19 @@
                         <f-button
                             is-icon
                             :class="[$style['c-megaModal-closeBtn'], {
-                                [$style['c-megaModal-closeBtn--fixed']]: isCloseFixed || isFullHeight
+                                [$style['c-megaModal-closeBtn--fixed']]: isCloseFixed || isFullHeight,
+                                [$style['c-megaModal-backBtn']]: hasBackButton,
+                                [$style['c-megaModal-modeRTL']]: hasModeRightToLeft
```
ashleynolan: You should be able to remove these two lines once you move the classes onto the `c-megaModal` element as above.

Can just target:

```scss
.c-megaModal-modeRTL {
  & .c-megaModal-closeBtn { … }
}
```

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -462,6 +499,14 @@ export default {
...
     margin: 0 f.spacing(f) 0 0;
 
+    &.c-megaModal-hasBackBtn {
+        margin: 0 0 0 f.spacing(f);
+    }
+
+    &.c-megaModal-modeRTL {
+        margin: 0 f.spacing(f) 0 0;
+    }
+
```
ashleynolan: Same here – will become:

```suggestion
    .c-megaModal--hasBackBtn & {
        margin: 0 0 0 f.spacing(f);
    }

    .c-megaModal--modeRTL & {
        margin: 0 f.spacing(f) 0 0;
    }

```

ashleynolan: Just a note – you'll likely need to remove the outer `.c-megaModal` in the sass here (it shouldn't be needed anyway, it's just making the styles more specific for no reason by the looks of it)

packages/components/molecules/f-mega-modal/stories/MegaModal.stories.js
```diff
@@ -23,6 +23,8 @@ export const MegaModalComponent = (args, { argTypes }) => ({
             :is-text-aligned-center="isTextAlignedCenter"
             :has-overlay="hasOverlay"
             :has-close-button="hasCloseButton"
+            :has-back-button="hasBackButton"
+            :has-mode-right-to-left="hasModeRightToLeft"
```
ashleynolan:
```suggestion
            :is-mode-right-to-left="isModeRightToLeft"
```

packages/components/molecules/f-mega-modal/stories/MegaModal.stories.js
```diff
@@ -49,6 +51,8 @@ MegaModalComponent.args = {
     isTextAlignedCenter: false,
     hasOverlay: true,
     hasCloseButton: true,
+    hasBackButton: false,
+    hasModeRightToLeft: false,
```
ashleynolan:
```suggestion
    isModeRightToLeft: false,
```

packages/components/molecules/f-mega-modal/stories/MegaModal.stories.js
```diff
@@ -89,6 +93,12 @@ MegaModalComponent.argTypes = {
     hasCloseButton: {
         control: { type: 'boolean' }
     },
+    hasBackButton: {
+        control: { type: 'boolean' }
+    },
+    hasModeRightToLeft: {
```
ashleynolan:
```suggestion
    isModeRightToLeft: {
```

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -436,6 +463,16 @@ export default {
         top: 22px;
         z-index: f.zIndex(high);
 
+        &.c-megaModal-backBtn {
```
ashleynolan: Same as below

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -147,14 +163,21 @@ export default {
...
             default: ''
         },
+
         titleHtmlTag: {
             type: String,
             default: 'h3',
             validator: value => ['h1', 'h2', 'h3', 'h4'].includes(value)
+        },
+
+        hasModeRightToLeft: {
```
ashleynolan:
```suggestion
        isModeRightToLeft: {
```

ashleynolan (commented):
Few comments on the styles/classnames – otherwise LGTM

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
kevinrodrigues: Did try this and it seems to break some of the colour styles.

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -147,14 +161,21 @@ export default {
...
             default: ''
         },
+
         titleHtmlTag: {
             type: String,
             default: 'h3',
             validator: value => ['h1', 'h2', 'h3', 'h4'].includes(value)
+        },
+
+        isModeRightToLeft: {
```
jamieomaguire, line 173: Should this be prop-driven or just dependent on the writing-mode/directionality of the consuming application?

In other words, would we expect a consumer that supports multiple languages to pass in a prop that changes depending on the currently selected language?

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -454,17 +479,45 @@ export default {
...
+            left: auto;
+            transform: scaleX(-1);
+        }
+    }
+
+    &.c-megaModal--hasBackBtn {
+        .c-megaModal-closeBtn {
+            left: f.spacing(d);
+        }
+    }
```
jamieomaguire: I think we can do all of this using logical properties rather than using a class to toggle it. For example, instead of `right` and `left` we could use [inset-inline](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline). This could reduce the amount of CSS we write, as long as the browser support matches our [support matrix](https://www.pie.design/engineers/guidelines/browser-support/).

When an RTL language is detected, I believe `inset-inline` will translate to right/left appropriately.

jamieomaguire: [This](https://rtlstyling.com/posts/rtl-styling#css-logical-properties) is also a good reference

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
kevinrodrigues: Yes so the consumer would be responsible for passing in a prop / boolean based on the tenant. Not sure if that answers the question?

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
kevinrodrigues: Just had a read. Isn't this relying heavily on the consuming application state? If so I think it is probably cleaner to allow the component to cater for different directions rather than it catering for a state which is assumed, if that state changed then the modal may break too. If the direction is controlled within the component and the consuming application can switch it on via a prop it is less prone to break. Let me know if i've misunderstood the main question though?

jamieomaguire: Yeah fully appreciate that. In my opinion, I think we should be relying on the consuming application but maybe this is a direction we should figure out as a team as it will also inform the WC work. I can set up a chat with the team, that might be easier than a PR thread :)

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -133,6 +142,11 @@ export default {
             default: true
         },
 
+        hasBackButton: {
```
xander-marjoram: IMO this has the potential to be a bit confusing. If someone passes in `hasCloseButton` and `hasBackButton`, the expected end result isn't obvious (without looking into the code); two buttons, or one?

I would prefer keeping the existing `hasCloseButton` but renaming this prop to something like `closeButtonStyle`, which controls whether or not the icon is rendered as a cross or a chevron.

kevinrodrigues: It will only display one or the other. If it has a back button the close button won't be displayed. I do see what you mean though, will take a look.

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -454,17 +479,50 @@ export default {
...
+    .c-megaModal--hasChevronIcon & {
+        margin: 0 0 0 f.spacing(f);
+    }
+
+    .c-megaModal-modeRTL & {
+        margin: 0 f.spacing(f) 0 0;
+    }
+
+    // Allow the title to sit flush when there is no close button in RTL mode.
+    .c-megaModal-modeRTL &:only-child {
```
kevinrodrigues: Could add a hook to the component if people prefer this. This only covers RTL as the title still has the margin if the consumer ops for no button and RTL which might be edge case but happy to add a prop or class somewhere to style this instead.

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
kevinrodrigues: Fixed.

packages/components/molecules/f-mega-modal/README.md
```diff
@@ -108,22 +108,23 @@ The following rudimentary example can be used as a guide for implementing this c
...
+| `is-scrollable`        | `Boolean` | `false`         | Makes the modal content scrollable.                                                                                                                               |
+| `is-close-fixed`       | `Boolean` | `false`         | Sets the modal close button position to `fixed`.                                                                                                                  |
+| `is-positioned-bottom` | `Boolean` | `false`         | Sets the modal position to the bottom of the viewport for all screen widths.                                                                                      |
+| `has-overlay`          | `Boolean` | `true`          | Controls whether or not to display an overlay behind the modal.                                                                                                   |
+| `close-button-style`   | `String`  | `close-small-icon`          | Controls the style of the button. Accepts `close-small-icon` & `chevron-left-icon`. If you pass an empty string '', no styles for the close button will be shown. |
```
xander-marjoram:
```suggestion
| `close-button-style`   | `String`  | `close-small-icon`          | Controls the style of the button. Accepts `close-small-icon` & `chevron-left-icon`. |
```

xander-marjoram: Sorry just another nitpick from me, I think if we're going to have the prop called `closeButtonStyle`, we would want to accept clearer, simpler strings like "cross" or "chevron", and then handle the icons within.

Alternatively, we could change the name of this prop to `closeButtonIcon` which indicates that we're passing in the icon name. This is definitely the simpler option to implement.

kevinrodrigues: Happy with either to be honest, i think with the first option it's doable but means we would have to modify the name a bit once passed in so it can be provided to the component.

ashleynolan: I'd probably go with the first option too if possible, just as it relies less on the user needing to know the icon naming.

kevinrodrigues: Agreed. Fixed now hopefully.

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -173,6 +191,14 @@ export default {
...
         showAriaLabel () {
             return this.ariaLabel === '' ? this.uid : this.ariaLabel;
+        },
+        hasChevronIcon () {
+            return this.closeButtonStyle === 'chevron';
+        },
+        setCloseButtonIconStyle () {
+            return this.closeButtonStyle.includes('cross')
+                ? 'close-small-icon'
+                : 'chevron-left-icon';
```
xander-marjoram, line 201: What do you think of this? Might make it easier to add other alternative icons in future?
```suggestion
            const iconMap = {
                chevron: 'chevron-left-icon',
                cross: 'close-small-icon'
            };
            return iconMap[this.closeButtonStyle];
```

kevinrodrigues: I did actually think about a similar approach (using switch tho) but left it as i think if we need more in future we could modify it then? I don't really seeing more than two icons ever being used in this situation to be honest.

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -454,17 +484,50 @@ export default {
...
         margin-left: f.spacing(e);
         margin-right: f.spacing(e);
     }
+
+    .c-megaModal--hasChevronIcon & {
+        margin: 0 0 0 f.spacing(f);
+    }
+
+    .c-megaModal-modeRTL & {
+        margin: 0 f.spacing(f) 0 0;
```
xander-marjoram, line 519: Is this effectively the same rule as L508?

kevinrodrigues: Ah yes. Good spot. 🦅 👁️

kevinrodrigues: Fixed.

packages/components/molecules/f-mega-modal/CHANGELOG.md
```diff
@@ -4,6 +4,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v7.4.0
+------------------------------
+*April 12, 2023*
+
+### Added
+- Prop `closeButtonStyle` allows either a chevron icon `chevron-left-icon` or close icon `close-small-icon` to be used.
+- `isModeRightToLeft` prop to allow a chevron icon to be used in a right to left tenant.
```
ashleynolan, line 13: I agree with @jamieomaguire  that it'd be better to look at doing this with logical values rather than positional values eventually, rather than setting RTL as a mode.

It's outside the scope of this ticket to do that though – could we create an EE task that could look at converting the modal over to RTL using logical values potentially? (as I think it'll need a bit of a revamp in the styling to do this)

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -2,8 +2,11 @@
     <div
         ref="megaModal"
         :class="['c-megaModal',
-                 $style['c-megaModal'],
-                 { 'u-overlay': showOverlay }]"
+                 $style['c-megaModal'], {
+                     'u-overlay': showOverlay,
+                     [$style['c-megaModal--hasChevronIcon']]: hasChevronIcon && hasCloseButton,
```
ashleynolan:
```suggestion
                     [$style['c-megaModal--hasBackButton']]: hasChevronIcon && hasCloseButton,
```

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -173,6 +191,14 @@ export default {
         },
         showAriaLabel () {
             return this.ariaLabel === '' ? this.uid : this.ariaLabel;
+        },
+        hasChevronIcon () {
+            return this.closeButtonStyle === 'chevron';
+        },
```
ashleynolan: I would move the logic out of the view for this property and make it:

```suggestion
        hasBackButton () {
            return (this.closeButtonStyle === 'chevron' && this.hasCloseButton);
        },
```

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
ashleynolan: Can remove the `&&` as well if you make the other suggested change below

packages/components/molecules/f-mega-modal/README.md
```diff
@@ -108,22 +108,23 @@ The following rudimentary example can be used as a guide for implementing this c
...
+| `is-full-height`       | `Boolean` | `false`         | Sets the modal content to full height of the screen.<br/><br/><blockquote>Note this only applies to small screen devices.</blockquote> |
+| `is-scrollable`        | `Boolean` | `false`         | Makes the modal content scrollable.                                                                                                    |
+| `is-close-fixed`       | `Boolean` | `false`         | Sets the modal close button position to `fixed`.                                                                                       |
+| `is-positioned-bottom` | `Boolean` | `false`         | Sets the modal position to the bottom of the viewport for all screen widths.                                                           |
+| `has-overlay`          | `Boolean` | `true`          | Controls whether or not to display an overlay behind the modal.                                                                        |
+| `close-button-style`   | `String`  | `close-small-icon`          | Controls the style of the button. Accepts `cross` & `chevron`.                                                               |
```
ashleynolan:
```suggestion
| `close-button-style`   | `String`  | `cross`          | Controls the style of the button. Accepts `cross` & `chevron`.                                                               |
```

ashleynolan (commented):
Couple of small things, but otherwise looks good

packages/components/molecules/f-mega-modal/CHANGELOG.md
kevinrodrigues: Yup i think it's a good idea. I was thinking maybe for the WC version if we're planning on porting this over there?

xander-marjoram (dismissed):

packages/components/molecules/f-mega-modal/src/components/MegaModal.vue
```diff
@@ -2,8 +2,11 @@
     <div
         ref="megaModal"
         :class="['c-megaModal',
-                 $style['c-megaModal'],
-                 { 'u-overlay': showOverlay }]"
+                 $style['c-megaModal'], {
+                     'u-overlay': showOverlay,
+                     [$style['c-megaModal--hasBackButton']]: hasBackButton,
+                     [$style['c-megaModal-modeRTL']]: isModeRightToLeft
```
ashleynolan:
```suggestion
                     [$style['c-megaModal--modeRTL']]: isModeRightToLeft
```

ashleynolan (commented):
One tiny naming thing – then LGTM

xander-marjoram (approved):

ashleynolan (approved):

PR #2268, opened 2023-04-12
f-segmented-control@v0.1.0 - initial component

jamieomaguire:
A Segmented Control component is a group of toggle buttons that behave like a single choice element. It is useful for displaying a set of options to the user when only one option can be selected.

## Props

| Prop | Type | Required | Default | Validator | Description |
| --- | --- | --- | --- | --- | --- |
| `screenreaderLabel` | `String` | `true` | N/A | N/A | The label that is read by screen readers to describe the segmented control |
| `options` | `Array` | `true` | N/A | Array must have 2-4 elements, and each element must have a `label` property of type `string`. The `iconName`, `disabled`, and `selected` properties are optional. `iconName` corresponds with the name of a PIE Icon found [here](https://github.com/justeattakeaway/pie/tree/main/packages/tools/pie-icons-vue). `selected` will set the default selected option. You can only set this to true for one option. | An array of objects representing each button in the segmented control |
| `size` | `String` | `false` | `'small'` | Must be either `'small'` or `'large'`. | The size of the segmented control |

## Events

The events that can be subscribed to are as follows:

| Event | Description |
| ----- | ----------- |
| input | Emitted when the user selects an option. The value emitted is the label of the selected option. |

## Accessibility
The Segmented Control component follows accessibility best practices and includes support for keyboard navigation. Users can use the arrow keys to navigate between options and the space or enter key to select an option. The component also includes ARIA attributes for screen readers.

## Usage
```
<template>
    <SegmentedControl
        :screenreaderLabel="screenreaderLabel"
        :options="options"
        :size="size"
    />
</template>

<script>
import SegmentedControl from '@justeat/f-segmented-control';

export default {
    components: { SegmentedControl },
    data() {
        return {
            screenreaderLabel: 'Select an option',
            options: [
                { label: 'Option 1', iconName: 'SomeIconName' },
                { label: 'Option 2' },
                { label: 'Option 3', selected: true },
                { label: 'Option 4', disabled: true }
            ],
            size: 'large', // Can be 'small' or 'large', defaults to 'small'
        };
    }
};
</script>

```

## Visual examples
Screenshot 2023-04-14 at 15 37 20
Screenshot 2023-04-14 at 15 37 13
Screenshot 2023-04-14 at 15 37 07

### LTR (Default)
Screenshot 2023-04-14 at 15 36 33
Screenshot 2023-04-14 at 15 36 21

### RTL
Screenshot 2023-04-14 at 15 36 41
Screenshot 2023-04-14 at 15 36 54


packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,254 @@
...
+    props: {
+        screenreaderLabel: {
+            type: String,
+            required: true
+        },
+        options: {
+            type: Array,
+            required: true,
+            validator: value => {
+                if (value.length < 2 || value.length > 4) {
```
kevinrodrigues: Could we move this logic to an external file and import it?

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,254 @@
...
+    outline: none; // Replace outline with box-shadow on focus due to iOS Safari not respecting border-radius
+
+    // Selected state
+    &--selected {
+        font-weight: f.$font-weight-bold;
+        background-color: f.$color-container-default;
+        box-shadow: f.$elevation-02;
+
+        // Hover state
+        &:hover:not(:disabled) {
```
kevinrodrigues: Does this need a focus state too?

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,254 @@
...
+    font-weight: f.$font-weight-regular;
+    background: none;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    gap: f.spacing(b);
+    outline: none; // Replace outline with box-shadow on focus due to iOS Safari not respecting border-radius
+
+    // Selected state
+    &--selected {
```
kevinrodrigues: Could you use the full selector here please as it makes it easier to find and change selectors in future.

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,254 @@
...
+    justify-content: space-between;
+    background-color: $sc-background-clr;
+    border-radius: $sc-border-radius;
+    gap: 2.5px; // Ensure outlines and box-shadows don't overlap
+    padding: $sc-padding-xs-modifier;
+    min-block-size: $sc-sm-height; // Small size by default
+    user-select: none;
+
+    // Large size modifier
+    &--large {
```
kevinrodrigues: Similar to below.

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,254 @@
+<template>
+    <div
+        :class="{
+            [$style['c-segmented-control']]: true,
+            [$style['c-segmented-control--large']]: size === 'large'
```
ashleynolan:
```suggestion
            [$style['c-segmentedControl']]: true,
            [$style['c-segmentedControl--large']]: size === 'large'
```

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,254 @@
...
+        @keydown="onKeyDown"
+    >
+        <button
+            v-for="(option, index) in options"
+            :key="option.label"
+            :data-test-id="`segmented-control-option-${index}`"
+            type="button"
+            :class="{
+                [$style['c-segmented-control__option']]: true,
+                [$style['c-segmented-control__option--selected']]: option.label === selectedLabel
```
ashleynolan: Small naming thing (use camelCase for naming and dashes for children)

```suggestion
                [$style['c-segmentedControl-option']]: true,
                [$style['c-segmentedControl-option--selected']]: option.label === selectedLabel
```

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
ashleynolan: I would split these validations up into functions and then you could call them (and test them) individually. i.e.

```js
validator: value => {
   isValid = isValidNumberOfOptions(value) && areOptionTypesValid(value);
   return isValid;
}

File: packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue [hunk 7]
```diff
@@ -0,0 +1,254 @@
...
+            tabIndex: selectedIndex > -1 ? selectedIndex : 0
+        };
+    },
+    computed: {
+        // dynamically set the tabindex attribute on each button, giving focus only to the currently selected button
+        focusedTabIndex () {
+            return this.options.map((_, index) => (index === this.tabIndex ? 0 : -1));
+        }
+    },
+    methods: {
```
ashleynolan, line 75: Can you add JSDoc comments to these methods?

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,254 @@
...
+    }
+};
+</script>
+
+<style lang="scss" module>
+@use '@justeat/fozzie/src/scss/fozzie' as f;
+
+// Variables
+$sc-sm-height: f.spacing(f);
+$sc-l-height: calc(f.spacing(b) * 6);
```
ashleynolan: Super readability NIT:

```suggestion
$sc-height-small: f.spacing(f);
$sc-height-large: calc(f.spacing(b) * 6);
```

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,254 @@
...
+$sc-border-radius: 9999px;
+$sc-padding-xs-modifier: calc(f.spacing(a) / 2);
+$sc-background-clr: f.$color-container-strong;
+$sc-background-clr-selected: f.$color-container-default;
```
ashleynolan: I'd go for slightly more verbose to aid with readability here:

```suggestion
$sc-background-color: f.$color-container-strong;
$sc-background-color-selected: f.$color-container-default;
```

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,254 @@
...
+        color: f.$color-content-disabled;
+        background-color: f.$color-disabled-01;
+        cursor: not-allowed;
+        pointer-events: none;
+    }
+
+    // Hover state
+    &:hover:not(:disabled) {
+        // Prevent hover effects from being applied on touch devices
+        @media(hover: hover) and (pointer: fine) {
```
ashleynolan: More a Q – would the media query on its own not be good enough to catch these devices here?

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,254 @@
...
+            }
+        }
+
+        // Active state
+        &:active:not(:disabled) {
+            background-color: darken($sc-background-clr-selected, f.$color-active-01);
+        }
+    }
+
+    // Not disabled state
```
ashleynolan, line 193: I would put all these descendent styles above the `--selected {` styling, to group all the base component state styling together

ashleynolan (commented):
Looks good 👍  Couple of small comments

ashleynolan:
Not sure if this request could be too much work potentially (if it is, then don't worry about it). Any chance you could PR in the initial generated component files (so just from running the yeoman generator) so that it's clear what was added/changed on top of those generated files. Just makes the config files really easy to review (and means less bloat in this PR then)

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
jamieomaguire: there is one, currently on line 250. There's a screenshot included in the PR description :)

jamieomaguire:
> Not sure if this request could be too much work potentially (if it is, then don't worry about it). Any chance you could PR in the initial generated component files (so just from running the yeoman generator) so that it's clear what was added/changed on top of those generated files. Just makes the config files really easy to review (and means less bloat in this PR then)

Ah sorry but I think that would be far too much work at this point, I take your point though. Mind if I leave it as is but promise to do that next time? 😂

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
jamieomaguire: I'm happy to do this but I personally disagree with the value it adds 😂

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
jamieomaguire: yes you're spot on! updating

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
kevinrodrigues: I do prefer the way you did it originally but it adds ambiguity to the code base imo. Otherwise i have no issues with it, i think this came up a few times over the years so we've always stuck to the full name for consistency.

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,240 @@
+<template>
+    <div
+        :class="{
+            [$style['c-segmentedControl']]: true,
```
kevinrodrigues: You can just use the style class without the boolean if it is always true? i.e move to outside of the object area inline.

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,240 @@
...
+        tabindex="-1"
+        @keydown="onKeyDown"
+    >
+        <button
+            v-for="(option, index) in options"
+            :key="option.label"
+            :data-test-id="`segmented-control-option-${index}`"
+            type="button"
+            :class="{
+                [$style['c-segmentedControl-option']]: true,
```
kevinrodrigues: Similar to above.

packages/components/atoms/f-segmented-control/src/components/helpers/_tests/validators.test.js
```diff
@@ -0,0 +1,77 @@
+import { validateOptions } from '../validators';
```
kevinrodrigues, line 1: Sorry i should have picked up on this before but would you mind adding in the AAA comments to each test spec please. `Arrange, Act & Assert etc`

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
ashleynolan: I like the syntax too – but it does make searchability harder when looking up class names in the codebase, which is why we don't use them

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
jamieomaguire: ah good spot! updated :)

packages/components/atoms/f-segmented-control/src/components/helpers/_tests/validators.test.js
jamieomaguire: oops just seen this. I can update shortly 👍

kevinrodrigues (dismissed):

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,244 @@
...
+    </div>
+</template>
+
+<script>
+import { validateOptions } from './helpers/validators';
+
+export default {
+    name: 'SegmentedControl',
+    props: {
+        screenreaderLabel: {
```
FayeCarter, line 44: Do we need this as a separate prop? Can the user can just apply `aria-label`?

jamieomaguire: I can see arguments for and against it. Using a prop means we can _require_ the consumer to provide a label which I quite like the idea of :) I can chat with Ash too when he's in!

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
```diff
@@ -0,0 +1,244 @@
...
+        }
+    }
+};
+</script>
+
+<style lang="scss" module>
+@use '@justeat/fozzie/src/scss/fozzie' as f;
+
+// Variables
+$sc-height-small: f.spacing(f);
```
FayeCarter, line 149: Should this be `$sc-height--small`?
I honestly don't know 😂

jamieomaguire: Not to my knowledge. I don't see much reason to use BEM style modifiers on SCSS variable names personally :)

ashleynolan: We've never defined this as a hard rule or not (we probably should either use it or not).

I think for fozzie it's fine doing either for now - we should probably decide when we start doing this on web components so it's more consistent

packages/components/atoms/f-segmented-control/src/components/SegmentedControl.vue
ashleynolan: Just to cover what Jamie and myself just chatted about in person – if it's expected for users to add an aria label or description (so the component won't work accessibly without it), we should add props that are named like the above so that we can make them required or validate them.

@FayeCarter If that's the case for the toggle component (where we expect the aria labels/descriptions to be included), then might be worth adding those in as required props where they're named `screenreaderLabel` or something similar.

ashleynolan (dismissed):

FayeCarter (dismissed):

ashleynolan (dismissed):

FayeCarter (dismissed):

kevinrodrigues (dismissed):

ashleynolan (dismissed):

FayeCarter (dismissed):

ashleynolan (approved):

PR #2269, opened 2023-04-13
f-toggle@0.1.0 - Create component

FayeCarter:
### Added
- Component structure and basic configuration (created using `generator-component`).
- Component markup and core functionality.
- Storybook config.

---

### States

|  | Checked | Unchecked |
| --- | --- | --- | 
| **default** | image | image |
| **disabled** | image | image |
| **rtl** | image | image |

---

### Transition

https://user-images.githubusercontent.com/24740015/231807819-c875ae4d-0549-499e-8e2b-701c8c048ad9.mov

---
### a11y

https://user-images.githubusercontent.com/24740015/231817946-da756aff-e677-4c68-ad21-2239f2fc19a7.mov


packages/components/atoms/f-toggle/src/components/Toggle.vue
```diff
@@ -0,0 +1,223 @@
...
+    name: 'FToggle',
+
+    components: {
+        CheckIcon
+    },
+
+    inheritAttrs: false,
+
+    props: {
+        isModeRightToLeft: {
```
jamieomaguire: have we agreed on this approach as a team? I personally think we shouldn't be passing an RTL prop to every component a RTL consumer might be using

packages/components/atoms/f-toggle/src/components/Toggle.vue
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+.c-toggle--modeRTL.c-toggle--checked {
+    .c-toggle-control {
+        @include toggle-transition(transform);
+
+        transform: translateX(0);
+    }
+}
```
jamieomaguire: this is what I'm hoping we can avoid if we don't use a boolean for RTL

packages/components/atoms/f-toggle/src/components/Toggle.vue
```diff
@@ -0,0 +1,223 @@
...
+.c-toggle-input {
+    appearance: none;
+    margin: 0;
+}
+
+.c-toggle-control {
+    @include toggle-transition(transform);
+
+    display: flex;
+
```
jamieomaguire: I don't think these empty lines are helping with readability, I'd probably opt to remove them personally

packages/components/atoms/f-toggle/src/components/Toggle.vue
```diff
@@ -0,0 +1,223 @@
...
+.c-toggle--modeRTL.c-toggle--checked {
+    .c-toggle-control {
+        @include toggle-transition(transform);
+
+        transform: translateX(0);
+    }
+}
+
+.c-toggle--disabled {
```
jamieomaguire: I don't think you need a class here. 

Just target `.c-toggle:disabled`

packages/components/atoms/f-toggle/src/components/Toggle.vue
```diff
@@ -0,0 +1,223 @@
...
+    height: $toggle-icon-size;
+
+    path {
+        @include toggle-transition(fill);
+        fill: $toggle-colour--checked;
+    }
+}
+
+
+.c-toggle--checked {
```
jamieomaguire: Do we need a class here: Can we use :checked?

packages/components/atoms/f-toggle/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+There may be props that allow you to customise its functionality.
+
+The props that can be defined are as follows (if any):
+
+| Prop  | Type  | Default | Description |
+| ----- | ----- | ------- | ----------- |
+| isModeRightToLeft | Boolean | false | sets reading direction to 'Right to Left' |
+| ariaLabelledBy | String | '' | The id for an element that will 'label' the toggle |
+| ariaDescribedBy | String | '' | The id for an element describing what the toggle will control |
```
jamieomaguire: 🔥

packages/components/atoms/f-toggle/src/components/Toggle.vue
```diff
@@ -0,0 +1,223 @@
...
+    transition-property: $property;
+    transition-duration: 0.15s;
+    transition-timing-function: cubic-bezier(0.4, 0, 0.9, 1);
+    transition-delay: 0s;
+}
+
+.c-toggle {
+    display: flex;
+
+    @include toggle-transition(background-color);
```
jamieomaguire, line 91: I'd wrap all animation in a media query to check the user doesn't have reduced motion enabled. This way people who do not like animations or suffer issues from them can opt-out.

A basic example can be seen [here](https://web.dev/prefers-reduced-motion/#working-with-the-media-query-from-css-and-javascript-contexts) :)

packages/components/atoms/f-toggle/README.md
```diff
@@ -0,0 +1,131 @@
...
+<img width="125" alt="Fozzie Bear" src="../../../../bear.png" />
+
+A component to toggle a single setting on/off.
+
+</div>
+
+---
+
+[![npm version](https://badge.fury.io/js/%40justeat%2Ff-toggle.svg)](https://badge.fury.io/js/%40justeat%2Ff-toggle)
+[![CircleCI](https://circleci.com/gh/justeat/fozzie-components.svg?style=svg)](https://circleci.com/gh/justeat/workflows/fozzie-components)
```
ashleynolan: Can remove the CircleCI line here

packages/components/atoms/f-toggle/src/components/Toggle.vue
```diff
@@ -0,0 +1,223 @@
...
+            if (this.$attrs.name) idArray.push(this.$attrs.name);
+            if (this.$attrs.id) idArray.push(this.$attrs.id);
+
+            idArray.push(this.uniqueIdentifierId);
+            return idArray.join('-');
+        },
+
+        isChecked () {
+            return this.$attrs.checked;
+        }
```
jamieomaguire, line 50: I could be mistaken, does this component currently allow the consumer to start with a checked component? I can't see how it uses data-binding as there's no checked prop

jamieomaguire:
Looks really good @FayeCarter, just a few comments/questions. I'm sure you already have but would be worth consulting the [W3 a11y guides on toggles](https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/) if you haven't already just to check you've got everything it mentions :)

packages/components/atoms/f-toggle/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:justeat/fozzie-components.git"
+  },
+  "bugs": {
+    "url": "https://github.com/justeat/fozzie-components/issues"
+  },
+  "license": "Apache-2.0",
+  "engines": {
+    "node": "^12 || ^14 || ^16"
```
ashleynolan: I'd probably just include 14 and 16 here now (and potentially include 18 as well)

```suggestion
    "node": "^14 || ^16 || ^18"
```

packages/components/atoms/f-toggle/src/components/Toggle.vue
```diff
@@ -0,0 +1,223 @@
...
+                 }]">
+        <input
+            :id="uniqueId"
+            role="switch"
+            type="checkbox"
+            v-bind="$attrs"
+            :class="$style['c-toggle-input']"
+            :aria-checked="isChecked"
+            :aria-labelledby="ariaLabelledBy"
+            :aria-describedby="ariaDescribedBy"
```
ashleynolan: More of a question rather than a request to change it – as you're binding `$attrs` to the input, if you set an `aria-labelledby` and  `aria-describedby` on the component, wouldn't those automatically get passed down and applied without using the prop?

I do like the way the props encourage the use of aria attributes, but just wondering if we're adding props that aren't needed

packages/components/atoms/f-toggle/src/components/Toggle.vue
ashleynolan: Not to pre-empt @FayeCarter 's response, but I'm guessing you need the class to style the parent label, as there's no way to target the parent of `:checked`?

packages/components/atoms/f-toggle/src/components/Toggle.vue
ashleynolan:
```suggestion
.c-toggle-control {
    .c-toggle--modeRTL & {
        transform: $toggle-translation;
    }
    .c-toggle--modeRTL.c-toggle--checked & {
       @include toggle-transition(transform);

        transform: translateX(0);
     }
}
```

packages/components/atoms/f-toggle/src/components/Toggle.vue
ashleynolan: Think this is the same as `--checked` potentially, where it applies styling to the label

packages/components/atoms/f-toggle/src/components/Toggle.vue
ashleynolan: Agree with this – would we be able to leverage the writing mode and use logical properties for this over a classname/prop?

(not sure how best to test this in storybook though – @jamieomaguire any ideas)

ashleynolan (commented):
Few comments from me

packages/components/atoms/f-toggle/test/visual/f-toggle.visual.spec.js
```diff
@@ -0,0 +1,25 @@
...
+            }
+        });
+
+        it('should display the f-toggle component', async () => {
+            // Act
+            await Toggle.load();
+
+            // Assert
+            await browser.percyScreenshot('f-toggle - Visual Test', device);
+        });
```
LTurns, line 23: should we have a visual or component test of the toggle in action? Just to make sure the right option is selected when checked

packages/components/atoms/f-toggle/src/components/Toggle.vue
jamieomaguire: you could quite trivially wrap the storybook template in a div with `dir="rtl` - I'm doing this on my PR soon :)

packages/components/atoms/f-toggle/src/components/Toggle.vue
jamieomaguire: that's a very fair point!

packages/components/atoms/f-toggle/src/components/Toggle.vue
FayeCarter: Yep if they use `f-toggle checked="true" />` the same way that they would be able to disable it.
It's the same thing we use form form filed checkbox

packages/components/atoms/f-toggle/src/components/Toggle.vue
FayeCarter: I think I've found a way. Not sure if it makes the CSS more complicated though

packages/components/atoms/f-toggle/src/components/Toggle.vue
FayeCarter: I based it off of what Kevin was doing in modal. I've changed it now to use `dir`

packages/components/atoms/f-toggle/src/components/Toggle.vue
FayeCarter: 👍

ashleynolan (dismissed):

packages/components/atoms/f-toggle/stories/Toggle.stories.js
```diff
@@ -0,0 +1,86 @@
...
+            </fields>
+        </div>`
+});
+
+ToggleComponent.storyName = 'f-toggle';
+
+ToggleComponent.args = {
+    checked: false,
+    disabled: false,
+    direction: 'ltr'
```
LTurns, line 68: super tiny but guessing you don't need this `direction` arg anymore if we're not passing the prop

LTurns (approved):

ashleynolan (approved):

packages/components/atoms/f-toggle/src/components/Toggle.vue
```diff
@@ -0,0 +1,184 @@
...
+
+    .c-toggle-input:disabled ~ & {
+        path {
+            @include toggle-transition(fill);
+            fill: $toggle-colour--disabled;
+        }
+    }
+}
+
+[dir="rtl"] {
```
jamieomaguire, line 168: would it be easier to group all the selectors using this transform mixin together?

I.e.
```
selector-1,
selector-2,
selector 3 {
  @include toggle-transition(transform);
}
```

jamieomaguire: I appreciate one uses `fill`

jamieomaguire (approved):

kevinrodrigues (approved):

packages/components/atoms/f-toggle/src/components/Toggle.vue
FayeCarter: I think it's best at this point to leave all the styles for each class together? I may be very wrong 😃

PR #2273, opened 2023-04-20
f-toggle-switch@0.1.0 - Rename f-toggle to f-toggle-switch

FayeCarter:
Rename `f-toggle` to `f-toggle-switch` - No new code added


packages/components/atoms/f-switch/package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "@justeat/f-toggle",
-  "description": "Fozzie Toggle - A component to toggle a single setting on/off",
+  "name": "@justeat/f-switch",
+  "description": "Fozzie Switch - A component to switch a single setting on/off",
```
jamieomaguire: I'm so sorry to say this but I think we should call it a toggle-switch based on the UX research from [Neilson Norman Group](https://www.nngroup.com/articles/toggle-switch-guidelines/).

What do you think?

FayeCarter: I feel sad, but it is done

jamieomaguire (approved):

ashleynolan (approved):

PR #2312, opened 2023-08-02
f-checkout@v4.11.0 - remove fozzie components as externals in Webpack config

zerosalinas:
v4.11.0
*August 2, 2023*

### Changed
- Remove fozzie components as externals in Webpack config.


ashleynolan:
Quick question – what's the reason for wanting to remove the `externals` config?

The reason this is implemented is so that files are cached across pages (rather than bundled multiple times), so would be good to know what issue you're getting from that being setup 🙂

zerosalinas:
> Quick question – what's the reason for wanting to remove the `externals` config?
> 
> The reason this is implemented is so that files are cached across pages (rather than bundled multiple times), so would be good to know what issue you're getting from that being setup 🙂

@ashleynolan, I've created a draft pull request to troubleshoot a problem I'm encountering with the pipelines. I've kept it as a draft because I don't plan to move forward with that. I plan on closing it.

PR #2329, opened 2023-09-08
f-toggle-switch@v0.2.0 - Fix ios safari bug.

kevinrodrigues:
Fixes safari IOS visual issue.

### Chrome

Screenshot 2023-09-08 at 13 09 02

### Iphone

Screenshot 2023-09-08 at 13 01 41


packages/components/atoms/f-toggle-switch/src/components/ToggleSwitch.vue
```diff
@@ -143,6 +144,8 @@ $toggle-switch-translation: calc($toggle-switch-width - $toggle-switch-control-s
 }
 
 .c-toggle-switch-control {
+    position: absolute;
+    left: 2px;
```
xander-marjoram, line 148: Can we use `inset-inline-start` to avoid needing left/right logic?

xander-marjoram (approved):

raoufswe (approved):

PR #2365, opened 2023-11-14
Ensure GitHub Actions chrome version is used when running tests #trivial

siggerzz:
- This PR updates to a new version of WebDriverIO that handles the automatic downloading of Chromedriver.
- I also had to add a new 'resolution for 'typesafe-i18n' to prevent f-loyalty's build from failing. - This was caused by an incompatible sub-dependency version. This new resolution forces the working version to be used.

### DIFF
--- a/test/configuration/local.config.js
+++ b/test/configuration/local.config.js
@@ -1,3 +1,4 @@
+const path = require('path');
 const { testType } = require('./shared.config');
 
 const configuration = {
@@ -20,7 +21,6 @@ const configuration = {
         },
         chromedriver: {
             headless: false,
-            path: '/',
             args: []
         }
     },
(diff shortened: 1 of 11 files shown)


JoshuaNg2332 (dismissed):

wdio-shared.conf.js
```diff
@@ -42,7 +41,7 @@ exports.config = {
     connectionRetryCount: configuration.connectionRetryTimeout,
     // Framework you want to run your specs with.
     // The following are supported: Mocha, Jasmine, and Cucumber
-    // see also: https://webdriver.io/docs/frameworks.html
+    // see also: https://webdriver.io/docs/frameworks tml
```
xander-marjoram, line 44:
```suggestion
    // see also: https://webdriver.io/docs/frameworks.html
```

test/configuration/local.config.js
```diff
@@ -1,3 +1,4 @@
+const path = require('path');
```
xander-marjoram, line 1:
```suggestion
```

PR #2370, opened 2023-11-21
fozzie-components@7.55.0 Update Supported Node engines with v20 and yarn v3.5.0 #globalconfig

zlatinivanov-je:
Updates supported Node engines 
-----

Change type: `#globalconfig`

**Note:** _With Node 17 and over `cryptoJS` requires the use of `NODE_OPTIONS=--openssl-legacy-provider`_

### Added

- support for Node version 18 and 20 and Yarn v3.5.0
  - Volta pin Node v20.3.1
  - Volta pin Yarn v3.5.0
- `--openssl-legacy-provider` in `package.json` scripts to fix issue with Node 17 and above
- Fix vulnerabilities with Yeoman component generator
- Bump vulnerable packages
  - `jest`, `jest-environment-jsdom`, `babel-jest`, `@vue/babel-preset-app`, `@vue/cli-plugin-babel`, `@vue/cli-plugin-unit-jest`, `stylelint`, `@vue/vue2-jest`
- `testURL` to `testEnvironmentOptions.url` and `vue-jest` to `@vue/vue2-jest` in `jest.config` for all components, tools and services
- Axios v.1.6.0 to `jest.config` `transformIgnorePatterns` (`@justeat/f-checkout`)
- `experimentalCSSCompile` to `jest.config` to resolve node-sass error (`@justeat/f-error-boundary`)

## Removed

- support for deprecated Node version 12

## Changed

- `build.yml` file to reflect setup for Node 20 and yarn 3.5.0

## Browsers Tested

- [x] Chrome (latest)
- [ ] Internet Explorer 11


CHANGELOG.md
```diff
@@ -4,6 +4,18 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 
+v7.55.0
+------------------------------
+*November 21, 2023*
+
+### Changed
+- Removes support for deprecated Node versions (12 and 14).
```
ashleynolan: This update is awesome!

I think we'll likely need to keep `v14` in the engines as I know some legacy Customer platforms use this (ones that are being deprecated by OneWeb and so won't be updating to latest versions of Node, like coreweb for instance)

zlatinivanov-je: And also just realised some legacy projects running on legacy build agents don't have access to Node >=16! Shame
Will revert back v14!

webbertakken:
Blocks https://github.je-labs.com/experimentation-platform/FeatureManagement_Portal/pull/1327#discussion_r558461

.yarnrc.yml
```diff
@@ -0,0 +1,11 @@
+nodeLinker: node-modules
+
+packageExtensions:
+  stylelint-config-recommended-scss@*:
+    peerDependencies:
+      postcss: 8
+  stylelint-config-standard-scss@*:
+    peerDependencies:
+      postcss: 8
```
siggerzz: If this was copied from PIE, I _think_ you can probably remove the stylelint-* stuff. It was a very old bug when the repo was first created, and I don't think it's present now, but the code just hasn't been removed yet 👍🏼

packages/components/atoms/f-form-field/src/components/_tests/FormDropdown.test.js
```diff
@@ -67,7 +67,7 @@ describe('FormDropdown', () => {
         });
 
         describe('value ::', () => {
-            it('should default to first dropdown option when omitted', () => {
+            it.skip('should default to first dropdown option when omitted', () => {
```
siggerzz, line 70: What was the reason for this test being skipped?

packages/components/molecules/f-searchbox/src/components/formElements/tests/FormSearchField.test.js
```diff
@@ -186,7 +188,7 @@ describe('`FormSearchField`', () => {
                     expect(spy).toHaveBeenCalledWith(false);
                 });
 
-                it('should call the `setInputTimeoutValue` action with the `inputTimeoutValue` so we can clear it later', () => {
+                it.skip('should call the `setInputTimeoutValue` action with the `inputTimeoutValue` so we can clear it later', () => {
```
siggerzz, line 191: ditto

.yarnrc.yml
LTurns: oh amazing, thank you! I'll remove and test 👍

packages/components/pages/f-self-exclusion/package.json
```diff
@@ -48,20 +48,20 @@
     "extends @justeat/browserslist-config-fozzie"
   ],
   "dependencies": {
-    "@justeat/f-globalisation": "1.2.0",
-    "@justeat/f-services": "1.x",
     "@justeat/f-alert": "6.x",
     "@justeat/f-button": "5.x",
     "@justeat/f-card": "4.x",
     "@justeat/f-form-field": "6.x",
+    "@justeat/f-globalisation": "1.2.0",
```
siggerzz: Maybe worth modifying this to `1.x` for consistency. It should still resolve the same version, just means the component doesn't need updating for any future minor versions of this package.

packages/components/atoms/f-form-field/src/components/_tests/FormDropdown.test.js
LTurns: ah yes these were the last remaining tests that I'd skipped 😂 error for this one is below - I'm unsure why this is failing since the component does default to the first option, so it shouldn't be blank - but I'll have another look now. If not, might be worth keeping skipped and I can revisit in a quick follow-up PR
Screenshot 2024-03-13 at 11 01 46

packages/components/molecules/f-searchbox/src/components/formElements/tests/FormSearchField.test.js
LTurns: error for this one below - I'll investigate more now but I think this might require a change in a jest function
Screenshot 2024-03-13 at 10 58 03

zlatinivanov-je (commented):
There is a lot to review here, but here is a point to discuss: 

Should we define Node engines a bit more liberally where newer version users are not forced to remain on a set of versions, e.g. current will not work for Node v.21 user. 

```json
  # Something like that ...
  "engines": {
    "node": ">=18"
  },
```

Screenshot 2024-03-13 at 15 27 30

LTurns:
> There is a lot to review here, but here is a point to discuss:
> 
> Should we define Node engines a bit more liberally where newer version users are not forced to remain on a set of versions, e.g. current will not work for Node v.21 user.
> 
> ```json
>   # Something like that ...
>   "engines": {
>     "node": ">=18"
>   },
> ```
> 
> Screenshot 2024-03-13 at 15 27 30

yeah good question, I agree it might be worth being more liberal here 😄 @ashleynolan wondering what you think about this?

ashleynolan:
@zlatinivanov-je @LTurns I'm fine with us doing that – I guess the main issue is if people think it'll automatically work in those versions (when in reality we aren't testing those newer versions yet, so things could break). I'm fine with us trying to add it and seeing what happens though

zlatinivanov-je:
> @zlatinivanov-je @LTurns I'm fine with us doing that – I guess the main issue is if people think it'll automatically work in those versions (when in reality we aren't testing those newer versions yet, so things could break). I'm fine with us trying to add it and seeing what happens though

I am thinking consuming apps would need to do their due diligence to ensure the app they support works as expected, otherwise we will end up blocking them and playing catch up. 

What I would be also curious to know is if we are happy to go with something like `>=18` and force people to update with next published version?

.github/workflows/build.yml
```diff
@@ -7,13 +7,13 @@ on:
...
     types: [assigned, opened, synchronize, reopened]
     paths-ignore:
-      - '.husky/**'
-      - 'stories/**'
-      - '.vscode/**'
-      - 'README.md'
-      - 'CONTRIBUTING.md'
-      - 'CHANGELOG.md'
-      - 'LICENSE'
+      - ".husky/**"
```
zlatinivanov-je, line 10: Maybe good to add rules to vscode settings or similar to avoid editor refactor quotes

package.json
```diff
@@ -138,13 +142,14 @@
     "Github contributors <https://github.com/justeat/fozzie-components/graphs/contributors>"
   ],
   "engines": {
-    "node": "^12 || ^14 || ^16"
+    "node": "^14 || ^16 || ^18 || ^20"
```
zlatinivanov-je: Not sure what is the effect on having node engines version defined here since this is not an installable package. Node 14 and 16 would not work. 

Screenshot 2024-03-13 at 14 53 24

And if required maybe we can remove deprecated versions.

packages/components/atoms/f-button/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "license": "Apache-2.0",
   "engines": {
-    "node": "^12 || ^14 || ^16"
+    "node": "^14 || ^16 || ^18 || ^20"
```
zlatinivanov-je: Should the package version go up here too? 
Will it be published?

packages/components/atoms/f-card/package.json
```diff
@@ -23,7 +23,7 @@
   },
   "license": "Apache-2.0",
   "engines": {
-    "node": "^12 || ^14 || ^16"
+    "node": "^14 || ^16 || ^18 || ^20"
```
zlatinivanov-je, line 26: Same point on package version. And won't be mentioning further down ...

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
```diff
@@ -60,7 +60,7 @@
                     <button-component
                         :class="$style['c-cookieBanner-cta']"
                         data-test-id="accept-all-cookies-button"
-                        button-size="small"
+                        button-size="small-productive"
```
zlatinivanov-je, line 63: Not sure of the reason behind that change. Please comment back, thanks

packages/components/pages/f-checkout/src/components/_tests/Selector.test.js
```diff
@@ -44,10 +44,10 @@ describe('Selector', () => {
                     propsData
                 });
 
-                const selector = wrapper.find('[data-test-id="formfield-order-time-dropdown-select"]');
+                const selector = wrapper.find('[data-test-id="formfield-order-time-label"]');
```
zlatinivanov-je, line 47: Would this be sufficient test coverage? 
I suppose the form field has comprehensive test coverage

packages/services/f-feature-management/package.json
```diff
@@ -34,11 +34,10 @@
     "build": "vite build",
     "lint": "eslint \"!(dist)/**/*.{js,vue}\"",
     "lint:fix": "yarn lint --fix",
-    "test": "jest",
+    "test": "jest --detectOpenHandles --forceExit",
```
zlatinivanov-je, line 37: > `--detectOpenHandles` ...has a significant performance penalty 
https://jestjs.io/docs/cli#--detectopenhandles 

Are we happy to accept that?

packages/tools/f-vue-icons/package.json
```diff
@@ -27,15 +32,16 @@
...
+    "node": "^14 || ^16 || ^18 || ^20"
   },
   "keywords": [
     "fozzie"
   ],
   "scripts": {
     "prepublishOnly": "yarn lint && yarn test && yarn build",
-    "build": "yarn build:icons && yarn build:dist",
-    "build:dist": "bili --format umd,es,cjs,umd-min --js buble --name f-vue-icons",
+    "clean": "run -T rimraf -rf ./icons/ ./dist/ ./esm/",
```
zlatinivanov-je, line 42: As above, imo warrants new package version

packages/tools/f-vue-icons/icons/AlcoholAgeIcon.js
```diff
@@ -1,26 +1,26 @@
-import _mergeJSXProps from "babel-helper-vue-jsx-merge-props";
+import _mergeJSXProps from "@vue/babel-helper-vue-jsx-merge-props";
```
zlatinivanov-je, line 1: Could not find this change reflected in the package dependencies. 
Need to remove `babel-helper-vue-jsx-merge-props` from component `package.json`

packages/tools/f-vue-icons/package.json
```diff
@@ -1,15 +1,20 @@
 {
   "name": "@justeat/f-vue-icons",
   "version": "3.13.2",
-  "main": "dist/f-vue-icons.cjs.js",
+  "main": "dist/index.cjs",
```
zlatinivanov-je, line 4: Changes like this would probably sit better in a new package version.

packages/tools/f-vue-icons/package.json
```diff
@@ -53,23 +59,14 @@
   },
```
zlatinivanov-je: Should the above package dependency be removed as per other component files changes

zlatinivanov-je (commented):
Happy to approve once clarified some small comments made. 
I was able to clean checkout, install dependencies, build and run tests without an issue for Node v.18 and v.20. It is a good step forward. 

_However using Node v.16 and v.21 ran into issues, guessing it will be the case for Node v.14_

packages/components/atoms/f-form-field/src/components/_tests/FormDropdown.test.js
LTurns: fixing forward with this and the below test in a follow up PR 👍

package.json
LTurns: I've defaulted fozzie to run on node 20.3.1 using volta so hopefully we shouldn't need to run these scripts on 14 and 16. However, for them to work, all that is needed is to remove the `NODE_OPTIONS=--openssl-legacy-provider` from the package.json script - this was implemented to fix a bug introduced in node version 17. I think it's worth looking into as a separate PR after this one, but so long as the consuming packages still work in node 14 or higher, I'm happy to merge 😄

packages/components/atoms/f-button/package.json
LTurns: yep we should! I'm going through them all now - will push up all the version bumps v shortly 👍

packages/components/organisms/f-cookie-banner/src/components/CookieBanner.vue
LTurns: This was to fix a visual test - f-cookie-banner was using the wrong button size due to the fact that the naming of the prop had changed. Updating this fixed the test 👍

packages/components/pages/f-checkout/src/components/_tests/Selector.test.js
LTurns: I think it is for now - I'm going to look through the tests I've altered in a separate PR after this in case there's anything else we need to do. But as it's a minor test and the code base is in maintenance mode anyway, happy to merge with this as the current solution 👍

packages/services/f-feature-management/package.json
LTurns: happy to accept but I'll look to remove in a follow up when I review the tests 👍 there hasn't been much of a performance penalty on the speed of tests so happy to merge for now

LTurns (approved):

siggerzz (approved):

PR #2372, opened 2023-11-23
f-self-exclusion f-button version update

m-axii:


jamieomaguire (approved):

packages/components/pages/f-self-exclusion/.eslintignore
```diff
@@ -0,0 +1,2 @@
+node_modules
```
ashleynolan, line 1: Could you rebase this PR against master? It looks like a lot of changes from the previous PR are in this one too (that are already in master)

PR #2408, opened 2023-12-07
f-self-exclusion@1.1.0 - Updated API response structure #trivial

m-axii:
Updated API response structure

### DIFF
--- a/packages/components/pages/f-checkout/jest.config.js
+++ b/packages/components/pages/f-checkout/jest.config.js
@@ -13,7 +13,7 @@ module.exports = {
     },
 
     transformIgnorePatterns: [
-        'node_modules/(?!(lodash-es)/)'
+        'node_modules/(?!(lodash-es|axios)/)'
     ],
 
     moduleNameMapper: {
--- a/packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
+++ b/packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
@@ -264,7 +264,6 @@ export default {
             await this.getExclusions({ api: this.selfExclusionApi, authToken: this.authToken });
             this.$log.info('Self exclusion status fetched successfully');
             this.selectedState = this.alcoholExclusion.state;
-            console.log(`selectedState: ${this.selectedState}`);
         } catch (error) {
             this.$log.error('Error getting self exclusion status', error);
             this.openAlertError();
--- a/packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
+++ b/packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
@@ -189,7 +189,7 @@ export default {
         },
         authToken: {
             type: String,
-            default: null
+            required: true
         },
         smartGatewayBaseUrl: {
             type: String,
--- a/packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
+++ b/packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
@@ -172,13 +172,16 @@ import tenantConfigs from '../tenants';
 
 export default {
     name: 'SelfExclusion',
+
     components: {
         FCard,
         FAlert,
         FButton,
         FFormField
     },
+
     mixins: [VueGlobalisationMixin],
+
     props: {
         locale: {
             type: String,
@@ -186,7 +189,7 @@ export default {
         },
         authToken: {
             type: String,
-            required: true
+            default: null
         },
         smartGatewayBaseUrl: {
             type: String,
@@ -198,9 +201,10 @@ export default {
         },
         privacyPolicyUrl: {
             type: String,
-            required: true
+            default: null
         }
     },
+
     data () {
         return {
             tenantConfigs,
@@ -215,6 +219,7 @@ export default {
             })
         };
     },
+
     computed: {
         ...mapGetters('fSelfExclusionModule', [
             'exclusions'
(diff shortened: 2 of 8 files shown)


packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
```diff
@@ -259,6 +264,7 @@ export default {
             await this.getExclusions({ api: this.selfExclusionApi, authToken: this.authToken });
             this.$log.info('Self exclusion status fetched successfully');
             this.selectedState = this.alcoholExclusion.state;
+            console.log(`selectedState: ${this.selectedState}`);
```
jamieomaguire: I expect this isn't wanted nor required in production. Are you happy to remove?

m-axii: Oh yes, forgotten. Lets remove.

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
```diff
@@ -346,6 +347,13 @@ export default {
 <style lang="scss" module>
 @use "@justeat/fozzie/src/scss/fozzie" as f;
 
+.c-selfExclusion {
+    display: flex;
+    justify-content: center;
+    width: 100%;
```
jamieomaguire, line 353: divs are 100% width by default. Is this really needed?

m-axii: Divs are display: block but aren't 100% of width. This one is changed to display: flex and width 100% is necessary.

jamieomaguire (dismissed):

ashleynolan (dismissed):

jamieomaguire (approved):

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
```diff
@@ -172,21 +172,24 @@ import tenantConfigs from '../tenants';
...
+
     props: {
         locale: {
             type: String,
             default: 'en-AU'
         },
         authToken: {
             type: String,
-            required: true
+            default: null
```
kevinrodrigues: Two question about this.
1.Do you not want to default this to an empty string? 
2. Isn't `authToken` required to make the call? So should required remain in place?

packages/components/pages/f-checkout/jest.config.js
```diff
@@ -13,7 +13,7 @@ module.exports = {
     },
 
     transformIgnorePatterns: [
-        'node_modules/(?!(lodash-es)/)'
+        'node_modules/(?!(lodash-es|axios)/)'
```
kevinrodrigues, line 16: More of a question but why was this change made in f-checkout? Does `f-self-exclusion` relate to checkout?

packages/components/pages/f-self-exclusion/src/store/selfExclusion.module.js
```diff
@@ -10,8 +10,8 @@ export default {
     actions: {
         async getExclusions ({ commit }, { api, authToken }) {
```
kevinrodrigues, line 11: Couldn't see any tests for this component. Worth adding tests (in a different pr is fine by me).

jamieomaguire (dismissed):

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
m-axii: For both, during testing with CoreWeb, the requests (that require the authToken) did not work with `required: true`. We consulted with the CoreWeb team and they advised to add `default: null`.

kevinrodrigues: OK. I think i'd need to look at why to answer properly but the reason i ask is because if you use this component elsewhere and you don't pass in a value for `authToken` (someone might not know this) the call will fail surely? Also you might want to update/fill out the prop table with those details now.
https://github.com/justeattakeaway/fozzie-components/blob/master/packages/components/pages/f-self-exclusion/README.md#props

packages/components/pages/f-checkout/jest.config.js
m-axii: No changes were made from our end to f-checkout. Not sure why this happened.

packages/components/pages/f-self-exclusion/src/store/selfExclusion.module.js
m-axii: We did not consider adding tests to the component. If required we can add in a different PR.

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
m-axii: We tested again and there seems to be no issue now with adding required: true.

packages/components/pages/f-checkout/jest.config.js
jamieomaguire: @siggerzz explained on our internal comms that he did this to fix the failing `f-checkout` tests that this PR uncovered

packages/components/pages/f-self-exclusion/src/store/selfExclusion.module.js
kevinrodrigues: There's some examples you can look at in the other components which you can follow so should help. Give me a shout if you need a hand with these.

kevinrodrigues (dismissed):

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
kevinrodrigues: Sorry I just noticed can you update the readme file too. You can do this in another pr if you need this in today. The prop table needs filling in, there's examples in other components too.

m-axii: Yes if we can update in another PR it would be great, so we can create the PR for CoreWeb.

jamieomaguire (approved):

kevinrodrigues (approved):

xander-marjoram (approved):

PR #2420, opened 2024-01-08
f-self-exclusion@1.1.1 - Remove Card and hide form when loading and when in error state #trivial 

m-axii:
Added f-spinner component to hide form before form is loaded and updated component version to v1.1.1

### DIFF
--- a/packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
+++ b/packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
@@ -166,8 +166,6 @@ import FButton from '@justeat/f-button';
 import '@justeat/f-button/dist/f-button.css';
 import FFormField from '@justeat/f-form-field';
 import '@justeat/f-form-field/dist/f-form-field.css';
-import FSpinner from '@justeat/f-spinner';
-import '@justeat/f-spinner/dist/f-spinner.css';
 import { TENANT_MAP } from '@justeat/f-checkout/src/constants';
 import { mapActions, mapGetters } from 'vuex';
 import SelfExclusionApi from '../services/providers/selfExclusion.api';
--- a/packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
+++ b/packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
@@ -1,47 +1,138 @@
 <template>
     <div
+        class="f-self-exclusion"
         :class="$style['c-selfExclusion']"
         data-test-id="self-exclusion-component">
-        <f-card
-            has-inner-spacing-large
-            card-size-custom="large"
-            has-outline
-            :card-heading="$t('heading')"
-            :class="$style['c-selfExclusion']">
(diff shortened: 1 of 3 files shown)


packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
```diff
@@ -1,47 +1,138 @@
 <template>
     <div
+        class="f-self-exclusion"
         :class="$style['c-selfExclusion']"
```
ashleynolan, line 4: Is this intended? There should already be a class of `c-selfExclusion` on this element that can be used for styling?

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
```diff
@@ -1,47 +1,138 @@
 <template>
     <div
+        class="f-self-exclusion"
```
ashleynolan: Is this intended? There should already be a class of c-selfExclusion on this element that can be used for styling?

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
```diff
@@ -348,17 +345,12 @@ export default {
...
+    position: relative;
 }
 
-.c-selfExclusion > div {
-    position: relative;
-    display: flex;
-    flex-direction: column;
-    align-items: stretch;
+.c-selfExclusion h1 {
+    margin-bottom: 16px;
```
ashleynolan: You can use `f.spacing(d)` here instead of `16px` so it uses the tokens

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
pttolev: Yes, it is, we need it for external use. In the component, we approach to use `$style` class.

ashleynolan (approved):

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
```diff
@@ -403,7 +394,11 @@ export default {
     align-items: center;
     justify-content: flex-end;
     gap: f.spacing(d);
-    padding: f.spacing(d);
+    margin-top: f.spacing(d);
+}
+
+.c-selfExclusion-buttons button {
+    margin: 0 5px 5px;
```
jamieomaguire, line 401: this margin seems a little irregular, what was the reason for 5px? I would have expected to use a value from our spacing set

pttolev: A confirmation dialog appears on top of the button once we select one of the options. The focus of the button appears and doesn't look nice. 8px is too big and 4px is not enough.

pttolev: Anyway, it won't be visible in this format to anyone, so I just removed it.

jamieomaguire (approved):

PR #2515, opened 2024-04-23
f-self-exclusion@1.4.0 - Handling of cases where the exclusion period has expired

m-axii:
## v1.4.0
_May 8, 2024_

### Added
Form functionality to handle cases where the exclusion period has expired.


packages/components/pages/f-self-exclusion/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## v1.3.1
```
xander-marjoram: You'll need to update the version in package.json too

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
```diff
@@ -310,6 +318,13 @@ export default {
             return `${dd}-${mm}-${yyyy}`;
         },
 
+        checkState (state) {
+            if (state === 'exclusionPeriodExpired') {
```
xander-marjoram: It looks like you're checking the same thing as your new computed property offers?
```suggestion
            if (this.exclusionPeriodExpired) {
```

packages/components/pages/f-self-exclusion/src/tenants/da-DK.js
```diff
@@ -19,6 +19,7 @@ const messages = {
         text1Show: 'We’ll show alcohol items.',
         text1Temporary: 'We\'ll exclude items containing alcohol from your menus on Menulog for six months.',
         text1Permanent: 'We\'ll permanently exclude items containing alcohol from your menus on Menulog.',
+        text1Expired: 'You have successfully removed the alcohol self-exclusion. It will take 24 hours for your change to take effect.',
```
xander-marjoram: I'm sure this is more a question for your PM, but this messaging doesn't appear to match the purpose? Expiry != removal

m-axii: This is as intended.

xander-marjoram (dismissed):

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
```diff
@@ -231,6 +233,7 @@ export default {
             isOpenAlertSuccess: false,
             isOpenAlertError: false,
             isFormVisible: false,
+            alcoholExclusionState: '',
```
kevinrodrigues: What's the purpose of this value out of interest?

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
kevinrodrigues, line 1: Will you be adding tests in a follow up PR?

m-axii: This feature has been on prod since February and will be replaced in OneWeb. Would we still need tests in this case?

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
```diff
@@ -253,6 +256,10 @@ export default {
             return this.exclusions.find(exclusion => exclusion.type === 'alcohol');
         },
 
+        exclusionPeriodExpired () {
```
kevinrodrigues, line 258: My Vue is a little rusty but shouldn't these two be moved to a getter and then you read there? Something like below. You can then read direct from the getter within this file? I think you can also move the method above into that too and reference the places it needs to be used via the getter instead.

```javascript
exclusionPeriodExpired: (state, getters) => getters.alcoholExclusion.state === 'exclusionPeriodExpired',

```

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
kevinrodrigues: Yes, we should add tests still.

m-axii: Ok, then we will add tests in a follow-up PR.

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
m-axii: Removed, as it was not used. Thank you for addressing.

packages/components/pages/f-self-exclusion/src/components/SelfExclusion.vue
m-axii: Is there any purpose for this change? It is implemented to follow up the pattern for the rest of the component.

kevinrodrigues: Happy for you to look in a different PR but i think it's a cleaner approach.

kevinrodrigues (approved):

ashleynolan (approved):

PR #2517, opened 2024-04-29
fozzie-components@7.56.2 - More security package updates (vue, storybook)

xander-marjoram:
Almost half of the open Snyk PRs are for upgrading @vue/plugin-cli-babel to v5.

I'm also updating storybook to the latest v6. A major version bump was not trivial and likely requires its own ticket.


xander-marjoram:
On balance I think resolving many vulnerabilities and introducing one is acceptable, especially considering the `@justeat/storybook` project is private and therefore not installable. To resolve the new vulnerability we would need to update storybook to v7, which is likely a large piece of work requiring its own ticket.

Undoing only the `@storybook/vue` upgrade causes other (compatibility) issues.

kevinrodrigues (approved):

fernandofranca (approved):

PR #2543, opened 2024-06-21
f-header@10.21.0 - Add support for tracking contexts

xander-marjoram:
### Added

- New prop `globalTrackingContexts` which are passed through to f-trak for analytics events.

---

## UI Review Checks

### DIFF
--- a/packages/components/organisms/f-footer/package.json
+++ b/packages/components/organisms/f-footer/package.json
@@ -54,7 +54,7 @@
     "v-click-outside": "3.1.2"
   },
   "peerDependencies": {
-    "@justeat/f-trak": ">=0.8.0"
+    "@justeat/f-trak": "1.x"
   },
   "devDependencies": {
     "@justeat/f-wdio-utils": "1.x"
--- a/packages/components/organisms/f-footer/package.json
+++ b/packages/components/organisms/f-footer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@justeat/f-footer",
-  "version": "8.8.0",
+  "version": "8.9.0",
   "main": "dist/f-footer.umd.min.js",
   "maxBundleSize": "80kB",
   "files": [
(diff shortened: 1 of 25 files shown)


kevinrodrigues (approved):

packages/components/organisms/f-footer/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@justeat/f-footer",
-  "version": "8.8.0",
+  "version": "8.9.0",
```
jamieomaguire, line 3: Again I am unsure if this is not a major change. Updating this would stop legacy tracking (if the newer solution was added)

xander-marjoram: It's a fair comment.

These changes should only affect consuming apps if they also do the major version bump of `f-trak` (which is a peer dependency of header and footer, rather than a dependency).

As far as header and footer are concerned it's just adding a new prop which (if they pass something into the new prop) updates the contents of the `data-trak` attributes. If they don't use the new prop or update `f-trak` there are effectively no changes in this version.

dandel10n (approved):

packages/components/organisms/f-footer/package.json
jamieomaguire: If this is allowed to be run without the non-breaking version of f-trak then I am happy with this. If it is going to update f-trak without consumers realising, or force them to, then I'd probably consider it a breaking change. Your call!

xander-marjoram: I took another look into this just to double check. Any consumers using `f-trak` should have it listed in their `package.json`, so the peer dependency version inside `f-header` and `f-footer` updating shouldn't be enough to override that. Even if they have it unpinned, it should be prevented from automatically jumping to the next major version of `f-trak`.

jamieomaguire (approved):

